PR #340, opened 2014-10-10
suite: add --subset option to allow us to schedule subsets of a suit...

athanatos:
...e

First, the old logic tended to do the following:

[0,1,2,3] x [a,b,c,d] -> [(0, a), (0, b), (0, c), ...]

which is a bummer since if we only schedule the first 1/4,
we only get combinations with 0.

Instead, we want:

[0,1,2,3] x [a,b,c,d] -> [(0, a), (1, b), (2, c), ...]

Second, the old logic tended to do the following:

[0,1,2,3] + [a,b,c,d] -> [0,1,2,3,a,b,c,d]

Which is a bummer since we need to all the way through the first
set before we get examples from the second.

Instead, we want:

[0,1,2,3] + [a,b,c,d] -> [0,a,1,b,2,c,3,d]

Lastly, if we have:

[0,1,2] + ([0,1,2],[a,b,c])

and want to schedule 1/3 of the suite, we don't want to
just schedule 0, we probably want to schedule all of
[0,1,2] and 1/3 of the second combo.  Thus, we want to
detect such cases and multiply out [0,1,2] into
[0,1,2,0,1,2,0,1,2].

Signed-off-by: Samuel Just <sam.just@inktank.com>

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -59,6 +59,12 @@ Scheduler arguments:
                               [default: 1]
   -l <jobs>, --limit <jobs>   Queue at most this many jobs
                               [default: 0]
+  --subset <index/outof>      Instead of scheduling the entire suite, break the
+                              set of jobs into <outof> pieces (each of which will
+                              contain each facet at least once) and schedule
+                              piece <index>.  Scheduling 0/<outof>, 1/<outof>,
+                              2/<outof> ... <outof>-1/<outof> will schedule all
+                              jobs in the suite (many more than once).
   -p <priority>, --priority <priority>
                               Job priority (lower is sooner)
                               [default: 1000]
--- a/teuthology/matrix.py
+++ b/teuthology/matrix.py
@@ -1,0 +1,276 @@
+import os
+from fractions import gcd
+
+
+class Matrix:
+    """
+    Interface for sets
+    """
+    def size(self):
+        pass
+
+    def index(self, i):
+        """
+        index() should return a recursive structure represending the paths
+        to concatenate for index i:
+
+        Result :: (PathSegment, Result) | {Result}
+        Path :: string
+
+        {Result} is a frozen_set of Results indicating that
+        the set of paths resulting from each of the contained
+        Results should be concatenated.  (PathSegment, Result)
+        indicates that PathSegment should be prepended to the
+        paths resulting from Result.
+        """
+        pass
+
+    def minscanlen(self):
+        """
+        min run require to get a good sample
+        """
+        pass
+
+    def cyclicity(self):
+        """
+        A cyclicity of N means that the set represented by the Matrix
+        can be chopped into N good subsets of sequential indices.
+        """
+        return self.size() / self.minscanlen()
+
+
+class Cycle(Matrix):
+    """
+    Run a matrix multiple times
+    """
+    def __init__(self, num, mat):
+        self.mat = mat
+        self.num = num
+
+    def size(self):
+        return self.mat.size() * self.num
+
+    def index(self, i):
+        return self.mat.index(i % self.mat.size())
+
+    def minscanlen(self):
+        return self.mat.minscanlen()
+
+
+class Base(Matrix):
+    """
+    Just a single item.
+    """
+    def __init__(self, item):
+        self.item = item
+
+    def size(self):
+        return 1
+
+    def index(self, i):
+        return self.item
+
+    def minscanlen(self):
+        return 1
+
+
+class Product(Matrix):
+    """
+    Builds items by taking one item from each submatrix.  Contiguous
+    subsequences should move through all dimensions.
+    """
+    def __init__(self, item, _submats):
+        assert len(_submats) > 0, \
+            "Product requires child submats to be passed in"
+        self.item = item
+
+        submats = sorted(
+            [((i.size(), ind), i) for (i, ind) in
+             zip(_submats, range(len(_submats)))], reverse=True)
+        self.submats = []
+        self._size = 1
+        for ((size, _), submat) in submats:
+            self.submats.append((self._size, submat))
+            self._size *= size
+        self.submats.reverse()
+
+        self._minscanlen = max([i.minscanlen() for i in _submats])
+
+    def minscanlen(self):
+        return self._minscanlen
+
+    def size(self):
+        return self._size
+
+    def _index(self, i, submats):
+        """
+        We recursively reduce the N dimension problem to a two
+        dimension problem.
+
+        index(i) = (lmat.index(i % lmat.size()), rmat.index(i %
+        rmat.size())) would simply work if lmat.size() and rmat.size()
+        are relatively prime.
+
+        In general, if the gcd(lmat.size(), rmat.size()) == N,
+        index(i) would be periodic on the interval (lmat.size() *
+        rmat.size()) / N.  To adjust, we increment the lmat index
+        number on each repeat.  Each of the N repeats must therefore
+        be distinct from the previous ones resulting in lmat.size() *
+        rmat.size() combinations.
+        """
+        assert len(submats) > 0, \
+            "_index requires non-empty submats"
+        if len(submats) == 1:
+            return frozenset([submats[0][1].index(i)])
+
+        lmat = submats[0][1]
+        lsize = lmat.size()
+
+        rsize = submats[0][0]
+
+        cycles = gcd(rsize, lsize)
+        clen = (rsize * lsize) / cycles
+        off = (i / clen) % cycles
+
+        def combine(r, s=frozenset()):
+            if type(r) is frozenset:
+                return s | r
+            return s | frozenset([r])
+
+        litems = lmat.index(i + off)
+        ritems = self._index(i, submats[1:])
+        return combine(litems, combine(ritems))
+
+    def index(self, i):
+        items = self._index(i, self.submats)
+        return (self.item, items)
+
+class Concat(Matrix):
+    """
+    Concatenates all items in child matrices
+    """
+    def __init__(self, item, submats):
+        self.submats = submats
+        self.item = item
+
+    def size(self):
+        return 1
+
+    def minscanlen(self):
+        return 1
+
+    def index(self, i):
+        out = frozenset()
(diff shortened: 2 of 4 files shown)


zmc:
@athanatos #463 added a pretty usable base for unit tests

zmc:
Also, we should probably drop the old codepath and use the new one since it is apparently compatible. If the existing unit tests aren't sufficient to prove the new codepath doesn't break things, let's add more.

zmc:
Another thing is that the modules should pass ``flake8`` and they don't:

``flake8 teuthology/matrix.py``

teuthology/matrix.py
```diff
@@ -0,0 +1,233 @@
...
+        return self.size() / self.minscanlen()
+
+class Cycle(Matrix):
+    """
+    Run a matrix multiple times
+    """
+    def __init__(self, num, mat):
+        self.mat = mat
+        self.num = num
+    def size(self):
```
zmc: These look like great candidates for the ``@property`` decorator: https://docs.python.org/2/library/functions.html#property

scripts/suite.py
```diff
@@ -59,6 +59,8 @@
                               [default: 1]
   -l <jobs>, --limit <jobs>   Queue at most this many jobs
                               [default: 0]
+  --subset <index/outof>      split suite into outof pieces and schedule index
```
andrewschoen: This help text is confusing to me.

teuthology/matrix.py
```diff
@@ -0,0 +1,270 @@
...
+
+        {Result} is a frozen_set of Results indicating that
+        the set of paths resulting from each of the contained
+        Results should be concatenated.  (PathSegment, Result)
+        indicates that PathSegment should be prepended to the
+        paths resulting from Result.
+        """
+        pass
+
+    def minscanlen(self):
```
andrewschoen: nitpick: this might be easier to read / understand as min_scan_len.

teuthology/matrix.py
```diff
@@ -0,0 +1,270 @@
...
+        return 1
+
+
+class Product(Matrix):
+    """
+    Builds items by taking one item from each submatrix.  Contiguous
+    subsequences should move through all dimensions.
+    """
+    def __init__(self, item, _submats):
+        assert len(_submats) > 0
```
andrewschoen: If we have to use assert please add a message to it detailing why the assert would have failed.

i.e. assert condition, message

teuthology/matrix.py
```diff
@@ -0,0 +1,270 @@
...
+        are relatively prime.
+
+        In general, if the gcd(lmat.size(), rmat.size()) == N,
+        index(i) would be periodic on the interval (lmat.size() *
+        rmat.size()) / N.  To adjust, we increment the lmat index
+        number on each repeat.  Each of the N repeats must therefore
+        be distinct from the previous ones resulting in lmat.size() *
+        rmat.size() combinations.
+        """
+        assert len(submats) > 0
```
andrewschoen: needs a message, please.

teuthology/matrix.py
```diff
@@ -0,0 +1,270 @@
...
+            for i in range(submat.size()):
+                out = out | frozenset([submat.index(i)])
+        return out
+
+class Sum(Matrix):
+    """
+    We want to mix the subsequences proportionately to their size.
+    """
+    def __init__(self, item, _submats):
+        assert len(_submats) > 0
```
andrewschoen: needs a message, please

teuthology/suite.py
```diff
@@ -59,6 +60,10 @@ def main(args):
     filter_in = args['--filter']
     filter_out = args['--filter-out']
 
+    subset = None
+    if args['--subset']:
+        subset = tuple(map(int, args['--subset'].split('/')))
```
andrewschoen: Is this just making sure that each subset is an int and then making a tuple of them?  If so, maybe leave a comment explaining what this does as it's initially difficult to parse.

teuthology/suite.py
```diff
@@ -699,60 +708,133 @@ def build_matrix(path, _isfile=os.path.isfile, _isdir=os.path.isdir,
...
+        if index == outof or index == outof - 1:
+            matlimit = mat.size()
+        else:
+            matlimit = (mat.size() / outof) * (index + 1)
+    else:
+        first = 0
+        mat = _build_matrix(path, _isfile, _isdir, _listdir)
+        matlimit = mat.size()
+
+    print mat.size(), first, matlimit
```
andrewschoen: stray print, please remove

teuthology/suite.py
```diff
@@ -699,60 +708,133 @@ def build_matrix(path, _isfile=os.path.isfile, _isdir=os.path.isdir,
...
+    return generate_combinations(path, mat, first, matlimit)
+
+def _build_matrix(path, _isfile=os.path.isfile,
+                  _isdir=os.path.isdir, _listdir=os.listdir, mincyclicity=0, item=''):
     if _isfile(path):
         if path.endswith('.yaml'):
-            return [(None, [path])]
+            return matrix.Base(item)
+        assert False
```
andrewschoen: Do we really need this?

teuthology/suite.py
```diff
@@ -699,60 +708,133 @@ def build_matrix(path, _isfile=os.path.isfile, _isdir=os.path.isdir,
...
+            submats = []
+            for fn in sorted(files):
+                submats.append(
+                    _build_matrix(
+                        os.path.join(path, fn),
+                        _isfile,
+                        _isdir,
+                        _listdir,
+                        mincyclicity,
+                        fn))
```
andrewschoen: Please remove this whitespace.

teuthology/suite.py
```diff
@@ -699,60 +708,133 @@ def build_matrix(path, _isfile=os.path.isfile, _isdir=os.path.isdir,
...
+                    mincyclicity,
+                    fn)
+                if submat.cyclicity() < mincyclicity:
+                    submat = matrix.Cycle(
+                        int(math.ceil(
+                            mincyclicity / submat.cyclicity())),
+                        submat)
+                submats.append(submat)
+            return matrix.Sum(item, submats)
+    assert False
```
andrewschoen: Is this needed?

teuthology/test/test_matrix.py
```diff
@@ -0,0 +1,70 @@
+from .. import matrix
+
+def do_test(res):
+    sz = res.size()
+    s = frozenset([matrix.generate_lists(res.index(i)) for i in range(sz)])
+    print sz, len(s)
```
andrewschoen: stray print, please remove

teuthology/test/test_matrix.py
```diff
@@ -0,0 +1,70 @@
...
+    if sz != len(s):
+        print sz
```
andrewschoen: stray print, please remove.  Actually, is this condition even needed at all?

teuthology/test/test_matrix.py
```diff
@@ -0,0 +1,70 @@
...
+    for i in range(res.size()):
+        print sorted([j for j in matrix.generate_lists(res.index(i))])
```
andrewschoen: stray print, please remove

teuthology/test/test_matrix.py
```diff
@@ -0,0 +1,70 @@
+from .. import matrix
+
+def do_test(res):
```
andrewschoen: I dislike this method.  It is very obfuscated and difficult to understand what is actually being asserted here.  For example, if test_product3 failed below I'd be very confused as to why it failed or what should have been expected.

teuthology/test/test_matrix.py
```diff
@@ -0,0 +1,70 @@
...
+        print sz
+    for i in range(res.size()):
+        print sorted([j for j in matrix.generate_lists(res.index(i))])
+        assert sz == len(s)
+
+def mbs(num, l):
+    return matrix.Sum(num*10, [matrix.Base(i + (100*num)) for i in l])
+
+class TestMatrix(object):
+    def test_simple(self):
```
alfredodeza: good naming conventions for unit tests is to have meaningful test methods, so that when they fail they can indicate what the failure is. 'test_matrix'  or 'test_simple' are obscure into what the intent of the test is.

This also allows for the test to become generic in the future because the signature allows it (since it is vague).

Also, like @andrewschoen mentioned, the `do_test` obscures further what the intent is here.

teuthology/matrix.py
athanatos: If you don't feel strongly, I'm going to ignore this one.

teuthology/suite.py
athanatos: Yes.

teuthology/suite.py
athanatos: What whitespace?

andrewschoen: The green highlight before _isfile, _isdir, etc.  Maybe I'm just misreading this diff.

teuthology/matrix.py
andrewschoen: @athanatos that's fine, it's nothing to block on.  It was just an observation.

scripts/suite.py
```diff
@@ -59,6 +59,12 @@
                               [default: 1]
   -l <jobs>, --limit <jobs>   Queue at most this many jobs
                               [default: 0]
+  --subset <index/outof>      Instead of scheduling the entire suite, break the
+                              set of jobs into <outof> pieces each of which will
+                              contain each facet at least once and schedule
+                              piece <index>.  Scheduling 0/<outof>, 1/<outof>,
+                              2/<outof> ... <outof>-1/<outof> will schedule all
+                              jobs in the suite (many more than once).
```
zmc: This is still hard to parse; I think that first sentence is missing a couple commas?

teuthology/suite.py
athanatos: That is what emacs thought was the right way to do the indent.  I'm not going to change it unless it really bothers you.

athanatos: Does it want all tabs?

athanatos: Oh, it's just highlighting because it thinks the versions are close enough for a word diff.

PR #401, opened 2015-01-09
Respect gitbuilder_host setting in install task

zmc:


andrewschoen:
Ship it

ktdreyer:
I have canceled the Jenkins run just now because it had hung :(

PR #411, opened 2015-01-19
Remove duplicate package list from install task

zmc:

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -258,9 +258,9 @@ def remove_installed_packages(ctx):
     install_task.remove_packages(
         ctx,
         conf,
-        {"deb": install_task.deb_packages['ceph'] +
+        {"deb": install_task.PACKAGES['ceph']['deb'] +
          ['salt-common', 'salt-minion', 'calamari-server'],
-         "rpm": install_task.rpm_packages['ceph'] +
+         "rpm": install_task.PACKAGES['ceph']['rpm'] +
          ['salt-common', 'salt-minion', 'calamari-server']})
     install_task.remove_sources(ctx, conf)
     install_task.purge_data(ctx)
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -740,7 +740,6 @@ def install(ctx, config):
 
     project = config.get('project', 'ceph')
 
-    global PACKAGES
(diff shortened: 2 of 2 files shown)


zmc:
test this please

teuthology/nuke.py
```diff
@@ -258,9 +258,9 @@ def remove_installed_packages(ctx):
     install_task.remove_packages(
         ctx,
         conf,
-        {"deb": install_task.deb_packages['ceph'] +
+        {"deb": install_task.PACKAGES['ceph']['deb'] +
          ['salt-common', 'salt-minion', 'calamari-server'],
```
andrewschoen: Can these be part of PACKAGES as well, seems weird to tack them on here.  Maybe a common_packages list of some sort that gets added to both the deb and rpm lists?

andrewschoen:
It looks like rpm_packages and deb_packages are being used in nuke.remove_installed_packages

https://github.com/ceph/teuthology/blob/master/teuthology/nuke.py#L255

zmc:
They are, in master. Not in my branch...

teuthology/nuke.py
zmc: I agree it's weird, but I don't think we want to install them whenever we use the install task. Not trying to fix that problem here.

andrewschoen: Fair enough

teuthology/task/install.py
```diff
@@ -776,10 +740,9 @@ def install(ctx, config):
 
     project = config.get('project', 'ceph')
 
-    global deb_packages
-    global rpm_packages
-    debs = deb_packages.get(project, [])
-    rpm = rpm_packages.get(project, [])
+    global PACKAGES
```
andrewschoen: This doesn't need to be global does it?

zmc: I don't think so, but I was trying to make as little impact as possible with the change.

andrewschoen: I can appreciate that concern.  At some point we should start removing cruft like this though.

FWIW, I see PACKAGES being used in install.upgrade_common, but it doesn't seem to depend on  any mutation that happens in install.install caused by it being global.

andrewschoen:
Thanks for removing that global, ship it.

PR #414, opened 2015-01-22
bug: teuthology-nuke with no args gives traceback instead of usage

crt004:
The usage sintaxis allows a call with no parameters that later gives an error, I change the sintax so if there is no parameter it print the usage.

http://tracker.ceph.com/issues/7930 Fixes: #7930

Signed-off-by Carlos <carlyt004@gmail.com>

### DIFF
--- a/scripts/nuke.py
+++ b/scripts/nuke.py
@@ -4,8 +4,26 @@ import teuthology.nuke
 
 doc = """
 usage: teuthology-nuke --help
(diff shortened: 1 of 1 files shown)


alfredodeza:
ok to test

zmc:
Thanks for the patch!

Unfortunately it adds too many different usage statements, and many of them are not valid. Luckily, the fix is very simple: https://github.com/ceph/teuthology/commit/d7faab7587ab16e392c5ed697dff21b854f41783

PR #417, opened 2015-01-23
Enable kernel installs from brew

andrewschoen:
If you have 'brew: <build_id>' in the 'kernel' stanza of your teuthology yaml config this will install the kernel from that build_id on the specified node.  This only works on rhel systems.

I've also tried to abstract the functions that talk with brew into their own functions in teuthology.packaging so we can reuse them in the future to download other things from brew.

For: http://tracker.ceph.com/issues/10561

(I might still want to write tests for this, looking mainly for feedback right now)

### DIFF
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -138,6 +138,8 @@ class TeuthologyConfig(YamlConfig):
         'src_base_path': os.path.expanduser('~/src'),
         'verify_host_keys': True,
         'watchdog_interval': 120,
+        'kojihub_url': 'http://koji.fedoraproject.org/kojihub',
+        'kojiroot_url': 'http://kojipkgs.fedoraproject.org/packages',
     }
 
     def __init__(self, yaml_path=None):
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -1069,12 +1069,14 @@ def task(ctx, config):
 
             # get information about this build from koji
             build_info = get_koji_build_info(build_id, role_remote, ctx)
-
-            need_install[role] = build_info
-            need_version[role] = "{ver}-{rel}.x86_64".format(
+            version = "{ver}-{rel}.x86_64".format(
                 ver=build_info["version"],
                 rel=build_info["release"]
             )
+
+            if need_to_install(ctx, role, version):
+                need_install[role] = build_info
+                need_version[role] = version
         else:
             package_type = role_remote.os.package_type
             larch = role_remote.arch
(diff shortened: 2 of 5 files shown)


teuthology/task/kernel.py
```diff
@@ -1021,11 +1049,35 @@ def task(ctx, config):
...
+                )
+                msg = msg.format(system_type=system_type)
+                log.error(msg)
+                ctx.summary["failure_reason"] = msg
+                ctx.summary["status"] = "dead"
+                raise ConfigError(msg)
+
+            # FIXME: this install should probably happen somewhere else
+            # but I'm not sure where, so we'll leave it here for now.
+            install_package('koji', role_remote)
```
andrewschoen: I thought about putting this call in packaging.get_brew_build_info.

Would that be terrible?

teuthology/config.py
```diff
@@ -138,6 +138,8 @@ class TeuthologyConfig(YamlConfig):
         'src_base_path': os.path.expanduser('~/src'),
         'verify_host_keys': True,
         'watchdog_interval': 120,
+        'brewhub_url': 'http://brewhub.devel.redhat.com/brewhub',
+        'brewroot_url': 'http://download.devel.redhat.com/brewroot/packages',
```
ktdreyer: discussed today in standup: we'll refer to Koji instead of brew, so "kojihub_url", "kojiroot_url".

The Koji packages URL is http://kojipkgs.fedoraproject.org/packages

teuthology/task/kernel.py
```diff
@@ -1021,11 +1049,37 @@ def task(ctx, config):
             if need_to_install_distro(ctx, role):
                 need_install[role] = 'distro'
                 need_version[role] = 'distro'
+        elif role_config.get("koji", None):
+            # installing a kernel from koji
+            build_id = role_config.get("koji")
+            if system_type != "rhel":
```
andrewschoen: I assumed we wanted to limit this to rhel.  Is that correct?

zmc: I think we'd want it to work for Fedora too.

andrewschoen: Cool.  Perhaps we should allow any rpm based systems?

zmc: Actually, yeah, sure. Any rpm systems.

teuthology/task/kernel.py
zmc: It looks like this is currently installing koji in all cases - is that right?

andrewschoen: Yeah, if the remote is rhel it'll try to install koji.

zmc: I think it's fine here at least for now

PR #418, opened 2015-01-23
Make teuthology-lock respect --os-type

andrewschoen:
This fixes the regression in issue #10621


zmc:
We should add a unit test covering this before we merge

andrewschoen:
Agreed, I added the test.  Well, fixed a test that should have caught this to begin with anyway.

PR #419, opened 2015-01-23
DaemonState: add wait()

liewegas:
This is needed for the failure injection tests.

### DIFF
--- a/teuthology/orchestra/daemon.py
+++ b/teuthology/orchestra/daemon.py
@@ -46,6 +46,23 @@ class DaemonState(object):
         self.proc = None
         self.log.info('Stopped')
 
+        """
+        Wait for daemon to exit
+
+        Wait for daemon to stop (but don't trigger the stop).  Pass up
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/daemon.py
```diff
@@ -46,6 +46,23 @@ def stop(self, timeout=300):
...
+        Wait for daemon to stop (but don't trigger the stop).  Pass up
+        any exception.  Mark the daemon as not running.
+        """
+    def wait(self, timeout=300):
+        self.log.debug('waiting for process to exit')
+        try:
+            run.wait([self.proc], timeout=timeout)
+            self.log.info('Stopped')
+        except:
+            self.log.info('Failed')
```
zmc: In these logging calls, might it be useful to know which one stopped or failed?

liewegas: IIRC the logger prefix has the daemon name

zmc: You're right, it does!

PR #420, opened 2015-01-28
deprecate support for a custom downburst config

andrewschoen:
We don't think this is being used anymore so we're gonna deprecate this feature and remove the unused code.

See, http://tracker.ceph.com/issues/10625

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1154,17 +1154,17 @@ def get_distro(ctx):
     """
     Get the name of the distro that we are using (usually the os_type).
     """
-    # ubuntu is our default distro choice
-    os_type = "ubuntu"
+    os_type = None
     if ctx.os_type:
         return ctx.os_type
 
     try:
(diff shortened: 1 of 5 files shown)


teuthology/misc.py
```diff
@@ -1154,26 +1154,16 @@ def get_distro(ctx):
...
     """
-    os_type = None
+    # ubuntu is our default distro choice
+    os_type = "ubuntu"
     if ctx.os_type:
         return ctx.os_type
 
     try:
-        os_type = ctx.config.get('os_type', None)
+        os_type = ctx.config.get('os_type', os_type)
```
andrewschoen: Do we need to worry about the possibility of ctx.config['os_type'] = None?

If for whatever reason that happens this function will return None and provisioning will fail.

Maybe default os_type=None at the top and then return os_type or "ubuntu" would be safer?

zmc: Yeah, for vps we need *some* value. ubuntu has been the default forever.

PR #422, opened 2015-01-29
If the version file is empty, complain.

zmc:
We *could* work around this, but I don't want to mask bugs in the build
system.

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -13,7 +13,9 @@ import urlparse
 from teuthology import misc as teuthology
 from ..orchestra import run
 from ..config import config as teuth_config
-from ..exceptions import UnsupportedPackageTypeError, ConfigError
+from ..exceptions import (UnsupportedPackageTypeError,
+                          ConfigError,
+                          VersionNotFoundError)
 from ..packaging import (
     install_package,
     get_koji_build_info,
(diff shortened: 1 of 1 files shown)


andrewschoen:
Agreed that masking bugs in the build system is something we don't want to do, but also failing jobs is a pain too.

Is there a way to make it complain loud enough that we'd see, but allow the job to continue and possibly finish by giving need_install the sha1 instead of version?

teuthology/task/kernel.py
```diff
@@ -1111,6 +1113,11 @@ def task(ctx, config):
                     log.debug('failed to get utsrelease string using url {url}'.format(
                               url=version_url))
 
+                if not version:
+                    raise VersionNotFoundError(
+                        "{url} is invalid; contains {ver}".format(
```
andrewschoen: considering that version is most likely an empty string, does it make sense to print it?

zmc:
Hmm, yeah, it'd be just as easy to log a warning. Let me think about that for a minute...

andrewschoen:
My only concern with the warning message is if the job passes with the sha1 instead of the version, we never see the warning message.  I surely don't make a habit of reading logs for jobs that pass.

zmc:
Yeah, exactly. I kind of want to force us to deal with build system issues like this. I'll bring it up in standup and see if we can get a consensus.

PR #424, opened 2015-02-03
Download ELFs (regular + debug) from remotes when coredump occurs.

SandonV:
Per issue #5629

Since we need the binaries (regular stripped + non-stripped with
debug symbols) to do debugging this will make it more convinient
for developers so they don't have to track down the files from
the specific build in question and instead they will just be
sitting in the archive folder with the coredump files.

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -389,16 +389,14 @@ def fetch_binaries_for_coredumps(path, remote):
                                                             dump_program))
 
             # Pull Debug symbols:
-            # RPM distro's append their non-stripped ELF's with .debug
-            # When deb based distro's do not.
-            debug_program = '{dump_program}.debug'.format(
-                dump_program=dump_program)
             debug_path = os.path.join('/usr/lib/debug', remote_path)
 
+            # RPM distro's append their non-stripped ELF's with .debug
+            # When deb based distro's do not.
             if remote.system_type == 'rpm':
                 debug_path = '{debug_path}.debug'.format(debug_path=debug_path)
 
-            remote._sftp_get_file(debug_path, os.path.join(coredump_path, debug_program))
+            remote.get_file(debug_path, coredump_path)
 
 
 @contextlib.contextmanager
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -359,18 +359,20 @@ def check_conflict(ctx, config):
     if failed:
         raise RuntimeError('Stale jobs detected, aborting.')
 
+
 def fetch_binaries_for_coredumps(path, remote):
     """
     Pul ELFs (debug and stripped) for each coredump found
     """
-    #Check for Coredumps:
+    # Check for Coredumps:
     coredump_path = os.path.join(path, 'coredump')
     if os.path.isdir(coredump_path):
         log.info('Transferring binaries for coredumps...')
         for dump in os.listdir(coredump_path):
             # Pull program from core file
             dump_path = os.path.join(coredump_path, dump)
-            dump_info = subprocess.Popen(['file', dump_path], stdout=subprocess.PIPE)
+            dump_info = subprocess.Popen(['file', dump_path],
+                                         stdout=subprocess.PIPE)
             dump_out = dump_info.communicate()
 
             # Parse file output to get program, Example output:
@@ -383,18 +385,21 @@ def fetch_binaries_for_coredumps(path, remote):
             remote_path = r.stdout.getvalue()
 
             # Pull remote program into coredump folder:
-            remote._sftp_get_file(remote_path, os.path.join(coredump_path, dump_program))
+            remote._sftp_get_file(remote_path, os.path.join(coredump_path,
+                                                            dump_program))
 
             # Pull Debug symbols:
             # RPM distro's append their non-stripped ELF's with .debug
             # When deb based distro's do not.
-            debug_program = '{dump_program}.debug'.format(dump_program=dump_program)
+            debug_program = '{dump_program}.debug'.format(
+                dump_program=dump_program)
             debug_path = os.path.join('/usr/lib/debug', remote_path)
 
             if remote.system_type == 'rpm':
                 debug_path = '{debug_path}.debug'.format(debug_path=debug_path)
 
             remote._sftp_get_file(debug_path, os.path.join(coredump_path, debug_program))
+
 
 @contextlib.contextmanager
 def archive(ctx, config):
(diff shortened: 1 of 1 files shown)


teuthology/task/internal.py
```diff
@@ -393,6 +393,35 @@ def archive(ctx, config):
...
+                        r = rem.run(args=['which', dump_program], stdout=StringIO())
+                        remote_path = r.stdout.getvalue()
+
+                        # Pull remote program into coredump folder:
+                        rem._sftp_get_file(remote_path, os.path.join(coredump_path, dump_program))
+
+                        # Pull Debug symbols:
+                        # RPM distro's append their non-stripped ELF's with .debug
+                        # When deb based distro's do not.
+                        debug_program = '{dump_program}.debug'.format(dump_program=dump_program)
```
andrewschoen: Do you want .debug here as well?  The comment makes it seem like you wouldn't.  It looks like it's added on later for on rpm based systems.

teuthology/task/internal.py
```diff
@@ -393,6 +393,35 @@ def archive(ctx, config):
                 path = os.path.join(logdir, rem.shortname)
                 misc.pull_directory(rem, archive_dir, path)
 
+                #Check for Coredumps:
```
zmc: This whole thing should be a separate function, or maybe even a subtask?

teuthology/task/internal.py
```diff
@@ -393,6 +393,35 @@ def archive(ctx, config):
...
+                coredump_path = os.path.join(path, 'coredump')
+                if os.path.isdir(coredump_path):
+                    log.info('Transferring binaries for coredumps...')
+                    for dump in os.listdir(coredump_path):
+                        # Pull program from core file
+                        dump_path = os.path.join(coredump_path, dump)
+                        dump_info = subprocess.Popen(['file', dump_path], stdout=subprocess.PIPE)
+                        dump_out = dump_info.communicate()
+                        dump_program = dump_out.split("from '")[1].split(' ')[0]
```
zmc: The above looks like magic. Could you put an example of the output you're parsing in a comment here?

teuthology/task/internal.py
SandonV: I can go ahead and move it to its own function. Only reason I didn't is the code is called only once

teuthology/task/internal.py
SandonV: Yeah sure. I unfortunately couldn't find a better way for figuring out the applocation for the coredump so basically parsing the output of the 'file' command, example output:

ubuntu@teuthology:/a/sage-2015-02-02_14:33:59-smoke-master-testing-basic-multi/736328/remote/plana68/coredump$ file 1422917770.7450.core
1422917770.7450.core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, from 'radosgw --rgw-socket-path /home/ubuntu/cephtest/apache/tmp.client.0/fastcgi_soc'

teuthology/task/internal.py
SandonV: There is a 'source' and a 'destination' filename when we do the copy. debug_program is used for the destination so yes we want the debug program to be named .debug always. The later logic is for the source file on weather or not the source file is expected to have .debug already appended or not.

teuthology/task/internal.py
tchaikov: @SandonV ,  not sure if following line has better readability, maybe a comment with a sample output is good enough:

```re.match(r".* from '(\S+) .*'", dump_out).group(1)```

Comment (SandonV) [inline teuthology/task/internal.py]: @zmc What do you think zack?

I think using split is probably slightly faster than re.match() and doesn't require doing an import re,  and actually easier to read (to me) but... I usually have to think about it a little when it comes to regular expressions. I am fine either way. I use re.match() quite a bit too depending on the situation.

Comment (zmc) [inline teuthology/task/internal.py]: I think the readability is fine; I just thought it'd be useful to put and example in the code comments; sorry if I was unclear!

File: teuthology/task/internal.py [hunk 4]
```diff
@@ -359,6 +359,43 @@ def check_conflict(ctx, config):
...
+            # Pull Debug symbols:
+            # RPM distro's append their non-stripped ELF's with .debug
+            # When deb based distro's do not.
+            debug_program = '{dump_program}.debug'.format(dump_program=dump_program)
+            debug_path = os.path.join('/usr/lib/debug', remote_path)
+
+            if remote.system_type == 'rpm':
+                debug_path = '{debug_path}.debug'.format(debug_path=debug_path)
+
+            remote._sftp_get_file(debug_path, os.path.join(coredump_path, debug_program))
```
zmc: Why not ``remote.get_file()``?

teuthology/task/internal.py
```diff
@@ -359,6 +359,43 @@ def check_conflict(ctx, config):
     if failed:
         raise RuntimeError('Stale jobs detected, aborting.')
 
+def check_coredumps(path, remote):
```
zmc: The name of this function doesn't really describe what it does. Maybe ``fetch_binaries_for_coredumps()`` ?

teuthology/task/internal.py
SandonV: @zmc because remote.get_file copies the file to /tmp and returns its path where as I just want to copy the file which the _sftp_sftp_get_file() function does whihch get_file() calls. Seems like its doing quite a bit if unnecessary steps in there.

teuthology/task/internal.py
SandonV: Sure, I will rename.

teuthology/task/internal.py
zmc: Ah, okay. Well, I agree the behavior of ``Remote.get_file()`` is a bit wonky but I don't want to necessarily support calling semi-private functions like this. I think it'd be worth me taking a few minutes to make ``Remote.get_file()`` behave in a more sane way so that you can use it.

What if it first tried to store the file locally in the ``dest_dir`` using the original filename, and if that was already taken, called ``mkstemp()``?

SandonV: well still going to have to do some additional checks and stuff like rename the file because the source filename and destination isn't the same file-name necessarily either. Maybe if get_file took a destination path instead of just a destination folder or a separate more simplified function? The _sftp_get_file() literally did exactly what I wanted which is why I used it.

zmc: As we discussed in IRC yesterday, here's the change I was proposing:
https://github.com/ceph/teuthology/pull/443

zmc:
test this please

PR #427, opened 2015-02-05
add Travis CI integration

ktdreyer:
This will allow Travis CI to run tox to check for test failures.

This should allow us to avoid depending on Jenkins to test pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python:
+    - 2.7
+
+install:
+    - pip install tox
+
+script: tox -rv


andrewschoen:
This is awesome. +1 from me.

andrewschoen:
Somebody with admin rights to this repo will have to create a profile at travis-ci.org so it'll get the necessary perms to update the status with a PR fails or passes tests.

ktdreyer:
(to clarify: we will disable Jenkins' PR tests once we can get this merged to master.)

ktdreyer:
@andrewschoen I think it already does have permissions, right? It just indicated that this PR itself passed.

andrewschoen:
@ktdreyer you're right, I guess somebody already did that.  One less thing to do then :)

zmc:
I didn't think we agreed on moving the unit tests; we still need Jenkins to build docs, don't we?

andrewschoen:
Docs builds only happen on merge to master, that's a different jenkins
build I believe.

zmc:
Still, it means we are still going to rely on Jenkins. I don't see the point of moving only some stuff away from it given that.

andrewschoen:
Fair enough, I can see that point.

ktdreyer:
Currently when Jenkins breaks, we encounter two issues:
1) We can't test PRs
2) We can't push teuthology docs to ceph.com

Today, with the status quo, issue 2 is solved by falling back to reading the docs via GitHub's web interface. It's not ideal, but at least there's some solution that's not incredibly painful. Issue 1 has no fallback (well, maybe "test PRs by hand").

Moving to Travis CI for testing PRs will allow us to have fallback solutions for both issues 1 and 2.

I'm not saying we'll get rid of Jenkins for everything; there are several use cases for which we'll still need Jenkins. I'm just saying that in the particular case of teuthology PR reviews, it doesn't make sense to run our own infrastructure for something that a "cloud" provider will do for free - particularly when we don't have the people resources to keep Jenkins going well enough to enable day-to-day teuthology development in the form of PR reviews.

zmc:
I'll merge this, but I don't want Jenkins disabled at this time. Thanks for the PR!

PR #428, opened 2015-02-05
Don't unlock VMs before destroying them

zmc:


andrewschoen:
Good catch.  Definitely could be dangerous to destroy after unlocking them.

PR #429, opened 2015-02-06
Adds a machine_type property to orchestra.remote.Remote

andrewschoen:
There is also some minor test clean up and I made the existing machine_type test use the new property.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -82,6 +82,15 @@ class Remote(object):
         return self._hostname
 
     @property
+    def machine_type(self):
+        if not getattr(self, '_machine_type', None):
+            remote_info = ls.get_status(self.hostname)
+            if not remote_info:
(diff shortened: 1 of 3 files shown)


zmc:
test this please

teuthology/orchestra/remote.py
```diff
@@ -82,6 +82,17 @@ def hostname(self):
         return self._hostname
 
     @property
+    def machine_type(self):
+        if not getattr(self, '_machine_type', None):
+            remote_info = ls.get_status(self.hostname)
+            if not remote_info:
+                log.info(
+                    "Failed to query lock server for status of {remote}".format(remote=self))
```
zmc: This should be at warning I think

andrewschoen: Agreed.  I just made that change and moved this logging statement to teuthology.lockstatus

PR #433, opened 2015-02-12
Add hang-on-error config option

dzafman:
Signed-off-by: David Zafman <dzafman@redhat.com>

### DIFF
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -1,5 +1,6 @@
 import sys
 import logging
+import time
 from .sentry import get_client as get_sentry_client
 from .job_status import set_status
 from .misc import get_http_log_path
@@ -106,6 +107,9 @@ def run_tasks(tasks, ctx):
             from .task import interactive
             log.warning('Saw failure during task execution, going into interactive mode...')
             interactive.task(ctx=ctx, config=None)
+        elif ctx.config.get('hang-on-error'):
+            while True:
+                time.sleep(300)
         # Throughout teuthology, (x,) = y has been used to assign values
         # from yaml files where only one entry of type y is correct.  This
         # causes failures with 'too many values to unpack.'  We want to
@@ -143,6 +147,9 @@ def run_tasks(tasks, ctx):
                         log.warning(
                             'Saw failure during task cleanup, going into interactive mode...')
                         interactive.task(ctx=ctx, config=None)
+                    elif ctx.config.get('hang-on-error'):
+                        while True:
+                            time.sleep(300)
                 else:
                     if suppress:
                         sys.exc_clear()
--- a/README.rst
+++ b/README.rst
@@ -283,6 +283,12 @@ chance to inspect the system -- both through Teuthology and via extra
 SSH connections -- and the cleanup completes only when you choose so.
 Just exit the interactive Python session to continue the cleanup.
 
+    hang-on-error: true
+
+will allow a scheduled job to be examined before machines are cleaned up.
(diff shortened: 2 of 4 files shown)


dzafman:
I ran teuthology with a yaml that had a test that would fail and saw this output and no prompt.

2015-02-11 15:51:59,609.609 WARNING:teuthology.contextutil:Saw failure, hanging...

andrewschoen:
I'm confused how this is any different than interactive-on-error.  It seems to do the same thing just without the prompt.

Even if we couldn't get to the prompt from a scheduled job wouldn't adding interactive-on-error to a scheduled jobs yaml do the same thing?  I suspect it might suspend the job until we teuthology-kill it.

dzafman:
I assumed that interactive-on-error wouldn't work for a scheduled job, but I have not tried it.

liewegas:
> I assumed that interactive-on-error wouldn't work for a scheduled job, but I have not tried it.

iirc it hangs?  or it fails out anyway because there's no stdin.. i forget

teuthology/run_tasks.py
```diff
@@ -143,6 +147,9 @@ def run_tasks(tasks, ctx):
                         log.warning(
                             'Saw failure during task cleanup, going into interactive mode...')
                         interactive.task(ctx=ctx, config=None)
+                    elif ctx.config.get('hang-on-error'):
+                        while True:
+                            time.sleep(300)
```
zmc: ``hang-on-error`` hangs... forever?!
https://www.youtube.com/watch?v=gvdf5n-zI14

zmc:
I understand wanting to be able to inspect a running cluster that's failed testing, but this seems dangerous to me; we have test throughput issues as it is. I wonder if there's a less obtrusive way to help you get the information you need.

andrewschoen:
Yeah, I'm not really familiar with the library that powers interactive-on-error so it might very well fail because of no stdin.

I agree with @zmc though, a while True seems dangerous. teuthology.contextutil.safe_while would be a better option if we can't figure out how to get at the info you need in another way.

dzafman:
The hang-on-error option wasn't intended to be used by a suite in ceph-qa-suite.  I created it to be able get at the machines for a hard to reproduce problem.  I was going to schedule multiple copies of a run and have it hang when a failure occurs.  What does teuthology.contextutil.safe_while do?

andrewschoen:
@dzafman it'll let you specify an exit condition for your while loop. 

http://ceph.com/teuthology/docs/teuthology.html#teuthology.contextutil.safe_while

dzafman:
I don't have a while loop.  I'm taking a yaml from a failed suite run generated by a test suite.  Modifying it to run as a stand alone run and adding hang-on-error: true.  Then doing a teuthology-schedule with -N 100.

zmc:
@dzafman you do have a while loop; it's where I commented a couple hours ago.

100 jobs using this feature as implemented will bring down the entire lab, potentially for days. I really don't think I can support this.

dzafman:
@zmc This isn't important right now, but I still don't have enough understanding to do what I had wanted to do.  Also, it sounds like I would have to modify the run_tasks machinery to keep re-running the job, until a failure is seen.  This way only a single job needs to be scheduled and only a single job could potentially tie up machines.

jdurgin:
@dzafman if you just want to run a job many times and pause on failure, you can easily do that with a yaml from a job by adding interactive-on-error: true and passing it to https://github.com/ceph/teuthology/blob/master/hammer.sh

PR #434, opened 2015-02-12
Hadoop 2.0 Task

noahdesu:
This isn't perfect yet, but it does use (1) Hadoop 2x series and (2) uses the gitbuilder bindings for the new hadoop-cephfs branch and (3) has been tested with the new wordcount.sh workunit in ceph branch 'hadoop' with outstanding pull request.


gregsfortytwo:
okay to test

PR #435, opened 2015-02-12
Safer vps unlocking

zmc:
Previously we'd see things like this:
```
07:44:30 vpm011 locked by job 749579
07:44:32 downburst created vpm011 for job 749579
07:46:01 downburst destroyed vpm011 for job 749584
07:46:02 vpm011 unlocked by job 749584
07:46:03 vpm011 locked by job 749554
07:46:05 downburst created vpm011 for job 749554
07:46:16 job 749579 lost connection to vpm011
07:46:16 downburst destroyed vpm011 for job 749579
07:46:17 vpm011 unlocked by job 749579
```

It was because:
1. Before destroying a VM, teuthology didn't check to see if it had the same ``description`` on the lock server, or even if it had the same ``locked_by`` value
2. teuthology didn't pass the ``description`` the the lock server when unlocking
3. Paddles didn't even care about ``description``s when unlocking single nodes

3 is fixed by https://github.com/ceph/paddles/pull/52. 1 and 2 are fixed by this; if we try to destroy or unlock a VM that doesn't have matching ``locked_by`` and ``description`` values, each action will fail.

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -427,13 +427,14 @@ def unlock_many(names, user):
     return response.ok
 
 
-def unlock_one(ctx, name, user=None):
+def unlock_one(ctx, name, user=None, description=None):
     if user is None:
         user = misc.get_user()
     name = misc.canonicalize_hostname(name, user=None)
-    if not provision.destroy_if_vm(ctx, name, user):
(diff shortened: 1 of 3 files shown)


zmc:
unit tests were broken by #434

PR #437, opened 2015-02-12
Fixes the issue of pytest failing with scheduled jobs

andrewschoen:
When you schedule a job it's in a background process with no terminal. When we invoked pytest it would die trying to report the status to the nonexistent terminal.  So, just unregister the TerminalReporter so it doesn't complain.

Also made it so if pytest dies the job reports as 'dead'.

### DIFF
--- a/teuthology/task/tests/__init__.py
+++ b/teuthology/task/tests/__init__.py
@@ -74,15 +74,21 @@ def task(ctx, config):
     and then executing them with the teuthology ctx and config args.
     Your tests must follow standard pytest conventions to be discovered.
     """
-    status = pytest.main(
-        args=[
-            '-q',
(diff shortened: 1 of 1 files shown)


teuthology/task/tests/__init__.py
```diff
@@ -38,6 +38,15 @@ def pytest_generate_tests(self, metafunc):
         # pass the teuthology ctx and config to each test method
         metafunc.parametrize(["ctx", "config"], [(self.ctx, self.config),])
 
+    @pytest.mark.trylast
+    def pytest_configure(self, config):
+        # removes the default pytest TerminalReporter
+        # this fixes failures with scheduled jobs; when ran by a worker
```
dmick: "when run"

PR #441, opened 2015-02-17
Bug #10803: Try apt-get -f install to workaround nuke failure.

SandonV:
Sometimes packages or dpkg gets in a weird state where a nuke will
fail to run and simply running an apt-get -f install on the
machine will fix things up so it can be nuked again. Currently
we only ran that being dependent on dpkg --configure completing
where as I changed it so it will still try apt-get -f install
to potentially require less manual intervention when cleaning up
machines. I kept the original behavior but adding some arguments
to insure it will not prompt you for a yes/no (like apt-get -f
install usually would do).

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -245,8 +245,9 @@ def dpkg_configure(ctx):
         proc = remote.run(
             args=[
                 'sudo', 'dpkg', '--configure', '-a',
-                run.Raw('&&'),
-                'sudo', 'apt-get', '-f', 'install',
+                run.Raw(';'),
+                'sudo', 'DEBIAN_FRONTEND=noninteractive',
+                'apt-get', '-y', '--force-yes', '-f', 'install',
                 run.Raw('||'),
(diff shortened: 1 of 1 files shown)


SandonV:
Simple change but I finally got a chance to test it on a machine that nuke was otherwise hanging on and with this change nuke was able to complete. Sometimes apt is in a state where an apt-get -f install does not fix but often times this can fix it and save manual intervention.

zmc:
Are you doing it twice on purpose?

SandonV:
Well the original code ran it if the dpkg --configure ran with a clean status I want to run it whether or not it does or doesn't. I wasn't sure why it as only clan exit status before but I am guessing there is a reason so I didn't want to necessarily change the default behavior except if the dpkg --configure fails.

zmc:
Oh, I didn't parse the call right. In that case, why not just use two different ``run()`` calls and pass ``check_status=False`` to the dpkg one?

PR #444, opened 2015-02-20
install: use newly-split ceph-devel package names

ktdreyer:
The ceph-devel package was split up on Ceph's hammer branch in https://github.com/ceph/ceph/pull/3744 .  This pull request updates Teuthology's install task to install the various sub-packages instead of the deprecated "ceph-devel" package.

Note that I'm not including `libradosstriper1-devel` here because the `libradosstriper1` package is not in this list. If that should be changed, please let me know and I'll amend this commit.


ktdreyer:
Note that the packages have *not* been split up anywhere other than the hammer branch. Not sure what that means for how Teuthology works - I'll need someone more well-versed in Teuthology to answer that.

andrewschoen:
Yeah, we won't be able to merge this as is.  If this merges without the packaging split in all branches of ceph then every teuthology run not pointed at hammer will fail.

dmick:
Maybe make install adaptive by trying the new names, and if they don't exist, fall back?

andrewschoen:
This feels like something the qa suite should be able to tell teuthology to do.  Ceph branch specific things typically happen there.

dmick:
Yeah, but this is a transition, and most suite users Just Don't Care.  Maybe it should key off specific versions in the code rather than dynamically testing, but I think it's overload to require a YAML option.

andrewschoen:
I was just thinking this might be a good use case to try out http://tracker.ceph.com/issues/10589

A new task that could let you install and/or remove arbitrary packages could be helpful along with the existing install task. It'd solve this use case I think and give us a chance to see how feasible #10589 is.

But yes, we could make this work with the existing install task.  However, in general I'd like to see teuthology start to be less ceph specific and more flexible in how it installs things.

ktdreyer:
In talking with Andrew, I realized that teuthology-nuke is going to continue to work even without this PR, because the nuke task runs `yum remove librados2`, so yum's dependency-solver is going to just clean out everything, including all the new -devel packages.

For the install task, well, `yum install ceph-devel` is going to continue to work as well, since we've put the compatibility bits into place in the RPM .spec file.

Eventually it would be nice to use the real package names and not rely on this backwards-compatibility stuff in the packaging, but since we need teuthology's `teuthology/task/install.py` to work across all branches of ceph, I think bigger changes are needed (like http://tracker.ceph.com/issues/10589 or something like it) before this PR can go in.

kshtsk:
This is a very old patch, more the 4 and half years old, is this still actual? If not can we close the PR?

smithfarm:
Please reopen if the patch is still relevant.

PR #449, opened 2015-03-03
Don't use yaml in FakeNamespace's repr()

zmc:


zmc:
Should fix http://tracker.ceph.com/issues/10750

PR #451, opened 2015-03-12
Verify that the correct ceph version is installed on updates and install

andrewschoen:
This adds a function to teuthology.packaging that can be reused to
fetch the installed version of a package for the given remote.


zmc:
test this please

PR #453, opened 2015-03-19
Allow running teuthology-lock --update on broken vps nodes

zmc:


zmc:
@dmick good point. I'll tweak.

PR #455, opened 2015-03-23
github compatible reference to the INSTALL documentation

dachary:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,7 +43,7 @@ Provided Utilities
 
 Installation
 ============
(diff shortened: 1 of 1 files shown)


dachary:
I actually have no clue how to do that properly :-(

andrewschoen:
@dachary I think the problem we ran into is that we couldn't find a solution that worked for both our hosted docs and github viewing.  We decided it'd be better if the linked worked on our hosted docs, for now anyway.

PR #457, opened 2015-03-26
valgrind: ignore lttng warning on un-cleaned-up state

gregsfortytwo:
This appears to have come from dca722ec7b2a7fc9214844ec92310074b5cb2faa,
which merged in support for use of lttng ust functions on fork. Valgrind
started warning on a Leak_StillReachable with lttng_ust_init in
the function call trace.

Hopefully this is specific enough that we don't mask out any other
lttng errors.

Signed-off-by: Greg Farnum <gfarnum@redhat.com>

### DIFF
--- a/teuthology/task/valgrind.supp
+++ b/teuthology/task/valgrind.supp
@@ -235,3 +235,12 @@
 	...
 	fun:exit
 }
+{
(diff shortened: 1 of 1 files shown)


liewegas:
looks okay to me!

PR #459, opened 2015-04-06
Refactor downburst provisioning

zmc:
This allows for better logging, easier debugging, and a proper way to give up attempting to start up instances that are somehow broken. I also intend for it to serve as a model to follow when adding support for e.g. Cobbler reimaging and OpenStack provisioning.

### DIFF
--- a/teuthology/provision.py
+++ b/teuthology/provision.py
@@ -5,6 +5,7 @@ import tempfile
 import yaml
 
 from .config import config
+from .contextutil import safe_while
 from .misc import decanonicalize_hostname, get_distro, get_distro_version
 from .lockstatus import get_status
 
@@ -12,7 +13,7 @@ from .lockstatus import get_status
 log = logging.getLogger(__name__)
 
 
-def _get_downburst_exec():
+def downburst_executable():
     """
     First check for downburst in the user's path.
     Then check in ~/src, ~ubuntu/src, and ~teuthology/src.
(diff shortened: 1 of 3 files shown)


zmc:
http://pulpito.front.sepia.ceph.com/zack-2015-04-02_21:06:32-teuthology-master---basic-multi/

teuthology/provision.py
```diff
@@ -29,73 +30,147 @@ def _get_downburst_exec():
     import pwd
     little_old_me = pwd.getpwuid(os.getuid()).pw_name
     for user in [little_old_me, 'ubuntu', 'teuthology']:
-        pth = os.path.expanduser("~%s/src/downburst/virtualenv/bin/downburst"
-                                 % user)
+        pth = os.path.expanduser(
+            "~%s/src/downburst/virtualenv/bin/downburst" % user)
```
andrewschoen: It kinda feels like this path should be in the teuthology config.

teuthology/provision.py
```diff
@@ -29,73 +30,147 @@ def _get_downburst_exec():
...
+
+    def remove_config(self):
+        if self.config_path and os.path.exists(self.config_path):
+            os.remove(self.config_path)
+            self.config_path = None
+            return True
+        return False
+
+    @property
+    def is_up(self):
```
andrewschoen: Did you mean to implement this?

andrewschoen:
Overall this is a huge improvement from what we had!  I would really like to see tests on the new code though, especially the new Downburst class.  Some additional documentation in that class would be helpful as well.

teuthology/provision.py
zmc: I think it should mostly just look in $PATH and not feel responsible for poking around in other locations, but I also think the scope of this PR is large enough ;)

teuthology/provision.py
zmc: Ha! Good question. It's not really necessary. I'll remove it.

zmc:
@andrewschoen I added some docs. I'm not entirely sure what the best way to go about writing tests for this would be, though. Maybe I could pick your brain sometime soon.

teuthology/provision.py
andrewschoen: Sure, that's reasonable.

andrewschoen:
@zmc The additional docs help, thanks.  I'd love to talk tests sometime.  I understand that some of the existing teuthology code is difficult to test, but any new code should include some test coverage if possible I think.

teuthology/test/test_provision.py
```diff
@@ -0,0 +1,54 @@
+from mock import Mock, MagicMock
+
+from teuthology import provision
+
+
+class TestDownburst(object):
+    def test_create_if_vm(self):
+        ctx = Mock()
```
andrewschoen: The creation of ctx should probably be in a setup method if it's going to be the same across the tests.

andrewschoen:
@zmc thanks for writing some tests! There are some conditions in destroy_if_vm we might want to test.  For example, you check user, description and status_info in destroy_if_vm.  It looks like create_if_vm does a check against status_info as well.

Also, if you wanted to test _run_create or destroy all I think it would take is a simple patch of subprocess.Popen and then just validate that the mock is getting called with the appropriate args.

teuthology/test/test_provision.py
zmc: It's common to these two tests, but I'm not sure new tests will need it at all.

andrewschoen:
@zmc here's a decent example of mocking subprocess.Popen and validating what args it was called with.

https://github.com/ceph/teuthology/blob/master/teuthology/test/test_worker.py#L146

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+        """
+        if not self.executable:
+            log.error("No downburst executable found.")
+            return False
+        self.build_config()
+        success = None
+        with safe_while(sleep=60, tries=3,
+                        action="downburst create") as proceed:
+            while proceed():
+                (returncode, stdout, stderr) = self._run_create()
```
alfredodeza: the parenthesis are not needed here

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+                if returncode == 0:
+                    log.info("Downburst created %s: %s" % (self.name,
+                                                           stdout.strip()))
+                    success = True
+                    break
+                elif stderr:
+                    # If the guest already exists first destroy then re-create:
+                    if 'exists' in stderr:
+                        success = False
```
alfredodeza: if `success` is `False` if `exists` is in `stderr` or not, then it is not needed to define it twice. Since the only place where it should be `True` is when there is a `returncode` that is `0` then it could just default to `True` (instead of `None`) unless that condition is not met.

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+                        break
+            return success
+
+    def _run_create(self):
+        """
+        Used by create(), this method is what actually calls downburst when
+        creating a virtual machine instance.
+        """
+        if not self.config_path:
+            raise ValueError("I need a config_path!")
```
alfredodeza: usually logs or error messages are better presented without using the first person in them, so in this case, it would look like:

    raise ValueError('a config_path is required')

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
         log.info("Provisioning a {distro} {distroversion} vps".format(
-            distro=os_type,
-            distroversion=os_version
+            distro=self.os_type,
+            distroversion=self.os_version
         ))
+        proc = subprocess.Popen(args, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        out, err = proc.communicate()
+        return (proc.returncode, out, err)
```
alfredodeza: parenthesis are not required here either

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+        if err:
+            log.error("Error destroying {machine}: {msg}".format(
+                machine=self.name, msg=err))
+            return False
+        elif proc.returncode == 0:
+            out_str = ': %s' % out if out else ''
+            log.info("Destroyed %s%s" % (self.name, out_str))
+            return True
+        else:
+            log.error("I don't know if the destroy of {node} succeded!".format(
```
alfredodeza: Same as the previous comment (better to avoid first person in logs/messages).

Also, if it is unknown if `destroy` worked, then why is this an error? Maybe it should be a warning?

alfredodeza: If this scenario where there is no `stderr` and the `returncode` is a positive integer, it might be useful to use `stdout` (if any)

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+        return (proc.returncode, out, err)
+
+    def destroy(self):
+        """
+        Destroy (shutdown and delete) a virtual machine instance.
+        """
+        executable = self.executable
+        if not executable:
+            log.error("No downburst executable found.")
+            return False
```
alfredodeza: I didn't see tests for this, shouldn't be difficult to add one for this condition

teuthology/provision.py
zmc: I dislike 'assume success' patterns; that's why this looks the way it does.

teuthology/provision.py
zmc: I'm returning False, so it should be an error. That said, to my knowledge this has never happened. The downburst log itself will be most helpful for debugging, if it ever does happen.

teuthology/provision.py
zmc: Honestly this won't happen unless someone introduces an obvious bug. I don't expect any user to see this.

teuthology/provision.py
zmc: Done

teuthology/provision.py
```diff
@@ -29,105 +30,200 @@ def _get_downburst_exec():
...
+            self.config_path = None
+            return True
+        return False
+
+    def __del__(self):
+        self.remove_config()
 
 
-def destroy_if_vm(ctx, machine_name, user=None, description=None):
+def create_if_vm(ctx, machine_name, _downburst=None):
```
andrewschoen: Would you please mention in the comments that the _downburst kwarg is used only for testing? The destroy method needs this as well.

andrewschoen:
@zmc tests are looking good.  If you can add that one comment about _downburst I'll be ready to merge.

teuthology/provision.py
alfredodeza: if it is for debugging then it should include as much information as possible. If there is not `stderr` available but there is `stdout` then it should be used.

Also, if something is an error then the wording should be changed to match that. "don't know if {action} succeeded" does not convey an error condition

teuthology/provision.py
alfredodeza: normalized wording will also make it easier to follow when adding other logic around the whole project. If it is something that is not expected then it should not be added. If it is going to be ever expected then a more suitable wording should be used.

This is an example of a lot of useful rules for wording on program messages that I find good to follow: http://www-01.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.nlsgdrf/write_styles.htm

teuthology/provision.py
alfredodeza: It is not assuming success, it would default to success unless an error condition occurs, which this code is defining twice needing it.

PR #461, opened 2015-04-07
Add teuthology-update-inventory command

zmc:
This ought to also replace teuthology-updatekeys, but might not be as
fast yet.

### DIFF
--- a/scripts/update_inventory.py
+++ b/scripts/update_inventory.py
@@ -1,0 +1,33 @@
+import docopt
+
+import teuthology
+import teuthology.lock
+import teuthology.orchestra.remote
+
+import logging
+
+doc = """
+usage: teuthology-update-inventory -h
+       teuthology-update-inventory [-v] REMOTE [REMOTE ...]
+
+Update the given nodes' inventory information on the lock server
+
+
+  -h, --help            show this help message and exit
+  -v, --verbose         be more verbose
+  REMOTE                hostnames of machines whose information to update
+
+"""
+
+
+def main():
+    args = docopt.docopt(doc)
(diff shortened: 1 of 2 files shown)


scripts/update_inventory.py
```diff
@@ -0,0 +1,35 @@
...
+def main():
+    args = docopt.docopt(doc)
+    if args['--verbose']:
+        teuthology.log.setLevel(logging.DEBUG)
+
+    remotes = args.get('REMOTE')
+    for rem_name in remotes:
+        remote = teuthology.orchestra.remote.Remote(rem_name)
+        inventory_info = remote.inventory_info
+        del inventory_info['user']
```
andrewschoen: Why are we deleting this key? Maybe could use a comment to clarify.

scripts/update_inventory.py
```diff
@@ -0,0 +1,35 @@
+import docopt
+
+import teuthology
+import teuthology.config
+import teuthology.lock
+import teuthology.orchestra.remote
```
andrewschoen: Are all these imports needed? Looks like you're using the full path to the modules being used below.

scripts/update_inventory.py
zmc: Ah, that turned out to be unnecessary. The user value would be incorrect, if paddles even used it. It doesn't yet, and when it does I'll have to fix it across the board. Removing that line.

scripts/update_inventory.py
zmc: The .config import is unnecessary so I'm taking it out. The other two are needed, yes.

PR #462, opened 2015-04-13
When modifying authorized_keys, store a backup

zmc:
The other changes are my text editor automatically removing end-of-line whitespace.

### DIFF
--- a/teuthology/task/ssh_keys.py
+++ b/teuthology/task/ssh_keys.py
@@ -6,6 +6,7 @@ import contextlib
 import logging
 import paramiko
 import re
+from datetime import datetime
 
 from cStringIO import StringIO
 from teuthology import contextutil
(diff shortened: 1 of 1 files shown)


teuthology/task/ssh_keys.py
```diff
@@ -15,6 +16,31 @@
...
 
+
+def timestamp(format_='%Y-%m-%d_%H:%M:%S:%f'):
+    """
+    Return a UTC timestamp suitable for use in filenames
+    """
+    return datetime.utcnow().strftime(format_)
+
+
+def backup_file(remote, path, sudo=False):
```
andrewschoen: Would be nice to have test coverage on this.

teuthology/task/ssh_keys.py
```diff
@@ -15,6 +16,31 @@
...
+    ]
+    if sudo:
+        args.insert(0, 'sudo')
+    remote.run(args=args)
+    return backup_path
+
+
 def generate_keys():
     """
     Generatees a public and private key
```
dmick: we gotta fix this while we're here.  it keeps bugging me.

zmc: Ha. And all the linter errors too...

dmick:
Otherwise looks like a good first step.

teuthology/task/ssh_keys.py
zmc: I'm not sure how this should be tested and, since it's likely a temporary addition, I don't care too much about making it the first task with unit tests.

zmc:
@dmick are we blocking on me fixing someone else's typo?

PR #463, opened 2015-04-14
Basic teuthology.suite.build_matrix() unit tests

zmc:


andrewschoen:
@zmc Looking good.  Could you add docs for the fake_* methods in the tests?  Something for TestBuildMatrix explaining how self.fs works would be nice as well.

PR #464, opened 2015-04-16
git.ceph.com

liewegas:


zmc:
The install task and internal task also reference ceph.com/git. They should all probably use ``config.ceph_git_base_url``

liewegas:
no ssl yet, do not merge

trhoden:
:+1: for the https->http changes

zmc:
Can these be squashed please?

PR #469, opened 2015-04-23
Adding missing files in setup.py

zhouyuan:
Some non-python scripts are missing when install.


zmc:
ok to test

zmc:
Thanks!

PR #470, opened 2015-04-23
Add timeouts to Remote connection functions

zmc:
This PR also splits ``task.internal.connect()`` into two subtasks.

Both of these changes are needed to support the provisioning work I'm doing in other branches.


zmc:
Integration tests: http://pulpito-rdu.front.sepia.ceph.com/zack-2015-04-23_14:42:39-teuthology-hammer---basic-typica/

PR #471, opened 2015-04-23
ssh_keys: save each remote's backup file name separately in a dict for u...

dmick:
...ndo


dmick:
tested with three remotes this time.

PR #472, opened 2015-04-23
lock.py: Add '--json-query' to allow complex search for --list/--brief

dmick:
--json-query <file> or --json-query <string> allows filtering search
results conveniently; the JSON must be a dictionary that is a subset
of the lock record to be matched.  The reason this was invented:
Find all VPSes on a particular vmhost:

teuthology-lock --list --all --json-query \
	'{"vm_host":{"name":"mira003.front.sepia.ceph.com"}}'

### DIFF
--- a/scripts/lock.py
+++ b/scripts/lock.py
@@ -158,5 +158,17 @@ def parse_args():
         default=None,
         help='OS (distro) version such as "12.10"',
     )
+    parser.add_argument(
+        '--json-query',
+        default=None,
+        help=textwrap.dedent('''\
+            JSON fragment, explicitly given, or a file containing
+            JSON, containing a query for --list or --brief.
+            Example: teuthology-lock --list --all --json-query
+            '{"vm_host":{"name":"mira003.front.sepia.ceph.com"}'
+            will list all machines who have a vm_host entry
+            with a dictionary that contains at least the name key
+            with value mira003.front.sepia.ceph.com.'''),
+    )
 
     return parser.parse_args()
--- a/teuthology/test/test_misc.py
+++ b/teuthology/test/test_misc.py
@@ -150,3 +150,20 @@ class TestMergeConfigs(object):
     def test_invalid_b_deep_merge(self):
         with pytest.raises(AssertionError):
             misc.deep_merge({"a": "b"}, "invalid")
+
+
+class TestIsInDict(object):
+    def test_simple_membership(self):
+        assert misc.is_in_dict('a', 'foo', {'a':'foo', 'b':'bar'})
+
+    def test_dict_membership(self):
+        assert misc.is_in_dict(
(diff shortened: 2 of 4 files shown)


scripts/lock.py
```diff
@@ -158,5 +158,10 @@ def parse_args():
         default=None,
         help='OS (distro) version such as "12.10"',
     )
+    parser.add_argument(
+        '--json-query',
+        default=None,
+        help='JSON frag/file of query parms for --list or --brief',
```
zmc: Could we get an example here? I think this PR is good to go if we have that.

dmick: I had an example in the commit message; I can add it to the code if you want?  Do you mean in the help text?

zmc: I meant in the help text, so users may discover how to use this (pretty cool) feature.

teuthology/test/test_misc.py
```diff
@@ -150,3 +150,15 @@ def test_missing_b_deep_merge(self):
     def test_invalid_b_deep_merge(self):
         with pytest.raises(AssertionError):
             misc.deep_merge({"a": "b"}, "invalid")
+
+
+class TestIsInDict(object):
+    def test_is_in_dict(self):
+        assert misc.is_in_dict('a', 'foo', {'a':'foo', 'b':'bar'})
+        assert misc.is_in_dict(
```
zmc: This ought to be four separate test methods, so that if one fails it's more clear which one failed.

teuthology/test/test_misc.py
```diff
@@ -150,3 +150,15 @@ def test_missing_b_deep_merge(self):
...
+            'a', {'sub1':'key1', 'sub2':'key2'},
+            {'a':{'sub1':'key1', 'sub2':'key2', 'sub3':'key3'}}
+        )
+
+        assert misc.is_in_dict('a', 'foo', {'a':'bar', 'b':'foo'}) == False
+        assert misc.is_in_dict('a', 'foo', {'a':{'a': 'foo'}}) == False
```
zmc: Nitpick, but linters will complain: with booleans, it's more Pythonic to test for ``if foo is False`` rather than use ``==``

andrewschoen: or maybe, ``assert not misc.is_in_dict``

teuthology/test/test_misc.py
dmick: OK, will do

teuthology/test/test_misc.py
dmick: sounds better, yes

dmick:
review commit added.  if it looks good I will squash before merge.

dmick:
@zmc: will squash at will.  LMK.

zmc:
@dmick sure, squash and we're good to go

dmick:
Squashed and rebased.

PR #475, opened 2015-04-24
New-style tasks

zmc:


teuthology/task/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+        if not hosts:
+            raise RuntimeError("All target hosts were excluded!")
+        self.hosts = list(hosts)
+        hostnames = [h.shortname for h in self.hosts]
+        self.log.debug("Restricting task {name} to hosts: {hosts}".format(
+            name=self.name, hosts=' '.join(hostnames))
+        )
+        return self.hosts
+
+    def setup(self):
```
andrewschoen: Could we have a _do_setup method that __enter__ would call instead of this and let this pass like being and end?  This _do_setup could call apply_overrides and find_hosts as well as call setup.  That way any subclasses won't need to call super if they override setup which I imagine every one of them will.

teuthology/task/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+        Can be skipped by setting 'skip_teardown' to True in self.config
+        """
+        pass
+
+    def __enter__(self):
+        """
+        When using an instance of the class as a context manager, this method
+        calls self.setup(), then calls self.begin() and returns self.
+        """
+        self.setup()
```
andrewschoen: as mentioned above. I think that a proxy function of sorts here that calls self.setup() would be better as subclasses will often want to override setup.  I suspect somebody will eventually forget to call super() and we wouldn't apply_overrides or find_hosts.

teuthology/task/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+        return self
+
+    def __exit__(self, type_, value, traceback):
+        """
+        When using an instance of the class as a context manager, this method
+        calls self.end() and self.teardown() - unless
+        self.config['skip_teardown'] is True
+        """
+        self.end()
+        if self.config.get('skip_teardown', False):
```
andrewschoen: so much win

teuthology/test/task/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_hosts_no_results(self):
+        self.ctx.cluster = Cluster()
+        self.ctx.cluster.add(Remote('remote1'), ['role1'])
+        task_config = dict(
+            hosts=['role2'],
+        )
+        with raises(RuntimeError):
+            with Task(self.ctx, task_config) as task:
+                task_hosts = task.hosts
+                assert len(task_hosts) == 0
```
andrewschoen: I'm confused by this assert, is this code even executed?  I'm assuming the with Task raises a RuntimeError.  In that case you should be able to just pass inside the context manager, right?

teuthology/test/task/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+            hostnames = [host.name for host in task_hosts]
+            assert sorted(hostnames) == ['remote1', 'remote3']
+
+    def test_hosts_two_hostnames(self):
+        self.ctx.cluster = Cluster()
+        self.ctx.cluster.add(Remote('remote1.example.com'), ['role1'])
+        self.ctx.cluster.add(Remote('remote2.example.com'), ['role2'])
+        self.ctx.cluster.add(Remote('remote3.example.com'), ['role3'])
+        task_config = dict(
+            hosts=['remote1', 'remote2.example.com'],
```
andrewschoen: It's great that this takes a hostname as well as the teuthology role.

teuthology/test/task/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_setup_called(self, m_setup):
+        task_config = dict()
+        with Task(self.ctx, task_config):
+            m_setup.assert_called_once_with()
+
+    @patch.object(Task, 'begin')
+    def test_begin_called(self, m_begin):
+        task_config = dict()
+        with Task(self.ctx, task_config):
+            m_begin.assert_called_once_with()
```
andrewschoen: I like the format you used below with a pass in the context manager and the assert after.

teuthology/task/__init__.py
zmc: Is there something problematic about calling ``super()``?

teuthology/test/task/__init__.py
zmc: You're right; I'd put those two lines in before I wrapped the block with ``with raises``.

teuthology/test/task/__init__.py
zmc: I want the asserts to be evaluated as soon as possible; this one can be done inside the ``with`` block but the other one cannot.

teuthology/test/task/__init__.py
zmc: Yeah, I'm not sure if that will ever be useful in practice, but it was easy to implement and test.

teuthology/task/__init__.py
andrewschoen: Not really, it's just easy to forget to do.  Why make users of this base class remember to do that when they don't have to for begin, end or teardown?

Also, when I think of a setup method I want to compare it to setup methods in pytest or unittest where I don't need to ever call super().

teuthology/test/task/__init__.py
andrewschoen: Might be useful if you're providing your own targets.yml and want to be very specific about the host.

teuthology/task/__init__.py
zmc: Well, these aren't pytest or unittest objects; this is just regular old inheritance. It's typically the right thing to do, to call methods of a class you're inheriting from, unless you have a specific reason not to.

andrewschoen: I just thought we could make usage of this base class easier if we went this direction. If I don't have to, I'd rather not have to call super every time I override setup, which I imagine almost every task will.

teuthology/task/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+
+class Task(object):
+    """
+    A base-class for "new-style" teuthology tasks.
+
+    Can be used as a drop-in replacement for the old-style task functions with
+    @contextmanager decorators.
+    """
+    def __init__(self, ctx=None, config=None):
+        self.log = log
```
andrewschoen: Does setting this here mean that log messages coming from subclasses will look like they originated in this file?

zmc: Nope!

PR #476, opened 2015-04-27
Log more output in need_to_install_distro()

zmc:


teuthology/task/kernel.py
```diff
@@ -597,31 +597,34 @@ def wait_for_reboot(ctx, need_install, timeout, distro=False):
...
+    role_remote.run(args=['uname', '-r'], stdout=output)
     current = output.getvalue().strip()
     if system_type == 'rpm':
-        role_remote.run(args=['sudo', 'yum', 'install', '-y', 'kernel'], stdout=output, stderr=err_mess )
+        role_remote.run(args=['sudo', 'yum', 'install', '-y', 'kernel'],
+                        stdout=output)
         if 'Nothing to do' in output.getvalue():
-            output.truncate(0), err_mess.truncate(0)
-            role_remote.run(args=['echo', 'no', run.Raw('|'), 'sudo', 'yum', 'reinstall', 'kernel', run.Raw('||'), 'true'], stdout=output, stderr=err_mess )
+            err_mess.truncate(0)
```
andrewschoen: Do you need to truncate here?  I don't see you using it after assignment, but I might have missed something.

PR #478, opened 2015-04-29
Don't assert status==up when using internal.check_lock.

andrewschoen:
internal.check_lock seems to only be used by nuke and it's sometimes helpful
to be able to nuke a node that is marked down.

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -497,7 +497,9 @@ def nuke_helper(ctx, should_unlock):
             log.info('console ready on %s' % cname)
 
     if ctx.check_locks:
-        check_lock(ctx, None)
+        # does not check to ensure if the node is 'up'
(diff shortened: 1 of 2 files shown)


zmc:
I don't think this is the right approach; this change would affect teuthology job codepaths and not just nuking

andrewschoen:
It's used in teuthology.run and in nuke.  Nuke shouldn't care if it's down
and with teuthology.run if internal.lock_machines (which was called before)
has locked a down machine we've already got trouble. I believe
internal.lock_machines already makes sure it's up before locking it anyway.

What approach would you take so that we can nuke down machines.  It's a
use-case I need quite a bit when cleaning up busted nodes.

andrewschoen:
Ok, I've updated this so that we only skip that check when nuking.

PR #479, opened 2015-04-30
Make the number machines to be reserved configurable

zhouyuan:

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -18,6 +18,10 @@ Here is a sample configuration with many of the options set and documented::
     # only used by teuthology-suite.
     default_machine_type: awesomebox
 
+    # Control how many machines need to be free in the cluster. 0 means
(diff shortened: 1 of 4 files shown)


andrewschoen:
ok to test

zmc:
I like the idea here. A few small things, though. 

One is that the name isn't specific enough; ``min_to_reserve`` -> min of what? Maybe ``reserve_machines`` or similar would be better.

Another is that, if this were set to a negative number, weird things would happen. There should be some sanity checking, and a value of ``0`` should disable the feature entirely.

Finally, please include some brief documentation in [siteconfig.rst](https://github.com/ceph/teuthology/blob/master/docs/siteconfig.rst).

Thanks!

zhouyuan:
@zmc any comments on the new version?

zmc:
@zhouyuan ah, didn't see the update. This looks great, thanks for making those changes!

PR #480, opened 2015-04-30
Changes for rh installer to work

vasukulkarni:

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -66,7 +66,6 @@ PACKAGES['ceph']['rpm'] = [
     'rbd-fuse',
 ]
 
-
 def _get_config_value_for_remote(ctx, remote, config, key):
     """
     Look through config, and attempt to determine the "best" value to use for a
@@ -927,6 +926,122 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
     )
 
 
+@contextlib.contextmanager
+def rh_install(ctx, config):
+    """
+    Installs rh ceph on all hosts in ctx.
+
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict
+    """
+    version = config['rhbuild']
+    rh_versions = ['1.3.0']
+    if version in rh_versions:
+        log.info("%s is a supported version", version)
+    else:
+        raise RuntimeError("Unsupported RH Ceph version %s", version)
+
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            if remote.os.name == 'rhel':
+                log.info("Installing on RHEL node: %s", remote.shortname)
+                p.spawn(rh_install_pkgs, ctx, remote, version)
+            else:
+                log.info("Node %s is not RHEL", remote.shortname)
+                raise RuntimeError("Test requires RHEL nodes")
+    try:
+        yield
+    finally:
+        if config.get('skip_uninstall'):
+            log.info("Skipping uninstall of Ceph")
+        else:
+            rh_uninstall(ctx=ctx, config=config)
+
+
+def rh_uninstall(ctx, config):
+    """
+     Uninstalls rh ceph on all hosts.
+     It actually spawns rh_uninstall_pkgs() on the remotes for uninstall.
+
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict
+    """
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            p.spawn(rh_uninstall_pkgs, ctx, remote)
+
+
+def rh_install_pkgs(ctx, remote, installed_version):
+    """
+    Installs RH build using ceph-deploy.
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    pkgs = ['ceph-deploy']
+    rh_version_check = {'0.94.1': '1.3.0'}
+    for pkg in pkgs:
+        log.info("Check if ceph-deploy is already installed on node %s", remote.shortname)
+        remote.run(args=['sudo', 'yum', 'clean', 'metadata'])
+        r = remote.run(
+             args=['yum', 'list', 'installed', run.Raw(pkg)],
+             stdout=StringIO(),
+             check_status=False,
+            )
+        if r.stdout.getvalue().find(pkg) == -1:
+            log.info("Installing %s " % pkg)
+            remote.run(args=['sudo', 'yum', 'install', pkg, '-y'])
+        else:
+            log.info("Removing and reinstalling ceph-deploy on %s", remote.shortname)
+            remote.run(args=['sudo', 'yum', 'remove', pkg, '-y'])
+            remote.run(args=['sudo', 'yum', 'install', pkg, '-y'])
+
+    log.info("Check if ceph is already installed on %s", remote.shortname)
+    r = remote.run(
+          args=['yum', 'list', 'installed','ceph'],
+          stdout=StringIO(),
+          check_status=False,
+        )
+    host = r.hostname
+    if r.stdout.getvalue().find('ceph') == -1:
+        log.info("Install ceph using ceph-deploy on %s", remote.shortname)
+        remote.run(args=['sudo', 'ceph-deploy', 'install', run.Raw('--no-adjust-repos'), host])
+        remote.run(args=['sudo', 'yum', 'install', 'ceph-test', '-y'])
+    else:
+        log.info("Removing and reinstalling Ceph on %s", remote.shortname)
+        remote.run(args=['sudo', 'ceph-deploy', 'uninstall', host])
+        remote.run(args=['sudo', 'ceph-deploy', 'purgedata', host])
+        remote.run(args=['sudo', 'ceph-deploy', 'install', host])
+        remote.run(args=['sudo', 'yum', 'remove', 'ceph-test', '-y'])
+        remote.run(args=['sudo', 'yum', 'install', 'ceph-test', '-y'])
+
+    # check package version
+    version = packaging.get_package_version(remote, 'ceph')
+    log.info("Node: {n} Ceph version installed is {v}".format(n=remote.shortname,v=version))
+    if rh_version_check[version] == installed_version:
+        log.info("Installed version matches on %s", remote.shortname)
+    else:
+        raise RuntimeError("Version check failed on node %s", remote.shortname)
+
+
+def rh_uninstall_pkgs(ctx, remote):
+    """
+    Removes Ceph from all RH hosts
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    log.info("uninstalling packages using ceph-deploy on node %s", remote.shortname)
+    r = remote.run(args=['date'], check_status=False)
+    host = r.hostname
+    remote.run(args=['sudo', 'ceph-deploy', 'uninstall', host])
+    time.sleep(4)
+    remote.run(args=['sudo', 'ceph-deploy', 'purgedata', host])
+    log.info("Uninstalling ceph-deploy")
+    remote.run(args=['sudo', 'yum', 'remove', 'ceph-deploy', '-y'], check_status=False)
+    remote.run(args=['sudo', 'yum', 'remove', 'ceph-test', '-y'], check_status=False)
+
+
 def _upgrade_rpm_packages(ctx, config, remote, pkgs):
     """
     Upgrade project's packages on remote RPM-based host
@@ -1248,6 +1363,8 @@ def task(ctx, config):
         project: samba
         branch: foo
         extra_packages: ['samba']
+    - install:
+        rhbuild: 1.2.3
 
     Overrides are project specific:
 
@@ -1256,6 +1373,8 @@ def task(ctx, config):
         ceph:
           sha1: ...
 
+    When passed 'rhbuild' as a key, it will attempt to install an rh ceph build using ceph-deploy
+
     :param ctx: the argparse.Namespace object
     :param config: the config dict
     """
@@ -1263,6 +1382,11 @@ def task(ctx, config):
         config = {}
     assert isinstance(config, dict), \
         "task install only supports a dictionary for configuration"
+
+    rhbuild = None
+    if config.get('rhbuild'):
+        rhbuild = config.get('rhbuild')
+        log.info("Build is %s " % rhbuild)
 
     project, = config.get('project', 'ceph'),
     log.debug('project %s' % project)
@@ -1277,17 +1401,24 @@ def task(ctx, config):
 
     ctx.summary['flavor'] = flavor
 
-    with contextutil.nested(
-        lambda: install(ctx=ctx, config=dict(
-            branch=config.get('branch'),
-            tag=config.get('tag'),
-            sha1=config.get('sha1'),
-            flavor=flavor,
-            extra_packages=config.get('extra_packages', []),
-            extras=config.get('extras', None),
-            wait_for_package=ctx.config.get('wait_for_package', False),
-            project=project,
-        )),
-        lambda: ship_utilities(ctx=ctx, config=None),
-    ):
-        yield
+    if config.get('rhbuild'):
+        with contextutil.nested(
+            lambda: rh_install(ctx=ctx, config=config),
+            lambda: ship_utilities(ctx=ctx, config=None)
+        ):
+            yield
+    else:
+        with contextutil.nested(
+            lambda: install(ctx=ctx, config=dict(
+                branch=config.get('branch'),
+                tag=config.get('tag'),
+                sha1=config.get('sha1'),
+                flavor=flavor,
+                extra_packages=config.get('extra_packages', []),
+                extras=config.get('extras', None),
+                wait_for_package=ctx.config.get('wait_for_package', False),
+                project=project,
+            )),
+            lambda: ship_utilities(ctx=ctx, config=None),
+        ):
+            yield


teuthology/task/install.py
```diff
@@ -520,6 +533,50 @@ def verify_package_version(ctx, config, remote):
...
+
+    TODO: split this into at least two functions.
+
+    :param ctx: the argparse.Namespace object
+    :param rpm: list of packages names to install
+    :param config: the config dict
+    """
+    pkgs=['ceph-deploy']
+    for pkg in pkgs:
+      log.info("Check if ceph-deploy is already installed on system")
```
andrewschoen: We have a function already in teuthology.packaging that should be able to tell you if a package is installed on a remote.  You could probably use that here. 

https://github.com/ceph/teuthology/blob/master/teuthology/packaging.py#L203

teuthology/task/install.py
```diff
@@ -63,6 +63,10 @@
     'python-ceph',
     'rbd-fuse',
 ]
+RHVERSION={}
+RHVERSION['ceph']={
```
andrewschoen: What's the purpose of this?  Are you just trying to validate that the version put in rhbuild is valid?

If so, maybe just a list of version would be better?

teuthology/task/install.py
```diff
@@ -425,6 +429,12 @@ def _update_rpm_package_list_and_install(ctx, remote, rpm, config):
     :param rpm: list of packages names to install
     :param config: the config dict
     """
+    rh_versions=RHVERSION['ceph']
+    if 'rhbuild' in config:
+        if config['rhbuild'] in rh_versions:
```
andrewschoen: We might want to have an else on this condition and log something if rhbuild was not in rh_versions.

teuthology/task/install.py
```diff
@@ -437,12 +447,11 @@ def _update_rpm_package_list_and_install(ctx, remote, rpm, config):
         proj=project, release=RELEASE, dist_release=dist_release)
     rpm_name = "{rpm_nm}.rpm".format(rpm_nm=proj_release)
     base_url = "{start_of_url}/noarch/{rpm_name}".format(
-        start_of_url=start_of_url, rpm_name=rpm_name)
+        start_of_url=start_of_url, rpm_name=rpm_name)
```
andrewschoen: nitpick: please remove the trailing whitespace.

teuthology/task/install.py
```diff
@@ -520,6 +533,50 @@ def verify_package_version(ctx, config, remote):
         )
 
 
+def _install_rh_build(ctx, remote, config):
+    """
+    Installs RH build using ceph-deploy.
+
+    TODO: split this into at least two functions.
```
zmc: d'oh

teuthology/task/install.py
```diff
@@ -520,6 +533,50 @@ def verify_package_version(ctx, config, remote):
...
+      )
+    host=r.hostname
+    if r.stdout.getvalue().find('ceph') == -1:
+        #Key in on Role: for better install
+        log.info("Install ceph using ceph-deploy")
+        remote.run(args = ['sudo', 'ceph-deploy','install',run.Raw('--no-adjust-repos'), host])
+    else:
+        log.info("Remove and Install ceph")
+        remote.run(args = ['sudo', 'ceph-deploy', 'purge', host])
+        remote.run(args = ['sudo', 'ceph-deploy', 'uninstall', host])
```
zmc: Why do you need the above two lines?

teuthology/task/install.py
```diff
@@ -520,6 +533,50 @@ def verify_package_version(ctx, config, remote):
...
+                 run.Raw(pkg)],
+          stdout=StringIO(),
+          check_status=False,
+      )
+      if r.stdout.getvalue().find(pkg) == -1:
+          log.info("Install ceph-deploy")
+          remote.run(args = ['sudo', 'yum', 'install', pkg, '-y'])
+      else:
+          log.info("Remove and Install ceph-deploy")
+          remote.run(args = ['sudo', 'yum', 'remove', pkg, '-y'])
```
zmc: Why is removing necessary?

teuthology/task/install.py
```diff
@@ -746,6 +814,11 @@ def remove_sources(ctx, config):
     :param ctx: the argparse.Namespace object
     :param config: the config dict
     """
+    
+    if config.get('rhbuild'):
+        log.info("Not required for rh")
+        return True
```
zmc: I don't understand the purpose of this block

teuthology/task/install.py
```diff
@@ -644,6 +701,17 @@ def _remove_rpm(ctx, config, remote, rpm):
...
+    
+    if config.get('rhbuild'):
+        log.info("Uninstalling packages using ceph-deploy")
+        r = remote.run(args=['date'],check_status=False)
+        host = r.hostname
+        remote.run(args = ['sudo', 'ceph-deploy', 'purge', host])
+        remote.run(args = ['sudo', 'ceph-deploy', 'uninstall', host])
+        log.info("Uninstalling ceph-deploy")
+        remote.run(args = ['sudo', 'yum', 'remove', 'ceph-deploy', '-y'],check_status=False)
+        return True
```
zmc: Wait, so if 'rhbuild' is in ``config`` you do something completely different?

I'm starting to wonder if this should be a separate task entirely, instead of heavily modifying the install task...

zmc:
This changeset heavily modifies existing codepaths in the install task. I'm wondering what, if anything, it *does* use. In other words, why not create a separate task? Is the answer simply that you want to still be able to use all the existing suites?

teuthology/task/install.py
vasukulkarni: Yes this is to differentiate between RH vs Fedora/Cent and keep others unmodified

teuthology/task/install.py
vasukulkarni: I can add else: with exception , since those versions will not work

teuthology/task/install.py
vasukulkarni: will fix

teuthology/task/install.py
vasukulkarni: yeah,  :) will fix.

teuthology/task/install.py
vasukulkarni: We shouldn't hit this if systems are in good state, but any bad runs its helpful the install works next time we run time.

teuthology/task/install.py
```diff
@@ -65,7 +65,10 @@
     'python-ceph',
     'rbd-fuse',
 ]
-
+RHVERSION={}
+RHVERSION['ceph']={
+    '1.3.0':1    
+}
```
zmc: @andrewschoen asked a while ago:
"What's the purpose of this? Are you just trying to validate that the version put in rhbuild is valid?

If so, maybe just a list of version would be better?"

vasukulkarni: @zmc this is to support for which versions are valid and also will be used in upgrades, list can do, but with dict i can hold additional info related related to each build and make use of it during upgrade.

zmc: I still don't understand; it's not even being used anywhere.

teuthology/task/install.py
```diff
@@ -926,6 +929,105 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    """
+    Installs rh ceph on all hosts in ctx.
+
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict 
+    """
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            distro = teuthology.get_system_type(remote,distro=True)
+            if distro == 'redhatenterpriseserver':
```
zmc: Instead of the above, please use ``remote.os.name`` - it will be 'rhel' for RHEL nodes.

teuthology/task/install.py
```diff
@@ -926,6 +929,105 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+def rh_install(ctx,config):
+    """
+    Installs rh ceph on all hosts in ctx.
+
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict 
+    """
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            distro = teuthology.get_system_type(remote,distro=True)
```
zmc: Please put a space before 'distro=True' here

teuthology/task/install.py
```diff
@@ -926,6 +929,105 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+            if distro == 'redhatenterpriseserver':
+                log.info("Installing on RHEL node")
+                p.spawn(rh_install_pkgs, ctx, remote)
+            else:
+                raise RuntimeError ( "Test requires RHEL nodes")
```
zmc: Please clean up this whitespace

teuthology/task/install.py
```diff
@@ -926,6 +929,105 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    try:
+       yield
+    finally:
+        if config.get('skip_uninstall'):
+           log.info("Skip uninstall of Ceph")
```
zmc: All your logging messages should be e.g. "Skipping uninstall", not "Skip install"

teuthology/task/install.py
vasukulkarni: @zmc 
 I will remove this for now and will add when we actually use it, I think its better to see this in use during upgrade tests.

teuthology/task/install.py
```diff
@@ -926,6 +929,105 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
         ] + ['%s=%s' % (d, version) for d in debs],
     )
 
+@contextlib.contextmanager
+def rh_install(ctx,config):
```
zmc: In any function call or signature, please separate arguments with spaces:
``def rh_install(ctx, config)``

zmc:
Despite all the nitpicks I might have, this is looking far better!

teuthology/task/install.py
vasukulkarni: will use this instead

teuthology/task/install.py
vasukulkarni: Might not need this function call at all, if I use the above remote.os.name

teuthology/task/install.py
vasukulkarni: will fix this

teuthology/task/install.py
vasukulkarni: will fix this

teuthology/task/install.py
```diff
@@ -1264,6 +1361,11 @@ def task(ctx, config):
     assert isinstance(config, dict), \
         "task install only supports a dictionary for configuration"
 
+    rhbuild=None
+    if config.get('rhbuild'):
+        rhbuild=config.get('rhbuild')
+        log.info("Build is %s " % rhbuild)
```
zmc: I can't actually find where ``rhbuild`` is used anywhere. What's the intention here?

teuthology/task/install.py
```diff
@@ -926,6 +925,104 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+        for remote in ctx.cluster.remotes.iterkeys():
+            p.spawn(rh_uninstall_pkgs, ctx, remote)
+    
+
+def rh_install_pkgs(ctx, remote):
+    """
+    Installs RH build using ceph-deploy.
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
```
zmc: At some point in here, shouldn't you check to see that the version installed matches what you expect?

teuthology/task/install.py
```diff
@@ -926,6 +925,104 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict 
+    """
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            if remote.os.name == 'rhel':
+                log.info("Installing on RHEL node")
+                p.spawn(rh_install_pkgs, ctx, remote)
+            else:
+                raise RuntimeError ( "Test requires RHEL nodes")
```
zmc: Please fix this whitespace

teuthology/task/install.py
```diff
@@ -926,6 +925,104 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    finally:
+        if config.get('skip_uninstall'):
+           log.info("Skipping uninstall of Ceph")
+        else:
+           rh_uninstall(ctx=ctx, config=config)
+
+           
+def rh_uninstall(ctx, config):
+    """
+    Uninstalls rh ceph on all hosts.
```
zmc: Let's mention here that it spawns ``rh_uninstall_pkgs()`` on the remotes, to avoid confusion between the two functions

teuthology/task/install.py
```diff
@@ -926,6 +925,104 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+          check_status=False,
+      )
+    host=r.hostname
+    if r.stdout.getvalue().find('ceph') == -1:
+        #Key in on Role: for better install
+        log.info("Install ceph using ceph-deploy")
+        remote.run(args = ['sudo', 'ceph-deploy','install',run.Raw('--no-adjust-repos'), host])
+        remote.run(args = ['sudo', 'yum','install', 'ceph-test', '-y'])
+    else:
+        log.info("Remove and Install ceph")
```
zmc: "Removing and reinstalling" please :)

teuthology/task/install.py
```diff
@@ -926,6 +925,104 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+          args = ['yum', 'list', 'installed',
+                 run.Raw(pkg)],
+          stdout=StringIO(),
+          check_status=False,
+      )
+      if r.stdout.getvalue().find(pkg) == -1:
+          log.info("Installing %s " % pkg )
+          remote.run(args = ['sudo', 'yum', 'install', pkg, '-y'])
+      else:
+          log.info("Remove and Install ceph-deploy")
```
zmc: "Removing and reinstalling" please

teuthology/task/install.py
vasukulkarni: This is the config dict (in yaml) that differentiates between rh testing and upstream. using this value the if statement above is taking different install approach.

teuthology/task/install.py
vasukulkarni: Successful install should be sufficient, since the ansible task will set the version to right repo, It all depends on what repo is being set on the nodes.

teuthology/task/install.py
vasukulkarni: sure

teuthology/task/install.py
vasukulkarni: sure

teuthology/task/install.py
vasukulkarni: sure

teuthology/task/install.py
dmick: assignments should also have whitespace surrounding '='

teuthology/task/install.py
vasukulkarni: fixed

teuthology/task/install.py
vasukulkarni: Added version check using get_version_check

teuthology/task/install.py
vasukulkarni: Both the comments have been addressed

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    """
+    Installs rh ceph on all hosts in ctx.
+
+    :param ctx: the argparse.Namespace object
+    :param config: the config dict
+    """
+    version = config['rhbuild']
+    rh_versions = ['1.3.0']
+    if version in rh_versions:
+        log.info("Supported version")
```
zmc: This isn't a particularly useful message: if you think it's necessary to log this, you might as well include the version in the message

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    else:
+        raise RuntimeError("Unsupported RH Ceph version %s " , version)
```
zmc: You have extra whitespace after the quote

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+
+    with parallel() as p:
+        for remote in ctx.cluster.remotes.iterkeys():
+            if remote.os.name == 'rhel':
+                log.info("Installing on RHEL node")
+                p.spawn(rh_install_pkgs, ctx, remote, version)
+            else:
+                raise RuntimeError("Test requires RHEL nodes")
```
zmc: It might be a good idea there to mention *which* remote caused the failure

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    pkgs = ['ceph-deploy']
+    rh_version_check = {'0.94.1':'1.3.0'}
+    for pkg in pkgs:
+        log.info("Check if ceph-deploy is already installed on system")
+        remote.run(args=['sudo', 'yum', 'clean', 'metadata'])
+        r = remote.run(
+             args=['yum', 'list', 'installed', run.Raw(pkg)],
```
zmc: Why not use ``get_package_version()`` here?

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+             stdout=StringIO(),
+             check_status=False,
+            )
+        if r.stdout.getvalue().find(pkg) == -1:
+            log.info("Installing %s " % pkg)
+            remote.run(args=['sudo', 'yum', 'install', pkg, '-y'])
+        else:
+            log.info("Removing and reinstalling ceph-deploy")
+            remote.run(args=['sudo', 'yum', 'remove', pkg, '-y'])
```
zmc: Curious, why not ``packaging.remove_package()`` and ``packaging.install_package()``?

teuthology/task/install.py
vasukulkarni: I think the log does indicate the last error message and its easier to know anyway

teuthology/task/install.py
vasukulkarni: I can use it here, but its just one liner here instead of a expensive function call

teuthology/task/install.py
vasukulkarni: They are just one liners, dont feel the need for function call here.

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    """
+    Installs RH build using ceph-deploy.
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    pkgs = ['ceph-deploy']
+    rh_version_check = {'0.94.1':'1.3.0'}
+    for pkg in pkgs:
+        log.info("Check if ceph-deploy is already installed on system")
```
zmc: Since this function will be run in parallel on multiple hosts, all the log messages should mention the hostname (``remote.shortname`` works well here), or else they won't be helpful at all

teuthology/task/install.py
```diff
@@ -927,6 +926,121 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+
+
+def rh_uninstall_pkgs(ctx, remote):
+    """
+    Removes Ceph from all RH hosts
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    log.info("Uninstalling packages using ceph-deploy")
```
zmc: As in ``rh_install_pkgs``, these log messages should mention the hostname

vasukulkarni: Fixed

teuthology/task/install.py
vasukulkarni: Fixed

teuthology/task/install.py
vasukulkarni: Fixed

teuthology/task/install.py
vasukulkarni: Added version info as well

teuthology/task/install.py
vasukulkarni: this is fixed

teuthology/task/install.py
```diff
@@ -927,6 +926,122 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    for pkg in pkgs:
+        log.info("Check if ceph-deploy is already installed on node %s", remote.shortname)
+        remote.run(args=['sudo', 'yum', 'clean', 'metadata'])
+        r = remote.run(
+             args=['yum', 'list', 'installed', run.Raw(pkg)],
+             stdout=StringIO(),
+             check_status=False,
+            )
+        if r.stdout.getvalue().find(pkg) == -1:
+            log.info("Installing %s " % pkg)
```
zmc: Any reason for only mentioning the node in one of the logging calls instead of all of the ones in this function like I asked?

teuthology/task/install.py
```diff
@@ -927,6 +926,122 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
...
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    """
+    log.info("uninstalling packages using ceph-deploy on node %s", remote.shortname)
+    r = remote.run(args=['date'], check_status=False)
+    host = r.hostname
+    remote.run(args=['sudo', 'ceph-deploy', 'uninstall', host])
+    time.sleep(4)
+    remote.run(args=['sudo', 'ceph-deploy', 'purgedata', host])
+    log.info("Uninstalling ceph-deploy")
```
zmc: Any reason for only mentioning the node in one of the logging calls instead of all of the ones in this function like I asked?

vasukulkarni: yeah sorry missed in some of the logs, will fix and resubmit.

PR #484, opened 2015-05-05
In add_remote method, first evaluate if the user has passed any value

qakapil:
In add_remote method, first evaluate if the user has passed any
test_user from .teuthology.yaml file. If yes, then pass that test_user
as the second parameter to misc.canonicalize_hostname method else
pass the default test_user which would be ubuntu.

Signed-off-by: Kapil Sharma ksharma@suse.com


qakapil:
user = None should have been user = 'ubuntu'
closing this pull request and will create a new one.

PR #485, opened 2015-05-05
Manage test_user in add_remote method

qakapil:
In add_remote method, first evaluate if the user has passed any
test_user from .teuthology.yaml file. If yes, then pass that test_user
as the second parameter to misc.canonicalize_hostname method else
pass the default test_user which would be ubuntu.

Signed-off-by: Kapil Sharma ksharma@suse.com

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -272,10 +272,13 @@ def add_remotes(ctx, config):
     """
     remotes = []
     machs = []
+    user = teuth_config.test_user
+    if not user:
+        user = 'ubuntu'
     for name in ctx.config['targets'].iterkeys():
         machs.append(name)
     for t, key in ctx.config['targets'].iteritems():
-        t = misc.canonicalize_hostname(t)
+        t = misc.canonicalize_hostname(t, user)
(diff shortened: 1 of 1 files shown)


zmc:
I support this general idea wholeheartedly, but:

1. Without adding support for this new ``test_user`` configurable everywhere else, how useful is this? It seems to me that a lot more work would be required here.
2. When adding a new setting to ``teuthology.yaml`` we need to have a default set in ``teuthology/config.py`` and documentation in ``docs/siteconfig.rst``

teuthology/task/internal.py
```diff
@@ -272,10 +272,13 @@ def add_remotes(ctx, config):
     """
     remotes = []
     machs = []
+    user = teuth_config.test_user
+    if not user:
+        user = 'ubuntu'
```
zmc: If you have a default set properly as I mentioned in my other comment, you won't need this block.

qakapil:
Makes sense. So maybe we can close this PR and I will submit a new PR keeping in mind the points you
have mentioned above ?

PR #487, opened 2015-05-08
Ansible task

zmc:


teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+
+from teuthology.exceptions import CommandFailedError
+from teuthology.repo_utils import fetch_repo
+
+from . import Task
+
+log = logging.getLogger(__name__)
+
+
+class LoggerFile(object):
```
andrewschoen: Why do we need this?  Maybe the docstring should explain that?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        hosts:
+          - client.0
+          - host1
+
+    tasks:
+    - ansible:
+        repo: /path/to/repo
+        inventory: /path/to/inventory
+        playbook: /path/to/playbook.yml
+        tags: my_tags
```
andrewschoen: tags is not documented above.  Does this take a comma separated list of tags?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+
+    """
+    def __init__(self, ctx, config):
+        super(Ansible, self).__init__(ctx, config)
+        self.log = log
+        self.generated_inventory = False
+        self.generated_playbook = False
+
+    def setup(self):
+        super(Ansible, self).setup()
```
andrewschoen: This isn't needed, right?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        self.get_inventory() or self.generate_hosts_file()
+        if not hasattr(self, 'playbook_file'):
+            self.generate_playbook()
+
+    def find_repo(self):
+        """
+        Locate the repo we're using; cloning it from a remote repo if necessary
+        """
+        repo = self.config.get('repo', '.')
+        if repo.startswith('http://') or repo.startswith('https://') or \
```
andrewschoen: nitpick: when using long conditions like this I like to either use a variable to hold the condition or create a small, easily tested function.  I find it helps with readability.

teuthology/task/ansible.py
zmc: Oh, you're right! I'll document it.

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def get_playbook(self):
+        """
+        If necessary, fetch and read the playbook file
+        """
+        playbook = self.config['playbook']
+        if isinstance(playbook, list):
+            # Multiple plays in a list
+            self.playbook = playbook
+        elif isinstance(playbook, str) and (playbook.startswith('http://') or \
```
andrewschoen: nitpick: checking again here to see if something is a url.  Maybe a utility method for these types of checks would be worth it?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        self.generated_inventory = True
+        self.inventory = hosts_file.name
+
+    def generate_playbook(self):
+        """
+        Generate a playbook file to use. This should not be called if we're
+        using an existing file.
+        """
+        for play in self.playbook:
+            play['hosts'] = 'all'
```
andrewschoen: I understand why you do this, but a comment might help the next reader of this task.

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        pb_buffer.seek(0)
+        playbook_file = NamedTemporaryFile(prefix="teuth_ansible_playbook_",
+                                           delete=False)
+        playbook_file.write(pb_buffer.read())
+        playbook_file.flush()
+        self.playbook_file = playbook_file
+        self.generated_playbook = True
+
+    def begin(self):
+        super(Ansible, self).begin()
```
andrewschoen: Not needed.

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        self.execute_playbook()
+
+    def execute_playbook(self, _logfile=None):
```
andrewschoen: I assume _logfile is for test injection purposes.  If so, can we doc that?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        Execute ansible-playbook
+        """
+        environ = os.environ
+        environ['ANSIBLE_SSH_PIPELINING'] = '1'
```
andrewschoen: \o/

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+            timeout=None,
+        )
+        if status != 0:
+            raise CommandFailedError(command, status)
+
+    def _build_args(self):
+        """
+        Assemble the list of args to be executed
+        """
+        fqdns = [r.name.split('@')[-1] for r in self.cluster.remotes.keys()]
```
andrewschoen: Don't you only want the fqdns for what's listed in hosts and not all the remotes that teuthology currently has locked?

teuthology/task/ansible.py
```diff
@@ -0,0 +1,225 @@
...
+        if tags:
+            args.extend(['--tags', tags])
+        return args
+
+    def teardown(self):
+        if self.generated_inventory:
+            os.remove(self.inventory)
+        if self.generated_playbook:
+            os.remove(self.playbook_file.name)
+        super(Ansible, self).teardown()
```
andrewschoen: Not needed.

teuthology/test/task/test_ansible.py
```diff
@@ -0,0 +1,306 @@
...
+            self.playbook_file = 'fake'
+
+        def fake_get_inventory(self):
+            pass
+
+        with patch.multiple(
+            ansible.Ansible,
+            find_repo=DEFAULT,
+            get_playbook=fake_get_playbook,
+            get_inventory=fake_get_inventory,
```
andrewschoen: Could just use Mock() here.

teuthology/test/task/test_ansible.py
```diff
@@ -0,0 +1,306 @@
...
+        def fake_get_playbook(self):
+            self.playbook_file = 'fake'
+
+        def fake_get_inventory(self):
+            pass
+
+        with patch.multiple(
+            ansible.Ansible,
+            find_repo=DEFAULT,
+            get_playbook=fake_get_playbook,
```
andrewschoen: Could use Mock(side_effect=Exception) here.

teuthology/test/task/test_ansible.py
```diff
@@ -0,0 +1,306 @@
...
+            get_inventory=fake_get_inventory,
+            generate_hosts_file=DEFAULT,
+            generate_playbook=fake_generate_playbook,
```
andrewschoen: Could use Mock(side_effect=Exception) here.

teuthology/test/task/test_ansible.py
```diff
@@ -0,0 +1,306 @@
...
+        def fake_generate_playbook(self):
+            raise Exception
+
+        def fake_get_playbook(self):
+            self.playbook_file = 'fake'
+
+        def fake_get_inventory(self):
+            pass
+
+        with patch.multiple(
```
andrewschoen: patch.multiple, very cool!

/me didn't know this was a thing

teuthology/task/ansible.py
zmc: ``self.cluster`` has been filtered already! Thanks, base ``Task``!

teuthology/task/ansible.py
zmc: Explanation added

teuthology/task/ansible.py
zmc: Done

zmc:
Addressed the docs-related comments.

As we discussed, the ``super()`` calls will stay at least for now.

I changed the long ``if`` statement to be more readable, but I don't want to move the content of the ``elif`` as I think that would actually be slightly more confusing.

teuthology/task/ansible.py
zmc: Done. And ``branch``.

teuthology/task/ansible.py
andrewschoen: That doesn't explain the why though.  Why can't ansible use the standard logger?

teuthology/task/ansible.py
andrewschoen: Ah, that's right.  We should probably document that behavior in the docstring of Task at some point.

andrewschoen:
This is looking really good.  Once we clean up the commits and maybe do some more docs clarification I think this is good to go. 

Concerning the calls to ``super``, I still feel pretty strongly that the ``setup``, ``begin``, ``end`` and ``teardown`` methods of ``Task`` should be reserved for use by subclasses of ``Task``, but for right now those are noops so I won't block on this PR as we discuss that.

Nice work!

teuthology/task/ansible.py
```diff
@@ -0,0 +1,239 @@
...
+        self.get_inventory() or self.generate_hosts_file()
+        if not hasattr(self, 'playbook_file'):
+            self.generate_playbook()
+
+    def find_repo(self):
+        """
+        Locate the repo we're using; cloning it from a remote repo if necessary
+        """
+        repo = self.config.get('repo', '.')
+        is_remote = (repo.startswith('http://') or
```
alfredodeza: `startswith` can take a tuple of strings to try, instead of doing `or` multiple times:

    is_remote = repo.startswith(('http', 'git', 'https'))

teuthology/task/ansible.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def get_playbook(self):
+        """
+        If necessary, fetch and read the playbook file
+        """
+        playbook = self.config['playbook']
+        if isinstance(playbook, list):
+            # Multiple plays in a list
+            self.playbook = playbook
+        elif isinstance(playbook, str) and (playbook.startswith('http://') or
```
alfredodeza: if the check 'http' or 'https' happens in more than one place you could just have a helper

teuthology/task/ansible.py
andrewschoen: I didn't know about that!  That would be nice here, +1.

teuthology/task/ansible.py
zmc: Well of course ansible itself doesn't know about python loggers. Normally what one would probably do is use ``subprocess.Popen`` to run the process, tell it to pipe stderr and stdout to a buffer, then call ``readlines()`` on the buffer in a loop and log the output. For whatever reason that doesn't work with ``ansible-playbook``; perhaps it writes directly to the terminal instead of using stdout/stderr.

``pexpect`` handles cases like this, and it takes a ``logfile`` argument, so I built a logger with a ``file``-like interface.

teuthology/task/ansible.py
zmc: Didn't know that @alfredodeza, thanks!

teuthology/task/ansible.py
zmc: Done

PR #488, opened 2015-05-11
Install scratch built kernels from koji

andrewschoen:
This allows us to install kernels from koji that have been scratch built, meaning we must install from a taskID not a buildID.

For reference: http://tracker.ceph.com/issues/11552

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -21,6 +21,8 @@ from ..packaging import (
     get_koji_build_info,
     get_kojiroot_base_url,
     get_koji_package_name,
+    get_koji_task_rpm_info,
+    get_koji_task_result,
 )
 
 log = logging.getLogger(__name__)
@@ -67,7 +69,7 @@ def normalize_config(ctx, config):
     """
     if config is None or \
             len(filter(lambda x: x in ['tag', 'branch', 'sha1', 'kdb',
-                                       'deb', 'rpm', 'koji'],
+                                       'deb', 'rpm', 'koji', 'koji_task'],
                        config.keys())) == len(config.keys()):
         new_config = {}
         if config is None:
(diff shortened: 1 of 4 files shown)


ktdreyer:
`downloadTaskOutput()` didn't work out? I thought that would obviate the need for a static `koji_task_url` ?

teuthology/test/test_packaging.py
```diff
@@ -204,3 +275,7 @@ def test_get_package_version_invalid_version(self):
         remote.run.return_value = proc
         result = packaging.get_package_version(remote, "httpd")
         assert result is None
+
+    @pytest.mark.parametrize("input, expected", KOJI_TASK_RPMS_MATRIX)
```
alfredodeza: this is fantastic, great use of `parametrize` here

andrewschoen:
I thought ``dowloadTaskOutput()`` didn't work on scratch builds as-is?
This is essentially doing what it does, only smarter because we only down
the rpm we need and no logs.

This way is much easier to integrate into how teuthology installs kernels
anyway.

ktdreyer:
@andrewschoen ok sounds good

zmc:
This looks really good. I love all the docs here. And tests!

teuthology/task/kernel.py
```diff
@@ -992,6 +1013,11 @@ def task(ctx, config):
         kojihub_url: http://koji.fedoraproject.org/kojihub
         kojiroot_url: http://kojipkgs.fedoraproject.org/packages
 
+    When installing from a koji task_id you also need to set koji_task_url,
+    which is the base url used to download rpms from koji task results::
+
+        koji_task_url: https://kojipkgs.fedoraproject.org/work/
```
zmc: I imagine you looked into this, but: is there really no way to figure this out based on ``kojiroot_url``?

andrewschoen: Yeah, as discussed in irc it is unfortunately necessary.

PR #489, opened 2015-05-12
lock.py: remove dead code trying to deal with locks

dmick:
Without this, lock.py keeps trying to get at the key in
case it might need it later...but there's nothing that even
uses that information left, so it just causes a delay.

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -13,7 +13,6 @@ import teuthology
 from . import misc
 from . import provision
 from .config import config
-from .contextutil import safe_while
 from .lockstatus import get_status
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


dmick:
I don't understand the test failures, but I don't think they're mine

zmc:
I told TravisCI to restart the build.

zmc:
Have you tested this to make sure this isn't some innocuous-looking magic that we actually need?

dmick:
Not super-extensively, but by inspection, I can't see any way status_info or ssh_key are used at all, and I can't see any other side effects.

dmick:
Anything else with this one, @zmc?

zmc:
I think you're right that this codepath is entirely useless. Let's hope we're both right!

PR #490, opened 2015-05-13
tests: set HOME env var

ktdreyer:
Travis CI's docs say that HOME is supposed to be set to `/home/travis` by default, however, it also says "do not depend on this value": http://docs.travis-ci.com/user/ci-environment/

Teuthology depends on the $HOME environment variable being set, and the test suites that depended on this recently failed to build in Travis CI.

Set the var explicitly so it's always available for the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+env: HOME=/home/travis
+
 language: python
 python:
     - 2.7
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist = docs, py27, py27-integration, flake8
 
 [testenv:py27]
+passenv = HOME
 sitepackages=True
 deps=
   -r{toxinidir}/requirements.txt
@@ -14,6 +15,7 @@ deps=
 commands=py.test --cov=teuthology --cov-report=term -v {posargs:teuthology scripts}
 
 [testenv:py27-integration]
+passenv = HOME
 sitepackages=True
 deps=
   -r{toxinidir}/requirements.txt


ktdreyer:
rats, this is still not fixed.

```
teuthology/test/task/__init__.py:12: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
teuthology/config.py:173: in __init__
    read_config(self)
teuthology/misc.py:1006: in read_config
    filename = os.path.join(os.environ['HOME'], '.teuthology.yaml')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = {'PYTHONHASHSEED': '1816339468', 'COV_CORE_DATA_FILE': '.coverage', 'COV_CORE_SOURCE': 'teuthology', 'COV_CORE_CONFIG'...s/.rvm/bin:/home/travis/.rvm/bin:/home/travis/.rvm/bin', 'VIRTUAL_ENV': '/home/travis/build/ceph/teuthology/.tox/py27'}
key = 'HOME'
    def __getitem__(self, key):
        if key in self.data:
            return self.data[key]
        if hasattr(self.__class__, "__missing__"):
            return self.__class__.__missing__(self, key)
>       raise KeyError(key)
E       KeyError: 'HOME'
```

ktdreyer:
This might actually be a tox 2.0 thing: http://tox.readthedocs.org/en/latest/example/basic.html#passing-down-environment-variables

ktdreyer:
:tada:

zmc:
Thanks @ktdreyer!

PR #492, opened 2015-05-13
Wip 11605

athanatos:


zmc:
If I take existing ``master`` and pull down ``test_suite.py``, ``test_disable_extension()`` doesn't fail. :-/

teuthology/suite.py
```diff
@@ -783,23 +783,27 @@ def _build_matrix(path, _isfile=os.path.isfile,
...
-                        mincyclicity,
-                        fn))
+                submat = _build_matrix(
+                    os.path.join(path, fn),
+                    _isfile,
+                    _isdir,
+                    _listdir,
+                    mincyclicity,
+                    fn)
+                if submat is not None:
```
andrewschoen: nitpick: why not simply ``if submat:``?  Do we need to allow other falsey values through this condition?

zmc:
Looks good! Thanks for the tests and splitting commits.

PR #493, opened 2015-05-14
Remove old kernels on RPM-based systems

zmc:
http://tracker.ceph.com/issues/11627

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -192,7 +192,7 @@ def install_firmware(ctx, config):
             log.info('Skipping firmware on distro kernel');
             return
         (role_remote,) = ctx.cluster.only(role).remotes.keys()
-        package_type = teuthology.get_system_type(role_remote)
+        package_type = role_remote.os.package_type
         if package_type == 'rpm':
             role_remote.run(args=[
                 'sudo', 'yum', 'upgrade', '-y', 'linux-firmware',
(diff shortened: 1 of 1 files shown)


ktdreyer:
yum is supposed to automatically remove old kernels and only keep the last three around (see `installonlypkgs` and `installonly_limit` in `yum.conf(5)` ). Are we using yum to install kernels?

zmc:
``installonly_limit`` is set, but I don't see ``installonlypkgs``

ktdreyer:
when `installonlypkgs` is unset, yum uses the default value specified in the yum.conf manpage (`kernel, kernel-bigmem, kernel-enter‐prise,  kernel-smp,  kernel-debug,  kernel-unsupported,  kernel-source, kernel-devel, kernel-PAE, kernel-PAE-debug`)

In other words it should be doing the right thing. I wonder what the kernel task is doing, though.

zmc:
The kernel task is using yum. I'm not sure why kernels are accumulating, but they clearly are...

ktdreyer:
(discussed in IRC, the problem is likely to be at https://github.com/ceph/teuthology/blob/master/teuthology/task/kernel.py#L414-L423)

```
            proc = role_remote.run(
                args=[
                    'sudo',
                    'rpm',
                    '-ivh',
                    '--oldpackage',
                    '--replacefiles',
                    '--replacepkgs',
                    remote_pkg_path(role_remote),
                ])
```

PR #495, opened 2015-05-15
Use kernel_flavor when polling gitbuilders

zmc:
http://tracker.ceph.com/issues/11645

### DIFF
--- a/teuthology/suite.py
+++ b/teuthology/suite.py
@@ -185,6 +185,7 @@ def create_initial_config(suite, suite_branch, ceph_branch, teuthology_branch,
         kernel_dict = dict(kernel=dict(kdb=True, sha1=kernel_hash))
     else:
         kernel_dict = dict()
+    kernel_dict['kernel']['flavor'] = kernel_flavor
 
     # Get the ceph hash
     ceph_hash = get_hash('ceph', ceph_branch, kernel_flavor, machine_type,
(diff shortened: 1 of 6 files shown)


zmc:
This is tested, but we may want to hold off on merging until Monday

andrewschoen:
This all looks pretty sane to me.  Waiting until Monday is fine as well.

zmc:
I *think* I'm happy with this, but I should probably test it via a scheduled run. Not sure this moment how that should best be done.

teuthology/suite.py
```diff
@@ -594,10 +617,29 @@ def schedule_suite(job_config,
     return count
 
 
-def get_package_versions(sha1, os_type, package_versions=None):
+def get_install_task_flavor(job_config):
```
andrewschoen: Tests on this method would be nice.

teuthology/task/install.py
```diff
@@ -1222,6 +1222,31 @@ def ship_utilities(ctx, config):
         )
 
 
+def get_flavor(config):
```
andrewschoen: Tests would also be nice here.

zmc: Added!

teuthology/suite.py
zmc: Added!

zmc:
So this should be ready to go.

I observed an issue where it wasn't doing what I wanted here:
http://qa-proxy.ceph.com/teuthology/zack-2015-05-20_15:39:30-fs:verify-master---basic-vps/902151/teuthology.log

But! That's only happening because of the weird vps-specific distro defaults in ``misc.get_distro_version()``. I want to fix that and probably consolidate it with ``suite.get_distro_defaults()``, but I think that's outside the (already large) scope of this PR.

PR #496, opened 2015-05-15
task/internal: account for locked machines acquired on previous loops

dmick:

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -159,6 +159,10 @@ def lock_machines(ctx, config):
             break
         elif not ctx.block:
             assert 0, 'not enough machines are available'
+        else:
+            how_many = how_many - len(newly_locked)
+            assert how_many > 0, "lock_machines: how_many counter went" \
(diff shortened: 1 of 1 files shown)


andrewschoen:
LGTM

teuthology/task/internal.py
```diff
@@ -159,6 +159,9 @@ def lock_machines(ctx, config):
             break
         elif not ctx.block:
             assert 0, 'not enough machines are available'
+        else:
+            how_many = how_many - len(newly_locked)
+            assert how_many > 0
```
andrewschoen: Is this needed?  If it is, can we add a message onto it explaining why the assertion failed?

zmc: Agreed; no bare asserts!

dmick: ahhh, ya bluenoses.  Fine.  FINE.

dmick: (the only reason the assert is there is that I could think of no good way to properly test this without inventing a lot more mechanism.)

PR #497, opened 2015-05-18
teuthology/ceph.conf.template: set 2 minutes mon scrub period

jecluis:
Meant to trigger periodic mon scrubs. The default is something like once a day or something, but it would be nice to have the runs testing this in the background.

Signed-off-by: Joao Eduardo Luis joao@suse.de

### DIFF
--- a/teuthology/ceph.conf.template
+++ b/teuthology/ceph.conf.template
@@ -48,6 +48,7 @@
 	debug paxos = 20
 	mon data avail warn = 5
 	mon reweight min pgs per osd = 4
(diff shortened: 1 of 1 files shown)


kshtsk:
@jecluis is this patch still actual? Can we close the PR and remove branch if it is not?

jecluis:
> @jecluis is this patch still actual? Can we close the PR and remove branch if it is not?

go ahead. I can't even remember what this was trying to address anymore. 4 years. yikes.

PR #498, opened 2015-05-19
Update kernel dependencies on Ubuntu

zmc:
http://tracker.ceph.com/issues/11684


idryomov:
Looks good to me in principle.  It looks like you are actually adding a stdout suppression rather than removing it - just wondering about the commit title?

zmc:
@idryomov ``Remote.run()`` logs stdout and stderr by default, unless it's passed a buffer, in which case it doesn't.

PR #499, opened 2015-05-22
Enablement for chef -> ansible transition

zmc:
This doesn't move us over yet, but adds all of the (teuthology-side) tweaks we need to make the real transition.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -7,6 +7,7 @@ import yaml
 from cStringIO import StringIO
 from tempfile import NamedTemporaryFile
 
+from teuthology.config import config as teuth_config
 from teuthology.exceptions import CommandFailedError
 from teuthology.repo_utils import fetch_repo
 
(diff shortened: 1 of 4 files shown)


zmc:
test this please

teuthology/task/ansible.py
```diff
@@ -213,9 +224,13 @@ def _build_args(self):
         """
         Assemble the list of args to be executed
         """
-        fqdns = [r.name.split('@')[-1] for r in self.cluster.remotes.keys()]
+        fqdns = [r.hostname for r in self.cluster.remotes.keys()]
+        # Assume all remotes use the same username
+        user = self.cluster.remotes.keys()[0].user
```
andrewschoen: Just a note: we should add the teuthology_user to teuthology.config sometime. We'll eventually need that.

zmc: I actually think we might want to get the user from paddles on a per-host basis. I'm not positive though.

andrewschoen: Yeah, maybe so. Either way, not an issue for this PR :)

PR #500, opened 2015-05-26
Support overrides: in the kernel task

idryomov:


idryomov:
Added Do Not Merge - need to abstract config parsing and possibly add some unit tests.

idryomov:
@jdurgin merged krbd unmap subsuite, which relies on overrides: support, so we need to merge this.

PR #502, opened 2015-05-26
Bump default OS versions

zmc:


zmc:
test this please

zmc:
@liewegas I'd like your input on this. Also @yuriw

liewegas:
:+1:

once we have fedora 21 or 22 we'll want to bump that one up too!

PR #503, opened 2015-05-27
Add --dry-run to teuthology-nuke

zmc:

### DIFF
--- a/scripts/nuke.py
+++ b/scripts/nuke.py
@@ -3,10 +3,11 @@ import docopt
 import teuthology.nuke
 
 doc = """
-usage: teuthology-nuke --help
-       teuthology-nuke [-v] [--owner OWNER] [-n NAME] [-u] [-i] [-r] [-s]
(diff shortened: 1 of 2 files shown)


scripts/nuke.py
```diff
@@ -19,6 +20,7 @@
                         archive path for a job to kill and nuke
   --stale               attempt to find and nuke 'stale' machines
                         (e.g. locked by jobs that are no longer running)
+  --dry-run             Don't actually nuke anything
```
andrewschoen: So what does --dry-run actually do then except make it a noop.  Does it have any output? I guess, what's the use case here?

zmc: It still outputs the list of targets that would be nuked.

andrewschoen: Ok, can we change the help text to mention that then? "Don't actually nuke anything, but output the list of targets to be nuked". Maybe?

zmc: Of course! I'll do that.

PR #504, opened 2015-05-27
ResultsReporter: add get_job()

zmc:

### DIFF
--- a/teuthology/report.py
+++ b/teuthology/report.py
@@ -349,18 +349,21 @@ class ResultsReporter(object):
         if os.path.exists(self.last_run_file):
             os.remove(self.last_run_file)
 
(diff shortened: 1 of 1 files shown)


teuthology/report.py
```diff
@@ -367,6 +367,26 @@ def get_jobs(self, run_name, fields=None):
         response.raise_for_status()
         return response.json()
 
+    def get_job(self, run_name, job_id, fields=None):
```
dmick: This seems so much like get_jobs that I wonder if it should be just a slight specialization of get_jobs(); maybe add a job_id=None parameter, None meaning "get them all"

zmc: Sure, I suppose I can do it that way.

zmc:
test this please

PR #505, opened 2015-05-28
Use requests locally instead of wget remotely

zmc:
... in _update_deb_package_list_and_install()

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -5,6 +5,7 @@ import copy
 import logging
 import time
 import os
+import requests
 import subprocess
 
 from teuthology.config import config as teuth_config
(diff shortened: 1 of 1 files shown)


zmc:
http://pulpito.ceph.com/zack-2015-05-28_09:26:50-teuthology:ceph-master---basic-multi/

ktdreyer:
this is the :bomb:

zmc:
test this please

dmick:
The pulpito jobs appear to still be using wget?...maybe I'm misunderstanding what those are supposed to represent

zmc:
@dmick ah, the first job is using RHEL; I didn't update that codepath. I will.

The second job on Ubuntu isn't using wget for this functionality.

teuthology/task/install.py
```diff
@@ -336,7 +335,8 @@ def _update_deb_package_list_and_install(ctx, remote, debs, config):
     remote.run(args=['sudo', 'apt-get', 'update'], check_status=False)
     remote.run(
         args=[
-            'sudo', 'DEBIAN_FRONTEND=noninteractive', 'apt-get', '-y', '--force-yes',
+            'sudo', 'DEBIAN_FRONTEND=noninteractive', 'apt-get', '-y',
+            '--force-yes',
```
ktdreyer: Looks like a separate change, here?

ktdreyer:
ticket http://tracker.ceph.com/issues/11630 is also related to this.

teuthology/task/install.py
zmc: Just a line-wrapping change to make the linter happier.

zmc:
http://pulpito.ceph.com/zack-2015-05-29_09:48:10-teuthology:ceph-master---basic-multi/

teuthology/task/install.py
ktdreyer: You know, if it was Nit-Picking Friday I would say it needs to be a separate commit :D

zmc: I guess I can just remove that change if you feel strongly about it?

zmc:
test this please

teuthology/task/install.py
dmick: overruled!  there's too damn many PRs already

andrewschoen:
This seems sane to me, are we waiting on anything else before merge on this?

zmc:
I'd been wondering the same thing.

PR #507, opened 2015-05-29
teuthology-lock usability

zmc:
http://tracker.ceph.com/issues/11812
http://tracker.ceph.com/issues/11813


alfredodeza:
Jenkins error was:     EnvironmentError: mysql_config not found

Will get that fixed - but, why do we *really* need it to run tests?

zmc:
@alfredodeza It's because we require MySQL in requirements.txt. Why do we do that? Because of the old teuthology-coverage utility that to my knowledge isn't even used :-/

PR #508, opened 2015-06-01
docker based integration tests

dachary:
If docker is available, running

   tox -e docker-delegate

will do the following:

   * create or re-use an Ubuntu 14.04 docker image with minimal python
     dependencies from docker-integration/ubuntu-14.04.dockerfile

   * run tox -e docker-integration in a container based
     on this image

The docker-integration environment runs docker-integration/setup.sh
which setup and run a paddles server using a clone of the paddles master
branch. If setup.sh succeeds, it returns on success.


dachary:
this needs work, I'll reopen whe it's ready

PR #509, opened 2015-06-01
expose the install._get_baseurl template in .teuthology.yaml

dachary:
Instead of hardcoding the template used by _get_baseurl to build the URL
of the repository where packages are to be found, make it a

   baseurl_template

configuration parameter in the teuthology.yaml file.

http://tracker.ceph.com/issues/11830 Fixes: #11830

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -68,3 +68,41 @@ Here is a sample configuration with many of the options set and documented::
     # How long a scheduled job should be allowed to run, in seconds, before 
     # it is killed by the worker process.
     max_job_time: 259200
+
+    # The template from which the URL of the repository containing packages
+    # is built.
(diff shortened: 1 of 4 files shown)


qakapil:
I think for baseurl_template, you would also want to add a default value in teuthology/config.py

qakapil:
Now the baseurl_template part looks fine. However,In install.py, _update_rpm_package_list_and_install function makes assumptions about the layout of the underlying rpm repository and these should probably be made more dynamic.
I opened an issue for this - http://tracker.ceph.com/issues/11846

zmc:
This looks really good. Has it been tested to make sure it won't break the current behavior?

dachary:
@zmc there are unit tests covering the current behavior

PR #510, opened 2015-06-01
Add extra_vars support for Ansible

zmc:

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -60,6 +60,8 @@ class Ansible(Task):
                     not, we generate a temporary file to use.
         tags:       A string including any (comma-separated) tags to be passed
                     directly to ansible-playbook.
+        vars:       A dict of vars to be passed to ansible-playbook via the
+                    --extra-vars flag
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


zmc:
http://tracker.ceph.com/issues/11614

teuthology/task/ansible.py
```diff
@@ -80,6 +82,9 @@ class Ansible(Task):
         inventory: /path/to/inventory
         playbook: /path/to/playbook.yml
         tags: my_tags
+        vars:
+            key1: value1
+            key2: value2
```
andrewschoen: Could you add an example of a list and maybe a hash here as well?

```
vars:
    var1: string_value
    var2:
        - list_item
    var3:
        key: value
```

zmc: Will do

teuthology/task/ansible.py
```diff
@@ -60,6 +60,8 @@ class Ansible(Task):
                     not, we generate a temporary file to use.
         tags:       A string including any (comma-separated) tags to be passed
                     directly to ansible-playbook.
+        vars:       A dict of vars to be passed to ansible-playbook via the
+                    --extra-vars flag
```
andrewschoen: Might be worth mentioning the possible values for a var are a yaml string, list or hash.

teuthology/task/ansible.py
andrewschoen: Awesome, thanks.

PR #511, opened 2015-06-01
lock.py: allow selection of machines to list by os-type/os-version

dmick:
Also, clean up the status selection a tiny bit with winnow()

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -151,6 +151,32 @@ def json_matching_statuses(json_file_or_str, statuses):
     return return_statuses
 
 
+def winnow(statuses, arg, status_key, func=None):
+    """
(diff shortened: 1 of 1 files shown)


liewegas:
works for me (to list only rhel 7.1 machines)!

teuthology/lock.py
```diff
@@ -151,6 +151,18 @@ def json_matching_statuses(json_file_or_str, statuses):
     return return_statuses
 
 
+def winnow(statuses, arg, status_key, func=None):
```
zmc: A docstring would be really nice here!

dmick: yeah...

dmick: forcepushed

zmc: thanks!

PR #512, opened 2015-06-02
do not canonicalize hostnames if canonicalize_hostname is false

dachary:
This is useful when running teuthology in a lab that does not enforce
the same naming conventions as the Ceph community lab.

http://tracker.ceph.com/issues/11844 Fixes: #11844

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -10,6 +10,12 @@ Here is a sample configuration with many of the options set and documented::
     # lab_domain: the domain name to append to all short hostnames
     lab_domain: example.com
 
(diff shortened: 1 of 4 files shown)


zmc:
Why is this necessary? I can see it breaking *lots* of things.

dachary:
Nothing works outside of sepia / typica if it's not possible to disable it.

zmc:
That's the thing: that's just not true. We have an entirely separate internal lab that inspired my writing [this](http://ceph.com/teuthology/docs/LAB_SETUP.html). Does ``lab_domain`` not help you?

dachary:
I see what you mean, my bad.

PR #513, opened 2015-06-02
Keep better count of locked nodes

zmc:
http://tracker.ceph.com/issues/11808

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -85,6 +85,7 @@ def lock_machines(ctx, config):
     # change the status during the locking process
     report.try_push_job_info(ctx.config, dict(status='waiting'))
 
+    all_locked = dict()
     while True:
         # get a candidate list of machines
         machines = lock.list_locks(machine_type=machine_type, up=True,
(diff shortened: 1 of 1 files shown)


teuthology/task/internal.py
```diff
@@ -116,9 +117,10 @@ def lock_machines(ctx, config):
                                       ctx.archive, os_type, os_version, arch)
         if not newly_locked and not isinstance(newly_locked, list):
             raise RuntimeError('Invalid parameters specified')
-        if len(newly_locked) == how_many:
+        all_locked.update(newly_locked)
+        if len(all_locked) == how_many:
             vmlist = []
-            for lmach in newly_locked:
+            for lmach in all_locked:
```
andrewschoen: nitpick: lmach is a terrible variable name.

zmc: I agree, but do you think I should change it here?

andrewschoen: Nah, it's nothing to block on. Just felt the need to point it out.

PR #514, opened 2015-06-02
teuthology-nuke improvements

zmc:


zmc:
It occurs to me that I should check that adding python-rados to the install list doesn't break anything.

zmc:
Looks fine:
http://pulpito.front.sepia.ceph.com/zack-2015-06-03_15:11:00-teuthology:ceph-master---basic-multi/918739

PR #515, opened 2015-06-08
teuthology: stop pip and setuptools from fighting over deps that exis…

GregMeno:
…t in the system site packages

Signed-off-by: Gregory Meno <gmeno@redhat.com>

@zmc I have tested this in a VERY limited case.
I ensured that teuthology --help ran

Please consider testing more.


zmc:
This is great! Thanks for the help here.

I tested manually and it was fine. I scheduled a couple test runs to exercise those codepaths, and they also behaved properly. Here is one:
http://pulpito.front.sepia.ceph.com/zack-2015-06-09_09:43:00-teuthology-master---basic-vps/

PR #519, opened 2015-06-14
support fedora 22

liewegas:


teuthology/provision.py
```diff
@@ -153,9 +161,19 @@ def build_config(self):
...
-        file_out = {'downburst': file_info, 'local-hostname': fqdn}
+        file_out = {
+            'downburst': file_info,
+            'local-hostname': fqdn,
+        }
         yaml.safe_dump(file_out, config_fd)
         self.config_path = config_fd.name
+
+        user_info = {
+            'user': 'ubuntu',
```
zmc: Instead of hardcoding this, could we add an arg to ``__init__`` and default it to 'ubuntu'?

Also, has this been tested with existing downburst images?

zmc:
@liewegas Has this been tested with existing downburst images?

liewegas:
tested against precise, trusty, and centos7 without problems.

PR #523, opened 2015-06-17
scripts/lock.py: fix help for --json-query

dmick:
Fixes: #11885


zmc:
Thanks @dmick!

PR #527, opened 2015-06-17
Let's test with SELinux

zmc:
This creates a new SELinux task that, on RPM-based systems, will do two things:
1. Set the SELinux mode before testing begins (default mode is permissive)
2. Before testing, scrape the audit log for denials
3. After testing, scrape the audit log for *new* denials - and fail the job if there are any
4. Restore the SELinux mode to the one used before testing began.

If run on a kernel which does not support SELinux, the job will fail. This is the reason for blocking.


zmc:
http://tracker.ceph.com/issues/12003 is resolved

PR #530, opened 2015-06-19
misc.py: sh helper to log command and output

dachary:
It is the same as subprocess.check_output, only it log.debug the command
being run as well as its output.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -10,6 +10,7 @@ import logging
 import configobj
 import getpass
 import socket
+import subprocess
 import sys
 import tarfile
 import time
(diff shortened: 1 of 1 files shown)


teuthology/misc.py
```diff
@@ -1257,3 +1258,16 @@ def is_in_dict(searchkey, searchval, d):
...
+
+
+def sh(command):
+    log.debug(command)
+    output = ''
+    try:
+        output = subprocess.check_output(command, stderr=subprocess.STDOUT,
+                                         shell=True)
+    except subprocess.CalledProcessError as e:
+        log.error(command + " error " + str(e.output))
```
zmc: ``log.exception()`` would be useful here; it is the same as ``log.error()`` but adds the traceback for you.

teuthology/misc.py
```diff
@@ -1257,3 +1258,16 @@ def is_in_dict(searchkey, searchval, d):
         return True
     else:
         return searchval == val
+
+
+def sh(command):
```
zmc: Could you please add a docstring here?

dachary:
Thanks for the review :-) I made the suggested changes and repushed.

PR #531, opened 2015-06-19
move ssh_keyscan from lock.py to misc.py

dachary:
Because it may be used by provision.py

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -638,33 +638,6 @@ def update_inventory(node_dict):
     return response.ok
 
 
-def ssh_keyscan(hostnames):
-    """
-    Fetch the SSH public key of one or more hosts
-    """
-    if isinstance(hostnames, basestring):
-        raise TypeError("'hostnames' must be a list")
-    hostnames = [misc.canonicalize_hostname(name, user=None) for name in
-                 hostnames]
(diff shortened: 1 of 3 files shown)


zmc:
I don't think "Because it may be used by provision.py" is a great reason to move this. I really don't want to pile more things into misc, and I think ``ssh_keyscan`` is more appropriate in ``lock`` than anywhere else. Am I missing something, though?

dachary:
@zmc the implementation of OpenStack needs ssh_keyscan. It cannot get it from lock.py because lock.py import provision and I don't think we can have circular references, can we ? Since ssh_keyscan is a general purpose function I would not be shocked to find it in the pile of ssh related tools (I'm not sure we have one beside what's in orchestra) or in a generic toolbox such as misc.

zmc:
@dachary ok, I didn't think that through enough. I can agree with moving ``ssh_keyscan`` but I still think ``get_status`` should go to ``lock``. This may belong in #532, but maybe I should keep the discussion in one place.

It looks like you want to move ``get_status`` so that your new implementation of ``is_vm`` will work.

What if we simply made the machine status a required arg for those things in ``provision`` that need it - as opposed to having them call ``get_status`` when it was missing?

zmc:
Since #532 is closed, we can disregard my previous comment.

PR #532, opened 2015-06-19
move get_status from lockstatus.py to misc.py

dachary:
Because it is generally useful and lockstatus.py only has that.


zmc:
See https://github.com/ceph/teuthology/pull/531#issuecomment-115384746 - I don't agree with moving this.

dachary:
I won't move this and find an alternate route.

PR #533, opened 2015-06-19
DNM: lock.py: is_vpm should test is_vm

dachary:
Waiting on #532

The is_vpm function relies on the machine name instead of the is_vm
field. Replace it with an is_vm function that tests the is_vm field of
the machine status returned by paddles.


zmc:
I like the intent of this PR, but it needs some fixing up :)

dachary:
@zmc what kind of fixing do you have in mind ?

zmc:
@dachary oh - it contains 97043a17ed67d4cb33ac1c0cbb402637f47e45d9

PR #534, opened 2015-06-19
implement teuthology_path similar to suite_path

dachary:
In case we want to reference a local directory instead of reseting from
a clone of teuthology.

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -57,6 +57,9 @@ Here is a sample configuration with many of the options set and documented::
     # Where teuthology and ceph-qa-suite repos should be stored locally
     src_base_path: /home/foo/src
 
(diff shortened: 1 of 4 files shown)


zmc:
Seems like we should have this documented somewhere...

dachary:
@zmc ooops forgot to update docs/siteconfig.rst accordingly. Done and repushed !

dachary:
test this please

zmc:
One difference between this PR and ``suite_path`` is that the latter is implemented as a flag for ``teuthology-suite`` whereas this ``teuthology_path`` is an item in ``teuthology.yaml``. I imagine you have reasons for this, and that's fine. What I'd like to see, though, is the default set in ``teuthology.config`` - would you mind doing that?

dachary:
updated & repushed !

PR #535, opened 2015-06-19
lock.py: not all virtual machines have a dedicated host

dachary:
Gracefully handle the case where a virtual machine has no dedicated
host. It's the case for OpenStack.

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -266,6 +266,9 @@ def main(ctx):
             for s in statuses:
                 if not s.get('is_vm', False):
                     continue
+                # with an OpenStack API, there is no host for a VM
+                if s['vm_host'] is None:
+                    continue
                 vm_host_name = s.get('vm_host', dict())['name']
                 if vm_host_name:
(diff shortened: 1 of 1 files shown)


zmc:
I think this deserves a comment noting it's a codepath for openstack as opposed to downburst, no?

dachary:
@zmc right ! comment added

dachary:
@zmc I think http://jenkins.ceph.com/job/teuthology-pull-requests/1546/console is a false negative ?

dachary:
test this please

teuthology/lock.py
```diff
@@ -265,6 +265,9 @@ def main(ctx):
             for s in statuses:
                 if not s.get('is_vm', False):
                     continue
+                # when provisioning with an API, the host of the VM is unknown
```
zmc: Could you word this more specifically, like "e.g. an OpenStack API" please?

dachary:
I amended the comment to be more specific and re-pushed.

zmc:
Thanks @dachary

PR #538, opened 2015-06-22
Fix InsecurePlatformWarning from urllib3

zmc:
See
https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning


andrewschoen:
The PR itself looks fine, but we should probably figure out how to fix jenkins before we merge this? Looks like libffi-dev needs installed on the slave running these tests.  I can take a look at converting the jenkins job that does this to jjb later.

Thoughts?

PR #539, opened 2015-06-26
Fix cleanup-and-unlock.sh

zhouyuan:
The bin dir does not exist in the source code tree


zmc:
I never even noticed this little script; is there any reason we should keep it around? Does it provide anything ``teuthology-nuke`` doesn't?

zhouyuan:
Hi Zack,

Well I'm having fun with these helper scripts ;-) They're some combination
of teuthology-nuke and teuthology-lock and etc.


Sincerely, Yuan

PR #540, opened 2015-06-29
Better logging when looking for a package version in the install task

andrewschoen:
I think this used to be a wget call, so the url we were requesting to
fetch the package version would be logged by that wget call.  Since
we've changed that call to use requests and lost that logging, this adds
some of that logging back.

I'm hoping to use this to debug: http://tracker.ceph.com/issues/12115

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -217,8 +217,10 @@ def _block_looking_for_package_version(remote, base_url, wait=False):
     :returns: str -- the version e.g. '0.67-240-g67a95b9-1raring'
     :raises: VersionNotFoundError
     """
+    url = "{0}/version".format(base_url)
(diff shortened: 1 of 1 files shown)


andrewschoen:
I'm unsure what's wrong with the jenkins build, but tox passes for me locally.

```
flake8 inst-nodeps: /Users/andrewschoen/dev/teuthology/.tox/dist/teuthology-0.1.0.zip
flake8 runtests: PYTHONHASHSEED='871529086'
flake8 runtests: commands[0] | flake8 --select=F,E9 teuthology scripts
______________________________________________________________________________________ summary ______________________________________________________________________________________
  docs: commands succeeded
  py27: commands succeeded
  py27-integration: commands succeeded
  flake8: commands succeeded
  congratulations :)
```

PR #542, opened 2015-06-29
list_locks must use 1 / 0 for True / False

dachary:
Depending on the backend being used (sqlite or another), paddles won't
interpret locked=false as being the same as locked=0 (see
http://tracker.ceph.com/issues/12183 for more context).

Convert True to 1 and False to 0 to always get the same behavior from
paddles, regardless of the backend.


dachary:
test this please

PR #544, opened 2015-06-29
do not run the kernel task unless there is a flavor

dachary:
The kernel task imposes requirements on virtual machines that can be
bypassed when no specific kernel is required.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -194,7 +194,7 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.vm_setup': None},
         ])
 
-    if 'kernel' in config:
+    if 'kernel' in config and config['kernel'].get('flavor'):
         init_tasks.append({'kernel': config['kernel']})
 
     init_tasks.extend([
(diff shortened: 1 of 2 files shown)


zmc:
This would break distro kernel installation, wouldn't it?

dachary:
@zmc the kernel is already installed, is there a point to re-installing it if it is not required ?

zmc:
In the beginning of each job, the kernel task installs a new kernel if necessary. It does not remove it later; it is up to the next job to ask for what it wants. In other words, if a job requiring an upstream kernel is run, that kernel stays on the machine. If the next job doesn't use the kernel task, it will use whatever the last job put on the machine.

Part of me really dislikes that, but it's the way things work at the moment.

dachary:
Since the OpenStack instances are re-created each time they are needed, this step is not necessary. I guess the "do not run the kernel task unless something non standard is required" thing should be an openstack only thing then, right ?

zmc:
Hrm. I'd really rather not special case one machine type. What all causes problems if the kernel task is run?

dachary:
I don't remember what breaks. The intention is to not always run the kernel task, because it's rarely needed, for all machines. This will reduce the maintenance of this task globally as it will have to be verified to run only when it's actually necessary. If your opinion is that it must run always and has to be debugged, I'll do that.

zmc:
@dachary and I just spoke about this; he's going to see how the OpenStack backend behaves with this commit removed. We can address the kernel task separately pending those results.

PR #545, opened 2015-06-29
task/install: rm {proj}.list is idempotent

dachary:
It does not matter if the {proj}.list file can't be removed because it
does not exist.


dachary:
test this please

PR #546, opened 2015-06-29
add stop_worker: as a mean to shutdown the worker

dachary:
It is convenient for integration tests as a way for the worker to be
instructed to shutdown cleanly.

### DIFF
--- a/teuthology/worker.py
+++ b/teuthology/worker.py
@@ -82,7 +82,8 @@ def main(ctx):
     fetch_teuthology('master')
     fetch_qa_suite('master')
 
-    while True:
+    keep_running = True
+    while keep_running:
         # Check to see if we have a teuthology-results process hanging around
         # and if so, read its return code so that it can exit.
         if result_proc is not None and result_proc.poll() is not None:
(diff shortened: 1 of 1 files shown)


dachary:
test this please

zmc:
This seems like it would be awkward to use. Why not intercept a signal telling the worker to stop after the current job is finished?

dachary:
This is how I'm using it in the context of integration tests for the OpenStack backend: https://github.com/dachary/teuthology/blob/wip-6502-openstack/teuthology/test/integration/noop.yaml . The sequence is (as implemented at https://github.com/dachary/teuthology/blob/wip-6502-openstack/teuthology/test/integration/openstack-integration.py#L76 )

* run the worker as a subprocess, 
* schedule the job including "stop_worker", 
* wait for the worker to finish

The alternative would be to poll the worker to figure out when the job finishes. I'm not sure there is a convenient way to do that ? Another option would be to add an option to the working to tell it to run X jobs at most and stop.

zmc:
@dachary could you rebase this please? I can merge it once that's done.

dachary:
rebased & repushed !

PR #547, opened 2015-06-29
a worker --verbose calls run --verbose

dachary:
When a worker runs with --verbose, it also runs the jobs with --verbose.

### DIFF
--- a/teuthology/test/test_worker.py
+++ b/teuthology/test/test_worker.py
@@ -73,7 +73,7 @@ class TestWorker(object):
         m_p.returncode = 0
         m_popen.return_value = m_p
         m_t_config.results_server = True
-        worker.run_job(config, "teuth/bin/path")
+        worker.run_job(config, "teuth/bin/path", verbose=False)
         m_run_watchdog.assert_called_with(m_p, config)
         expected_args = [
             'teuth/bin/path/teuthology',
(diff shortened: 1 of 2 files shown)


teuthology/worker.py
```diff
@@ -222,7 +222,7 @@ def run_with_watchdog(process, job_config):
         report.try_push_job_info(job_info, dict(status='dead'))
 
 
-def run_job(job_config, teuth_bin_path):
+def run_job(job_config, teuth_bin_path, verbose):
```
zmc: If you don't provide a default value for the verbose arg, this will break:
https://github.com/dachary/teuthology/blob/wip-worker-verbose/teuthology/worker.py#L165

teuthology/test/test_worker.py
```diff
@@ -73,7 +73,7 @@ def test_run_job_with_watchdog(self, m_tempfile, m_safe_dump, m_environ,
         m_p.returncode = 0
         m_popen.return_value = m_p
         m_t_config.results_server = True
-        worker.run_job(config, "teuth/bin/path")
+        worker.run_job(config, "teuth/bin/path", False)
```
zmc: If you end up adding a default value for ``verbose`` to ``run_job()``, I'd encourage either dropping these args or naming them, i.e. ``verbose=False``

dachary:
Thanks for catching this mistake. I introduced the problem while working on fixing flake8 messages. This should be ok now, and I added verbose=False as suggested.

PR #549, opened 2015-06-29
bootstrap: apt/yum install the list of dependencies if missing

andrewschoen:
Adds an 'install' flag to the bootstrap that if given will install any missing packages needed for the bootstrap. This requires the user running ./bootstrap have sudo privileges.


andrewschoen:
test this please

PR #550, opened 2015-06-29
Work around BZ 1233329

zmc:
https://bugzilla.redhat.com/show_bug.cgi?id=1233329
http://tracker.ceph.com/issues/12005


dmick:
LGTM.

PR #556, opened 2015-07-01
Skip SELinux operations on downburst VMs

zmc:


zmc:
http://pulpito.ceph.com/zack-2015-07-01_09:15:11-teuthology-master---basic-vps/

PR #557, opened 2015-07-01
lock.py: is_vpm should test is_vm

dachary:
Testing the name of a machine to figure out if it is a virtual machine
is unreliable. Use the is_vm field of get_status instead.

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -17,7 +17,9 @@ from .lockstatus import get_status
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


zmc:
I had to revert this:
https://github.com/ceph/teuthology/commit/097e11c04a40e013e91a7a64836c98413af6a71a

zmc:
I was mistaken in reverting this, so I reverted the revert:
https://github.com/ceph/teuthology/commit/48e80be73348f5207eefe432f2c8abc02a42366a

dachary:
@zmc :-D

PR #559, opened 2015-07-02
Fix Task, Ansible and CephLab unittests and overrides

zmc:


zmc:
Hrm. Tests passed locally, but Jenkins and TravisCI failed. I'll look into this Monday.

zmc:
test this please

teuthology/task/__init__.py
```diff
@@ -13,9 +13,11 @@ class Task(object):
     Can be used as a drop-in replacement for the old-style task functions with
     @contextmanager decorators.
     """
+
+    name = 'task'
```
andrewschoen: Do all class-based tasks require this 'name' property now?  If so, maybe some docs about this are necessary?

PR #560, opened 2015-07-07
Drop Chef in favor of Ansible

zmc:


teuthology/task/internal.py
```diff
@@ -726,19 +727,20 @@ def vm_setup(ctx, config):
     Look for virtual machines and handle their initialization
     """
     all_tasks = [x.keys()[0] for x in ctx.config['tasks']]
-    need_chef = False
-    if 'chef' in all_tasks or 'kernel' in all_tasks:
-        need_chef = True
-    with parallel() as p:
-        editinfo = os.path.join(os.path.dirname(__file__),'edit_sudoers.sh')
+    need_ansible = False
+    if 'kernel' in all_tasks and 'ansible.cephlab' not in all_tasks:
```
andrewschoen: This condition seems strange to me, if only because it doesn't match the logic we did here for chef.  So, we have to be running the kernel task and not have explicitly asked for ansible.cephlab for it to run here?  Should this be an 'or' instead of 'and'?

zmc: Initially I effectively had it as an 'or'. Then I noticed ``ansible`` was being run twice on all scheduled jobs (because it is inserted by ``teuthology-suite`` just like ``chef`` was). Then I realized we'd been running ``chef`` twice on all scheduled ``vps`` jobs!!

andrewschoen: Ok, gotcha.  Now that you mention it I do remember seeing chef being ran twice on some jobs but never really tracked down that it was vps specific.

andrewschoen:
This is looking good to me, just let me know when it's ready to go.

PR #561, opened 2015-07-07
Ensures that task list is a python list

ivotron:
http://tracker.ceph.com/issues/12226

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -151,7 +151,7 @@ def get_summary(owner, description):
 
 def validate_tasks(config):
     """
-    Ensures that config tasks do not include 'kernel'.
+    Ensures that config tasks is a list and doesn't include 'kernel'.
 
     Returns the original tasks key if found.  If not, returns an
     empty list.
@@ -161,8 +161,11 @@ def validate_tasks(config):
         # return the default value for tasks
         return []
 
+    msg = "Expected list in 'tasks'; instead got: {0}".format(config['tasks'])
+    assert isinstance(config['tasks'], list), msg
+
     for task in config['tasks']:
-        msg = ('kernel installation shouldn be a base-level item, not part ' +
+        msg = ('kernel installation should be a base-level item, not part ' +
                'of the tasks list')
         assert 'kernel' not in task, msg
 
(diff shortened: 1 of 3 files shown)


teuthology/run.py
```diff
@@ -161,9 +161,12 @@ def validate_tasks(config):
         # return the default value for tasks
         return []
 
+    msg = 'Expecting (bullet) list of tasks instead of unordered map'
```
andrewschoen: Maybe print out config['tasks'] here as well?  Maybe: msg = "Expected a list of tasks and instead got: {0}".format(config['tasks'])

andrewschoen: Or even just log.info it somewhere before the assertion.

teuthology/test/test_run.py
```diff
@@ -83,7 +83,7 @@ def test_validate_task_no_tasks(self):
         assert result == []
 
     def test_validate_tasks_valid(self):
-        expected = {"foo": "bar"}
```
andrewschoen: We should have another test where you pass in a dict and assert that an AssertionError is raised.  We have a test below for something similar (test_get_invalid_tasks_invalid), but now instead of making sure targets isn't in the list it's failing on it being a dict and not a list.  Two different failure conditions there.

ivotron: true. Will add that.

teuthology/run.py
```diff
@@ -161,9 +161,12 @@ def validate_tasks(config):
         # return the default value for tasks
         return []
 
+    msg = "Expected a list of tasks; instead got: {0}".format(config['tasks'])
+    assert isinstance(config['tasks'], list), msg
+
     for task in config['tasks']:
-        msg = ('kernel installation shouldn be a base-level item, not part ' +
-               'of the tasks list')
+        msg = ("kernel installation shouldn't be a base-level item, not part " +
```
zmc: I think this should actually read "should be a base-level item"

ivotron: thought it was "shouldn't" . fixed below

zmc:
This looks good to me; maybe squashing some commits (at least the most recent one) would be good

ivotron:
how does squashing work for PRs on github? can I force an update to this
branch? or should I create a new branch?

zmc:
Yeah, you can force push this branch

ivotron:
done. switched back to single quotes

ivotron:
OK. Done with last modifications. Thanks!

PR #563, opened 2015-07-08
Allows other users besides ubuntu

ivotron:
http://tracker.ceph.com/issues/10086

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -38,41 +38,33 @@ is_vm = lambda x: x.startswith('vpm') or x.startswith('ubuntu@vpm')
 is_arm = lambda x: x.startswith('tala') or x.startswith(
     'ubuntu@tala') or x.startswith('saya') or x.startswith('ubuntu@saya')
 
-hostname_expr_templ = '(?P<user>.*@)?(?P<shortname>.*)\.{lab_domain}'
-
 
 def canonicalize_hostname(hostname, user='ubuntu'):
-    hostname_expr = hostname_expr_templ.format(
-        lab_domain=config.lab_domain.replace('.', '\.'))
-    match = re.match(hostname_expr, hostname)
-    if match:
-        match_d = match.groupdict()
-        shortname = match_d['shortname']
-        if user is None:
-            user_ = user
-        else:
-            user_ = match_d.get('user') or user
-    else:
-        shortname = hostname.split('.')[0]
-        user_ = user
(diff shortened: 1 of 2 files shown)


ivotron:
1) You're right, I shouldn't said simpler. It's the most straightforward way I found to implement the new functionality, which is to allow other usernames instead of always assuming "ubuntu".

2) The new tests won't pass if they run against master. For example, given `foo@bar` the `canonicalize_hostname` function returns `ubuntu@bar.front.sepia.ceph.com`

dmick:
OK, fair enough; I see that failure in the original code after reviewing the original in situ.  Will this also work for "ubuntu@plana33.front.sepia.ceph.com" -> <same>?

ivotron:
Depends on whether `lab_domain` is empty. If empty, it will do `ubuntu@plana33.front.sepia.ceph.com -> ubuntu@plana33.front.sepia.ceph.com`. If not (say `lab_domain='ceph.com'`), then this will do `ubuntu@plana33.front.sepia.ceph.com -> ubuntu@plana33.front.sepia.ceph.com.ceph.com`. Is this OK?

dmick:
I see that.  OK.  I think it wouldn't hurt to add that test case, but I'm OK with it.

ivotron:
It's already there. `test_decanonicalize_hostname_nolab()` covers that case

dmick:
Sigh.  I apologize for being so thick about this. 

I think that user@plana33.front.sepia.ceph.com, with lab_domain set to front.sepia.ceph.com, should return user@plana33.front.sepia.ceph.com, and I think the current code will instead return 
user@plana33.front.sepia.ceph.com.front.sepia.ceph.com.  Does that sound right, and can we make the code handle the case of "the lab domain is already supplied and correct, so don't modify it"?

ivotron:
No worries. Sure, I will add that.

zmc:
I had to revert this:
https://github.com/ceph/teuthology/commit/57e73e729fc7a345537f09a99887199574075db9

ivotron:
I was looking at the the [log] and couldn't figure out what was the problem in `decanonicalize_hostname` that caused this. If you have time, can you describe the use case? Are there any others that are not covered by the unit tests?

[log]: http://qa-proxy.ceph.com/teuthology/teuthology-2015-07-10_11:29:23-upgrade:hammer:point-to-point-wip-12021-hammer---basic-vps/968166/teuthology.log

PR #565, opened 2015-07-09
Fixing small typo

ivotron:
Changing from `dsa` to `dss` in error message.

### DIFF
--- a/teuthology/orchestra/connection.py
+++ b/teuthology/orchestra/connection.py
@@ -34,7 +34,7 @@ def create_key(keytype, key):
     elif keytype == 'ssh-dss':
         return paramiko.dsskey.DSSKey(data=base64.decodestring(key))
     else:
-        raise ValueError('keytype must be ssh-rsa or ssh-dsa')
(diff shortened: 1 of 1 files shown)


dmick:
was surprised to see this; found http://security.stackexchange.com/questions/51567/why-are-dsa-keys-referred-to-as-dss-keys-when-used-with-ssh; I actually think ssh is contributing to confusion here.  I'd advise using "or ssh-dss (DSA)"  to avoid confusing people with a similar level of partial knowledge, since the key types are really RSA and DSA, despite the label ssh-dss.

What do you think?

ivotron:
I think either is fine, as long as there's consistency between the message
and the config. I agree that DSA makes more sense but changing it will
break existing yaml config files.

dmick:
?  No, I'm talking about changing the error message from your suggested 'keytype must be ssh-rsa or ssh-dss' to 'keytype must be ssh-rsa or ssh-dss (DSA)'

ivotron:
Sorry Dan, I somehow misread your first message (too tired I guess). I
agree, will change it the way you suggest.

ivotron:
done

PR #566, opened 2015-07-09
Pass a targets var when using the Ansible task

andrewschoen:
Passing targets will allow playbooks to use that var
in the 'hosts' declaration.  This way teuthology
does not need to somehow look at the playbook being used
and parse it to figure out what group to assign hosts to
when creating a dynamic inventory.

In the ansible playbook, you can do something like this to
utilize this targets var:

  - hosts: "{{ targets | default('a_group') }}"


andrewschoen:
corresponding ceph-cm-ansible pr: https://github.com/ceph/ceph-cm-ansible/pull/71

PR #567, opened 2015-07-10
Allows other users besides ubuntu (2nd round)

ivotron:
[1st attempt](https://github.com/ceph/teuthology/pull/563) to address [this](http://tracker.ceph.com/issues/10086) was reverted due to a buggy commit.

I think I managed to capture with a test the use case that observed the buggy behavior. If you have time @zmc and @andrewschoen , could you please point me to other use cases in teuthology or ceph-qa-suite that are not covered by the unit tests?

PR #568, opened 2015-07-11
Pins mock to 1.0.1

ivotron:
[`mock`](https://pypi.python.org/pypi/mock) was updated on 07/10/15 and it seems that the update introduced an issue I'm experiencing in https://github.com/ceph/teuthology/pull/567. Reverting to previous version (`mock==1.0.1`) fixes the issue. In order to avoid this type of issues in the future, I propose we pin dependencies to specific versions we know are working OK.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,7 @@ passenv = HOME
 sitepackages=True
 deps=
   -r{toxinidir}/requirements.txt
-  mock
+  mock==1.0.1
   fudge
   nose
   pytest-cov==1.6
@@ -19,7 +19,7 @@ passenv = HOME
 sitepackages=True
 deps=
   -r{toxinidir}/requirements.txt
-  mock
+  mock==1.0.1
   fudge
   nose
   pytest-cov==1.6


andrewschoen:
+1

ktdreyer:
I think we should only pin to fix *specific* bugs, and each git commit that adds pinning should document the specific bugs as reasons for pinning. Otherwise, maintenance becomes a guessing game - "did we pin to avoid a specific bug? Or did we pin to avoid an unknown unknown?"

andrewschoen:
@ktdreyer That's a decent middle ground. Bugs introduced because of a dependency that upgraded because it wasn't pinned can sometimes be very hard to find. So yes, sometimes it is can be good pinning to avoid an unknown unknown.

zmc:
@andrewschoen Sure, it can be good to pin in order to avoid problems we don't know if we will have. However, I agree with @ktdreyer in that tracking down the reasons for pinning can be very difficult. By contrast, pinning to fix a regression is fairly simple as we can just back down to the last version we were using before a particular problem presented itself.

I think we should pin to avoid specific issues, and document those, to reduce the burden of attempting to bring ourselves up-to-date in the future.

andrewschoen:
That's fine with me.  My only point was that sometimes it's hard to know
that the regression introduced was because of a dependency upgrade.

So, in this case we want one commit that only pins mock with some sort of
information in the commit message mentioning why it was pinned.

zmc:
There's still a lot more in this one commit than pinning mock to 1.0.1, and there's no comment explaining why the pin is necessary.

The actual error doesn't seem to be mentioned in this PR, but I dug it up:
```
teuthology/orchestra/test/test_cluster.py:4: in <module>
    from mock import patch, Mock
.tox/py27/local/lib/python2.7/site-packages/mock/__init__.py:2: in <module>
    import mock.mock as _mock
.tox/py27/local/lib/python2.7/site-packages/mock/mock.py:68: in <module>
    from six import wraps
E   ImportError: cannot import name wraps
```

... could we just upgrade ``six``?

ivotron:
When trying to find the root cause of this issue (last friday) I tried by upgrading six but another issues appear:

```
2015-07-13 13:11:23,998.998 ERROR:teuthology.provision:Tried to destroy vpm999 as user@b but it is locked by user@a
____________________________ TestDownburst.test_destroy_if_vm_wrong_description _____________________________

self = <teuthology.test.test_provision.TestDownburst object at 0x7f6295ca0250>

    def test_destroy_if_vm_wrong_description(self):
        name = self.name
        ctx = self.ctx
        status = self.status
        status['description'] = 'desc_a'

        dbrst = provision.Downburst(name, ctx.os_type, ctx.os_version, status)
        dbrst.destroy = MagicMock(name='destroy')
        dbrst.destroy.return_value = True

        result = provision.destroy_if_vm(ctx, name, description='desc_b',
                                         _downburst=dbrst)
        assert result is False

>       dbrst.destroy.assert_called_once()

teuthology/test/test_provision.py:83:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <MagicMock name='destroy' id='140061396567504'>, name = 'assert_called_once'

    def __getattr__(self, name):
        if name in {'_mock_methods', '_mock_unsafe'}:
            raise AttributeError(name)
        elif self._mock_methods is not None:
            if name not in self._mock_methods or name in _all_magics:
                raise AttributeError("Mock object has no attribute %r" % name)
        elif _is_magic(name):
            raise AttributeError(name)
        if not self._mock_unsafe:
            if name.startswith(('assert', 'assret')):
>               raise AttributeError(name)
E               AttributeError: assert_called_once

virtualenv/local/lib/python2.7/site-packages/mock/mock.py:714: AttributeError
```

The problematic lib is `mock`, which got upgraded on Friday. AFAIK, no other 3rd party lib that teuthology depends on got upgraded from Thursday (when everything worked OK) to Friday. I did include a brief description of the problem and the error message (but not the entire trace) in the commit message. I've updated the commit so it includes the complete trace and removed the synchronization of `setup.py` and `tox.ini` (will create a separate PR for that).

let me know if more info should be added to the commit message

andrewschoen:
LGTM

zmc:
This is looking more and more like something we should have filed as a bug, to determine the appropriate fix.

Things I'm wondering:
1. Where did ``assert_called_once`` go?
2. Where is ``mock``'s ChangeLog? I can't seem to find it.

zmc:
Also @ivotron I did see your commit message, and it's helpful, but I really think we should be putting a *comment* in when we pin dependencies so that in the future we don't have to dig through commit history to find the rationale.

ivotron:
Oh, I see, I thought you meant commit message. Will try to look further and find the root cause within `mock`. Will file an issue on their end if I find it.

ivotron:
So `mock.assert_called_once()` got removed and instead there is `assert mock.called` or `assert mock.call_count == 1` but I'm having a hard time having those pass on the existing code. I propose pinning for now.

PR #569, opened 2015-07-13
task.ansible: add nodes to a 'testnodes' group when creating a dynami…

andrewschoen:
…c inventory

Without this, teuthology can not run the testnodes.yml playbook against
it's locked nodes using a dynamic inventory. That playbook is restricted to
only operate against nodes in the testnodes group.

We might want to make that group name configurable through teuthology.yml. I could do that in
this PR or another if that's wanted.  I'd prefer following up with a separate PR if it's all the same.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -94,6 +94,9 @@ class Ansible(Task):
                 key: value
 
     """
+    # set this in subclasses to provide a group to
+    # assign hosts to for dynamic inventory creation
+    inventory_group = None
 
     def __init__(self, ctx, config):
         super(Ansible, self).__init__(ctx, config)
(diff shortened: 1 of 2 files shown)


andrewschoen:
Test failures here are because of mock.  Proposed fix here, https://github.com/ceph/teuthology/pull/568

zmc:
``Ansible`` shouldn't know anything about our specific playbooks or even repos; that's why I created the ``CephLab`` subclass. Perhaps ``Ansible.generate_hosts_file()`` should look for a config item called e.g. ``inventory_group`` which is only used when generating the hosts file. ``CephLab`` could provide a default value of ``testnodes``.

zmc:
@andrewschoen Ah, ok. The task works as-is without specifying a group name, so maybe the default should be to just not use one.

andrewschoen:
@zmc yeah, that's what I thought as well.  Best to not change default
behavior here.

I'll squash this down if you're cool with it.

zmc:
Sounds good to me!

teuthology/task/ansible.py
```diff
@@ -177,7 +180,11 @@ def generate_hosts_file(self):
         hosts = self.cluster.remotes.keys()
         hostnames = [remote.hostname for remote in hosts]
         hostnames.sort()
-        hosts_str = '\n'.join(hostnames + [''])
+        inventory = []
+        if self.inventory_group:
+            inventory.extend(['[{0}]'.format(self.inventory_group)])
```
zmc: Curious, why use ``list.extend()`` here?

andrewschoen: This might be refactoring cruft. It would be clearer as ``list.append()`` here as it's the first thing going in the list, no?

andrewschoen:
@zmc I think I've got all your feedback incorporated here and everything squashed up.  Let me know if there's anything else.

zmc:
@andrewschoen I just want to see unit tests pass; #571 should enable that.

PR #573, opened 2015-07-17
ansible: do not use colors in the teuthology log

andrewschoen:
I tested this locally and the environment variable does work.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -226,6 +226,7 @@ class Ansible(Task):
         environ = os.environ
         environ['ANSIBLE_SSH_PIPELINING'] = '1'
         environ['ANSIBLE_ROLES_PATH'] = "%s/roles" % self.repo_path
+        environ['ANSIBLE_NOCOLOR'] = '1'
         args = self._build_args()
         command = ' '.join(args)
         log.debug("Running %s", command)


dmick:
OTOH, I downloaded the log and looked at it with "less -r", and it was actually quite helpful having the color codes in the file....I wonder if there's a way to render them in a browser with the color?

zmc:
I actually started working this morning on a way to filter it from the logfile but not from stdout.

I'd love to see colored logfiles in the browser though!

dmick:
http://256.io/escapes.js/

andrewschoen:
This is also this ``force_color`` config option that looks promising.

http://docs.ansible.com/intro_configuration.html#force-color

If I'm understanding this right it would color the stdout in a terminal, but not for the saved log?

dmick:
Looking at the code in color.py, it seems to me that doesn't apply to the version of ansible we're running....but also it seems like it has the opposite effect, to always add color to the output, regardless of "terminal or not"

andrewschoen:
@dmick the default is set to 1 though for force_color.  So, maybe if we set it to 0 it would not force color mode even when running without a TTY.

dmick:
ah, finally found it in our source; yes, it's there, and I see the default:
ANSIBLE_FORCE_COLOR            = get_config(p, DEFAULTS, 'force_color', 'ANSIBLE_FORCE_COLOR', None, boolean=True)

But I assumed there was no problem about how to disable color, just whether we want to or not.

dmick:
Could we just spit out escapes.js for log views from pulpito, and spread the word about less -r?

andrewschoen:
@dmick I think we'd have to make a log viewer thing for pulpito for that to work, but I like the idea.  IIRC, the logs are just served directly from the filesystem now.

dmick:
I was hoping we could do something like wrap the existing file inside a call to escape.js, and ship escape.js and a simple "call its rendering function", whatever it is, so that the page is now JS, but just looks like colored text (because its now html)

dmick:
I think this is better: https://pypi.python.org/pypi/ansi2html  -l -s xterm is the closest to right, although I despise the gray background; I'd hack in FFFFFFF for the AAAAAA were it me.

zmc:
I don't think we should remove the colors entirely.

andrewschoen:
Agreed, ansi2html seems like a good option.  I'll go ahead and close this PR.

PR #574, opened 2015-07-17
Ansible: reconnect to remotes after playbook runs

zmc:
To take advantage of the updated ssh_config shipped by the testnodes
role

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -66,6 +66,10 @@ class Ansible(Task):
                     again during teardown with a 'cleanup' var set to True.
                     This will allow the playbook to clean up after itself,
                     if the playbook supports this feature.
+        reconnect:  If set to True (the default), then reconnect to hosts after
+                    ansible-playbook completes. This is in case the playbook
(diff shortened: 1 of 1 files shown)


zmc:
http://tracker.ceph.com/issues/9908

A run from this branch:
http://pulpito.ceph.com/zack-2015-07-17_16:01:57-teuthology:no-ceph-master---basic-vps/
A run from a branch with this change plus a couple others:
http://pulpito.ceph.com/zack-2015-07-17_13:46:38-upgrade:firefly-x-hammer-distro-basic-vps/

teuthology/task/ansible.py
```diff
@@ -66,6 +66,8 @@ class Ansible(Task):
                     again during teardown with a 'cleanup' var set to True.
                     This will allow the playbook to clean up after itself,
                     if the playbook supports this feature.
+        reconnect:  If set to True (the default), then reconnect to hosts after
```
dmick: we need a mention of why one might want this; specifically something like "The playbook may change targets' sshd config, so we want the new config to take effect."

PR #578, opened 2015-07-21
ansible: decode('utf-8', 'ignore') before logging

dachary:
b834dd6660a77379b2c693a36487352941a8e52d added decode('utf-8') which
breaks when a substring inadevertendly cuts in the middle of a valid
utf-8 sequence. Add the 'ignore' so the offending sequence is discarded
instead of raising an error.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -29,7 +29,7 @@ class LoggerFile(object):
         self.level = level
 
     def write(self, string):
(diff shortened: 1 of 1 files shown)


alfredodeza:
as long as we are OK with ignoring errors, this looks good to me

dachary:
These are rare occurences, only at the end of a string, when it is cut at the wrong place as explained in the commit message, therefore I don't see how it could possibly create a problem. The worst that can happen is that we miss one character from the log message.

PR #580, opened 2015-07-23
Introduces teuthology.packaging.GitbuilderProject

andrewschoen:
teuthology.suite, teuthology.task.install and teuthology.task.internal.check_packages all use different methods to check for packages in gitbuilder.  This provides a standard interface that all places in teuthology should be able to use to interact with packages on gitbuilder.

I'm looking for feedback on the class and the approach.  I've implemented it's usage in teuthology.task.internal.check_packages and that tests out with manual teuthology runs.  I've scheduled a suite, but it hasn't completed yet.

I'm hoping we can agree on  the approach and usage in check_packages in this PR and then submit others to use this in suite and install.

I've also done a bit of refactoring and testing to install here as well.

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -396,7 +396,7 @@ def get_package_version(remote, package):
 def _get_config_value_for_remote(ctx, remote, config, key):
     """
     Look through config, and attempt to determine the "best" value to use
-    for a given key. For example, given:
+    for a given key. For example, given::
 
         config = {
             'all':
@@ -430,6 +430,8 @@ class GitbuilderProject(object):
     def __init__(self, project, job_config, ctx=None, remote=None):
         self.project = project
         self.job_config = job_config
+        #TODO: we could get around the need for ctx by using a list
+        # of roles instead, ctx is only used in _get_config_value_for_remote.
         self.ctx = ctx
         self.remote = remote
         # avoiding circular imports
@@ -475,12 +477,28 @@ class GitbuilderProject(object):
 
     @property
     def version(self):
+        """
+        Performs a call to gitubilder to retrieve the version number for the
+        project. The returned value is cached so that this call only happens
+        once.
+
+        :returns: The version number of the project as a string.
+        """
         if not hasattr(self, '_version'):
             self._version = self._get_package_version()
         return self._version
 
     @property
     def base_url(self):
+        """
+        The base url that points at this project on gitbuilder.
+
+        For example::
+
+            http://gitbuilder.ceph.com/ceph-deb-raring-x86_64-basic/ref/master
+
+        :returns: A string of the base url for this project
+        """
         return self._get_base_url()
 
     @property
@@ -541,6 +559,16 @@ class GitbuilderProject(object):
         )
 
     def _get_codename(self, distro, version):
+        """
+        Attempts to find the codename for a given distro / version
+        pair.  Will first attempt to find the codename for the full
+        version and if not found will look again using only the major
+        version.  If a codename is not found, None is returned.
+
+        The constant DISTRO_CODENAME_MAP is used to provide this mapping.
+
+        :returns: The codename as string or None if not found.
+        """
         major_version = version.split(".")[0]
         distro_codes = DISTRO_CODENAME_MAP.get(distro)
         if not distro_codes:
@@ -552,6 +580,14 @@ class GitbuilderProject(object):
         return codename
 
     def _get_version(self):
+        """
+        Attempts to find the distro version from the job_config.
+
+        If not found, it will return the default version for
+        the distro found in job_config.
+
+        :returns: A string distro version
+        """
         version = self.job_config.get("os_version")
         if not version:
             version = DEFAULT_OS_VERSION.get(self.os_type)
@@ -560,7 +596,17 @@ class GitbuilderProject(object):
 
     def _get_uri(self):
         """
-        Set the uri -- common code used by both install and debian upgrade
+        Returns the URI that identifies what build of the project we'd like
+        to use.
+
+        If a remote is given, it will attempt to read the config for the given
+        remote to find either a tag, branch or sha1 defined. If there is no
+        remote, the sha1 from the config will be used.
+
+        If a tag, branch or sha1 can't be found it will default to use the
+        build from the master branch.
+
+        :returns: A string URI. Ex: ref/master
         """
         tag = branch = sha1 = None
         if self.remote:
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -502,8 +502,20 @@ class GitbuilderProject(object):
         return self._get_base_url()
 
     @property
-    def uri(self):
-        return self._get_uri()
+    def uri_reference(self):
+        """
+        The URI reference that identifies what build of the project
+        we'd like to use.
+
+        For example, the following could be returned::
+
+            ref/<branch>
+            sha1/<sha1>
+            ref/<tag>
+
+        :returns: The uri_reference as a string.
+        """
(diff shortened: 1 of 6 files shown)


andrewschoen:
The commits are a bit of a mess, lots of refactoring happened during the course of this ticket.  I'll clean those up, of course.

teuthology/packaging.py
```diff
@@ -369,3 +381,229 @@ def get_package_version(remote, package):
         )
 
     return installed_ver
+
+
+def _get_config_value_for_remote(ctx, remote, config, key):
+    """
+    Look through config, and attempt to determine the "best" value to use
+    for a given key. For example, given:
```
alfredodeza: if going for proper RST you would need `given::` here

teuthology/packaging.py
```diff
@@ -369,3 +381,229 @@ def get_package_version(remote, package):
...
+            # FIXME: 'version' as retreived from the repo is actually the
+            # RPM version PLUS *part* of the release. Example:
+            # Right now, ceph master is given the following version in the
+            # repo file: v0.67-rc3.164.gd5aa3a9 - whereas in reality the RPM
+            # version is 0.61.7 and the release is 37.g1243c97.el6 (centos6).
+            # Point being, I have to mangle a little here.
+            if version[0] == 'v':
+                version = version[1:]
+            if '-' in version:
+                version = version.split('-')[0]
```
alfredodeza: I don't think you need the if condition here, because the split will give you one item if `'-'` is not there so it is safe to always split for it.

Not that this needs to change though, checking for it is OK too.

zmc:
I like this approach - it's actually really similar to something I've sort-of daydreamed about but never got around to. I can't help but think it would be worth splitting the generic parts out into a parent class, though, in case we want to create e.g. a ``PuddleProject`` using the same interface. Thoughts?

teuthology/packaging.py
```diff
@@ -26,6 +27,17 @@
     'httpd': {'deb': 'apache2', 'rpm': 'httpd'}
 }
 
+DISTRO_CODENAME_MAP = {
+    "ubuntu": {
+        "14": "trusty",
+        "12": "precise",
+        "15": "vivid",
```
dmick: well.....there are two releases per year; 12.04 is precise, 12.10 is quantal, 13.04 is raring, 13.10 saucy, etc.  Within our limited sphere the major-only compare for the only minor we care about works, but it's confusing; I'd use the full version, even if you leave holes.

teuthology/packaging.py
```diff
@@ -369,3 +381,231 @@ def get_package_version(remote, package):
...
+    for a given key. For example, given:
+
+        config = {
+            'all':
+                {'branch': 'master'},
+            'branch': 'next'
+        }
+        _get_config_value_for_remote(ctx, remote, config, 'branch')
+
+    would return 'master'.
```
dmick: I guess I don't understand the rule for "best" here.  I woudl have naively thought that 'branch' was better, because it was higher-level, and despite the name of the key being 'all', it's a subset of config.

teuthology/packaging.py
```diff
@@ -369,3 +381,231 @@ def get_package_version(remote, package):
...
+        return self._get_uri()
+
+    def _parse_version(self, version):
+        """
+        Parses a distro version string and returns a modified string
+        that matches the format needed for the gitbuilder url.
+
+        Minor version numbers are ignored if they end in a zero. If they do
+        not end in a zero the minor version number is included with a dash as
+        the separtor instead of a period.
```
dmick: separator

teuthology/task/install.py
```diff
@@ -114,75 +114,72 @@ def _get_uri(tag, branch, sha1):
...
+
+
+def _get_gitbuilder_version(version):
+    """
+    Parses a distro version string and returns a modified string
+    that matches the format needed for the gitbuilder url.
+
+    Minor version numbers are ignored if they end in a zero. If they do
+    not end in a zero the minor version number is included with a dash as
+    the separtor instead of a period.
```
dmick: same comment?  hang on..

teuthology/packaging.py
andrewschoen: That makes sense.

teuthology/packaging.py
andrewschoen: This code was ported from teuthology.task.install and had had this logic for quite some time. I can't say that I truly understand the logic behind this either, but I'm hesitant to change it.

teuthology/task/install.py
andrewschoen: Yeah, I originally wrote this function here in install and then ported it over to GitbuilderProject. :)

ktdreyer:
Yay to the overall idea here - thanks for implementing this

zmc:
I do still really think that this should have its interface abstracted before merging, rather than refactoring this refactor later... unless we know we'll never need to do it. Maybe we won't?

andrewschoen:
@zmc I disagree. I don't think we should abstract an interface without having an already existing use case for it. It just seems like a premature optimization at this point and what if we "imagine" the wrong interface.  You have more context than me here though. What do you see this interface looking like?

zmc:
@andrewschoen I care less now, after thinking more realistically about the likelihood that we'd want to reuse this code for something like Puddle. We can cross that bridge when we reach it.

teuthology/packaging.py
```diff
@@ -369,3 +391,243 @@ def get_package_version(remote, package):
...
+    def version(self):
+        if not hasattr(self, '_version'):
+            self._version = self._get_package_version()
+        return self._version
+
+    @property
+    def base_url(self):
+        return self._get_base_url()
+
+    @property
```
zmc: I think it would be worth adding brief docs for these properties

teuthology/packaging.py
andrewschoen: Yup, I'll make this change.

teuthology/packaging.py
andrewschoen: Agreed, I'll do that.

andrewschoen:
@zmc yeah, we can always rework later if necessary.  Let me get those docs added and I'd like to try removing the need for ``ctx`` as well.  I think that either passing a list of roles or ``ctx.cluster`` there would be sufficient.

zmc:
@idryomov revived #500 recently, and pointed out that we're missing the fact that the kernel task talks to gitbuilders as well.

Looks like we have a bit more to do here after all...

https://github.com/ceph/teuthology/blob/master/teuthology/misc.py#L197

andrewschoen:
@zmc yeah, the install task isn't included in this patch either.

I was hoping to only change internal.check_packages with this PR and follow
up with the others.  Do you think this PR should find all the places we
talk to gitbuilder and replace it with this?

teuthology/task/install.py
```diff
@@ -114,75 +114,78 @@ def _get_uri(tag, branch, sha1):
...
+    return result
+
+
+def _get_gitbuilder_version(version):
+    """
+    Parses a distro version string and returns a modified string
+    that matches the format needed for the gitbuilder url.
+
+    Minor version numbers are ignored if they end in a zero. If they do
+    not end in a zero the minor version number is included with a dash as
```
zmc: You say 'dash' but then you use an underscore!

zmc:
So good. I just noticed one tiny docs issue but I think we're good after that.

teuthology/task/install.py
andrewschoen: Ah, yeah.  Originally I had used a dash, but during testing I figured out
that was wrong.  These docs just never got updated.  I'll do that now.

PR #581, opened 2015-07-24
transparent OpenStack provisioning for teuthology-suite

dachary:
http://tracker.ceph.com/issues/6502


README.rst
```diff
@@ -317,6 +317,106 @@ specified in ``$HOME/.teuthology.yaml``::
 
     test_path: <directory>
 
+OpenStack backend
+=================
+
+The ``teuthology-openstack`` command is a wrapper around
+``teuthology-suite`` that transparently creates the teuthology cluster
+using OpenStack virtual machine.
```
dmick: "machines"

README.rst
```diff
@@ -317,6 +317,106 @@ specified in ``$HOME/.teuthology.yaml``::
...
+
+Prerequistes
```
dmick: "Prerequisites"

README.rst
```diff
@@ -317,6 +317,106 @@ specified in ``$HOME/.teuthology.yaml``::
...
+------------
+
+An OpenStack tenant with access to the nova and cinder API (for
+instance http://entercloudsuite.com/). If the cinder API is not
+available (for instance https://www.ovh.com/fr/cloud/), some jobs
+won't run because they expect volumes attached to each instances.
```
dmick: instance

dachary:
@dmick fixed the typos, thanks :-)

README.rst
```diff
@@ -317,6 +317,106 @@ specified in ``$HOME/.teuthology.yaml``::
...
+
+* `create an account <https://www.ovh.com/fr/support/new_nic.xml>`_
+* get $HOME/openrc.sh from `the horizon dashboard <https://horizon.cloud.ovh.net/project/access_and_security/?tab=access_security_tabs__api_access_tab>`_
+
+Setup
+-----
+
+* Get and configure teuthology::
+
+    $ git clone -b wip-6502-openstack-v2 http://github.com/dachary/teuthology
```
andrewschoen: We don't want to keep this branch here do we?  These docs should read like they've already been merged.

openstack-user-data.txt
```diff
@@ -0,0 +1,16 @@
...
+bootcmd:
+  - touch /tmp/init.out
+system_info:
+  default_user:
+    name: ubuntu
+packages:
+ - python-virtualenv
+ - git
+runcmd:
+ - su - -c '(set -x ; git clone -b wip-6502-openstack-v2 http://github.com/dachary/teuthology && cd teuthology && ./bootstrap install)' ubuntu >> /tmp/init.out 2>&1
```
andrewschoen: another instance of wip-6502 and your github repo

scripts/openstack.py
```diff
@@ -0,0 +1,156 @@
+import argparse
```
andrewschoen: We had started to prefer docopt over argparse.  Is there a reason to not use docopt here?

dachary: The inability of docopt to issue a useful error message when the wrong option is being used makes it extremely painful to use, on a daily basis. (yes, today, once more I faced a cryptic usage display because I repeated an option twice by mistake...).

README.rst
dachary: if that's ok with you, I'll make this change last. The reason is that wip-6502-openstack-v2 is currently been used by a few people. Maybe I shouldn't have done that but these brave early adopters were instrumental in getting this patch to work and figure out the proper interface. The alternative would have been to maintain another branch for that purpose but ... I did not forsee this patch would take more than a week to complete and got carried away :-)

setup.py
```diff
@@ -40,7 +40,7 @@
                       'boto >= 2.0b4',
                       'bunch >= 1.0.0',
                       'configobj',
-                      'six',
+                      'six >= 1.9', # python-openstackclient won't work properly with less
```
andrewschoen: This was recently changed in 9d59d6bd93de47664c939fea331ced6593e8ae16

I believe because of an issue with mock. Would this change reintroduce that?

dachary: I don't know and I could not find a test that fails because of it. Is there one ?

README.rst
```diff
@@ -317,6 +317,106 @@ specified in ``$HOME/.teuthology.yaml``::
...
+* follow the `OpenStack API Quick Start <http://docs.openstack.org/api/quick-start/content/index.html#cli-intro>`_
+* source $HOME/openrc.sh
+* verify the OpenStack client works::
+
+    $ openstack server list
+    +----+------------+--------+------------+-------------+-------------------------+
+    | ID | Name       | Status | Task State | Power State | Networks                |
+    +----+------------+--------+------------+-------------+-------------------------+
+    +----+------------+--------+------------+-------------+-------------------------+
+* upload your ssh public key with::
```
zmc: This just uploads your *existing* key, right?

openstack-user-data.txt
```diff
@@ -0,0 +1,16 @@
...
+ - echo 'export OPENRC' | tee /home/ubuntu/teuthology/teuthology/test/integration/openrc.sh
```
zmc: Can this be a ``$HOME`` instead of ``/home/ubuntu`` ?

openstack-user-data.txt
zmc: So, what if we want to use a different teuthology branch? Will that be possible somehow?

scripts/lock.py
```diff
@@ -149,6 +149,11 @@ def parse_args(argv):
         help='machines to operate on',
     )
     parser.add_argument(
+        '--openstack-cluster',
+        default=None,
+        help='OpenStack cluster as defined in .teuthology.yaml',
```
zmc: What sort of value will this be? A name as a string? Something else?

scripts/openstack.py
zmc: @dachary has some pretty good arguments (ha) against docopt and in favor of argparse. As much as I prefer docopt as a developer I can't argue against him using argparse here.

scripts/openstack.py
```diff
@@ -0,0 +1,156 @@
...
+        '--integration-tests',
+        action='store_true', default=None,
+        help='run integration tests when creating the cluster',
+    )
+    parser.add_argument(
+        '--teardown',
+        action='store_true', default=None,
+        help='destroy the cluster, if it exists',
+    )
+    # copy/pasted from scripts/suite.py
```
zmc: I'm concerned that when we make changes to ``teuthology-suite`` we'll have to duplicate effort here. I wonder what we can do to avoid that.

setup.py
zmc: I think we're fine requiring a newer six here; if this was a ``<=`` I'd be opposed.

teuthology/lock.py
```diff
@@ -385,6 +405,11 @@ def lock_many(ctx, num, machine_type, user=None, description=None,
     machine_types_list = misc.get_multi_machine_types(machine_type)
     if machine_types_list == ['vps']:
         machine_types = machine_types_list
+    elif machine_types_list == ['openstack']:
```
zmc: What if ``machine_types_list`` is e.g. ``['plana', 'mira', 'openstack']`` ?

teuthology/openstack.py
```diff
@@ -0,0 +1,375 @@
...
+import tempfile
+import teuthology
+
+from .contextutil import safe_while
+from .config import config as teuth_config
+from teuthology import misc
+
+log = logging.getLogger(__name__)
+
+class OpenStack(object):
```
zmc: I would really like to see docstrings for methods of this class explaining what the args are supposed to be

teuthology/openstack.py
```diff
@@ -0,0 +1,375 @@
...
+                            str(providers))
+
+
+    def image(self):
+        if self.provider == 'ovh':
+            return 'Ubuntu 14.04'
+        elif self.provider == 'redhat':
+            return 'ubuntu-server-14.04-x86_64'
+        elif self.provider == 'entercloudsuite':
+            return 'GNU/Linux Ubuntu Server 14.04 Trusty Tahr x64'
```
zmc: What's the rationale for hardcoding all these distros? And, why specifically Ubuntu Trusty?

teuthology/openstack.py
```diff
@@ -0,0 +1,375 @@
...
+        # and 172.16.0.0/12 and 10.0.0.0/8 and only allow 80/8081/22
+        # for the rest.
+        misc.sh("""
+openstack security group create teuthology
+openstack security group rule create --dst-port 1:10000 teuthology
+#openstack security group rule create --dst-port 8081 teuthology # pulpito
+#openstack security group rule create --dst-port 443 teuthology # https
+#openstack security group rule create --dst-port 115 teuthology # sftp
+#openstack security group rule create --dst-port 80 teuthology # http
+#openstack security group rule create --dst-port 22 teuthology # ssh
```
zmc: Do you intend to uncomment these, or remove them - or, if you want to leave them present but commented, why?

teuthology/test/integration/openstack-integration.py
```diff
@@ -0,0 +1,275 @@
...
+import teuthology.schedule
+import teuthology.suite
+import scripts.schedule
+import scripts.lock
+import scripts.suite
+from teuthology import provision
+from teuthology.config import config
+
+
+class Integration(testtools.TestCase):
```
zmc: We're pretty much standardized on ``py.test``; is there some reason to use ``testtools`` here instead?

teuthology/test/integration/setup-openstack.sh
```diff
@@ -0,0 +1,572 @@
...
+    local network=$1
+    local subnet=$2
+    local private_key=$3
+    local nameserver=$4
+    local labdomain=$5
+    local ip=$6
+
+    local where=$(dirname $0)
+
+    cat > ~/.teuthology.yaml <<EOF
```
zmc: Uhoh, this will just nuke my ``~/.teuthology.yaml`` ?!

zmc:
Would it be possible to put ``teuthology/openstack-*.txt`` in a subdir? I wonder if it might be good to actually make ``teuthology.openstack`` a subpackage rather than just a module.

README.rst
dachary: yes

openstack-user-data.txt
dachary: It's hardcoded on purpose, there is zero portability on that part and that's intentional.

openstack-user-data.txt
dachary: we could change that to be configurable, yes

scripts/lock.py
dachary: I'll get rid of that parameter, it's too complicated.

scripts/openstack.py
dachary: We could if both shared the same argument parsing method. However the variance is very low and updating options in two places when one changes every three month is probably not a big burden.

teuthology/lock.py
dachary: I'm pretty sure things would go sideways if we tried to mix openstack and non openstack machines in the same cluster.

teuthology/openstack.py
dachary: Ok, i'll add docstrings.

teuthology/openstack.py
dachary: I'll change that for something more flexible.

teuthology/openstack.py
dachary: I'll remove the comments.

teuthology/test/integration/openstack-integration.py
dachary: I did not realize py.test was required. I used testtools because I know about it. I've used py.test for the other file though and I can switch over.

teuthology/test/integration/setup-openstack.sh
dachary: I would happily do otherwise if there was a --config option but I don't think there is. In any case this is run on a machine that is created for the sole purpose of running the cluster. You're not supposed to run setup-openstack yourself. That would mess a lot more than just the ~/.teuthology.yaml ;-)

dachary:
> Would it be possible to put teuthology/openstack-*.txt in a subdir?

Yes.

> I wonder if it might be good to actually make teuthology.openstack a subpackage rather than just a module.

what would that entail ?

dachary:
Moving on to https://github.com/ceph/teuthology/pull/592 so that github does not loose all the comments in a rebase.

PR #585, opened 2015-08-04
 - Use random distro/version instead of ubuntu as default

vasukulkarni:
- Use random distro/version when testing on vps without any os/verison option

### DIFF
--- a/teuthology/lock.py
+++ b/teuthology/lock.py
@@ -8,6 +8,7 @@ import collections
 import os
 import requests
 import urllib
+import random
 
 import teuthology
 from . import misc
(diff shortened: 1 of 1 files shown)


teuthology/lock.py
```diff
@@ -16,21 +17,7 @@
...
-def get_distro_from_downburst():
-    """
-    Return a table of valid distros.
-
-    If downburst is in path use it.  If either downburst is unavailable,
-    or if downburst is unable to produce a json list, then use a default
-    table.
-    """
-    default_table = {u'rhel_minimal': [u'6.4', u'6.5'],
+default_table = {u'rhel_minimal': [u'6.4', u'6.5'],
```
zmc: If this is going to be top-level in the module it needs a more specific name

teuthology/lock.py
```diff
@@ -226,6 +226,13 @@ def main(ctx):
     if ctx.num_to_lock:
         assert ctx.machine_type, \
             'must specify machine type to lock'
+    if ctx.machine_type == 'vps':
+        if ctx.os_type is None:
+            ctx.os_type=random.choice(default_table.keys())
+	    ctx.os_version=random.choice(default_table[ctx.os_type])
```
zmc: What you're doing here is adding an unrelated, unexpected side-effect to a function designed to simply "Return a table of valid distros". It is not the place where decisions about *what* distro to use should be made.

It is currently made here for downburst:
https://github.com/ceph/teuthology/blob/master/teuthology/provision.py#L210-L211

To properly implement this would require a little bit of refactoring I think.

vasukulkarni: @zmc: I think the function does more than just return table of valid distros, if you use the teuthology-lock cli it creates VMs. After I pushed this PR i realized this is not the best place since it handles only  "teuthology-lock" cli case, probably Downburst class may be the right place if no other function is setting default values before they call the init or init can set the default values to be more clear and should also own supported_table dict. I will close this and reopen a different one a

PR #586, opened 2015-08-04
packaging: ignore minor version numbers in gitbuilder urls

andrewschoen:
We don't care about minor version numbers anymore for gitbuilder.

### DIFF
--- a/teuthology/test/test_packaging.py
+++ b/teuthology/test/test_packaging.py
@@ -352,7 +352,9 @@ class TestGitbuilderProject(object):
 
     GITBUILDER_DISTRO_MATRIX = [
         ('rhel', '7.0', None, 'centos7'),
-        ('centos', '6.5', None, 'centos6_5'),
+        ('centos', '6.5', None, 'centos6'),
+        ('centos', '7.0', None, 'centos7'),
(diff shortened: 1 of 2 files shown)


teuthology/test/test_packaging.py
```diff
@@ -352,7 +352,7 @@ def test_get_package_version_not_found(self, m_get, m_get_config_value,
 
     GITBUILDER_DISTRO_MATRIX = [
         ('rhel', '7.0', None, 'centos7'),
-        ('centos', '6.5', None, 'centos6_5'),
+        ('centos', '6.5', None, 'centos6'),
```
liewegas: shoudl centos7 be in this list?  what is it used for?

liewegas:
lgtm

teuthology/test/test_packaging.py
andrewschoen: yeah, it should be.  I'll add both centos 7.0 and centos 7.1.

This is used to feed my test that uses pytest.parameterize below.  This makes sure I can generate the correct distro name given distro, version and codename.

ktdreyer:
LGTM

PR #588, opened 2015-08-04
install: ignore minor version when building a gitbuilder url

andrewschoen:
We don't care about minor version in gitbuilder anymore.


ktdreyer:
:+1:

PR #591, opened 2015-08-05
make a colored version of teuthology.log

andrewschoen:
This uses ansi2html to convert the teuthology.log to an html file that
shows the ansi colors.

See an example here: 

http://qa-proxy.ceph.com/teuthology/ubuntu-2015-08-05_13:51:24-teuthology-master---basic-multi/1001742/teuthology.log.html

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -5,6 +5,7 @@ import contextlib
 import sys
 import logging
 from traceback import format_tb
+from ansi2html import Ansi2HTMLConverter
 
 import teuthology
 from . import report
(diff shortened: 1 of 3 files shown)


teuthology/run.py
```diff
@@ -243,13 +244,26 @@ def report_outcome(config, archive, summary, fake_ctx):
 
     report.try_push_job_info(config, summary)
 
+    teuth_log = os.path.join(archive, 'teuthology.log')
     if passed:
         log.info(status)
+        make_html_log(teuth_log)
     else:
         log.info(str(status).upper())
```
andrewschoen: I don't see why we couldn't always make the status line uppercase and simplify this block. Thoughts?

Something like:

```
teuth_log = os.path.join(archive, 'teuthology.log')
log.info(str(status).upper())
make_html_log(teuth_log)
if not passed:
    sys.exit(1)
```

zmc:
My concern about this is performance-related; we do sometimes generate very large logfiles and ``teuthology.front`` is borderline overloaded as it is.

andrewschoen:
@zmc I'll see if I can test how well ansi2html handles a large log file.  Also, we could make this feature configurable through the teuthology config so that we could easily turn it off if we notice it's burning down ``teuthology.front``.

andrewschoen:
@zmc I found a large log file (307M) and wrote a script that does exactly what ``make_html_log`` is doing.  The process took ~8 seconds and did spike the CPU to about 99% for the last few seconds according to top.  That first process here is my script.

screen shot 2015-08-06 at 9 50 39 am

smithfarm:
This PR hasn't been touched in years, so I'll close it now. Please reopen if it's still relevant.

PR #592, opened 2015-08-05
transparent OpenStack provisioning for teuthology-suite

dachary:
http://tracker.ceph.com/issues/6502


dachary:
@zmc I think I addressed your comments, except the transition from testtool to py.test. How does that look ?

dachary:
@zmc the files were reorganized to be in the openstack subpackage and the ubuntu user made configurable. The transition from testtools to py.test is not done yet. How does that look ?

README.rst
```diff
@@ -317,6 +317,126 @@ specified in ``$HOME/.teuthology.yaml``::
...
+    | name        | myself                                          |
+    | user_id     | 5cf9fa21b2e9406b9c4108c42aec6262                |
+    +-------------+-------------------------------------------------+
+
+Usage
+-----
+
+* Run the dummy suite as a test::
+
+    $ teuthology-openstack --key-name myself --suite dummy
```
zmc: Might want to clarify that 'myself' is "your openstack username"

dachary: It's not the OpenStack user name, it's one of the keypair you registered for your user.

zmc: Actually maybe it's not (OS1 internal):
``
2015-08-07 15:38:54,990.990 ERROR:teuthology.misc:openstack server create  --image 'teuthology-ubuntu-14.04'  --flavor 'm1.small'   --key-name zcerza --user-data /tmp/tmpPySLGY --security-group teuthology --wait teuthology error ERROR: openstack Invalid key_name provided. (HTTP 400) (Request-ID: req-dc994e8c-c85d-4a47-a12f-5dbfc5ff34ea)
``

teuthology/openstack/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+                        action="cloud_init_wait " + name_or_ip) as proceed:
+            success = False
+            # CentOS 6.6 logs in /var/log/clout-init-output.log
+            # CentOS 7.0 logs in /var/log/clout-init.log
+            all_done = ("tail /var/log/cloud-init*.log ; " +
+                        " test -f /tmp/init.out && tail /tmp/init.out ; " +
+                        " grep '" + self.up_string + "' " +
+                        "/var/log/cloud-init*.log")
+            while proceed():
+                client = paramiko.SSHClient()
```
zmc: Curious why you're not using ``Remote`` here - is it just inappropriate?

It does seem like this ``while`` block might want to be split out as a psuedo-private method.

teuthology/provision.py
```diff
@@ -195,6 +200,168 @@ def __del__(self):
         self.remove_config()
 
 
+class ProvisionOpenStack(OpenStack):
```
zmc: Do you think it would be feasible/desirable to implement an abstract base ``Provision`` class that ``Downburst`` and ``ProvisionOpenStack`` (should maybe just be called ``OpenStack``) could inherit from? 

Just to formalize a bit more that they intend to provide a similar interface: ``create()``, ``destroy()``, ``build_config()``, ``remove_config()`` - though the last two are currently implemented by differently-named methods in ``ProvisionOpenStack``

teuthology/openstack/setup-openstack.sh
```diff
@@ -0,0 +1,548 @@
+#!/bin/bash
```
zmc: I think you mentioned yesterday on the phone that you intend to (or intend to help) obsolete this with ansible or even just python; perhaps a comment at the top noting that would be helpful in case we both disappear :)

README.rst
zmc: Hrm, I'm seeing a lot of:
``cloud_init_wait AuthenticationException Authentication failed.``

zmc:
@dachary The reorganization looks very good - thanks!

The user appears to be mostly configurable, save for ``teuthology/openstack/openstack-teuthology.init`` and ``teuthology/openstack/test/user-data-test1.txt``; I don't intend to block this PR on that though.

teuthology/provision.py
dachary: create() and destroy(), definitely. The build_config / remove_config are not shared, I think.

dachary:
@zmc I'm guessing the troubles you have with authentication via paramiko are related to https://github.com/paramiko/paramiko/issues/387

README.rst
dachary: @zmc I'm guessing the troubles you have with authentication via paramiko are related to paramiko/paramiko#387

A simple test is http://paste2.org/ZagzFYZX

Maybe orchestra would work but I would have to add support for the key_filename option and figure out how to catch the exceptions properly.

teuthology/openstack/__init__.py
dachary: The part that tracks the progress of the tail until a given string is found could probably be using Remote.

dachary:
> teuthology/openstack/openstack-teuthology.init and teuthology/openstack/test/user-data-test1.txt 

I modified teuthology/openstack/openstack-teuthology.init but did not changed teuthology/openstack/test/user-data-test1.txt  because I assume we don't care since it's a test file

dachary:
I can't make sense of this error:
<pre>
import file mismatch:
imported module 'teuthology.test.task.test_selinux' has this __file__ attribute:
  /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/py27/local/lib/python2.7/site-packages/teuthology/test/task/test_selinux.py
which is not the same as the test file we want to collect:
  /home/jenkins-build/build/workspace/teuthology-pull-requests/teuthology/test/task/test_selinux.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
</pre>
what is it about ?

dachary:
@zmc you can change the name of the instance created to run the teuthology cluster with **--name**. I only used it for test purposes and was not sure if it was usable. I tried it and all is well. Note that all instances created by a given teuthology cluster have property=IP-of-the-teuthology-instance. When the cluster is teardown, it does so by destroying the instances that have that property. It allows multiple cluster to co-exist within the same tenant.

dachary:
@zmc testtools has been replaced with py.test

README.rst
```diff
@@ -317,6 +317,144 @@ specified in ``$HOME/.teuthology.yaml``::
...
+
+It is cheaper than EnterCloudSuite but does not provide volumes (as
+of August 2015) and is therefore unfit to run teuthology tests that
+require disks attached to the instance. Each instance has a public IP
+by default.
+
+* `create an account <https://www.ovh.com/fr/support/new_nic.xml>`_
+* get $HOME/openrc.sh from `the horizon dashboard <https://horizon.cloud.ovh.net/project/access_and_security/?tab=access_security_tabs__api_access_tab>`_
+
+Setup
```
andrewschoen: It's not entirely clear to me if these steps are happening on your host or the vm you just created.

dachary: All the steps happen on your host. Only integration testing in the chapter below happen on the vm that was created. This is indeed confusing and I'm assuming the person willing to run the integration tests will have enough motivation to sort that out. 

Any idea how to clarify that there is no need to ssh to something for the steps below ?

andrewschoen: I'm working through getting this setup now.  After I get it work I most likely will have some suggestions, I hope at least.

I'm also trying a different openstack provider than OVH, so there might be some differences there as well.

zmc: Maybe just put something like "On your own workstation:" just above the steps?

zmc:
@dachary I'm still seeing the same authentication issues I saw last time I tried to use this.

If this is merged to master as-is, there may be many people that have the same issue as me; since I don't know how to fix it myself, I'd have no idea what to tell them.

It seems like ``teuthology.openstack.OpenStack.cloud_init_wait()`` should be using ``teuthology.orchestra.connection``

dachary:
@zmc --verbose will now also set the paramiko debug flag and give us more information about the error you're seeing. Could you please upload the output somewhere for me to look at ? 

Although orchestra would most certainly fix the specific problem you currently have, it has other issues such as not exposing the key_filename argument and maybe others. If it was just a matter of switching, I would happily do it. But it's more a matter of patching orchestra to fit the needs of the OpenStack backend and I feel the cons currently outweight the pros.

To be entirely honest I have to admit that part of my resistance is that switching to orchestra would require  two or three days of work on my part.

teuthology/openstack/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+        Return the private IP of the OpenStack instance_id. The network,
+        if not the empty string, disambiguate multiple networks attached
+        to the instance.
+        """
+        return re.findall(network + '=([\d.]+)',
+                          self.get_addresses(instance_id))[0]
+
+class TeuthologyOpenStack(OpenStack):
+
+    def __init__(self, args, config, argv):
```
zmc: What is the difference between ``args`` and ``argv`` here? A docstring might be useful for this method.

zmc:
@dachary it looks like it's using the wrong SSH key; it's ignoring my ``~/.ssh/config``, which tells everything else which one to use.

I think it would be worthwhile to enable this PR to use orchestra; I might be willing to do the work myself if I knew what all was needed. It looks like exposing ``key_filename`` would be trivial, for example.

teuthology/openstack/__init__.py
dachary: argv is the orginal arv list so that it can be displayed as is should an error occur. Using args is more confusing to the reader.

dachary:
@zmc the other thing that's non trivial (at least to me) is handling the errors in the right way at the various stages of the virtual machine boot (which may involve a reboot). It looks like orchestra is not transparent when it comes to bubbling the errors from lower levels. I did not investigate much though.

dachary:
@zmc it occurs to me that there may be a need to explicitly ask paramiko to handle the ~/.ssh/config file ? I though that was implicit but maybe it's not ?

zmc:
@dachary It looks like ``orchestra.connection.connect()`` used to bubble up exceptions but I changed that last May to add retries (sadly I left the comment in place, which is now a lie). It would be easy to make retries optional, and if they were disabled, the caller would be exposed to any exceptions raised by ``paramiko.SSHClient.connect()``.

dachary:
I'll rework orchestra and the OpenStack backend to work together.

dachary:
It will take a few weeks (hollidays etc.)

zmc:
@dachary I can put some time into it if you think it would be worthwhile

teuthology/provision.py
zmc: Well, ``init_user_data()`` is "building a config" and ``__del__()`` is "removing a config"

scripts/openstack.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+    parser.add_argument(
+        '--upload',
+        action='store_true', default=False,
+        help='upload archives to an rsync server',
+    )
+    parser.add_argument(
+        '--archive-upload',
+        help='rsync destination to upload archives',
+        default='ubuntu@integration.ceph.dachary.org:./',
```
zmc: This should not be hardcoded

README.rst
```diff
@@ -317,6 +317,144 @@ specified in ``$HOME/.teuthology.yaml``::
...
+* Visit the web interface (the URL is displayed at the end of the
+  teuthology-openstack output) to monitor the progress of the suite.
+
+* The virtual machine running the suite will persist for forensic
+  analysis purposes. To destroy it run::
+
+    $ teuthology-openstack --key-filename myself.pem --key-name myself --teardown
+
+* The test results can be uploaded to a publicly accessible location
+  with the ``--upload`` flag::
```
zmc: Apparently this is a flag you want to pass when scheduling; that's confusing since it's mentioned *after* the docs explaining how to create and destroy a cluster. This should also mention that ``--upload`` by itself won't be useful unless ``--archive-upload`` has a useful value.

zmc:
@dachary Have you seen this before?
```
(virtualenv)zack@magna002:~/teuthology$ teuthology-lock --lock-many 1 --machine-type openstack
  File "/home/zack/teuthology/virtualenv/bin/teuthology-lock", line 9, in <module>
    load_entry_point('teuthology==0.1.0', 'console_scripts', 'teuthology-lock')()
  File "/home/zack/teuthology/scripts/lock.py", line 17, in main
    sys.exit(teuthology.lock.main(parse_args(sys.argv[1:])))
  File "/home/zack/teuthology/teuthology/lock.py", line 331, in main
    ctx.desc, ctx.os_type, ctx.os_version)
  File "/home/zack/teuthology/teuthology/lock.py", line 408, in lock_many
    arch=arch)
  File "/home/zack/teuthology/teuthology/lock.py", line 381, in lock_many_openstack
    num, os_type, os_version, arch, resources_hint)
  File "/home/zack/teuthology/teuthology/provision.py", line 298, in create
    self.init_user_data(os_type, os_version)
  File "/home/zack/teuthology/teuthology/provision.py", line 226, in init_user_data
    template_path = config['openstack']['user-data'].format(
TypeError: 'NoneType' object has no attribute '__getitem__'
```

dachary:
@zmc yeah, it's when you try to use the OpenStack backend and ~/.teuthology.yaml has no information about it

zmc:
@dachary That's what I guessed - I don't see docs on how to fill that in. I did run ``teuthology-openstack`` beforehand and the teuthology node there does have a config with an ``openstack`` stanza, as expected. I'm currently trying to figure out how to construct one that I can use without a separate teuthology node inside OpenStack...

teuthology/openstack/openstack-user-data.txt
```diff
@@ -0,0 +1,16 @@
...
+  - touch /tmp/init.out
+system_info:
+  default_user:
+    name: TEUTHOLOGY_USERNAME
+packages:
+ - python-virtualenv
+ - git
+ - rsync
+runcmd:
+ - su - -c '(set -x ; git clone -b wip-6502-openstack-v3 http://github.com/dachary/teuthology && cd teuthology && ./bootstrap install)' TEUTHOLOGY_USERNAME >> /tmp/init.out 2>&1
```
zmc: This should not be hardcoded.

teuthology/openstack/test/test_openstack.py
```diff
@@ -0,0 +1,131 @@
...
+    def test_create(self, capsys):
+        teuthology_argv = [
+            '--suite', 'upgrade/hammer',
+            '--dry-run',
+            '--ceph', 'master',
+            '--kernel', 'distro',
+            '--flavor', 'gcov',
+            '--distro', 'ubuntu',
+            '--suite-branch', 'hammer',
+            '--email', 'loic@dachary.org',
```
zmc: This should not be hardcoded.

zmc:
I attempted to use ``teuthology-openstack`` to run the 'teuthology' suite. It's not going very well :(

http://149.202.167.19:8081/ubuntu-2015-08-27_17:48:34-teuthology-master---basic-openstack/
1. It used 'ubuntu' as the username whereas I ran the command as 'zack'.
2. Several tests are failing. The most common is: ``[Errno 2] No such file or directory: 'teuthology/openstack/openstack-rhel-7.0-user-data.txt'``
4. Only two worker processes are running for a 25-job run on a providor with a very large quota
3. This is more of an issue with the suite I guess, but I have two jobs stuck in 'waiting' for ~2h because they are looking for 'vps' machines.
5. Logs are not stored in ``/a/`` or ``~teuthworker/archive`` but ``/usr/share/nginx/html`` inexplicably.
6. Logs are served as ``Content-Type: application/octet-stream``, so browsers refuse to display them

I'll attempt to address some of these, but I'll admit I'm having a very hard time navigating the codebase. A lot of them would be solved by simply not attempting to set up an entirely separate teuthology cluster. But I haven't gotten ``--lock-many`` working either.

dachary:
> 1. It used 'ubuntu' as the username whereas I ran the command as 'zack'.

Which command did you run ?

> 2. Several tests are failing. The most common is: ``[Errno 2] No such file or directory: 'teuthology/openstack/openstack-rhel-7.0-user-data.txt'``

rhel 7.0 is not supported, only centos 7 and ubuntu 14.04

> 4. Only two worker processes are running for a 25-job run on a providor with a very large quota

The number of workers is defined by --simultaneous-jobs regardless of the quotas. Which provider are you using ?

> 3. This is more of an issue with the suite I guess, but I have two jobs stuck in 'waiting' for ~2h because they are looking for 'vps' machines.

I would need details to figure this one out.

> 5. Logs are not stored in ``/a/`` or ``~teuthworker/archive`` but ``/usr/share/nginx/html`` inexplicably.

The teuthology cluster created by teuthology-openstack has that hardcoded at https://github.com/dachary/teuthology/blob/wip-6502-openstack-v3/teuthology/openstack/openstack-teuthology.init#L58

> 6. Logs are served as Content-Type: application/octet-stream, so browsers refuse to display them

That's nginx default.

dachary:
> I'll attempt to address some of these, but I'll admit I'm having a very hard time navigating the codebase. A lot of them would be solved by simply not attempting to set up an entirely separate teuthology cluster. 

I agree. But how to run integration tests if not by setting up an entire cluster ? Things would be infinitely better if the cluster deployment was done via ansible instead of a shell script. Do you think we would be better off without the integration tests and we should wait for ansible to be able to deploy a full cluster instead ?

scripts/openstack.py
dachary: this is a sensible default for everyone to use

README.rst
dachary: I suspect --upload won't be used most of the time. Only after error / trial, when someone wants to permanently advertise and publish success. We're still experimenting with the use cases and I'm not sure how it should be documented just yet.

Note that the user of teuthology-openstack does not distinguish the cluster creating from scheduling. From his point of view scheduling is the only thing that exists and the cluster creation is transparent.

teuthology/openstack/test/test_openstack.py
dachary: this is a test value only

teuthology/openstack/openstack-user-data.txt
dachary: agreed

dachary:
Rebased. The git clone command is now configurable. The configuration options are documented and sensible defaults were added.

dachary:
Rebased. The git clone command is now configurable. The configuration options are documented and sensible defaults were added.

dachary:
Rebased. The git clone command is now configurable. The configuration options are documented and sensible defaults were added.

zmc:
@dachary Thanks for the config change. I had a similar work in progress in a branch of mine, but yours is more complete.

zmc:
@dachary You may want to cherry-pick https://github.com/zmc/teuthology/commit/97e9b3954ace6a7e5ae6deadcd635dd352ec336a . Let me know if PRs work better for this.

zmc:
@dachary RE: testing:

While I like that there are "integration" tests here (my quotes will make sense in a second), I'd like to get some unit tests into this branch. 

In looking at how we might bring the existing and future tests in line with the way the rest of teuthology tries to stay organized, I noticed that ``test_openstack.py`` looks more like an integration test than a unit test. This is because it hits the network, creates VMs, etc. It also doesn't want to pass for me locally.

What we're currently calling "integration" tests look more like system tests. I think we should organize tests to reflect this. If these seem like meaningless words, the main point is that we should have a set of tests that are able to execute very quickly (sub-second runtimes) that we can expand on as we make changes.

Regarding the concept of always setting up a teuthology cluster inside of OpenStack, it's not that I oppose that functionality; I just want to make sure we aren't *tied* to it.

As with most of the changes I'm suggesting, I'm willing to do most or all of the work, assuming I'm able :)

dachary:
cherry-picked  810cce8 (a PR works fine too)

dachary:
@zmc I understand your concern regarding integration tests being mandatory.

scripts/openstack.py
liewegas: i set up a dns alias teuthology-logs.public.ceph.com

scripts/openstack.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+    parser.add_argument(
+        '--upload',
+        action='store_true', default=False,
+        help='upload archives to an rsync server',
+    )
+    parser.add_argument(
+        '--archive-upload',
+        help='rsync destination to upload archives',
+        default='ubuntu@teuthology-logs.public.ceph.com:./',
```
dachary: Updated with the ceph.com URL

teuthology/openstack/__init__.py
zmc: Sorry if I'm being thick here, but I don't quite understand. Could we just get a docstring here please?

dachary:
@zmc commented on args + argv and rebased and repushed

zmc:
@dachary Thanks, that's a bit more clear.

To repeat what I've said to you privately, I don't wait to wait until this particular PR feels "perfect" to get it into master - that could drag on endlessly and not help anyone. I think the major issues are taken care of, so I'll merge it now.

Sometime in the near future I'll look at adapting the feature to be usable seamlessly with the current workflow we use in our labs. In the meantime this is hugely useful for external contributors! Thanks so much for all your work and patience :)

PR #595, opened 2015-08-13
fix ansible error reporting

andrewschoen:
When moving the error log between file systems of two different types os.rename will throw an error while shutil.move correctly handles that case.  See: http://tracker.ceph.com/issues/12656

I've also added some exception handling incase we fail to parse the error log.  All it does is fail silently and log the exception so that we can inspect the log after the fact. See: http://tracker.ceph.com/issues/12675

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -4,6 +4,7 @@ import requests
 import os
 import pexpect
 import yaml
+import shutil
 
 from cStringIO import StringIO
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


teuthology/task/ansible.py
```diff
@@ -263,7 +264,10 @@ def execute_playbook(self, _logfile=None):
     def _handle_failure(self, command, status):
         failures = None
         with open(self.failure_log.name, 'r') as log:
-            failures = yaml.safe_load(log)
+            try:
+                failures = yaml.safe_load(log)
+            except yaml.parser.ParserError:
+                log.exception("Failed to load failure log...")
```
alfredodeza: it would be useful to add the log path here

andrewschoen: Agreed. The failure log will get moved into $archive_dir/ansible_failures.yaml but they might not know that.

dmick: What he said, otherwise lgtm.  Have we seen failures to parse?

andrewschoen: @dmick yeah, http://tracker.ceph.com/issues/12675

andrewschoen: @dmick @alfredodeza I've got that logging updated.

PR #596, opened 2015-08-13
refactor the install task to use packaging.GitbuilderProject

andrewschoen:
I've tested this on Ubuntu, Debian and Centos.  It might be smart to run an upgrade suite with this branch before merging just to be safe though.

This should fix: http://tracker.ceph.com/issues/12668

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -443,6 +443,8 @@ class GitbuilderProject(object):
         else:
             self._init_from_config()
 
+        self.dist_release = self._get_dist_release()
+
     def _init_from_remote(self):
         """
         Initializes the class from a teuthology.orchestra.remote.Remote object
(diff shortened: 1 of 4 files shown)


andrewschoen:
@yuriw tested this with an upgrade suite and it fixes the issue we saw in: http://tracker.ceph.com/issues/12668

I also did quite a bit of manual testing with this. I think it's good to go after review.

yuriw:
@andrewschoen here is the run results "-t wip-12668" branch http://pulpito.ceph.com/teuthology-2015-08-14_13:38:32-upgrade:hammer-hammer-distro-basic-vps/

andrewschoen:
@yuriw lots of failures there, but I'm pretty sure I've fixed most of these with the latest version of this branch, but let me look over them again.

 "'file' object has no attribute 'exception'" fixed with https://github.com/ceph/teuthology/pull/597

Most (if not all) of the ansible failures are related to a timeout on this or other dl.fedoraproject.org timeouts: http://dl.fedoraproject.org/pub/epel/6/x86_64/repodata/93e31187cc7a93b4f37ba4872591d83669a87d6addaa4ce50ab698b9248a271a-comps-el6.xml.bz2

"{'vpm078.front.sepia.ceph.com': "error while evaluating conditional: ssh_key_update.state == 'present'"}" is typically a sign that we used ansible < 1.9.2. How did you schedule this?

'Failed to validate the SSL certificate for raw.githubusercontent.com:443" I forget what causes this, but I've seen it before.

It would probably be a good idea to schedule this suite again with the wip-12668 branch to see if my latest commits fix a few things.  Hopefully we'll see less timeouts on a second run too.

andrewschoen:
@yuriw the good news about that test run is that I didn't see anything that looked like http://tracker.ceph.com/issues/12668

PR #597, opened 2015-08-14
ansible: rename log to fail_log to not conflict with ansible.log

andrewschoen:
Fixing a bug introduced with https://github.com/ceph/teuthology/pull/595.  When I opened the failure_log I gave it the name ``log`` which conflicts with ``ansible.log`` which is what I was really wanting to use with ``log.exception``.


dmick:
Oops!  Looks right to me.

PR #598, opened 2015-08-18
task.ansible: set the ansible_failure.yaml file mode to 0664

andrewschoen:
Without the permissions set correctly apache will not serve this file

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -288,6 +288,7 @@ class Ansible(Task):
             self.failure_log.name,
             archive_path
         )
+        os.chmod(archive_path, 0664)
 
     def _build_args(self):
         """


ktdreyer:
Things that would be cool to put into the commit log:

* What were the old permissions that were broken?
* Does the file have to be group-writable as well as user-writable?

andrewschoen:
@ktdreyer I've updated the commit message.  To answer your questions, 0600 was the original mode and I've set this to 0664 instead of 0644 to match the rest of the archived files from a teuthology run.  I'm unsure if group writeable is really needed, but I thought consistency would be best just in case.

ktdreyer:
Cool, I was wondering about that import, too. I see it's gone now.

PR #603, opened 2015-08-26
kernel: use packaging.GitbuilderProject to talk with gitbuilder.

andrewschoen:
This fixes issue: http://tracker.ceph.com/issues/12795

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -7,8 +7,6 @@ import logging
 import os
 import re
 import shlex
-import urllib2
-import urlparse
 
 from teuthology import misc as teuthology
 from ..orchestra import run
@@ -23,6 +21,7 @@ from ..packaging import (
     get_koji_package_name,
     get_koji_task_rpm_info,
     get_koji_task_result,
+    GitbuilderProject,
 )
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


teuthology/task/kernel.py
```diff
@@ -354,31 +353,19 @@ def download_kernel(ctx, config):
...
-                arch=larch,
-                dist=ldist,
-                )
+            gitbuilder = GitbuilderProject(
+                'kernel',
+                ctx.config.get('kernel'),
+                ctx=ctx,
+                remote=role_remote,
+            )
+            baseurl = gitbuilder.base_url + "/"
```
zmc: This strikes me as odd. Should ``base_url`` maybe supply the trailing ``/``?

andrewschoen: It could, I suppose. I'm unsure if it's worth it though. The url template for that is currently tweakable through teuthology.yaml, but I'm unsure if the labs actually use it or if they just depend on the default value.  I'll have to take a look.

andrewschoen: Doesn't look like the labs override ``baseurl_template`` from what I see.

zmc: So, does that mean it should be safe to supply the trailing ``/``?

On second thought, I don't want to block this bugfix for this. We should look into this at some point though.

PR #605, opened 2015-08-28
suite: pass dev-commit: <sha1> to ceph-deploy

liewegas:
Use the new ceph-deploy install --dev-commit <sha1> option to inform the
install.

### DIFF
--- a/teuthology/suite.py
+++ b/teuthology/suite.py
@@ -1006,7 +1006,7 @@ dict_templ = {
         },
         'ceph-deploy': {
             'branch': {
-                'dev': Placeholder('ceph_branch'),
+                'dev-commit': Placeholder('ceph_hash'),
             },
             'conf': {
                 'client': {


zmc:
Any reason to care about old ceph-deploy versions here?

liewegas:
I don't think so..w e don't test them!

note that the ceph-deply pr hasn't merged yet, tho..

https://github.com/ceph/ceph-deploy/pull/348

zmc:
@liewegas Oh! Glad I didn't merge yet then :)

trhoden:
:+1: 

Are there things in the suites that will have to change to figure out a SHA1 instead of a branch name?  That's probably already done somewhere...

liewegas:
On Fri, 28 Aug 2015, Travis Rhoden wrote:
> Are there things in the suites that will have to change to figure out a SHA1 instead of a branch name? That's probably already done somewhere...

the only change to teh suite will be to remove the install task, i 
believe.  once this slow-ass dist-upgrade finishes i can test and verify!

liewegas:
tested this maually and dev-commit: sha1 is working!

PR #609, opened 2015-09-02
Use lsb_release -a as the first method to gather info about a remote

andrewschoen:
Before this change debian would use the python method to gather information,
but that does not include codename.  Using lsb_release first and then
checking /etc/os-release and lastly python's platform module should give
us more complete information about many distros.

See: http://tracker.ceph.com/issues/12669


zmc:
Closing this as you were going to rework it with a different approach

PR #610, opened 2015-09-02
Add version check for 1.3.1, add ansible task

vasukulkarni:
Version check for 1.3.1  bit of restructure for overrides and add ansible task
 
### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -12,6 +12,7 @@ from teuthology import misc as teuthology
 from teuthology import contextutil, packaging
 from teuthology.parallel import parallel
 from ..orchestra import run
+from . import ansible
 
 log = logging.getLogger(__name__)
 
@@ -1143,7 +1144,12 @@ def task(ctx, config):
         branch: foo
         extra_packages: ['samba']
     - install:
-        rhbuild: 1.2.3
+        rhbuild: 1.3.0 
+        playbook: downstream_setup.yml
+        vars:
+           yum_repos:
+             - url: "http://location.repo"
+               name: "ceph_repo"
 
     Overrides are project specific:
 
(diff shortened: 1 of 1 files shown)


zmc:
The commit message isn't clear (I can't parse it). Overrides don't seem to be affected by this commit either.

vasukulkarni:
@zmc: commit message should be fixed now

dmick:
This is at least three commits, and each of those separate commits should have messages explaining the problem clearly and why this code resolves the problem

vasukulkarni:
@dmick: if you want 3 different commits and message for each commit, I can force push another.

dmick:
3 commits is definitely better, but please put some real information in the commit messages, like I asked: explaining the problem clearly and why this code resolves the problem.  It only takes you a few tens of seconds, but can save a lot of time in six months when we revisit this

vasukulkarni:
@dmick : I already pushed 3 commits in the morning, Each commit now explains what it does.

zmc:
@vasukulkarni @dmick is asking you for commit messages "explaining the problem clearly and why this code resolves the problem". Your commit messages say what they do, yes, but they don't say what the problem they're solving is. This seems like a reasonable request.

vasukulkarni:
@dmick: I have added the required info in the additional lines of commit message it should be clear now what it solves.

teuthology/task/install.py
```diff
@@ -1175,17 +1176,26 @@ def task(ctx, config):
...
 
     if config.get('rhbuild'):
-        with contextutil.nested(
-            lambda: rh_install(ctx=ctx, config=config),
-            lambda: ship_utilities(ctx=ctx, config=None)
-        ):
-            yield
+        if config.get('playbook'):
+            ansible_config=dict(config)
+            #remove key not required by ansible task
```
dmick: still no space here?

teuthology/task/install.py
```diff
@@ -1175,17 +1176,26 @@ def task(ctx, config):
...
+    if config.get('rhbuild'):
+        rhbuild = config.get('rhbuild')
+        log.info("Build is %s " % rhbuild)
+
     flavor = get_flavor(config)
     log.info("Using flavor: %s", flavor)
 
     ctx.summary['flavor'] = flavor
+    nested_tasks = [lambda: rh_install(ctx=ctx, config=config),
+                     lambda: ship_utilities(ctx=ctx, config=None)]
```
dmick: the lambdas should still align?

PR #611, opened 2015-09-03
Modifies check of stale data by checking if /var/lib/ceph is empty

ivotron:
Checks whether `/var/lib/ceph` is empty, instead of checking whether the folder exists.


kshtsk:
@smithfarm @gregsfortytwo is this is something we need or we have to close this PR?

smithfarm:
@kshtsk This PR was opened four years ago. In the meantime, the code in master changed like so: https://github.com/ceph/teuthology/blob/f40fd3fd75825607d021e582babd73a059bd2724/teuthology/task/internal/__init__.py#L256-L260

So, I think this PR is obsolete.

PR #612, opened 2015-09-03
Adds option to override the default SSH port

ivotron:
Currently, if a test node's `sshd` binds to another port, it is not possible to configure teuthology so that it communicates via this port (22 is "hard-coded"). This PR add support for alternate ports by specifying them in `targets`, for example:

``` yaml
targets:
  'ubuntu@localhost:2222': ssh-dss sshkeytolocalhost
```

### DIFF
--- a/teuthology/orchestra/connection.py
+++ b/teuthology/orchestra/connection.py
@@ -38,7 +38,7 @@ def create_key(keytype, key):
 
 
 def connect(user_at_host, host_key=None, keep_alive=False, timeout=60,
-            _SSHClient=None, _create_key=None):
+            _SSHClient=None, _create_key=None, port=22):
     """
     ssh connection routine.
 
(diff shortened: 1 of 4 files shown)


zmc:
This looks like it would require manual assembly of the targets each time. Have you looked to see if it would be possible to simply create ssh config entries that we could ask paramiko to look up? It could go here: 
https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/connection.py#L89-L93

If that works, we could potentially use a single ssh config stanza to set the port for multiple hosts.

ivotron:
I'm not sure I understand what you mean by "manual assembly", do you mean that this is not backwards-compatible? The problem with using `~/ssh/.config` is that, for [the use case we intend to use this](https://github.com/ivotron/docker-cephdev/blob/master/teuth/README.md#multiple-targets), multiple ssh daemons might be running on the same host.

liewegas:
@ivotron is this still needed/relevant?

PR #614, opened 2015-09-03
do not install libpython-dev by default because it's not on precise

dachary:


dachary:
tested to ./bootstrap install successfully on precise, testing on trusty

dachary:
tested to work on ubuntu 14.04, not sure why I felt the need of adding this initially

zmc:
@dachary I pushed https://github.com/ceph/teuthology/commit/4d1d8d93b4cc8f7ea246109b9b2d7038cd397130 earlier - do we still need this?

PR #616, opened 2015-09-04
use modern paramiko

liewegas:
We need >= 1.15 to connect to newer OpenSSH, such as the version used by
debian jessie.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
                       'configobj',
                       'six',
                       'httplib2',
-                      'paramiko < 1.8',
+                      'paramiko >= 1.15.1',
                       'pexpect',
                       'requests >= 2.3.0',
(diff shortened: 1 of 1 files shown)


andrewschoen:
I remember at one point there was an attempt to upgrade paramiko and the result was lots of tests breaking, I could be remembering wrong though. @zmc did that work.

Have you tested this on other distros?

I think this was the ticket where we attempted to upgrade paramiko for, http://tracker.ceph.com/issues/11035

setup.py
```diff
@@ -42,7 +42,7 @@
                       'configobj',
                       'six',
                       'httplib2',
-                      'paramiko < 1.8',
+                      'paramiko >= 1.5.1',
```
zmc: Do you mean 1.15.1 here?

PR #618, opened 2015-09-07
openstack: throttle job scheduling by default

dachary:
http://tracker.ceph.com/issues/12977

### DIFF
--- a/teuthology/suite.py
+++ b/teuthology/suite.py
@@ -12,6 +12,7 @@ import subprocess
 import smtplib
 import socket
 import sys
+from time import sleep
 import yaml
 import math
 from email.mime.text import MIMEText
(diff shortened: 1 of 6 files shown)


teuthology/suite.py
```diff
@@ -618,6 +624,9 @@ def schedule_suite(job_config,
             verbose=verbose,
             log_prefix=log_prefix,
         )
+        if throttle:
+            log.info("pause between jobs : --throttle " + str(throttle))
+            sleep(int(throttle))
```
zmc: This should probably no-op if ``--dry-run`` is passed

dachary:
@zmc updated to not wait if dry_run, thanks for catching this :-)

PR #619, opened 2015-09-08
openstack: add jessie / debian 8 support

dachary:

### DIFF
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -57,6 +57,7 @@ class OpenStack(object):
         'centos-6.5': 'http://cloud.centos.org/centos/6/images/CentOS-6-x86_64-GenericCloud.qcow2',
         'centos-7.0': 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-20150628_01.qcow2',
         'ubuntu-14.04': 'https://cloud-images.ubuntu.com/trusty/current/trusty-server-cloudimg-amd64-disk1.img',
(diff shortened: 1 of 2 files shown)


dachary:
tested with
<pre>
teuthology-lock --lock-many 1 --os-type debian --os-version 8.0 --machine-type openstack
</pre>

teuthology/openstack/__init__.py
```diff
@@ -191,12 +192,12 @@ def cloud_init_wait(self, name_or_ip):
...
-                    if 'Unknown server' in str(e):
-                        continue
-                    else:
-                        raise e
+                    transients = ('Incompatible ssh peer', 'Unknown server')
+                    for transient in transients:
+                        if transient in str(e):
+                            continue
+                    log.exception('cloud_init_wait ' + name_or_ip)
+                    raise e
```
andrewschoen: Would you please ``raise`` instead of ``raise e`` here. Re-raising the named exception obfuscates the traceback a bit.  Notice the line numbers, the ``raise e`` shows the exception happening on that line, which is not true.

```
>>> try:
...     2/0
... except Exception:
...     raise
...
  File "<stdin>", line 2, in <module>
ZeroDivisionError: integer division or modulo by zero
>>> try:
...     2/0
... except Exception as e:
...     raise e
...
  File "<stdin>", line 4, in <module>
ZeroDivisionError: integer division or modulo by zero
```

dachary:
Thanks for educating me, much appreciated :-) Repushed !

PR #620, opened 2015-09-08
Fix bug in using priority 0

zmc:
http://tracker.ceph.com/issues/12995


vasukulkarni:
:+1:

PR #622, opened 2015-09-12
misc: make wait_for_healthy dump pgs if it times out

liewegas:
Signed-off-by: Sage Weil sage@redhat.com

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -873,23 +873,38 @@ def wait_until_healthy(ctx, remote):
     """
     testdir = get_testdir(ctx)
     with safe_while(tries=(900 / 6), action="wait_until_healthy") as proceed:
-        while proceed():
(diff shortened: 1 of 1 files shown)


teuthology/misc.py
```diff
@@ -873,23 +873,38 @@ def wait_until_healthy(ctx, remote):
...
+                    ],
+                    stdout=StringIO(),
+                    logger=log.getChild('health'),
+                )
+                out = r.stdout.getvalue()
+                log.debug('Ceph health: %s', out.rstrip('\n'))
+                if out.split(None, 1)[0] == 'HEALTH_OK':
+                    break
+                time.sleep(1)
+        except:
```
zmc: No bare ``except:`` statements, ever:
https://docs.python.org/2/howto/doanddont.html#except

PR #623, opened 2015-09-14
internal: move buildpackages before install task

dachary:
If a buildpackages task is found, ensure it is always before the install
task because it is intended to produce the packages that will be used by
the install task.

http://tracker.ceph.com/issues/13031 Refs: #13031

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -234,6 +234,12 @@ def check_packages(ctx, config):
 
     If there are missing packages, fail the job.
     """
+    for task in ctx.config['tasks']:
+        if task.keys()[0] == 'buildpackages':
+            log.info("Checking packages skipped because "
+                     "the task buildpackages was found.")
+            return
+
     log.info("Checking packages...")
     os_type = ctx.config.get("os_type")
     sha1 = ctx.config.get("sha1")
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -342,6 +342,51 @@ def push_inventory(ctx, config):
     except Exception:
         log.exception("Error pushing inventory")
 
+BUILDPACKAGES_SWAPPED = 0
+BUILDPACKAGES_OK = 1
+BUILDPACKAGES_REMOVED = 2
+BUILDPACKAGES_NOTHING = 3
+
+def buildpackages_prep(ctx, config):
+    """
+    Make sure the 'buildpackages' task happens before
+    the 'install' task.
+
+    Return:
+
+    BUILDPACKAGES_NOTHING if there is no buildpackages task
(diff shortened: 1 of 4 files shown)


zmc:
This looks like a mistake - the language is about "moving" the task... but it hasn't been merged yet

dachary:
> This looks like a mistake - the language is about "moving" the task... but it hasn't been merged yet

It's being worked on, hence the DNM in the title.

dachary:
Sorry, I confused the two pull requests :-) The idea is to have the buildpackages task implemented in ceph-qa-suite, not in teuthology. This is the teuthology part supporting this task: there is no way around it. My intention was to minimize the teuthology changes to the strict minimum. But maybe I should do things differently ?

dachary:
Note that this change does nothing at all unless there is a buildpackages task somewhere. Reason why I felt it was appropriate to submit it for review even before the buildpackages tasks is complete.

zmc:
So, you say it won't do anything unless a ``buildpackages`` task is present, but it *is* adding it to the list of tasks in ``run.py``

dachary:
@zmc it adds the **internal.buildpackages** task which supports the **buildpackages** task, if present.

teuthology/task/internal.py
```diff
@@ -342,6 +348,50 @@ def push():
...
+    if buildpackages_index is not None and install_index is not None:
+        if buildpackages_index > install_index:
+            log.info('buildpackages moved before the install task')
+            buildpackages = ctx.config['tasks'].pop(buildpackages_index)
+            ctx.config['tasks'].insert(install_index, buildpackages)
+            return BUILDPACKAGES_SWAPPED
+        else:
+            log.info('buildpackages is before the install task')
+            return BUILDPACKAGES_OK
+    elif buildpackages_index is not None and install_index is None:
```
dachary: does nothing... needs fixing

teuthology/task/internal.py
```diff
@@ -342,6 +348,50 @@ def push():
     except Exception:
         log.exception("Error pushing inventory")
 
+BUILDPACKAGES_SWAPPED = 0
+BUILDPACKAGES_OK = 1
+BUILDPACKAGES_REMOVED = 2
+BUILDPACKAGES_NOTHING = 3
+
+def buildpackages(ctx, config):
```
zmc: I was massively confused about the difference between ``buildpackages`` and ``internal.buildbackages``. Would it make sense perhaps to rename the latter to ``internal.buildpackages_prep`` or something similar?

dachary:
@zmc sorry about the confusion, I should have chosen the names more wisely :-( The internal function has been renamed ```buildpackages_prep``` as you suggested

zmc:
@dachary no worries! Sorry for the slow review :)

PR #624, opened 2015-09-14
move the buildpackages task before the install task

dachary:
http://tracker.ceph.com/issues/13031

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -342,6 +342,50 @@ def push_inventory(ctx, config):
     except Exception:
         log.exception("Error pushing inventory")
 
+BUILDPACKAGES_SWAPPED = 0
+BUILDPACKAGES_OK = 1
+BUILDPACKAGES_REMOVED = 2
+BUILDPACKAGES_NOTHING = 3
+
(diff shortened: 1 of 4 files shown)


teuthology/task/internal.py
```diff
@@ -342,6 +342,50 @@ def push():
     except Exception:
         log.exception("Error pushing inventory")
 
+BUILDPACKAGES_SWAPPED = 0
+BUILDPACKAGES_OK = 1
+BUILDPACKAGES_REMOVED = 2
+BUILDPACKAGES_NOTHING = 3
+
+def buildpackages(ctx, config):
```
zmc: Curious... why isn't this implemented as a class-based ``Task`` with https://github.com/ceph/teuthology/blob/master/teuthology/task/__init__.py#L9 ?

dachary: I'll use that class template, thanks ! This is not a task though, just a function shuffling tasks.

dachary:
@zmc ping ?

zmc:
@dachary this looks like a duplicate of #623?

dachary:
@zmc my bad indeed

PR #626, opened 2015-09-14
openstack: upgrade to debian 8.2.0

dachary:


zmc:
It seems wrong that 'debian-8.0' would get you debian 8.1 *or* debian 8.2.

If you only want to keep the latest versions available, would it be reasonable to call it 'debian-8'?

dachary:
If we change 8.0 to 8 we'll have to also change it in other places as well as all ceph-qa-suite jobs that refer to it. We have the same issue with CentOS 7.0 being CentOS 7.1 as well, probably for the same reason. 

I agree with you that there should be a better way to deal with this but I don't feel confident enough to address that problem.

zmc:
Hrm, but CentOs 6.5 is correctly referred to as CentOS 6.5. Kind of a mess.

I suppose you're right that you're not inventing this broken convention :)

PR #631, opened 2015-09-17
suite: fixes a bug that causes the email results job to not be scheduled

andrewschoen:
See: http://tracker.ceph.com/issues/13125


vasukulkarni:
thanks good catch :+1:

PR #633, opened 2015-09-20
openstack: get rid of failed attempt to use ephemerals

dachary:
It's a hack at best.


zmc:
Jenkins passed.

PR #637, opened 2015-09-23
suite: add config.suite_verify_ceph_hash when no gitbuilder

dachary:


zmc:
Should this new logic just be built in to ``get_hash()``?

dachary:
I can certainly do that and add the relevant tests. Do you agree on the principle ? This is a behavior change. It makes sense to me but I don't have the big picture :-)

zmc:
Hrm, actually, I might not agree on it now that I really think about it. If you want behavior to change when using the buildpackages task, you should probably special-case it.

Oh, also: no bare ``except:`` statements. https://docs.python.org/2/howto/doanddont.html#except

dachary:
@zmc I understand your reservations and, to be honest, I was not too happy about the current implementation myself :-) The current semantic of --ceph argument when used with teuthology-suite and its interactions with the gitbuilders, the implicits of the --distro flag and the priority implementation of the install and install.upgrade tasks are very tricky. I'm not sure what would break if this is changed, even slightly. 

I will rework an implementation that makes a special case if the buildpackages task is involved, thanks for your feedback.

dachary:
@zmc that approach works for **teuthology-openstack** and is non-intrusive (I think). I'm happy about it. What do you think ?

zmc:
@dachary I think this is sane. There's at lease one failing test though.

dachary:
@zmc that was a false negative, rebased & repushed to verify

dachary:
@zmc this eludes me completely : when running locally it always succeeds. Any idea ?

dachary:
@zmc works now, cool :-)

PR #638, opened 2015-09-24
openstack: do not rely on gitbuilder.ceph.com by default

dachary:
Make it so the default when using the OpenStack backend is to build the
packages transparently using the OpenStack cluster instead of relying on
http://ceph.com/gitbuilder.cgi.


dachary:
tested to work at http://167.114.234.211:8081/ubuntu-2015-09-24_07:06:48-ceph-disk-pr-6063---basic-openstack/

PR #640, opened 2015-09-26
config: ~/teuthology.yaml check_package_signatures hint

dachary:
http://tracker.ceph.com/issues/13249


zmc:
Could we call this ``check_package_signatures`` and default to ``True``?

dachary:
@zmc I'll do that

dachary:
re-pushed !

PR #641, opened 2015-09-26
config: add ceph_git_url and ceph_qa_suite_git_url

dachary:
The ~/.teuthology.yaml ceph_git_base_url configuration does not allow to
modify the URL of the Ceph repository without also modifying the URL of
the teuthology repository. Although it is frequently needed to point to
an alternate ceph or ceph-qa-suite repository, it is rarely necessary to
point to an alternate teuthology repository.

This is not a blocker: it is enough to mirror the teuthology,
ceph-cm-ansible, ceph-deploy and maybe a few other repositories to
satisfy this requirement. This is however inconvenient because the
exact list of repositories that need to be mirrored is not easily
accessible. In addition, unless the user is careful about updating the
mirrors prior to running teuthology, there is a good chance that an
obsolete version of the repository will be used and this may lead to
problems difficult to diagnose.

The git_ceph_url and git_ceph_qa_suite_url configuration variables are
added to specify the URL of the ceph and ceph-qa-suite repositories
without modifying the git_ceph_base_url value so that all other
repositories retain their default location.

For easier consumption within teuthology and ceph-qa-suite, the
get_git_ceph_url() and get_git_ceph_qa_suite_url() accessors are added
to the config class. They use the user provided value, if available, and
otherwise fallback to constructing the URL with git_ceph_base_url which
is the legacy behavior.

http://tracker.ceph.com/issues/11883 Fixes: #11883

### DIFF
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -131,6 +131,8 @@ class TeuthologyConfig(YamlConfig):
         'automated_scheduling': False,
         'reserve_machines': 5,
         'ceph_git_base_url': 'https://github.com/ceph/',
+        'ceph_git_url': None,
+        'ceph_qa_suite_git_url': None,
         'gitbuilder_host': 'gitbuilder.ceph.com',
         'lab_domain': 'front.sepia.ceph.com',
         'lock_server': 'http://paddles.front.sepia.ceph.com/',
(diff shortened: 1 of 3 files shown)


teuthology/config.py
```diff
@@ -131,6 +131,8 @@ class TeuthologyConfig(YamlConfig):
         'automated_scheduling': False,
         'reserve_machines': 5,
         'ceph_git_base_url': 'https://github.com/ceph/',
+        'ceph_git': None,
+        'ceph_qa_suite_git': None,
```
zmc: These names don't match the docs or the rest of the code

dachary: Ouch ! nice catch, thanks :-)

dachary:
rebased & fixed & repushed !

PR #650, opened 2015-10-08
install: split the upgrade_common function

dachary:
The upgrade_common function implements a non trivial logic that defines
how overrides are applied to the install.upgrade task, as well as the
way upgrades are applied to the desired targets.

The function is split in two:

  upgrade_common which remains the entry point
  upgrade_remote_to_config which encapsulates the logic

This allows other parts of teuthology to obey the same logic by calling
the function instead of replicating it.


zmc:
I'm thrilled about this even if only because you included tests! :)

PR #651, opened 2015-10-09
run_tasks.py: fix Sentry URL

dmick:
I don't know when or why it changed, but the existing URL format, which
uses '/search?q=<id>', fails; what works, by observing the web UI's URL
submission and by testing, is to omit the 'search' part of the path:
'/?q=<id>'


dmick:
BTW I *defy* you to find any documentation of this, or any, web UI URL format.

zmc:
Great catch @dmick! I'd wondered why the links were broken but never found time to look into it.

PR #655, opened 2015-10-13
Add the option to enable or disable epel during install or upgrade

vasukulkarni:
Option to enable_epel during install task to resolve package dependency
using epel repo

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -706,6 +706,54 @@ def _upgrade_deb_packages(ctx, config, remote, debs):
     )
 
 
+
+@contextlib.contextmanager
+def enable_epel(ctx, enable_epel):
+    """
+    Enables epel repo on all rpm host type
+
+    :param ctx: the argparse.Namespace object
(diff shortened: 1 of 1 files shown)


zmc:
Months ago, @andrewschoen made this possible using ansible. Why not use that?

andrewschoen:
@vasukulkarni Yeah, I'd say using the downstream setup playbook here is the right method. Unless you're having some troubles with it I'm not aware of?

vasukulkarni:
@zmc @andrewschoen This is for different issue, downstream.yml is already being used for a very long time, we have issues running in magna lab on rhel due to some package dependency for firefly, hammer or master branch, check 13203 where we ended up adding fcgi to ansible. With enable_epel override the runs will be clean

zmc:
This is what I was talking about earlier:
https://github.com/ceph/ceph-cm-ansible/blob/master/roles/common/tasks/epel.yml

vasukulkarni:
I this this is not required for now, would submit a pr for ansible to install some dependent packages.

PR #657, opened 2015-10-16
adding configuration item to control whether to fetch repo in real time

caibo2014:
So we can use teuthology without network.

Signed-off-by: Bo Cai <cai.bo@h3c.com>


caibo2014:
@zhouyuan 周渊能review下这块代码不

caibo2014:
好的，我清楚了，谢谢。

PR #664, opened 2015-10-23
openstack: resources hint is the max of all hints

dachary:
Exactly one OpenStack resources hint can be included in a given job, as
part of an existing facet. It is error prone because it is sometimes not
trivial to figure out how a given job is composed and if two resources
hint are included only one of them will be taken into account which can
lead to problems difficult to diagnose. Another undesirable side effect
is to artificially increase resources usage. It is easier and more
reliable (from the test maintainer point of view) to increase the
resources of all jobs when a few need more RAM or disk rather than
trying to figure where to write the hints so that they are used by these
jobs and these jobs only.

Instead of being a fixed hint for a given job, the max of all hints
found in each facet is used. For instance, rados/thrash can have a facet
requiring that all jobs are given 3 devices.

cat rados/thrash/cluster/openstack.yaml
  openstack:
    - volumes:
        count: 3

If one task in rados/thrash needs 16GB RAM instead of the default of 8GB
RAM, it can have:

cat rados/thrash/tasks/bigworkunit.yaml

  task:
    - workunit: highmemoryusage.sh
  openstack:
    - machine:
        ram: 16 # GB

And a job composed of rados/thrash/{cluster/openstack.yaml
tasks/bigworkunit.yaml} is aggregated as the max of all resources,
including the default, that is:

  task:
    - workunit: highmemoryusage.sh
  openstack:
    - machine:
        disk: 20 # GB
        ram: 16 # GB
        cpu: 1
      volumes:
        count: 3
        size: 1 # GB

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -487,6 +487,84 @@ test. Login wih the ssh access displayed at the end of the
     ...
     ========= 1 passed in 204.35 seconds =========
 
+Defining instances flavor and volumes
+-------------------------------------
+
+Each target (i.e. a virtual machine or instance in the OpenStack
+parlance) created by the OpenStack backend are exactly the same. By
+default they have at least 8GB RAM, 20GB disk, 1 cpus and no disk
+attached. It is equivalent to having the following in the
+`~/.teuthology.yaml <https://github.com/ceph/teuthology/blob/master/docs/siteconfig.rst>`_ file::
+
+    openstack:
+      ...
+      machine:
+        disk: 20 # GB
+        ram: 8000 # MB
+        cpus: 1
+      volumes:
+        count: 0
+        size: 1 # GB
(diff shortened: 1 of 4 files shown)


dachary:
An example of the new style at https://github.com/ceph/ceph-qa-suite/blob/master/suites/ceph-disk/basic/tasks/ceph-disk.yaml 

* A run with jobs using the new style http://167.114.239.68:8081/ubuntu-2015-10-23_16:37:01-ceph-disk-loic-infernalis---basic-openstack
* A run with a job using the old style http://167.114.239.68:8081/ubuntu-2015-10-23_16:18:37-rados-loic-infernalis---basic-openstack

zmc:
This is great!

PR #669, opened 2015-10-23
openstack: implement OpenStack.set_provider

dachary:
Setting the provider name depending on the OS_AUTH_URL content is
generally useful and moved to the OpenStack base class. There is a need
to cope with public OpenStack providers special cases, even if
temporarily (i.e. no volumes on OVH, no security group on RackSpace
etc.).


dachary:
rebased and repushed

PR #674, opened 2015-10-27
openstack: server show returns a dict or an array

dachary:
The returned json from server show may either be an array of
'Field'/'Value' or a dictionary, depending.


dachary:
See https://bugs.launchpad.net/python-openstackclient/+bug/1510546

PR #677, opened 2015-10-28
openstack: server show must always return array

dachary:
If cliff-tablib is not present, the output of openstack server show is a
dictionary, otherwise it is a list of Field/Value pairs. Since
cliff-tablib is not a required dependency of python-openstack client,
the output format of openstack server show may vary. Add cliff-tablib as
a dependency so that the output format is always the same.

https://bugs.launchpad.net/python-openstackclient/+bug/1510546

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,10 @@ setup(
                       'pyopenssl>=0.13',
                       'ndg-httpsclient',
                       'pyasn1',
+                      # the following may be removed when
+                      # https://bugs.launchpad.net/python-openstackclient/+bug/1510546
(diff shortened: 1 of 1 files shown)


zmc:
Wouldn't we rather have a dict? We could change OpenStack.get_value() to accept dicts easily (in fact I've done this in a branch).

dachary:
@zmc I'm not sure how to make it so we will always have a dict. Do you ?

zmc:
@dachary I don't actually. This is probably the right way to go right now.

PR #680, opened 2015-10-28
misc: reduce misc.sh verbosity

dachary:
Only shows the first 128 characters of the output when the command
succeeds and show all of it only when it fails.


dachary:
@zmc is this better ?

PR #684, opened 2015-10-29
suite: implement git_branch_exists with ls-remote

dachary:


dachary:
@zmc I do not understand why the test for git_ls_remote fails on travis-ci. When I run it locally it is fine. Any idea ?

zmc:
@dachary it looks like the test in question tries to hit the network - that might be the problem, but even if it isn't I'd advocate using mock to avoid doing so

dachary:
@zmc it does not use the network, that's what confuses me. dbeeb67393c2288b4ab8deac89731b7f4d933963 has a fixture creating a local git repository to not be network dependent.

dachary:
@zmc the problem was with git author / email settings, fixed & repushed !

PR #685, opened 2015-10-30
openstack: more robust server deletion

dachary:
When removing a server and the volumes attached to it, the following can
happen:

  * the server goes in ERROR status instead of being deleted
  * if removed before the server is deleted, the volume may fail detach
    and consequently the deletion can fail with "still attached, detach
    volume first"
  * some volumes go in ERROR status instead of being deleted

This will cause problem when and if an instance is assigned the same IP
as an instance in ERROR state. They will both have the same name and
confuse teuthology.

Instead of assuming deletion is a reliable operation, assume it is a
best effort:

   * rename the server and volumes to REMOVE-ME
   * delete the server and then the volumes: ignore failures

The names instances and volumes in error state no longer conflict with
the name of running instances and can be dealt with at a later time,
and kept around for the cloud provider support to investigate.

### DIFF
--- a/teuthology/provision.py
+++ b/teuthology/provision.py
@@ -362,18 +362,13 @@ class ProvisionOpenStack(OpenStack):
         if not self.exists(name_or_id, server_info=server_info):
             return True
         volumes = self.list_volumes(name_or_id, server_info=server_info)
-        if self.get_value(server_info, 'status').lower() == 'error':
-            log.info(
-                "Instance %s is in an error state; skipping volume detachment",
-                name_or_id
-            )
-        else:
-            for volume in volumes:
-                misc.sh("openstack server remove volume %s %s" %
-                        (name_or_id, volume))
-        misc.sh("openstack server delete " + name_or_id)
+        server_id = self.get_value(server_info, 'ID')
+        misc.sh("openstack server set --name REMOVE-ME-" + name_or_id +
+                " " + server_id)
+        misc.sh("openstack server delete --wait " + server_id + " || true")
         for volume in volumes:
-            misc.sh("openstack volume delete " + volume)
+            misc.sh("openstack volume set --name REMOVE-ME " + volume)
+            misc.sh("openstack volume delete " + volume + " || true")
         return True
 
 


dachary:
testing

teuthology/provision.py
```diff
@@ -362,18 +362,12 @@ def destroy(self, name_or_id):
...
-                name_or_id
-            )
-        else:
-            for volume in volumes:
-                misc.sh("openstack server remove volume %s %s" %
-                        (name_or_id, volume))
-        misc.sh("openstack server delete " + name_or_id)
+        server_id = self.get_value(server_info, 'ID')
+        misc.sh("openstack server set --name REMOVE-ME " + server_id)
+        misc.sh("openstack server delete --wait " + server_id + " || true")
```
zmc: I wonder, should ``misc.sh()`` take a ``_raise`` arg that defaults to ``True``?

teuthology/provision.py
```diff
@@ -362,18 +362,12 @@ def destroy(self, name_or_id):
...
-                "Instance %s is in an error state; skipping volume detachment",
-                name_or_id
-            )
-        else:
-            for volume in volumes:
-                misc.sh("openstack server remove volume %s %s" %
-                        (name_or_id, volume))
-        misc.sh("openstack server delete " + name_or_id)
+        server_id = self.get_value(server_info, 'ID')
+        misc.sh("openstack server set --name REMOVE-ME " + server_id)
```
zmc: What are your thoughts on using "REMOVE-ME_" + ``<name>`` or ``<name_or_id>`` here?

teuthology/provision.py
dachary: I can implement that if you want.

dachary:
<pre>
teuthology-suite --machine-type openstack '--verbose' '--ceph' 'hammer'  '--suite' 'dummy' 
....
2015-10-30 23:26:48,141.141 DEBUG:teuthology.misc:openstack server set --name REMOVE-ME-target245207 68126824-763b-467e-98eb-908a7438b7ef
2015-10-30 23:26:49,726.726 DEBUG:teuthology.misc:openstack server delete --wait 68126824-763b-467e-98eb-908a7438b7ef || true
2015-10-30 23:26:56,759.759 INFO:teuthology.lock:unlocked target245207.teuthology
2015-10-30 23:26:56,762.762 INFO:teuthology.run:Summary data:
{description: 'dummy/{all/nop.yaml}', duration: 466.8103280067444, owner: scheduled_ubuntu@teuthology,
  success: true}

2015-10-30 23:26:56,762.762 DEBUG:teuthology.report:Pushing job info to http://localhost:8080/
2015-10-30 23:26:56,802.802 INFO:teuthology.run:pass
2015-10-30 23:28:08,476.476 INFO:teuthology.worker:Success!
</pre>

PR #693, opened 2015-11-03
run: do not lock machines if there are no roles

dachary:
The worker runs the job, even if there are no roles: it is legitimate to
run a job that does not need machines.

Run only locks machines if there are roles and init_tasks is only
populated with tasks that make sense depending on the presence of roles
or not.

Move the internal task added in suite.py to run.py : there is no reason
to not have them with teuthology-run or teuthology-schedule (ansible +
clock).

### DIFF
--- a/scripts/test/script.py
+++ b/scripts/test/script.py
@@ -11,6 +11,6 @@ class Script(object):
         assert out.startswith('usage')
 
     def test_invalid(self):
-        args = (self.script_name, 'INVALID')
+        args = (self.script_name, '--invalid-option')
         with raises(subprocess.CalledProcessError):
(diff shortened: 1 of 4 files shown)


zmc:
I like this - has it been tested?

Also, it appears the unit test needs updating.

dachary:
@zmc I've used during a week or so. Since all current jobs have roles, it does not change anything for them really.

dachary:
@zmc I can't make sense of this failure. It looks like a false negative but it's not (pushed twice and it happened twice). The error relates to argument parsing but ... this commit does not do anything with argument parsing. Any idea you may have would be greatly appreciated :)

PR #694, opened 2015-11-03
openstack: add support for dreamcompute

dachary:


dachary:
http://tracker.ceph.com/issues/13684 is causing problems

PR #698, opened 2015-11-04
openstack: poll every 30 seconds instead of 2

dachary:
No cloud provider boots an instance within 30 seconds, no need to poll
aggressively. Also wait more than a total of 1200 seconds (3000 seconds)
because some providers may take more than 20 minutes to complete the
boot sequence.


zmc:
test this please

PR #699, opened 2015-11-05
teuthology-results: Be aware of all job statuses

zmc:
We used to claim that every job that didn't pass or fail had 'hung'. That was a holdover from the pre-paddles era.

Now we can report 'dead', 'running', 'waiting', and 'queued'.

``teuthology-results`` will now require a running paddles instance, but then again, so does scheduling jobs in the first place.

### DIFF
--- a/teuthology/results.py
+++ b/teuthology/results.py
@@ -175,7 +175,7 @@ def format_job(run_name, job):
     job_id = job['job_id']
     status = job['status']
     description = job['description']
-    duration = int(job['duration'] or 0)
+    duration = seconds_to_hms(int(job['duration'] or 0))
 
     # Every job gets a link to e.g. pulpito's pages
     info_url = misc.get_results_url(run_name, job_id)
@@ -239,6 +239,12 @@ def format_job(run_name, job):
         return email_templates['fail_templ'].format(**format_args)
 
 
+def seconds_to_hms(seconds):
+    (minutes, seconds) = divmod(seconds, 60)
+    (hours, minutes) = divmod(minutes, 60)
+    return "%02d:%02d:%02d" % (hours, minutes, seconds)
+
+
 email_templates = {
     'body_templ': dedent("""\
         Test Run: {name}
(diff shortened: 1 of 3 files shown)


teuthology/results.py
```diff
@@ -36,21 +36,26 @@ def main(args):
...
-    archive_base = os.path.split(archive_dir)[0]
-    serializer = ResultsSerializer(archive_base)
     starttime = time.time()
 
     if timeout:
         log.info('Waiting up to %d seconds for tests to finish...', timeout)
-    while serializer.running_jobs_for_run(name) and timeout > 0:
+
+    reporter = ResultsReporter()
+    while timeout > 0:
```
andrewschoen: Do you believe this condition here to be safe? Should we use safe_while instead?

teuthology/test/test_results.py
```diff
@@ -74,19 +134,19 @@ class TestResultsEmail(object):
     def setup(self):
         config.results_ui_server = "http://example.com/"
         config.archive_server = "http://qa-proxy.ceph.com/teuthology/"
-        self.archive = FakeArchive()
-        self.archive.setup()
-        self.archive_base = self.archive.archive_base
 
     def teardown(self):
```
andrewschoen: Why not just remove this method?

zmc: Just an oversight - will do that

teuthology/results.py
zmc: ``timeout`` is a value that doesn't change. This block will only run if we want ``teuthology-results`` to block for a time waiting for the run to finish.

zmc:
Example of output: http://paste2.org/NFkENFBY

teuthology/results.py
andrewschoen: Cool, just wanted to check.

andrewschoen:
@zmc that output looks much better! Perhaps we should format ``time`` to be H:M:S instead of just seconds? An extra line break between the jobs in the ``running`` and ``waiting`` sections would make that read easier as well, IMHO.

zmc:
@andrewschoen good suggestions; updated!

PR #703, opened 2015-11-10
Unpin os-client-config

dachary:
The bug found in 1.10.1 was fixed and version 1.10.2 is now current.

https://bugs.launchpad.net/os-client-config/+bug/1513919


dachary:
Testing with
<pre>
teuthology-openstack --verbose --key-name myself  --key-filename ~/Downloads/myself --suite dummy
</pre>

PR #704, opened 2015-11-13
DNM - Fix yum installs on firefly

wusui:
Add -x librados2-devel and -x librbd1-devel parameters to ceph yum
installs.  Make sure that librbd1, librados2, and libcephfs1 are
not installed prior to installing the list of ceph packages.

Signed-off-by: Warren Usui warren.usui@inktank.com

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -261,7 +261,7 @@ def _update_rpm_package_list_and_install(ctx, remote, rpm, config):
         args=[
             'sudo', 'yum', 'clean', 'all',
         ])
-
+    remote.run(args = ['sudo', 'yum', 'remove', 'librados2', 'libcephfs1', '-y',])
     ldir = _get_local_dir(config, remote)
     for cpack in rpm:
         pkg = None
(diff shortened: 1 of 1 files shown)


ktdreyer:
(Note: this will break runs on Hammer and above)

yuriw:
@wusui does this wip has to be marked DNM ?

branto1:
I'd suggest splitting this into two patches:

- one that adds (we really need that one to be sure that the packages will install correctly on centos/rhel and fedora)

`remote.run(args = ['sudo', 'yum', 'remove', 'librados2', 'libcephfs1', '-y',])`

- one that adds the -x workarounds and can be reverted on its own at some point (it should no longer be needed by next week, hopefully starting from Monday)

btw: Is there any way to add options for yum to the ceph install task? If yes then we could skip the second patch, modify the .yaml files to add the -x options to the yum calls and skip the second patch altogether.

dillaman:
FYI -- removing the librados and librbd packages will cause qemu-kvm to be removed as well.  This will cause several RBD test cases to fail due to QEMU being yanked.

teuthology/task/install.py
```diff
@@ -261,7 +261,7 @@ def _update_rpm_package_list_and_install(ctx, remote, rpm, config):
         args=[
             'sudo', 'yum', 'clean', 'all',
         ])
-
+    remote.run(args = ['sudo', 'yum', 'remove', 'librados2', 'libcephfs1', '-y',])
```
zmc: Why is this necessary? We remove ceph packages during ansible:
https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/vars/yum_systems.yml#L15-L16

zmc:
Please put the tracker URL in the commit message.
If this breaks hammer onward, we can't merge it can we?

teuthology/task/install.py
```diff
@@ -274,16 +274,16 @@ def _update_rpm_package_list_and_install(ctx, remote, rpm, config):
                 args = ['if', 'test', '-e',
                         run.Raw(pkg), run.Raw(';'), 'then',
                         'sudo', 'yum', 'remove', pkg, '-y', run.Raw(';'),
-                        'sudo', 'yum', 'install', pkg, '-y',
+                        'sudo', 'yum', 'install', pkg, '-y', '-x', 'librbd1-devel', '-x', 'librados2-devel',
                         run.Raw(';'), 'fi']
             )
         if pkg is None:
-            remote.run(args=['sudo', 'yum', 'install', cpack, '-y'])
+            remote.run(args = ['sudo', 'yum', 'install', cpack, '-y', '-x', 'librbd1-devel', '-x', 'librados2-devel',])
```
zmc: There shouldn't be spaces around the `=` here

tmuthamizhan:
This is a "DNM" branch. we have fixed the install task to make it work for the firefly branch, inorder to get some regression runs on firefly for centos and rhel to be able to sign off ceph v0.80.11 release. so we would continue to use this branch for 'firefly' only but NOT merge the fix to teuthology master.

kshtsk:
Seems like this is very old, the last record is 4 years ago. The original ticket (https://tracker.ceph.com/issues/13622) is close 3 years ago. Wiping it out.

PR #709, opened 2015-11-18
Remove epel packages during downstream testing if present on the node

vasukulkarni:
Remove epel packages during downstream testing if present on the node and install these packages using the downstream repo

Dependent on https://github.com/ceph/ceph-cm-ansible/pull/167

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -760,6 +760,8 @@ def rh_install_pkgs(ctx, remote, installed_version):
     """
     pkgs = ['ceph-deploy']
     rh_version_check = {'0.94.1': '1.3.0', '0.94.3': '1.3.1'}
+    log.info("Remove any epel packages installed on node %s", remote.shortname)
+    remote.run(args=['sudo', 'yum', 'remove', run.Raw("leveldb xmlstarlet fcgi"), '-y'],check_status=False)
     for pkg in pkgs:
         log.info("Check if ceph-deploy is already installed on node %s", remote.shortname)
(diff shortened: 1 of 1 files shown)


zmc:
ceph/ceph-cm-ansible#167 is marked as DNM, so I'll make this DNM too

zmc:
Whoops, sorry if I stepped on your toes there

vasukulkarni:
@zmc  no problem, will correct

zmc:
Will this cause problems on CentOS/Fedora? Should we make it only affect RHEL?

teuthology/task/install.py
```diff
@@ -760,6 +760,8 @@ def rh_install_pkgs(ctx, remote, installed_version):
     """
     pkgs = ['ceph-deploy']
     rh_version_check = {'0.94.1': '1.3.0', '0.94.3': '1.3.1'}
+    log.info("Remove any epel packages installed on node %s", remote.shortname)
+    remote.run(args=['sudo', 'yum', 'remove', run.Raw("leveldb xmlstarlet fcgi"), '-y'],check_status=False)
```
zmc: Why is this being removed if the PR it depends upon is adding it specifically for RHEL?

vasukulkarni:
@zmc: those os should be safe, this will only affect downstream runs(only rhel for now)

teuthology/task/install.py
vasukulkarni: we have  packages with different names in upstream and downstream, so during downstream testing it removes the upstream packages present and uses the one from the downstream repo,

PR #711, opened 2015-11-21
suite: do not cancel --dry-run if double --verbose

dachary:
This behavior is confusing and will lead to unexpected suite scheduling
if the --verbose flag is in the command line twice by accident.

http://tracker.ceph.com/issues/13854 Fixes: #13854


zmc:
``-vv --dry-run`` is supposed to call ``teuthology-schedule --dry-run``. Are you saying ``teuthology-schedule`` is not respecting ``--dry-run`` ?

dachary:
I misunderstood. All is well then, sorry for the noise.

PR #713, opened 2015-11-23
prune: Process log dirs in order of ctime

zmc:
Most recent first, of course.


dmick:
Have no memory of this.  Should we revive it?

PR #715, opened 2015-11-24
suite: use --subset X/Y as hints

dachary:
http://tracker.ceph.com/issues/13855


athanatos:
I'm a bit confused, iirc, the mincyclicity argument to _build_matrix was supposed to ensure that chopping the output into <outof> pieces retained the property.  I'll take another look after a good night's sleep :)

dachary:
The problem is that dividing the resulting matrix into outof slices may not obey the minscanlen() constraint. Maybe the right fix would be that generate_combinations detects when that happens and throw an error ?

PR #717, opened 2015-11-25
openstack: enforce hints as an array

dachary:
All OpenStack hints have been converted to arrays (hammer, infernalis,
master), raise an error if a hint is added old style to avoid mixing
style.

http://tracker.ceph.com/issues/13581 Fixes: #13581


dachary:
stupid duplicate

PR #718, opened 2015-11-25
misc: get_pkg_type to infer the package time from the os_type

dachary:

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1218,6 +1218,11 @@ def get_system_type(remote, distro=False, version=False):
         return "rpm"
     return system_value
 
+def get_pkg_type(os_type):
+    if os_type in ('centos', 'fedora', 'opensuse', 'rhel', 'sles'):
+        return 'rpm'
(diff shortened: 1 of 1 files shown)


dachary:
@zmc there is a lot of duplication and variations in naming dist, distro, types etc. None of the existing function I could find has a way to translate the os_type into the distribution type. This is useful when trying to figure out just by reading the yaml file, what the distribution type is (as opposed to asking an already provisioned target about it).

My incentive by adding this function near a function that does almost exactly the same thing but in a very different context is that it does not go unnoticed when someone brave enough finds time to unify the naming conventions.

This is to be used by the buildpackages tasks which as the same function, only internaly, which is bad.

zmc:
@dachary ``misc.get_system_type()`` does something similar; maybe to avoid a little repetition that function could use your new ``misc.get_pkg_type()`` ?

dachary:
@zmc I don't see how that would work because os_type is different from ```lsb_release', '-rs'``` ?

PR #733, opened 2015-12-02
openstack: use OpenStackInstance in TeuthologyOpenStack

dachary:
The setup() function of TeuthologyOpenStack creates an OpenStackInstance
datamember to cache results and avoid running numerous duplicate
commands.


zmc:
Awesome!

PR #736, opened 2015-12-02
Ignore few se denials in log parsing

vasukulkarni:
dmidecode intermittently appears in denial list causing the test to fail

### DIFF
--- a/teuthology/task/selinux.py
+++ b/teuthology/task/selinux.py
@@ -6,6 +6,7 @@ from cStringIO import StringIO
 from teuthology.exceptions import SELinuxError
 from teuthology.misc import get_archive_dir
 from teuthology.orchestra.cluster import Cluster
+from teuthology.orchestra import run
 from teuthology.lockstatus import get_status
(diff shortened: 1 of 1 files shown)


zmc:
This should include a bugzilla URL in both the commit message and a comment

vasukulkarni:
Fixed and repushed

zmc:
I still want URLs in both the commit message and a comment, please.

Seems like we should have a BZ for the chronyd issue, no?

vasukulkarni:
@zmc I have added the url in both commit and in comments, there is only tracker for chronyd which was originally seen on centOS not a bz.

zmc:
Thanks @vasukulkarni 

In the future if we have to add more of these, I'd like to do it in a nicer way. And put the comments *before* the code in question. But I won't block on those issues now so this doesn't get stalled for another month :)

vasukulkarni:
thanks @zmc :) sure i think it makes sense to read comments first.

PR #738, opened 2015-12-07
amendment misunderstood lock server and results server address

caibo2014:
In fact, results server and lock server are function of the paddles, written pulpito is misleading, although pulpito and paddles always installed on one machine.


caibo2014:
@zhouyuan

caibo2014:
@zhouyuan 看下这个PR呗，谢谢。#_#

PR #740, opened 2015-12-07
bootstrap: fixes for openSUSE

smithfarm:
When a package is found not to exist, we need to also check if there
is another package installed that provides a capability of the same
name.

Also, the python-devel package must be installed, or the pip phase
fails.

http://tracker.ceph.com/issues/14002 Fixes: #14002

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -66,9 +66,11 @@ Linux)
 	fi
 	;;
     "openSUSE project"|"SUSE LINUX")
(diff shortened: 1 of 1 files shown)


smithfarm:
Manually tested on openSUSE 13.2 - without this patch, bootstrap fails. With this patch, it succeeds.

zmc:
Thanks! Looks good!

PR #741, opened 2015-12-07
Stop hardcoding apt-mirror's IP

zmc:
We will point to apt-mirror.sepia.ceph.com, which is
publicly-accessible.


zmc:
Depends on https://github.com/ceph/ceph-cm-ansible/pull/175

PR #749, opened 2015-12-15
dnsmasq is not actually restarting

dachary:
The init.d fres, but does nothing.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>


dachary:
@robbat2 FYI

zmc:
Ouch, that's weird. Might be good to file a bug against the distro you're seeing this on. I won't block the PR on that though.

dachary:
@zmc this has not be confirmed but it does not do any harm either

PR #751, opened 2015-12-15
openstack: keep image private

dachary:
Ensure a newly created image is private to this tenant by default,
regardless of whatever the openstack defaults are.

Signed-off-by: Robin H. Johnson <robin.johnson@dreamhost.com>


dachary:
@robbat2 FYI

PR #752, opened 2015-12-15
openstack: switch the default json output to dictionary

dachary:
The python-openstackclient output may either be a list of Value/Field or
a dictionary. Use the dictionary instead of the Value/Field form
because it is easier to deal with. The former was chosen because of a
bug that made it impossible to control the output. The bug still exists
but is believed to be rare enough to ignore. Should it happen, a message
recommending a corrective action and a comment on:

https://bugs.launchpad.net/python-openstackclient/+bug/1510546

is displayed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,10 +55,6 @@ setup(
                       'pyopenssl>=0.13',
                       'ndg-httpsclient',
                       'pyasn1',
-                      # the following may be removed when
-                      # https://bugs.launchpad.net/python-openstackclient/+bug/1510546
-                      # is resolved
-                      'cliff-tablib', # required to get a Field/Value output from openstack server show
                       'python-openstackclient',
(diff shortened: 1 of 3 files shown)


zmc:
Looks good, but you misspelled "dictionary" :)

dachary:
@zmc reppushed with not doubble lettters this timme ;-)

zmc:
I was about the merge this but then realized it might break automated testing everywhere without manual intervention. Holding off for now.

dachary:
@zmc indeed: upgrading to a version of teuthology with this requires **pip uninstall cliff-tablib** as well

zmc:
@dachary on the other hand, this will only affect openstack deployments; I can easily make the change in our ovh cluster right before merging this, so I'll do that.

PR #756, opened 2015-12-17
openstack: support CentOS 7.1

dachary:
It uses the same cloud image as centos-7.0 because it already was 7.1
anyway, only misnamed 7.0 because changing it is inconvenient.

### DIFF
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -182,6 +182,7 @@ class OpenStack(object):
     image2url = {
         'centos-6.5': 'http://cloud.centos.org/centos/6/images/CentOS-6-x86_64-GenericCloud-1508.qcow2',
         'centos-7.0': 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1508.qcow2',
+        'centos-7.1': 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1508.qcow2',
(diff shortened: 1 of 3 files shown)


dachary:
@zmc https://github.com/ceph/ceph-qa-suite/commit/d4d3787609f351b7efa0d52120858d090d1cf80e replaces 7.0 with 7.1 and requires this for OpenStack to run

dachary:
@yuriw FYI

teuthology/openstack/__init__.py
```diff
@@ -183,6 +183,7 @@ class OpenStack(object):
     image2url = {
         'centos-6.5': 'http://cloud.centos.org/centos/6/images/CentOS-6-x86_64-GenericCloud-1508.qcow2',
         'centos-7.0': 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1508.qcow2',
+        'centos-7.1': 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1508.qcow2',
```
zmc: Why not at least use:
http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1510.qcow2
?

dachary: @zmc because it has not been tested yet.

zmc: Ok. Feels weird to point '7.0' and '7.1' to the same image... but it's not like centos.org is giving us a ton of choice here.

dachary: And 7.0 was already pointing to 7.1 to begin with.

dachary:
@zmc repushed & rebased

PR #759, opened 2015-12-18
Block requests==2.9.0

zmc:
Because of https://bugs.launchpad.net/nova/+bug/1526413


dmick:
ah, we live and die by upstream...

Reviewed-by: Dan Mick <dmick@redhat.com>

PR #763, opened 2016-01-05
openstack: nuke must use ID and not name

dachary:
Otherwise it will fail if there are multiple instances by the same name.

http://tracker.ceph.com/issues/14238 Fixes: #14238

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -370,7 +370,7 @@ def synch_clocks(remotes):
 
 
 def stale_openstack(ctx):
-    targets = dict(map(lambda i: (i['Name'], i),
+    targets = dict(map(lambda i: (i['ID'], i),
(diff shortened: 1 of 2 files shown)


dachary:
@andrewschoen testing with a real use case

andrewschoen:
@dachary I see that it did clean up those stale nodes in the dashboard. It might be beneficial to write a unit test validating that this fixes the "multiple instances with the same name" bug though. Thoughts?

dachary:
@andrewschoen the code itself has no issue with multiple instances with the same name. It's openstack that fails because it can't show details for more than one instance.

dachary:
@andrewschoen I've fixed the existing unit tests so they reflect the new assumption, but it's mostly about consistency for the reader because the logic does not change much.

PR #764, opened 2016-01-05
openstack: flock timeout on openstack server create

dachary:
The global lock used to run openstack server create must timeout
otherwise it may hang forever. The timeout is set to 8h. Assuming one
server create --wait takes ~1mn to return, that should be more than
enough to absorbe a burst of 200 jobs provisioning instances at the same
time.

The goal here is to ensure the queue does not stay blocked forever, not
to timeout jobs because they cannot get a server in a given delay.

http://tracker.ceph.com/issues/14243 Fixes: #14243


andrewschoen:
LGTM. I think we may want to eventually make that timeout value configurable though.

PR #765, opened 2016-01-07
packaging.GitbuilderProject: support the wait_for_package feature

andrewschoen:
This fixes: http://tracker.ceph.com/issues/14125

When trying to fetch a package version, GitbuilderProject will now wait and retry for 2 1/2 minutes if the status_code returned is not a 200 and the config option ``wait_for_package: true`` is set for either the install task or install.upgrade.

The 2 1/2 minute wait time is fairly arbitrary, I'm open to better suggestions there. We could also make it a teuthology config option (in teuthology.yaml) if we wanted to get fancy.

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -428,7 +428,16 @@ def _get_response(url, wait=False, sleep=15, tries=10):
     with safe_while(sleep=sleep, tries=tries, _raise=False) as proceed:
         while proceed():
             resp = requests.get(url)
-            if resp.ok or not wait:
+            if resp.ok:
+                log.info('Package found...')
+                break
+
+            if not wait:
+                log.info(
+                    'Package is not found at: %s (got HTTP code %s)...',
+                    url,
+                    resp.status_code,
+                )
                 break
 
             log.info(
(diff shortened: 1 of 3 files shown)


teuthology/packaging.py
```diff
@@ -422,6 +424,30 @@ def _get_config_value_for_remote(ctx, remote, config, key):
...
+            if not wait:
+                log.info(
+                    'Package is not found at: %s (got HTTP code %s)...',
+                    url,
+                    resp.status_code,
+                )
+                break
+
+            log.info(
+                'Package not there yet (got HTTP code %s), waiting...',
```
alfredodeza: if `wait` is True and we never are able to retrieve the package it would be useful to know the `url` here.

andrewschoen: Yeah, I could. The teuthology logs already print out the url before these would hit though. And example would look like:

```
Looking for package version: http://gitbuilder.ceph.com/ceph-deb-trusty-x86_64-basic/ref/hammer/version
Package not there yet (got HTTP code 404), waiting...
Package not there yet (got HTTP code 404), waiting...
Package not there yet (got HTTP code 404), waiting...
... (snip)

    raise VersionNotFoundError(url)
VersionNotFoundError: Failed to fetch package version from http://gitbuilder.ceph.com/ceph-deb-trusty-x86_64-basic/ref/hammer/version
```

dmick:
lgtm.

PR #768, opened 2016-01-12
Use remote.inventory_info['os_version']

zmc:
... instead of remote.os.version.


zmc:
test this please

PR #769, opened 2016-01-12
valgrind.supp: suppress helgrind warnings for external libraries

dillaman:
Fixes: #14163
Signed-off-by: Jason Dillaman <dillaman@redhat.com>


dillaman:
verified correct results: http://qa-proxy.ceph.com/teuthology/jdillaman-2016-01-12_16:46:19-rbd-wip-14163-jewel---basic-multi/26127/teuthology.log

PR #773, opened 2016-01-21
fix a --subset bug that misses facets in some cases

dachary:


dachary:
@zmc this is rebased and tox -e flake8 has no issue. The tostr() helper is added to support __str__() because it is implemented recursively with an additional depth argument to please the eye.

Note that I also include a workaround for a pip 8.0.0 bug preventing bootstrap to work, so that automated checks work. It showed up yesterday and I'm not sure what's the proper way to deal with it.

dachary:
@zmc ping ?

zmc:
@dachary hmm, I'd added a comment or two in the other PR before it was closed and now I can't find it. I remember that one of my issues was the strange ``tostr()`` methods - why not use ``__str__()`` since it's a standard?

zmc:
Also, the bootstrap fix shouldn't be in here - it should be in a separate PR (I'm attempting to fix that and similar issues in #775)

dachary:
@zmc The tostr() helper is added to support str() because it is implemented recursively with an additional depth argument to please the eye.

dachary:
@zmc I removed the extra pip 8 related bug fix.

zmc:
@dachary hrm, ok. Could you rebase this on master so unit tests can execute properly?

dachary:
@zmc rebased & repushed !

dachary:
@zmc fixed the failed test (because a prototype of the test helper changed in december). Should be good now !

PR #777, opened 2016-01-26
Map locking failures to 'dead' status

zmc:
Also, add a specific exception for quota-exceeded errors. This will help us identify general OpenStack breakage sooner and more easily.

This is http://tracker.ceph.com/issues/14515


dmick:
other than one comment, lgtm

zmc:
Good catch! Repushed

PR #745, opened 2015-12-11
misc: ignore some unusable block devices

robbat2:
If a block device is read-only, or is a cloud-init datasource, we should
not change the content on it!

This change detects most documented datasources as of 2015/12/10 from
http://cloudinit.readthedocs.org/en/latest/topics/datasources.html

It cannot detect datasources that are only detectable based on files on
volume (eg config-drive v1)

Signed-off-by: Robin H. Johnson <robin.johnson@dreamhost.com>

PR #780, opened 2016-01-27
Revert "misc: ignore some unusable block devices"

zmc:
Reverts ceph/teuthology#745


zmc:
http://tracker.ceph.com/issues/14532

PR #782, opened 2016-02-03
Task timing

zmc:


zmc:
Tested here:
http://pulpito.ceph.com/zack-2016-02-04_14:56:04-smoke-master---basic-multi/

If you visit the log link for each job and look at the parent dir, timing.yaml is present and correctly-formatted

zmc:
http://tracker.ceph.com/issues/14655

PR #783, opened 2016-02-04
Move sentry event submission to sentry module

zmc:
This is just cleanup.


kshtsk:
@zmc do we still need this?

PR #786, opened 2016-02-11
Support all types of nodes for selinux

vasukulkarni:
By default vps nodes are disabled for selinux, we should enable them it there as well.


zmc:
I'd like to see some test coverage on this :)

vasukulkarni:
Tests are running here, will take some time for it to run: http://pulpito.ceph.com/vasu-2016-02-11_09:35:28-selinux-v10.0.3---basic-vps/

vasukulkarni:
closing this because somehow the vps nodes seems to have issues when testing with selinux, the osd path in vm's internally comes up as a long uuid instead of /var/lib/ceph/... causing lots of unlabeled selinux issues

PR #787, opened 2016-02-17
suite.py: tell you that the suite you asked for doesn't exist

dmick:
I know it's not in fashion to issue useful error messages, but...


dmick:
seriously..."Invalid path seen"?...

dmick:
Sigh.  This breaks the kind-of-amazing tests.

PR #789, opened 2016-02-18
Add ceph_ansible task

zmc:


zmc:
Regression testing here:
http://pulpito.ceph.com/zack-2016-02-18_15:01:21-smoke-master---basic-vps/

zmc:
Ansible execution is passing. Looks like no regressions.

dmick:
One of the suite runs failed, but see http://tracker.ceph.com/issues/14813; I think it has nothing to do with this code.

LGTM as a start.

PR #790, opened 2016-02-20
Add whitelist option and fix for cephtest denials

vasukulkarni:
http://tracker.ceph.com/issues/14244

Fixes 14244 and adds whitelist option to ignore known future denials using overrides

### DIFF
--- a/teuthology/task/selinux.py
+++ b/teuthology/task/selinux.py
@@ -20,6 +20,22 @@ class SELinux(Task):
     must first be enabled and the filesystem must have been labeled.
 
     On teardown, also checks the audit log for any denials.
+    By default selinux will ignore few known denials(listed below). The test
+    will fail for any other denials seen in audit.log. For the test not to
+    fail for other denials one can add the overrides with appropriate escapes
+    overrides:
+       selinux:
+         whitelist:
+         - 'name="cephtest"'
+         - 'dmidecode'
+         - 'comm="logrotate"'
+         - 'comm="idontcare"'
+
+    Known denials which are ignored:
+       comm="dmidecode"
+       chronyd.service
+       name="cephtest"
+
 
     Automatically skips hosts running non-RPM-based OSes.
     """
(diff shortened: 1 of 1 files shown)


teuthology/task/selinux.py
```diff
@@ -91,8 +100,13 @@ def get_denials(self):
         all_denials = dict()
         ignore_known_denials = '\'\(' + 'comm="dmidecode"'
```
zmc: This should probably be a ``list``. You can then use ``list.join()`` in combination with ``str.format()`` to assemble the regex

teuthology/task/selinux.py
```diff
@@ -20,6 +20,15 @@ class SELinux(Task):
     must first be enabled and the filesystem must have been labeled.
 
     On teardown, also checks the audit log for any denials.
+    By default the test will fail for any denials seen in audit.log, for the
```
zmc: This is actually not quite true. Since you're providing a default whitelist (which appears to be mandatory), it won't fail for *any* by default.

teuthology/task/selinux.py
vasukulkarni: I will use list here and flatten it to string using join.

teuthology/task/selinux.py
vasukulkarni: The wording could be wrong, whitelist is still optional like log-whitelist, so that's why i meant by default it may fail unless included in se-whitelist.

zmc: Yeah, I think it's just a matter of improving the wording. Just please document the defaults and note that they are always used regardless of if se-whitelist is passed in via a config.

liewegas:
@branto1 can you look at this?

branto1:
@liewegas those three look ok to me to filter out, we do not run chronyd (afaik, it is run teuthology), the cephtest denials are caused by the way teuthology stores its (log) files and looking at the bugzilla the dmidecode denials seem to be a rhel issue

teuthology/task/selinux.py
```diff
@@ -20,6 +20,22 @@ class SELinux(Task):
...
+    with appropriate escapes
+    overrides:
+       selinux:
+         se-whitelist:
+         - 'name="cephtest"'
+         - 'dmidecode'
+         - 'comm="logrotate"'
+         - 'comm="idontcare"'
+
+    Known denails which is ignored are:
```
zmc: Typo :-/

zmc: Also "which are ignored"

teuthology/task/selinux.py
```diff
@@ -20,6 +20,22 @@ class SELinux(Task):
     must first be enabled and the filesystem must have been labeled.
 
     On teardown, also checks the audit log for any denials.
+    By default selinux will ignore few known denials(listed below), the test will fail for any other
```
zmc: [...] "a few known denials (listed below). The test" [...]

teuthology/task/selinux.py
```diff
@@ -20,6 +20,22 @@ class SELinux(Task):
     must first be enabled and the filesystem must have been labeled.
 
     On teardown, also checks the audit log for any denials.
+    By default selinux will ignore few known denials(listed below), the test will fail for any other 
+    denials seen in audit.log, for the test not to fail for other denials one can add the overrides
```
zmc: [...] "denials seen in audit.log. For the test" [...]

zmc: Also these lines should be wrapped to 80chars

zmc:
Since the option already goes under ``selinux``, shouldn't we just call it ``whitelist`` ?

vasukulkarni:
@zmc comments should be fixed now

zmc:
Thanks @vasukulkarni !

PR #794, opened 2016-02-27
openstack: add --wait to block until the suite finishes

dachary:


zmc:
Is the wait feature fully implemented?

dachary:
@zmc it is implemented in teuthology-suite. Adding the option here in teuthology-openstack means it will be transparently passed to teuthology-suite. That's why it looks so magical ;-)

zmc:
Ah of course! Thanks for jogging my memory.

PR #797, opened 2016-02-27
openstack: allow override of openstack:clone:

dachary:
When running from a wip teuthology branch, it is inconvenient to tweak
~/.teuthology.yaml. Add --teuthology-git-url and --teuthology-branch to
take precendence and override the defaults.


dachary:
rebased & repushed

PR #798, opened 2016-02-27
openstack: encode instance name with the full IP

dachary:
Using only the last two numbers is problematic when mixing multiple
OpenStack clusters.

### DIFF
--- a/teuthology/openstack/openstack-centos-6.5-user-data.txt
+++ b/teuthology/openstack/openstack-centos-6.5-user-data.txt
@@ -3,7 +3,7 @@ bootcmd:
  - echo nameserver {nameserver} | tee /etc/resolv.conf
  - echo search {lab_domain} | tee -a /etc/resolv.conf
  - sed -ie 's/PEERDNS="yes"/PEERDNS="no"/' /etc/sysconfig/network-scripts/ifcfg-eth0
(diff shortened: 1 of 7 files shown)


dachary:
@zmc this is a non-backward compatible change. Before applying it, the DNS of sepia.ovh must be updated so that target064003 is a CNAME of target158069064003 which is the name using the full IP instead of the last two digits ( 158.69.64.3 ). By doing so we can update the DNS whenever convenient. And then this patch can be merged and when the update happens, all should be fine.

I think this is an important change because it allows us to cluster together any number of OpenStack tenants with no risk of name collisions.

What do you think ?

zmc:
@dachary apologies for not responding earlier. I did see this and I agree that we should do it. I just haven't had time to look into what the DNS migration will take. I imagine it'd be doable with a shell script?

dachary:
@zmc yes, a shell script identical to the one your presumably used to populate the DNS initially, only using the full IP should do fine.

dachary:
ping ?

smithfarm:
We (SUSE) are very keen on getting suites to run on openSUSE targets in OpenStack. This work is going forward against Loic's openstack branch [1] so we are not blocked. But it would be very nice if that could be merged into master as soon as possible.

[1] https://github.com/dachary/teuthology/tree/openstack

kshtsk:
@smithfarm seems like this is a very old PR with conflicts for fixing some dns thing for teuthology-openstack. Can we just close this PR as outdated?

kshtsk:
Closing this PR in favor of #1310

PR #803, opened 2016-02-28
bootstrap: install python-libvirt from package

dachary:
Otherwise Ubuntu 14.04 will try to install python-libvirt 1.3.X from
pypi and fail because only libvirt-devel 1.2.x is packaged.

http://tracker.ceph.com/issues/14917 Fixes: #14917

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -16,7 +16,7 @@ case "$(uname -s)" in
 Linux)
     case "$(lsb_release --id --short)" in
     Ubuntu|Debian)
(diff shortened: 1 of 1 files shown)


zmc:
I'm skeptical of this issue:

* teuthology.front runs 14.04 and has ``libvirt-python==1.3.2`` installed in the main virtualenv used by workers
* PyPI still has ``libvirt-python<1.3``.
* This won't work anyway because we don't use the system site-packages anymore.

dachary:
fixed already

PR #807, opened 2016-03-04
repo_utils: Call git set remote-url

zmc:
This will cause git repos on-disk to be updated to point at the correct
remote based on config.ceph_git_base_url


dachary:
no tests ?

PR #810, opened 2016-03-06
pin gevent 1.0.2

dachary:
Signed-off-by: Loic Dachary <loic@dachary.org>

PR #812, opened 2016-03-07
Unpin gevent and avoid patching subprocess

zmc:
Recently, gevent updated to 1.1.0, breaking a few things like misc.sh().
PR #810 pinned gevent to 1.0.2, but I wanted to properly investigate the
issue to not be stuck on an old version. Some git bisects later, I
discovered that gevent began patching the subprocess module by default
in:

https://github.com/gevent/gevent/commit/fc73e9afa34d811ede5d62531025d2e3876b8c01

Unpinning gevent and also telling it not to patch subprocess fixes the
issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
         'Topic :: System :: Filesystems',
     ],
     install_requires=['setuptools',
-                      'gevent == 1.0.2',
+                      'gevent',
                       'MySQL-python == 1.2.3',
                       'PyYAML',
                       'argparse >= 1.2.1',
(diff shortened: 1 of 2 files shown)


zmc:
More details about the problem:
http://tracker.ceph.com/issues/14990#note-3

dmick:
/me shakes head in fear at monkeypatching, but agrees this is a better solution

zmc:
For some level of posterity, the ``dns=False`` comes from:
https://github.com/ceph/teuthology/commit/99ac6b0b3e7e6016dbc354b268e8aa90a3af2b3e

dachary:
@zmc @dmick thanks a lot for looking into this !

PR #814, opened 2016-03-08
install task: Make package lists configurable

zmc:
This is aimed at addressing:
http://tracker.ceph.com/issues/15014

That will also require a suite change to apply a different package list.

There's also a bit of cleanup here.

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -6,6 +6,7 @@ import logging
 import time
 import os
 import subprocess
+import yaml
 
 from teuthology.config import config as teuth_config
 from teuthology import misc as teuthology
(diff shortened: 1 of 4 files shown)


teuthology/task/install.py
```diff
@@ -1179,6 +1116,28 @@ def task(ctx, config):
...
+
+    Debug packages may optionally be installed:
+
+    overrides:
+      install:
+        ceph:
+          debuginfo: true
+
+
+    Default package lists may be overridden:
```
gregsfortytwo: So if we override the list, we have to replace the contents wholesale?
So the full fix for the bug will include setting these overrides in ceph-qa-suite by copying the entire package list and then updating them?

zmc: That's correct, yes

teuthology/task/install.py
```diff
@@ -577,6 +529,23 @@ def remove_sources(ctx, config):
             p.spawn(remove_fn, remote, project)
 
 
+def get_package_list(config):
+    debug = config.get('debuginfo', False)
+    project = config.get('project', 'ceph')
+    yaml_path = os.path.join(os.path.dirname(__file__), 'packages.yaml')
```
vasukulkarni: packages.yaml should exist in qa-suites, so that the construction can be /hammer/packages.yaml or similar and we dont have to overrides lots of install tasks for previous branches.

zmc: This PR makes that possible. But we still need a default somewhere.

vasukulkarni: yeah so the default should  be per branch and should exist in ceph-qa-suites. since teuthology isn't branched

zmc: teuthology doesn't currently depend on ceph-qa-suite, and what you're proposing would change that.

vasukulkarni: I dont understand, I think we want to make it easy for install task to be configurable per branch, keeping this file here will only make it difficult again when it can easily be in qa-suites and its easier for one to configure it per ceph branch, Is there anything wrong with that approach?

zmc: The mere presence of this file, which avoids teuthology having a hard dependency on ceph-qa-suite, is not problematic. This feature allows an alternate package list to be provided; one way to do that would be shipping one in whichever ceph-qa-suite branches need it. Currently, as I understand it, that is just jewel and master right now.

What you're proposing, if I understand it correctly, is to remove the default entirely. So for that reason alone, teuthology would become dependent on ceph-qa-suite. And further, we would have to add a package list to every ceph-qa-suite branch that currently exists.

I honestly don't see how that would be a good thing.

vasukulkarni: can we have 2 files then packages-default which can stay here, and packages-qa.yaml in qa-suites so that  install task yaml will look nicer without big overrides and just use the packages-qa.yaml if it exists?

zmc: That might be possible, though I'd have to think about how to do it in a way that wasn't too ugly :)

vasukulkarni: Thanks @zmc

zmc: Just had an idea that might work. Haven't implemented or tested it yet. I'll see how @dmick feels about it and whether we should block this PR on it.

zmc:
The commits I just added are intended to work with something like what's in https://github.com/ceph/ceph-qa-suite/pull/859

vasukulkarni:
:+1:

PR #818, opened 2016-03-11
valgrind: add suppression for tcmalloc in libboost_thread-mt.so.1.53.0

dachary:
http://tracker.ceph.com/issues/14799 Fixes: #14799

### DIFF
--- a/teuthology/task/valgrind.supp
+++ b/teuthology/task/valgrind.supp
@@ -35,6 +35,20 @@
    ...
 }
 {
+   tcmalloc: libboost_thread-mt.so.1.53 is linked with tcmalloc
+   Memcheck:Param
+   msync(start)
+   obj:/usr/lib64/libpthread-2.17.so
+   obj:/usr/lib64/libunwind.so.8.0.1
+   obj:/usr/lib64/libunwind.so.8.0.1
+   obj:/usr/lib64/libunwind.so.8.0.1
+   obj:/usr/lib64/libunwind.so.8.0.1
+   ...
+   fun:_ZN8tcmalloc11ThreadCache
+   ...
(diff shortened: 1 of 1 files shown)


dachary:
@dmick note that I hand crafted the suppression, including the name mangling, which seems like the hard way of doing it. Do you know how to create a suppression from the XML report ?

dachary:
<pre>
filter='fs/verify/{clusters/fixed-3-cephfs.yaml debug/mds_client.yaml fs/btrfs.yaml overrides/whitelist_wrongly_marked_down.yaml tasks/cfuse_workunit_suites_dbench.yaml validater/valgrind.yaml}'
teuthology-suite --priority 101 --suite fs --filter="$filter" --suite-branch hammer --email loic@dachary.org --ceph hammer-backports --machine-type smithi
</pre>

* **running** http://pulpito.ceph.com/loic-2016-03-10_19:56:52-fs-hammer-backports---basic-smithi/

dmick:
I don't know anything in particular about suppressions, no, but I do note in the docs the --gen-suppressions flag?...

gregsfortytwo:
Did you run this through the test triggering issues and see if it suppresses properly? ;)

I think you don't need to do the mangling, though...our more recent functions are all just pulling out of valgrind xml log, which in the instance referenced here doesn't seem to be mangled — if you're worried about it you can also just include the important parts of the function name and use * for all the mangling bits.

dachary:
> Did you run this through the test triggering issues and see if it suppresses properly? ;)

Yes (see above) and it does pass. If you're also doing things the hard way (i.e. manual crafting instead of --gen-suppressions (which is not in the valgrind run unfortunately), then I'm happy to follow your lead.

PR #819, opened 2016-03-14
Enable epel on centos and fedora

vasukulkarni:
In octo lab the octo-secrets ansible disables epel for all distros, this will enable epel for centos/fedora.

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -177,6 +177,17 @@ def _yum_unset_check_obsoletes(remote):
                check_status=False)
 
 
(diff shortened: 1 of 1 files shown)


zmc:
It doesn't seem appropriate to forcibly enable EPEL like this, but maybe I just don't understand why it's needed. Putting that information here would be helpful.

Also, does EPEL even exist for Fedora?

vasukulkarni:
@zmc Other than rhel we want this enabled to get the extra packages, without that it will fail, here are the logs without epel enabled http://fpaste.org/339752/57976009/raw/

PR #821, opened 2016-03-16
install: Fix non-idempotent check_obsoletes tweak

zmc:


dmick:
At first I was confused by the absence of run.Raw() around the sed regexp, because it contains a shell metacharacter, but on reflection, it's actually appropriate that it get quoted so that the remote shell doesn't try to glob it against filenames; indeed this is the whole point of quoting the args in run().  Managed to finally unconfuse myself.

PR #823, opened 2016-03-17
Task module refactoring

zmc:
Back in March I thought it would be good to attempt to better organize our many tasks, subtasks, and their helper functions.

I thought a good step to take would be taking the larger tasks and making them subpackages of `teuthology.task`. I had to modify `run_tasks` to make this work. I started with `task.internal`.


zmc:
http://pulpito.ceph.com/zack-2016-11-01_16:07:44-smoke-jewel-distro-basic-smithi/
Currently in progress but looking good

PR #830, opened 2016-03-29
task/kernel: run apt-get clean before updating to avoid 15090

athanatos:
Signed-off-by: Samuel Just <sjust@redhat.com>


zmc:
nitpick: could you add the issue URL to the commit message? :)

athanatos:
Yes

athanatos:
@zmc updated

PR #831, opened 2016-03-29
task/parallel: accept lists of tasks in referenced sections

jdurgin:
This syntax is consistent with task configuration everywhere else.

The subtley of different syntax for this task made modifying suites
error-prone. Keep the existing syntax so we don't have to change all
the upgrade suites in each ceph-qa-suite branch.

### DIFF
--- a/teuthology/task/parallel.py
+++ b/teuthology/task/parallel.py
@@ -19,14 +19,23 @@ def task(ctx, config):
        - tasktest:
        - tasktest:
 
-    You can also reference the job from elsewhere:
+    You can also define tasks in a top-level section outside of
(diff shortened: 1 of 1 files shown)


yuriw:
Tests passed here @jdurgin @zmc 

http://pulpito.ceph.com/yuriw-2016-03-29_16:09:39-upgrade:infernalis-x:parallel-jewel-distro-basic-vps/
http://pulpito.ceph.com/yuriw-2016-03-29_16:09:27-upgrade:hammer-x:parallel-jewel-distro-basic-vps/

zmc:
test this please

PR #836, opened 2016-04-01
Changes for bootstrap to work on RHEL

vasukulkarni:
Changes for Redhat Linux
http://fpaste.org/348380/36955145/raw/

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -35,6 +35,38 @@ Linux)
             fi
         fi
         ;;
+    RedHatEnterpriseWorkstation|RedHatEnterpriseServer)
+        if ! rpm -q mysql-community-devel ; then
+             cat << EOM
+             Please install mysql-community devel package before running bootstrap
+             Reference document: http://dev.mysql.com/doc/refman/5.7/en/linux-installation-yum-repo.html
+             mysql Yum repo's exist in following location
+             http://dev.mysql.com/downloads/repo/yum/
+             and the Repo can be installed using (note below example is for v5.7)
+             yum localinstall http://dev.mysql.com/get/mysql57-community-release-el7-8.noarch.rpm
+             yum update metadata
+             yum install mysql-community-devel"
+EOM
+	    exit 1
+        fi
+        for package in python-pip python-virtualenv libev-devel libvirt-devel libffi-devel; do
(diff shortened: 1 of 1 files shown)


zmc:
"redhat linux" - do you mean "Red Hat Enterprise Linux"/"RHEL" ? ;)

vasukulkarni:
@zmc just added the enterprise server as well, they all same flavors and the commit now says RHEL :)

zmc:
@vasukulkarni did you check that the package name change is needed for / won't break fedora?

vasukulkarni:
@zmc  this will not break fedora, https://paste.fedoraproject.org/357536/1625146/raw/

dmick:
So this does two separate things:

1) changes from community-mysql-devel to mysql-community-devel
2) adds a new openssl-devel

What's the reason for each change?   

1) community-mysql-devel still appears to exist in Fedora:  https://apps.fedoraproject.org/packages/community-mysql-devel ; however, mysql-community-devel  does not:  https://apps.fedoraproject.org/packages/s/mysql-community-devel

2) what adds the new dependency on openssl-devel?

vasukulkarni:
@dmick 
1) I am using the *official* rpm's which has clear documentation for all distro's(rhel/fedora/centos) and how to install those packages: http://dev.mysql.com/doc/refman/5.7/en/linux-installation-rpm.html
2) This is not a *new* dependency that is needed, The old bootstrap just assumes that package is installed which is can be false for fedora/rhel and the bootstrap can fail.

zmc:
@vasukulkarni I'm not sure I understand fully. You claim this PR won't break Fedora, but your paste shows the package being installed from an external repository.

Is this correct: You want teuthology's bootstrap to attempt to install ``mysql-community-devel  ``, which doesn't exist on Fedora, instead of ``community-mysql-devel``, which does? 

Won't that make bootstrap fail on stock Fedora?

vasukulkarni:
@zmc updated, and fedora is untoched in this pr, so it should be like before.

dmick:
So, MySQL support in RHEL appears to be based on the MariaDB fork.  Both user and devel packages are available.  It seems clear to me that we should base teuthology support on bundled packages.

I'll file a followup.

PR #837, opened 2016-04-05
Rebase 2.0 to 10.1.0

vasukulkarni:
Rebase 2.0 to 10.1.0


dmick:
Appears to be changing a crosscheck inside teuthology to expect Ceph version 10.1.0 when installing what-will-become RHCS 2.0.  I can't help but think there's a better solution than hardcoding the source, but it's the way it's written now.

PR #839, opened 2016-04-05
Bootstrap

ErwanAliasr1:
A small PR to fix one issue and make an improvement.

As we look for english sentences from rpm, let's enforce the locale to avoid any false-positive effect.
A small improvement to make dnf the default package installation tool for fedora > 22.

Enjoy,


zmc:
Thanks!

PR #841, opened 2016-04-06
support multiple ceph clusters

jdurgin:
Tested against master, hammer and infernalis ceph-qa-suites, running a couple kinds of upgrade tests, plus subsets of the fs, rados, rbd, and rgw suites. The impetus for this is being able to test rbd mirroring, which uses multiple ceph clusters. There's a companion PR for ceph-qa-suite: https://github.com/ceph/ceph-qa-suite/pull/937

The general idea is that roles can have a prefix of the ceph cluster name. Since roles are only interpreted by some tasks (mainly the ceph task), this localizes the required changes somewhat, and makes it simple to stay compatible with existing single-cluster suites.

### DIFF
--- a/teuthology/ceph.conf.template
+++ b/teuthology/ceph.conf.template
@@ -1,6 +1,6 @@
 [global]
 	chdir = ""
-	pid file = $name.pid
+	pid file = /var/run/ceph/$cluster-$name.pid
         auth supported = cephx
 
 	filestore xattr use omap = true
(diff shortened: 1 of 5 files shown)


dillaman:
lgtm

jdurgin:
fixed the flake8 error

jdurgin:
@zmc ping - would like to merge this next week when I'm back and can deal with any fallout this causes

PR #844, opened 2016-04-14
Reimplementing get_scratch_devices()

ErwanAliasr1:
This PR is about re-implementing get_scratch_devices() while keeping the same input/outputs and adding news features
- listing all present block devices and not only sda or vda
- supporting more than 26 disks per host : sdaa vs sda
- auto-detecting the backing device of the root system by reading /proc/cdmline
- inspecting dm devices to detect more used devices
- expanding UUIDs & symlinks to understand the underlying block devices
- return no free block devices if known was detected as used to avoid damaging
  data
- return no free block devices if we cannot get intersections between the list
  of available devices and used devices to avoid damaging data
- compute the list of free devices by removing the detected used device from the
  complete list

As a result, a block reported to be free is a detected block devices which is
proven not being used by : the root fs, a dm device or a user mounted
filesystem.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -985,6 +985,24 @@ def translate_block_UUID(remote, devices):
     return devs
 
 
+def partitions_to_block_device(devices):
+    """
+    From a list of used partitions, let's compute the list of used block devices
+    """
+    devs = []
+    for device_name in devices:
+        # /dev/dm-0 should not be touched
+        if "-" not in device_name:
+            # Extracting device name "/dev/sda1" -> "/dev/sda"
+            matching_device = re.match('(/dev/[a-z]+)(\d?)', device_name, re.M | re.I)
+            if matching_device:
+                device_name = matching_device.group(1)
+
+        if device_name not in devs:
+            devs.append(device_name)
+    return devs
+
+
 def get_scratch_devices(remote):
     """
     Read the scratch disk list from remote host
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -873,6 +873,14 @@ def get_block_devices(remote):
     return devs
 
 
+def get_free_block_devices(available_block_devices, used_block_devices):
+    """
+    Compute the list of free block devices by comparing the list
+    of block devices we have and the one we consider as used
+    """
+    return list(set(available_block_devices) - set(used_block_devices))
+
+
 def get_used_block_devices(remote):
     """
     From the actual mount points, let's found out
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -834,6 +834,8 @@ def get_block_devices(remote):
         devices = r.stdout.getvalue().strip().split('\n')
         for device in devices:
             # Let's split the /proc/partitions in fields
+            #major minor  #blocks  name
+            #   8        0  500107608 sda
             matching_line = re.match('\s+(\d+)\s+(\d+)\s+(\d+)\s+(.*)$', device, re.M | re.I)
             if matching_line:
                 major = matching_line.group(1)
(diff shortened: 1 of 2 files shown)


zmc:
ok to test

teuthology/misc.py
```diff
@@ -818,53 +819,269 @@ def get_wwn_id_map(remote, devs):
...
+    return devs
+
+
+def expand_dm_devices(remote, devices):
+    """
+    Extract the real device under the DM devices
+    """
+    devs = []
+    for device in devices:
+        if device.startswith("/dev/dm-"):
```
vasukulkarni: I am not sure whats happening here, for multipath devices, /dev/dm* are for internal use only, one shouldn't manipulate any of those internal devices within, users should only deal with /dev/mapper

ErwanAliasr1: The main idea is here to detect what are the block devices currently used by the system. So I'm opening here dm-* if and only if they are present in /proc/partitions then read from the output of "dmsetup ls --tree -o blkdevname" command what are the underlying block device tied to it. So I think that should be ok but if you have a system running such configuration I'd be interested having the /proc/partitions & the result of dmsetup command listed before. If that case should be treated differently, I'll update the code & the tests accordingly.

vasukulkarni: @ErwanAliasr1 Thanks for the explanation, multipath/multipathd commands properly deal with devices based on the multipath.conf, you can safely remove the dm logic since I think multipath devices nodes are not available for you to test anywhere in the lab  and also I guess the imaging of new nodes ends up removing any multipath packages from the system even for root devices.

ErwanAliasr1: But I need the DM logic. On my local system, I'm using dm-0 without any multipath.

zmc:
I'd like to see the results of some jobs executed using this teuthology branch, just to be a little more sure that there isn't a regression. Probably need to see smithi, mira and vps

ErwanAliasr1:
@zmc Good point. I tested my code locally against a mira and it worked but let me check simith & vps before ... I should I done that before pushing the PR.

ErwanAliasr1:
Tested a smithi and found a bug in my code but the protection code worked and would have avoid issues.

But that brings a question on my side : if we have a /scratch_device file, we should consider the content without touching it and could provide partitions.

If we don't have that scratch file then we should detect the free block devices. Shall consider the actual partition scheme and report each partition individually or shall I report the associate block itself ?
if /dev/sda is featuring /dev/sda1, /dev/sda2, /dev/sda3  : shall I report /dev/sda as free or each individual partition ?

zmc:
@ErwanAliasr1 re: devices vs. partitions: I believe the current behavior is to detect and report whole devices as opposed to partitions. We should be preserving the existing behavior in this PR. It's good to add the capability to detect different types of devices, but the hosts we use daily in our labs should continue to behave as they have before. That's part of why I mentioned wanting to see test results :)

ErwanAliasr1:
@zmc Ok, I update the PR this way. I also fixed the nvme support and discovered that I badly handled the /scratch_devs

ErwanAliasr1:
Looks good for me, local test ran fine

zmc:
@ErwanAliasr1 any word on scheduled tests using our hardware?

wusui:
@ErwanAliasr1 Hi.  I can provide some teuthology suite setup and running help you need it.  What tests do you want to run?  Looking at the change and what calls it, I assume that you want to run this against ceph-ansible and ceph-deploy tests.

ErwanAliasr1:
thx @wusui Any test would be fine if we pass via all kind of hardware types.

kshtsk:
Hm... Looks like these changes deserves a separate module like `blockdevices` or `blockdev` or `dev`, and can be placed for example by one of that path:
- `teuthology.blockdevices`
- `teuthology.orchestra.blockdevices`
- `teuthology.orchestra.blockdev`
- `teuthology.orchestra.dev`

Are there any plans to continue efforts?

PR #845, opened 2016-04-15
ansible.CephLab: touch ~/.vault_pass.txt

zmc:
To avoid ansible bailing with an OSError. Normally we don't even need
the vault password for cephlab.yml.


dmick:
empty file is probably the best option, if a little dissatisfying.

PR #847, opened 2016-04-20
docs: add SUSE to list of bootstrapable operating systems

kmroz:
Signed-off-by: Karol Mroz kmroz@suse.com


alfredodeza:
OK to test

PR #848, opened 2016-04-21
Restrict filter to not scan too far up a path.

wusui:
Directory names closer to the root than /suites are not used by the filter.
Fixes 15470

### DIFF
--- a/teuthology/suite.py
+++ b/teuthology/suite.py
@@ -594,6 +594,16 @@ def get_branch_info(project, branch, project_owner='ceph'):
     if resp.ok:
         return resp.json()
 
+def strip_fragment_path(original_path):
+    """
+    Given a path, remove the text before '/suites/'.  Part of the fix for
+    http://tracker.ceph.com/issues/15470
+    """
+    scan_after = '/suites/'
+    scan_start = original_path.find(scan_after) 
+    if scan_start > 0:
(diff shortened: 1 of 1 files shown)


teuthology/suite.py
```diff
@@ -594,6 +594,16 @@ def get_branch_info(project, branch, project_owner='ceph'):
     if resp.ok:
         return resp.json()
 
+def remove_front_of_path(original_path):
+    """
+    Given a path, remove the text before '/suites/'.  Part of fix for
+    Bug 15470
```
zmc: Please use a URL for the ticket! Thanks!

teuthology/suite.py
```diff
@@ -594,6 +594,16 @@ def get_branch_info(project, branch, project_owner='ceph'):
     if resp.ok:
         return resp.json()
 
+def remove_front_of_path(original_path):
```
zmc: Could we call this, maybe, ``strip_fragment_path`` perhaps?

zmc:
Looks good so far. Hopefully it works well with @yuriw's workflow. Would you mind also putting the tracker URL in the commit message?

zmc:
Thanks @wusui. @yuriw - if this doesn't break things for you, feel free to merge it!

PR #849, opened 2016-04-22
Fix ceph-deploy 'branch' override

vasukulkarni:
Default ceph-deploy override template defines devcommit which makes
'branch' overrides in the qa-suite useless since it only picksup one option
Remove the devcommit overrides from the template and use that as
default in case no other 'branch' override exist.

### DIFF
--- a/teuthology/suite/placeholder.py
+++ b/teuthology/suite/placeholder.py
@@ -73,9 +73,6 @@ dict_templ = {
             'sha1': Placeholder('ceph_hash'),
         },
         'ceph-deploy': {
-            'branch': {
-                'dev-commit': Placeholder('ceph_hash'),
-            },
             'conf': {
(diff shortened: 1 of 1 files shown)


vasukulkarni:
Depends on https://github.com/ceph/ceph-qa-suite/pull/970

vasukulkarni:
@alfredodeza can you look at this one and its associated qa pr? I would like to add a test for the last stable branch as well in addition to the current tip of the master.

alfredodeza:
This looks good to me

vasukulkarni:
http://qa-proxy.ceph.com/teuthology/vasu-2016-06-22_13:52:51-ceph-deploy-master---basic-vps/271405/teuthology.log failure is unrelated to this change, uses stable jewel branch

PR #851, opened 2016-04-29
prune: also remove mon data in jobs' 'data' subdir

zmc:
This will likely free up a large amount of data on sepia's log storage cluster


zmc:
@sage, you added the bits to archive this data in 3abc4128; do you think it's ok to prune the old mon data with the same rules as remote daemon logs?

zmc:
``[13:32:55]  <sage>	zackc: lgtm``

PR #832, opened 2016-03-30
 fix nuke

vasukulkarni:
Tested this on some of the nodes which weren't cleaning up before due to bad package state, this works fine on both rhel and ubuntu, Since I have updated paramiko I will be testing this with few test suites to ensure everything works.

PR #857, opened 2016-05-11
Fix the nuke cli behavior to not reboot on default

vasukulkarni:
cli users have to pass -r to reboot like before, still testing...

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -24,7 +24,6 @@ from .misc import decanonicalize_hostname
 from .misc import merge_configs
 from .misc import get_testdir
 from .misc import get_user
-from .misc import reconnect
 from .misc import sh
 from .parallel import parallel
 from .task import install as install_task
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@idryomov that is not required, the -r flag is more clear, just tested this on cli as well as in smoke.

idryomov:
Are you sure?  Did you try nuking a kernel client box with a cephfs filesystem mounted?  With this patch applied, if you don't pass --reboot-all, it'll leave the mount behind. 

If you pass --reboot-all, it'll most likely get stuck trying to sync a filesystem whose OSDs have already been nuked.  That's why we need to pass -f -n to reboot.

vasukulkarni:
I am going to try that on my test node, although the case is very hard to hit in kcephfs teuthology tests.

idryomov:
Well, the premise of #832 was that teuthology-nuke can be invoked at any time...  You should also try nuking a box with a local filesystem mounted on top of krbd.  In fact I'd try it first - the "umount -f doesn't work" comment in remove_kernel_mounts() is from 2011, I think it's been fixed since then and is working for cephfs now.

Doing a normal reboot with a local filesystem mounted on top of krbd and OSDs gone has always been a sure way to lockup the box.  It's probably going to be fixed in the near future, but for now the only way to nuke in this case is to force reboot.

vasukulkarni:
I will add a system reset here

vasukulkarni:
closing, many things changed after this.

PR #859, opened 2016-05-13
task/hadoop: fix for multi cluster is_type() change

jdurgin:
This is the only non-ceph task that uses custom roles, and it happens
to put a '.' as a delimiter between hadoop and the rest of the
role. Since the multi-cluster branch, is_type() is ceph-specific and
interprets this as a cluster named hadoop.

To avoid needing to change several branches in ceph-qa-suite, just use
a local is_hadoop_type() function instead of relying on the now
ceph-specific misc.is_type().

Fixes: http://tracker.ceph.com/issues/15862

### DIFF
--- a/teuthology/task/hadoop.py
+++ b/teuthology/task/hadoop.py
@@ -20,10 +20,13 @@ def dict_to_hadoop_conf(items):
     out += "</configuration>\n"
     return out
 
+def is_hadoop_type(type_):
+    return lambda role: role.startswith('hadoop.' + type_)
+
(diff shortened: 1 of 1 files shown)


jdurgin:
Passed a manual wordcount run.

gregsfortytwo:
This looks good to me, thanks @jdurgin. (Not merging as the build isn't done, and it's good to get extra eyes on teuthology repo changes.)

jcsp:
:+1:

PR #861, opened 2016-05-14
packaging.py: make _get_distro() support openSUSE

smithfarm:
Add an explicit "elif" block for opensuse - without this, it falls through into
the else clause and dies because it cannot determine a codename.


smithfarm:
@zmc @dachary

PR #863, opened 2016-05-18
suite.py: allow specifying ceph sha1

dmick:

### DIFF
--- a/teuthology/suite.py
+++ b/teuthology/suite.py
@@ -23,7 +23,7 @@ import teuthology
 import matrix
 from . import lock
 from .config import config, JobConfig
-from .exceptions import BranchNotFoundError, ScheduleFailError
+from .exceptions import BranchNotFoundError, CommitNotFoundError, ScheduleFailError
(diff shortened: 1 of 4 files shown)


yuriw:
👍

zmc:
I like this - but the ``--help`` text should be updated as well.

I also wonder if we should have a unit test for the new behavior.

athanatos:
lgtm

dmick:
I'll investigate help and unit test

dmick:
I think I'm done.

teuthology/suite.py
```diff
@@ -252,12 +253,23 @@ def create_initial_config(suite, suite_branch, ceph_branch, teuthology_branch,
     else:
         kernel_dict = dict()
 
-    # Get the ceph hash
-    ceph_hash = git_ls_remote('ceph', ceph_branch)
+    # Get the ceph hash: if --cephsha1/-S is supplied, use it if
```
zmc: Looks like it's ``--sha1``, not ``--cephsha1``

PR #865, opened 2016-05-25
Prevent workers from becoming zombies when repo operations fail

zmc:
The problem only shows up when all retries are exhausted.

Most of this PR is refactoring to enable unit testing.

http://tracker.ceph.com/issues/13925

### DIFF
--- a/teuthology/contextutil.py
+++ b/teuthology/contextutil.py
@@ -3,7 +3,9 @@ import sys
 import logging
 import time
 import itertools
+
 from .config import config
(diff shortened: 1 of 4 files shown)


dmick:
Couple of flake8 errors, and I sorta half-assed the review of the tests (forgive me), but it all kinda makes sense.  If there's something you want harder review on, I'll focus if you ask; otherwise it seems to make sense.

zmc:
@dmick whoops. Repushed to fix the flake errors. I think this should be good to go now

PR #866, opened 2016-05-25
Give a better error message for "suite not found"

dmick:
This involves a little work in the tests.


zmc:
@dmick there's an IndentationError

dmick:
Yeah, I'm still in process on this anyway

zmc:
Thanks for reworking the tests here! I like this much better.

PR #867, opened 2016-05-31
pcp: Don't fail a job if graph downloading fails

zmc:
http://tracker.ceph.com/issues/16049
Fixes: 16049

### DIFF
--- a/teuthology/test/task/test_pcp.py
+++ b/teuthology/test/task/test_pcp.py
@@ -1,7 +1,8 @@
 import os
+import requests
 import urlparse
 
-from mock import patch, DEFAULT, Mock, MagicMock
+from mock import patch, DEFAULT, Mock, MagicMock, call
 from pytest import raises
 
 from teuthology.config import config, FakeNamespace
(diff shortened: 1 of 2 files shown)


teuthology/test/task/test_pcp.py
```diff
@@ -360,3 +361,18 @@ def test_end(self, m_grafana, m_graphite, m_makedirs):
...
+    def test_end_16049(self, m_grafana, m_graphite, m_makedirs):
+        # http://tracker.ceph.com/issues/16049
+        # Jobs were failing if graph downloading failed. We don't want that.
+        self.ctx.archive = '/fake/path'
+        with self.klass(self.ctx, self.task_config) as task:
+            task.graphite.download_graphs.side_effect = \
+                requests.ConnectionError
+        # Even though downloading graphs failed, we should have called
+        # write_html() a second time
+        assert len(task.graphite.write_html.call_args_list) == 2
```
dmick: I realize there's always overkill, but, given that it's easy to check call args, would you want to do that here (i.e. static the first time, no static the second)?  That could subsume the "number of calls" check

teuthology/test/task/test_pcp.py
```diff
@@ -353,11 +353,10 @@ def test_begin(self, m_grafana, m_graphite, m_makedirs):
...
-        assert isinstance(task.stop_time, int)
-        return
-        self.task_config['graphite'] = True
         self.ctx.archive = '/fake/path'
         with self.klass(self.ctx, self.task_config) as task:
-            task.graphite.write_html.assert_called_once_with()
+            # begin() should have called write_html() once by now
+            assert len(task.graphite.write_html.call_args_list) == 1
+        # end() should have called write_html() a second time by now
+        assert len(task.graphite.write_html.call_args_list) == 2
```
dmick: See comment on other commit

teuthology/test/task/test_pcp.py
zmc: Good idea; done

PR #868, opened 2016-06-02
Bump downburst VMs to 4GB

zmc:


dmick:
This is to address http://tracker.ceph.com/issues/15052

PR #871, opened 2016-06-03
pin paramiko to 1.17.0

tchaikov:
http://tracker.ceph.com/issues/16142
Fixes: 16142

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,9 @@ setup(
                       'configobj',
                       'six >= 1.9', # python-openstackclient won't work properly with less
                       'httplib2',
-                      'paramiko',
+                      # 2.0.0 switched to cryptography, which causes
+                      # http://tracker.ceph.com/issues/16142
+                      'paramiko == 1.17.0',
                       'pexpect',
                       'requests',
                       'raven',


tchaikov:
@zmc is there anyway i can test this fix?

setup.py
```diff
@@ -60,7 +60,9 @@
                       'configobj',
                       'six >= 1.9', # python-openstackclient won't work properly with less
                       'httplib2',
-                      'paramiko',
+                      # 2.0.0 switched to cryptography, which causes
+                      # http://tracker.ceph.com/issues/16142
+                      'paramiko = 1.17.0',
```
zmc: This needs to use ``==``

tchaikov:
@zmc, fixed and repushed.

zmc:
@tchaikov have you been able to test that this fixes the problem?

tchaikov:
@zmc not yet. is there any way i can test it without merging this change?

zmc:
@tchaikov well, it really depends on how the issue can be reproduced. So far I have not been able to reproduce it. So far I've tried using ``teuthology.orchestra.remote.Remote`` to connect to a few machines that had exhibited the problem. I've also used ``teuthology-nuke --stale`` to clean up machines that were left locked because of the problem.

vasukulkarni:
@tchaikov here's the change log http://www.paramiko.org/changelog.html and I haven't seen any issue after update to latest version with PyCA change,  The tracker which you mentioned for this issue is not related to this change.

tchaikov:
> here's the change log http://www.paramiko.org/changelog.html

thanks for the link. maybe it's not a known issue when paramiko 2.0.0 was released?

> I haven't seen any issue after update to latest version with PyCA change

@vasukulkarni now the first thing i do everyday is to `teuthology-nuke --stale` the locked machine, which helps to move the queue. this is the issue i've being seeing recently.

>  The tracker which you mentioned for this issue is not related to this change.

i am not sure about this. but neither do i have strong evidence for my change. all i know is: this is a regression and this happens in cryptography. maybe we can switch to a test branch with this change and give it a trial run for one day or two? before we can not constantly reproduce it as @zmc pointed out.

zmc:
This hasn't been seen since we last talked about it; going to close this. If it pops up again we can revisit!

PR #872, opened 2016-06-03
Suite refactor

zmc:
This PR:
* Moves a lot of ``suite``'s functionality into a new ``Run`` class
* Makes use of ``GitbuilderProject``
* Makes ``suite`` a subpackage

It doesn't intend to change behavior otherwise.


zmc:
http://pulpito.ceph.com/zack-2016-06-03_14:08:14-smoke-jewel---basic-vps/

teuthology/suite/run.py
```diff
@@ -0,0 +1,411 @@
...
+                verbose=self.args.verbose,
+                log_prefix="Results email: ",
+            )
+            results_url = get_results_url(self.base_config.name)
+            if results_url:
+                log.info("Test results viewable at %s", results_url)
+
+    def schedule_suite(self):
+        """
+        Schedule the suite-run. Returns the number of jobs scheduled.
```
dmick: no hyphen necessary here

zmc: Really? That's how everyone seems to use the term. You'd prefer "suite run" ?

teuthology/suite/test/test_init.py
```diff
@@ -0,0 +1,186 @@
...
+    suite.Run.WAIT_PAUSE = 1
+
+    in_progress = deepcopy(results)
+    assert 0 == suite.wait('name', 1, 'http://UPLOAD_URL')
+    assert m_get_jobs.called_with('name', fields=['job_id', 'status'])
+    assert 0 == len(in_progress)
+    assert 'fail http://UPLOAD_URL/name/2' in caplog.text()
+
+    in_progress = deepcopy(results)
+    in_progress = deepcopy(results)
```
dmick: Is this subtle, or an error?

teuthology/suite/test/test_init.py
```diff
@@ -0,0 +1,186 @@
...
+    results.append([{'status': 'queued', 'job_id': '2'}])
+
+    def get_jobs(name, **kwargs):
+        return results.pop(0)
+    m_get_jobs.side_effect = get_jobs
+    suite.Run.WAIT_PAUSE = 1
+    suite.Run.WAIT_MAX_JOB_TIME = 1
+    with pytest.raises(suite.WaitException) as error:
+        suite.wait('name', 1, None)
+        assert 'abc' in str(error)
```
dmick: I don't get this line at all?...

teuthology/suite/util.py
```diff
@@ -0,0 +1,453 @@
...
+    return gp.sha1
+
+
+def get_distro_defaults(distro, machine_type):
+    """
+    Given a distro (e.g. 'ubuntu') and machine type, return:
+        (arch, release, pkg_type)
+
+    This is used to default to:
+        ('x86_64', 'trusty', 'deb') when passed 'ubuntu' and 'plana'
```
dmick: indenting

teuthology/suite/test/test_init.py
zmc: That certainly looks like an error!

teuthology/suite/run.py
dmick: yeah.  it's a run, and the type of run is suite.  clearly a nit.

teuthology/suite/test/test_init.py
zmc: Me either; I hadn't noticed that before. I wonder what it does...

teuthology/suite/util.py
zmc: Good catch

PR #875, opened 2016-06-08
modify initial of help in teuthology of explain function

yonghengdexin735:
I run the help for teuthology(like: teuthology-lock -h), the arguments for helps of explain function didn't distinguish between upper case and lower case in the first letter.

Signed-off-by: zhang.zezhu <zhang.zezhu@zte.com.cn>

### DIFF
--- a/scripts/coverage.py
+++ b/scripts/coverage.py
@@ -5,17 +5,17 @@ Analyze the coverage of a suite of test runs, generating html output with
 lcov.
 
 options:
-  -h, --help            show this help message and exit
+  -h, --help            Show this help message and exit
   -o LCOV_OUTPUT, --lcov-output LCOV_OUTPUT
-                        the directory in which to store results
+                        The directory in which to store results
   --html-output HTML_OUTPUT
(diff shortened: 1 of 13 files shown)


zmc:
@yonghengdexin735, why not use ``--owner`` ?

yonghengdexin735:
@zmc , this view is another pull request for modify initial of help in teuthology of explain function, I think this modify can optimize our help of teuthology. 
Thanks.

yonghengdexin735:
@zmc , would you please review for me, thanks!

dmick:
I don't see how this helps anything at all, and it just seems like pointless churn.  Why does it matter whether the first character of the help string is capitalized?

yonghengdexin735:
@dmick ,because I find the first character of the help string some times for upper or lower, I think the unification is better.

dmick:
Yeah, that's not worth it to me, but I've said my opinion.

zmc:
I agree with @dmick that this seems unnecessary. It also introduces typos.

PR #878, opened 2016-06-15
Install python 2.7 on ubuntu downburst VMs

zmc:
http://tracker.ceph.com/issues/16302
Fixes: 16302


zmc:
Tested.

PR #883, opened 2016-06-21
Add teuthology-suite --newest

dmick:
This adds a simple webserver (Flask, currently manually deployed on git.ceph.com) to allow querying git history from teuthology-suite in order to implement "--newest", a facility to search for the latest-built version of a branch and schedule it if the branch tip hasn't yet been built.  There's also a --newest-limit parameter to limit the amount of backtracking allowed (10 by default).  The number of commits backtracked is logged.

This uses GitbuilderProject to check for 'builds done' on the gitbuilders in the same way suite has been doing, but adds a loop to retry if newest is set, which uses util.find_git_parent() to find the sha1 of the parent of the requested branch/sha1 and retry.  There's some refactoring of suite.run to make this happen.

There could probably be more tests.

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -35,6 +35,11 @@ Standard arguments:
                               If both -S and -c are supplied, -S wins, and
                               there is no validation that sha1 is contained
                               in branch
+  -n <newest>, --newest <newest>
+                              Search for the newest revision built on all
+                              required distro/versions, starting from
+                              either --ceph or --sha1, backtracking
+                              up to <newest> commits [default: 10]
   -k <kernel>, --kernel <kernel>
                               The kernel branch to run against; if not
                               supplied, the installed kernel is unchanged
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -273,6 +273,7 @@ class Run(object):
         log.info('Suite %s in %s generated %d jobs (not yet filtered)' % (
             suite_name, suite_path, len(configs)))
 
+    def collect_jobs(self, arch, configs):
         # used as a local cache for package versions from gitbuilder
         package_versions = dict()
         jobs_to_schedule = []
@@ -369,7 +370,9 @@ class Run(object):
                     jobs_missing_packages.append(job)
 
             jobs_to_schedule.append(job)
+        return jobs_missing_packages, jobs_to_schedule
 
+    def schedule_jobs(self, jobs_missing_packages, jobs_to_schedule, name):
         for job in jobs_to_schedule:
             log.info(
                 'Scheduling %s', job['desc']
@@ -378,8 +381,10 @@ class Run(object):
             log_prefix = ''
             if job in jobs_missing_packages:
                 log_prefix = "Missing Packages: "
-                if (not self.args.dry_run and not
-                        config.suite_allow_missing_packages):
+                if (
+                    not self.args.dry_run and
+                    not config.suite_allow_missing_packages
+                ):
                     util.schedule_fail(
                         "At least one job needs packages that don't exist for "
                         "hash {sha1}.".format(sha1=self.base_config.sha1),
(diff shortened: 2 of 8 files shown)


dmick:
gitserver moved to ceph/gitserver by request

teuthology/suite/run.py
```diff
@@ -345,31 +333,40 @@ def schedule_suite(self):
...
+                except VersionNotFoundError:
+                    pass
                 if not util.has_packages_for_distro(sha1, os_type, flavor,
-                                                    package_versions):
+                                                    Run.package_versions):
                     m = "Packages for os_type '{os}', flavor {flavor} and " + \
                         "ceph hash '{ver}' not found"
                     log.error(m.format(os=os_type, flavor=flavor, ver=sha1))
                     jobs_missing_packages.append(job)
+                    # optimization: one missing causes backtrack in newest mode,
```
zmc: can't parse this

dmick: if one package is missing, that's enough to cause backtracking if you're in newest mode, so if we're in newest mode, stop right now rather than continuing to look for more missing packages.  It's pretty terse.  Would it scan if I added "package" after "missing"?

scripts/suite.py
```diff
@@ -35,6 +35,11 @@
                               If both -S and -c are supplied, -S wins, and
                               there is no validation that sha1 is contained
                               in branch
+  -n, --newest                Search for the newest revision built on all
+                              required distro/versions, starting from
+                              either --ceph or --sha1.
+  --newest-limit <backtrack>  backtrack is the number of commits to try
+                              in --newest mode  [default: 10]
```
zmc: Are these separate because you want to be able to pass ``-n`` and not e.g. ``-n 10`` all the time?

teuthology/suite/run.py
zmc: I'd say that and: ``s/,/;/ s/searching/search/``

scripts/suite.py
dmick: Right, and I couldn't figure out how to make docopt have an optional arg with a default.

Sent from Nine
________________________________

teuthology/suite/run.py
```diff
@@ -345,31 +333,40 @@ def schedule_suite(self):
...
                 # present in package_versions and gitbuilder will not be asked
                 # again for them.
-                package_versions = util.get_package_versions(
-                    sha1,
-                    os_type,
-                    flavor,
-                    package_versions
-                )
+                try:
+                    Run.package_versions = util.get_package_versions(
```
zmc: Why set the class attribute here? Why not ``self``?

teuthology/suite/util.py
```diff
@@ -453,3 +453,34 @@ def teuthology_schedule(args, verbose, dry_run, log_prefix=''):
...
+        return sha1s
+
+    # XXX don't do this every time?..
+    refresh(project)
+    # we want the one just before sha1; list two, return the second
+    sha1s = get_sha1s(project, sha1, 2)
+    if len(sha1s) == 2:
+        return sha1s[1]
+    else:
+        return None
```
zmc: Just a note that ``return None`` is never strictly necessary; it's implicit. But if you want to leave it there, that's fine

teuthology/suite/run.py
dmick: The comment is wrong; I was thinking in general that "package version presence caching is global", but I was thinking "global across methods", which of course self would accomplish.  But since you mentioned it, using the class attr also makes it global across Run instances, which is still valid, although less useful because I don't think we ever instantiate more than one Run.

I can move it back into the instance.

teuthology/suite/run.py
dmick: ok

teuthology/suite/util.py
dmick: good reminder.  I think I slightly prefer the explicitness here but noted.

teuthology/suite/util.py
```diff
@@ -453,3 +453,34 @@ def teuthology_schedule(args, verbose, dry_run, log_prefix=''):
         ))
     if not dry_run or (dry_run and verbose > 1):
         subprocess.check_call(args=args)
+
+
+def find_git_parent(project, sha1):
+
+    base_url = config.get('githelper_base_url', 'http://git.ceph.com:8080')
```
zmc: Ah. I should have noticed this before. Could you put the default for this directly in ``teuthology.config``? Bonus points if you want to add an entry to ``docs/siteconfig.rst``, but I certainly won't block on that.

dmick:
OK, I think this is done.

zmc:
🍻

PR #884, opened 2016-06-24
teuthology-suite: Spelling error in web API + minor test improvement

dmick:
In my defense, "committish" isn't a word, and I'm not the only one:  https://git.kaarsemaker.net/git/commit/a8a5406ab32fccabf8ed08001d50b5373e18ff1a/


dmick:
This goes with one last change to gitserver to match; see https://github.com/ceph/gitserver/pull/1

PR #889, opened 2016-06-28
Insure left-over ceph-release doesn't break a machine.

zmc:
We had a situation where a machine had a left-over ceph-release
file. Possibly due to a bad download at the time causing the
initial failure of the install command or a connection loss but
the rm that runs after the install command never ran and thus
wget just kept appending a download number to the file names
leaving around a blank/broken ceph-release RPM causing failures
in its wake.

This should prevent that by running a rm -f on the file first.


ktdreyer:
I forget why we don't just `yum install base_url` directly?

dmick:
likely because whoever wrote that code was unaware yum could do that

zmc:
#892

PR #890, opened 2016-06-29
Fix resolve_role_list in daemon.py

wusui:
resolve_role_list did not handle cases where roles specified contained wild-cards or
were not specified.  ceph.role.a would end up being resolved as ceph.rolea.

Fixes: tracker issue 16053

### DIFF
--- a/teuthology/orchestra/daemon.py
+++ b/teuthology/orchestra/daemon.py
@@ -238,7 +238,7 @@ class DaemonGroup(object):
                         prefix = type_
                         if cluster_aware:
                             prefix = daemon.role
-                        resolved.append(prefix + daemon.id_)
+                        resolved.append(prefix + '.' + daemon.id_)
         else:
             # Handle explicit list of roles or wildcards
             for raw_role in roles:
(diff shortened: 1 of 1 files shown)


zmc:
Interesting. I thought the bug was that the test in question was just using a broken regexp, but now I see it's not using the ``re`` module at all.

I do still think we should put a unit test in place for this bug. Can we be sure it won't break existing tests?

Also, please use a format like this when mentioning redmine issues:
https://github.com/ceph/teuthology/pull/879/commits/73a64f728eff08b243646ad609f0c147436a04bf

wusui:
"Can we be sure it won't break existing tests?" -- I looked through the existing tests, and it looks like none of them use wild-cards.  I will sanity check a couple of the other tests.

zmc:
Ok, let me know what you find!

wusui:
It looks like this does not break existing tests:

http://pulpito.ceph.com/wusui-2016-07-05_11:18:48-upgrade:jewel-x:stress-split-master---basic-vps/

I think that this is good to go.

zmc:
Thanks!

PR #892, opened 2016-06-29
Install -release RPMs directly

zmc:
This avoids having to dance around with wget and rm -f.
Fixes: 10481
http://tracker.ceph.com/issues/10481


ktdreyer:
:+1: nice

zmc:
@ktdreyer good point on the approach :beers:

PR #893, opened 2016-06-29
GitbuilderProject: respect branch in config

zmc:

PR #894, opened 2016-06-30
Complete --newest feature, add some cleanups

dmick:
two major fixes for --newest:

1) rewrite the /tmp yaml file that holds base_config when base_config is finally done (backtracking rewrites this info)
2) make sure that the copy of the job configuration doesn't share refs with the base_config dict so that it doesn't corrupt base_config while backtracking

The other two commits are cleanup in arg processing.

### DIFF
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -5,7 +5,6 @@
 import logging
 import os
 import time
-from tempfile import NamedTemporaryFile
 
 import teuthology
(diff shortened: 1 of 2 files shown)


dmick:
rebased after #893

zmc:
flake8

zmc:
'nanagement'? :)

zmc:
Other than the above comments, LGTM

PR #896, opened 2016-06-30
Add test requirements

zmc:


zmc:
I never really got why we were leaving them out. I think it's an unnecessary barrier for new contributors.

dmick:
given that there is a requirements.test.txt at least, I'm not militant, but this seems fine

dmick:
heh, oh, right, requirements.test.txt is something I invented locally.  Doh!

PR #898, opened 2016-06-30
Add 'tag' and 'ref' to GitbuilderProject's init, warn if more than one set, and tests

dmick:


dmick:
rebased to master again

teuthology/packaging.py
```diff
@@ -643,19 +644,34 @@ def _get_uri_reference(self):
...
+
+        def warn(attrname):
+            if len([b for b in ref, tag, branch, sha1]) > 1:
+                log.warning(
+                    'More than one of ref, tag, branch, or sha1 supplied; using %s',
+                     attrname
+                )
 
-        if tag:
+        if ref:
```
zmc: I'm curious: was this order of precedence used for any particular reason?

dmick: no.  If I were designing from scratch I would have used "ref" and *not* used 'tag' or 'branch', so probably just personal bias ('ref' means 'you know what you're doing')

PR #901, opened 2016-07-06
Add teuthology git version query/logging

dmick:

### DIFF
--- a/teuthology/__init__.py
+++ b/teuthology/__init__.py
@@ -14,10 +14,20 @@ monkey.patch_all()
 
 import logging
 import os
+import subprocess
 
+__version__ = '1.0.0'
 
-__version__ = '0.1.0'
+# do our best, but if it fails, continue with above
 
+try:
+    __version__ += '-' + subprocess.check_output(
+        'git rev-parse --short HEAD'.split(),
+        cwd=os.path.dirname(os.path.realpath(__file__))
+    ).strip()
+except Exception as e:
+    # before logging; should be unusual
+    print >>sys.stderr, 'Can\'t get version from git rev-parse', e
 
 # If we are running inside a virtualenv, ensure we have its 'bin' directory in
 # our PATH. This doesn't happen automatically if scripts are called without
@@ -38,6 +48,8 @@ logging.basicConfig(
     format='%(asctime)s.%(msecs)03d %(levelname)s:%(name)s:%(message)s')
 log = logging.getLogger(__name__)
 
+log.info('teuthology version: %s', __version__)
+
 
 def setup_log_file(log_path):
     root_logger = logging.getLogger()
@@ -54,3 +66,4 @@ def setup_log_file(log_path):
     handler = logging.FileHandler(filename=log_path)
     handler.setFormatter(formatter)
     root_logger.addHandler(handler)
+    root_logger.log.info('teuthology version: %s', __version__)


teuthology/__init__.py
```diff
@@ -14,10 +14,17 @@
...
+mydir = os.path.dirname(os.path.realpath(__file__))
 
-__version__ = '0.1.0'
-
+try:
+    __version__ = subprocess.check_output(
+        'git rev-parse --short HEAD'.split(),
+        cwd=mydir,
+    ).strip()
+except:
```
zmc: https://docs.python.org/2/howto/doanddont.html#except

teuthology/__init__.py
```diff
@@ -14,10 +14,17 @@
...
 import logging
 import os
+import subprocess
 
+mydir = os.path.dirname(os.path.realpath(__file__))
 
-__version__ = '0.1.0'
-
+try:
+    __version__ = subprocess.check_output(
```
zmc: I wonder if we should use strings like ``0.1.0-$hash``; just in case we ever do real releases...

teuthology/__init__.py
dmick: yyyyyyeaahhhhh....  ok

teuthology/__init__.py
dmick: I'm amenable.  Maybe we should call this 1.0.0 now since if this isn't production I don't know what is

dmick:
k.  1) changed to 1.0.0-<shorthash> or 1.0.0 if failure 2) better exception handling and a message if failure

teuthology/__init__.py
```diff
@@ -14,10 +14,22 @@
 
 import logging
 import os
+import subprocess
 
+mydir = os.path.dirname(os.path.realpath(__file__))
```
zmc: This is pedantic, but why store this in a variable? Could just put it down on line 28

dmick: no good reason; tiny bit of documentation as to why we're calculating it, less chance of line overrun.  I can move it.

zmc:
Awesome, thanks!

zmc:
This broke everything:
https://paste.fedoraproject.org/388672/14679031/

zmc:
A more readable traceback:
```
2016-07-07 09:02:11,897.897 INFO:teuthology:teuthology version: 1.0.0-6514f9e
  File "/Users/zack/inkdev/teuthology/virtualenv/bin/teuthology", line 9, in <module>
    load_entry_point('teuthology', 'console_scripts', 'teuthology')()
  File "/Users/zack/inkdev/teuthology/scripts/run.py", line 34, in main
    teuthology.run.main(args)
  File "/Users/zack/inkdev/teuthology/teuthology/run.py", line 306, in main
    set_up_logging(verbose, archive)
  File "/Users/zack/inkdev/teuthology/teuthology/run.py", line 30, in set_up_logging
    teuthology.setup_log_file(os.path.join(archive, 'teuthology.log'))
  File "/Users/zack/inkdev/teuthology/teuthology/__init__.py", line 69, in setup_log_file
    root_logger.log.info('teuthology version: %s', __version__)
AttributeError: 'function' object has no attribute 'info'
```

dmick:
How the....I ran this code!...  sigh.

PR #902, opened 2016-07-06
Add tests for schedule_suite(); fix some schedule_fail arguments

dmick:


zmc:
flake8

dmick:
sigh.  fixed and repushed

teuthology/suite/test/test_run_.py
```diff
@@ -111,3 +114,195 @@ def test_sha1_nonexistent(
...
+        m_build_matrix.return_value = build_matrix_output
+        m_file.side_effect = [StringIO('field: val\n') for i in xrange(11)]
+        m_get_install_task_flavor.return_value = 'basic'
+        m_get_package_versions.return_value = dict()
+        m_has_packages_for_distro.side_effect = \
+            [False for i in xrange(5)] + [True]
+
+        m_find_git_parent.side_effect = lambda proj, sha1: sha1 + '^'
+
+        self.args.newest = 10
```
zmc: I'm confused about why this value is ``10``, but most (but not all) of the ``xrange()`` calls use ``5``

dmick: just boilerplated from 'failure' case.  It only has to be >=5.

zmc: Where does the 5 comes from though?

dmick: it's less than 10.  It tries for some number of parent commits and then succeeds.

PR #904, opened 2016-07-07
teuthology: log version number

dmick:
Round 2, this time with testing

### DIFF
--- a/teuthology/__init__.py
+++ b/teuthology/__init__.py
@@ -14,10 +14,20 @@ monkey.patch_all()
 
 import logging
 import os
+import subprocess
 
+__version__ = '1.0.0'
 
-__version__ = '0.1.0'
+# do our best, but if it fails, continue with above
 
+try:
+    __version__ += '-' + subprocess.check_output(
+        'git rev-parse --short HEAD'.split(),
+        cwd=os.path.dirname(os.path.realpath(__file__))
+    ).strip()
+except Exception as e:
+    # before logging; should be unusual
+    print >>sys.stderr, 'Can\'t get version from git rev-parse', e
 
 # If we are running inside a virtualenv, ensure we have its 'bin' directory in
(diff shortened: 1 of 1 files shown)


dmick:
So I removed the extraneous 'log.' from the info() in setup_log_file(), and temporarily added 'setup' to that log message, and invoked "teuthology -a archive --lock --machine-type vps minimal.yaml' to test that both log messages appear to the console without error, and the latter appears in archive/teuthology.log.  I think that is actually a full-coverage test.

teuthology/__init__.py
```diff
@@ -38,6 +48,8 @@
     format='%(asctime)s.%(msecs)03d %(levelname)s:%(name)s:%(message)s')
 log = logging.getLogger(__name__)
 
+log.info('teuthology version: %s', __version__)
```
zmc: I like everything about this PR except for this line. Do we *really* want this output for every single invocation of any teuthology command?

dmick: Well, the other one is only to the log, so not in anyone's face, particularly if they're doing -suite or something.  I was aiming for "in your face" as a goad to get people to verify that they're running the latest.  But I can definitely change it or remove it.  Some ideas, some of which we discussed:

1) at least not for help/usage, but only on a valid command (don't know how hard this is; might involve touching every entry point).  Although this does give us a version-check-only for free.
2) only on --version/-V?
3) .debug instead of .info
4) ???
5) Profit

I'll bow to your recommendation; I don't feel that strongly and we can change minds easily later.

zmc: I don't have strong enough feelings to want to get too fancy here, so maybe using using ``log.debug`` is the right thing. If we later on feel like we really wish we had the version logged more frequently, we can revisit

dmick:
forcepushed with change to debug

PR #907, opened 2016-07-13
task/packages: install libcephfs1-dev

jcsp:
The way we currently build the cephfs java
bindings depends on the unversioned .so file
which is in the -dev package, not the
main one.

Fixes: http://tracker.ceph.com/issues/16640


ktdreyer:
@jcsp are we covering up a packaging bug with this change?

jcsp:
@ktdreyer I prefer "working around", but yes :-) -- as you say on the ticket, the solution is probably to persuade the JNI lib to link differently

PR #914, opened 2016-07-20
orchestra.run logging

zmc:


zmc:
passed: http://pulpito.front.sepia.ceph.com/zack-2016-07-26_13:01:43-smoke-master---basic-vps/

teuthology/orchestra/run.py
```diff
@@ -102,6 +102,15 @@ def wait(self):
 
         status = self._get_exitstatus()
         self.exitstatus = self.returncode = status
+        for stream in ('stdout', 'stderr'):
+            if hasattr(self, stream):
+                stream_obj = getattr(self, stream)
+                # Despite ChannelFile having a seek() method, it raises
+                # "IOError: File does not support seeking."
```
dmick: I guess it's friendlier/points more to the error than just a traceback with attr not found.  go paramiko.

teuthology/orchestra/run.py
```diff
@@ -102,6 +102,15 @@ def wait(self):
...
+                if hasattr(stream_obj, 'seek') and \
+                        not isinstance(stream_obj, ChannelFile):
+                    stream_obj.seek(0)
```
dmick: What is it that necessitates this seek, again?

teuthology/orchestra/run.py
```diff
@@ -370,9 +379,13 @@ def run(
...
-        if stderr is None:
-            stderr = logger.getChild(name).getChild('stderr')
-        g_err = gevent.spawn(copy_file_to, r.stderr, stderr)
+        # Log stderr
+        stderr_logger = logger.getChild(name).getChild('stderr')
+        # If the stderr arg is a stream, have the logging module write to it as
+        # well
+        if stderr is not None:
+            stderr_logger.addHandler(logging.StreamHandler(stdout))
+        g_err = gevent.spawn(copy_file_to, r.stderr, stderr_logger)
```
dmick: What happens if stderr_logger has no handler?  the copy just does nothing?

teuthology/orchestra/run.py
zmc: After the process completes, ``stream_obj.read()`` will always return ``''`` since it's still pointed at the end. The ``seek(0)`` points it back at the beginning, which is what callers will expect.

teuthology/orchestra/run.py
dmick: answering my own question: yes.

dmick:
I'm happy, but want to wait to merge this until tomorrow

smithfarm:
Purported to cause regression: http://tracker.ceph.com/issues/17102

PR #917, opened 2016-07-27
Fix regression when using teuthology-suite with --kernel

zmc:
http://tracker.ceph.com/issues/16838

First I created a reproducer, then I discovered two ways to fix the bug.

They both seem like good ideas, and appear to not cause further regressions, so I think we may want both commits.

### DIFF
--- a/teuthology/suite/test/test_run_.py
+++ b/teuthology/suite/test/test_run_.py
@@ -4,7 +4,7 @@ import requests
 import yaml
 
 from datetime import datetime
-from mock import patch, call, ANY
+from mock import patch, call, ANY, DEFAULT
 from StringIO import StringIO
 
 from teuthology.config import config, YamlConfig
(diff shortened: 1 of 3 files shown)


teuthology/suite/test/test_run_.py
```diff
@@ -130,6 +130,45 @@ def test_sha1_nonexistent(
...
+            'kernel_flavor': 'basic',
+            'kernel_branch': 'testing',
+            'suite': 'krbd',
+        }
+        self.args = YamlConfig.from_dict(self.args_dict)
+        with patch.multiple(
+            'teuthology.suite.util.GitbuilderProject',
+            _get_package_sha1=DEFAULT,
+        ) as m:
+            assert m != dict()
```
dmick: is this asserting that patch.multiple() worked?

zmc: Yeah. It's left over from development and I forgot to remove it. I can remove it if you want?

zmc:
In addition to the unit test I used:
```
teuthology-suite -vv -c master -m smithi -s krbd -k testing --dry-run -l 1 -n 7
```

``-n 7`` is present because master HEAD was not built at the time.

PR #920, opened 2016-07-29
Fix an obvious bug in sentinel()

zmc:


dmick:
man, your reviewer sucks.

The auto-check won't really help here, so going ahead

PR #922, opened 2016-08-03
install: Allow installing python3 packages universally

BlaXpirit:
...by replacing `python3-*` with `python34-*` on CentOS.

This is required so python3-based packages can be specified cleanly in `extra_packages` (like [here](https://github.com/ceph/ceph-qa-suite/compare/wip-ceph-python3#diff-9b1ef2f4528109b07046d9dec615dd08R3)), because packages for CentOS will have a `python34-` prefix and not `python3-`, to follow the existing convention.

Related to https://github.com/ceph/ceph/pull/10208

### DIFF
--- a/teuthology/task/install.py
+++ b/teuthology/task/install.py
@@ -186,6 +186,24 @@ def _yum_unset_check_obsoletes(remote):
                check_status=False)
 
 
+def _rpm_package_overrides(pkgs, os):
+    """
+    Replaces some package names with their distro-specific equivalents
+    (currently "python3-*" -> "python34-*" for CentOS)
+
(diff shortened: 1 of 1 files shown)


jdurgin:
lgtm - a simple approach is all that's needed now

 @zmc what do you think?

dmick:
oy.  what's next, python-342?  :(  one questions the wisdom of the package naming.

BlaXpirit:
(By existing convention I mean https://dl.fedoraproject.org/pub/epel/7/x86_64/p/ *python34*)

BlaXpirit:
ping

PR #923, opened 2016-08-03
valgrind: wildcard boost supression

jcsp:
...so that it will match mangled names.

Fixes: http://tracker.ceph.com/issues/14794

### DIFF
--- a/teuthology/task/valgrind.supp
+++ b/teuthology/task/valgrind.supp
@@ -303,9 +303,9 @@
 	thread_local memory is falsely detected (https://svn.boost.org/trac/boost/ticket/3296)
 	Memcheck:Leak
 	...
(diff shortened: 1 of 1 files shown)


jcsp:
I don't actually know if this is going to work!  Testing:
http://pulpito.ceph.com/jspray-2016-08-03_04:48:17-fs:verify-master-distro-basic-mira
http://pulpito.ceph.com/jspray-2016-08-03_04:47:50-fs:verify-master-distro-basic-mira

jcsp:
The tests look promising: they are catching the OSD valgrind errors that happen to exist in master at the moment, but the client is not complaining about anything.

gregsfortytwo:
👍

PR #925, opened 2016-08-08
misc.sh(): Don't capture stderr by default

zmc:
If not capturing it, log it.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1331,19 +1331,21 @@ def is_in_dict(searchkey, searchval, d):
         return searchval == val
 
 
-def sh(command, log_limit=1024):
+def sh(command, log_limit=1024, include_stderr=False):
     """
-    Run the shell command and return the output in ascii (stderr and
-    stdout).  If the command fails, raise an exception. The command
-    and its output are logged, on success and on error.
+    Run the shell command and return the output in ascii (stdout and optionally
+    stderr). If the command fails, raise an exception. The command and its
+    output are logged, on success and on error.
     """
     log.debug(":sh: " + command)
     proc = subprocess.Popen(
         args=command,
         stdout=subprocess.PIPE,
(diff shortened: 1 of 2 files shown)


zmc:
The reason for this is:
```
$ openstack image list -f json --property name='teuthology-ubuntu-14.04'
Discovering versions from the identity service failed when creating the password plugin. Attempting to determine version from URL.
[
  {
    "Status": "active",
    "ID": "209f81d9-b96e-42e2-98f0-577725424538",
    "Name": "teuthology-ubuntu-14.04"
  }
```
See https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/auth/identity/generic/base.py#L143-L145

teuthology expected the output to be JSON; the first line in that output clearly breaks that. This commit causes that first line to be logged instead of captured.

dmick:
:thumbsup: just waiting for the check

dachary:
This will effectively hide error messages by default and I'm quite sure that's going to make it very difficult to figure out whatever problem happens next.

dmick:
perhaps it should be log.warning()

zmc:
I'm not sure what you mean, @dachary - stderr is still logged.

dachary:
Hum, strike that, it will display stderr instead of returning it. But the reason why it was not done this way in the first place is because it blocs until the process is finished. Which not only breaks the functionality that allows lines to be displayed as they are printed but will also probably deadlock if the process does not finish because the stdout is full and is not consumed by the parent because it is only waiting on the stderr.

zmc:
I see your new concern and I agree. I can put the stderr processing after the stdout loop.

My issue with using -q on all the openstack calls is that if something more important pops up, we won't see it.

dachary:
The **-q** does not suppress errors, only warnings. And if you put the stderr after the stdout you have another potential deadlock, only the other way around.

PR #927, opened 2016-08-09
openstack: do not display openstack client warnings

dachary:
Because they are concatenated with the STDOUT by the sh function despite
being output on STDERR and that breaks json parsing.

Fixes: http://tracker.ceph.com/issues/16972

### DIFF
--- a/teuthology/nuke.py
+++ b/teuthology/nuke.py
@@ -450,7 +450,7 @@ def stale_openstack_volumes(ctx, volumes):
     for volume in volumes:
         volume_id = volume.get('ID') or volume['id']
         try:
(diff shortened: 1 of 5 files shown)


dmick:
but....misc.sh() was just changed to separate stdout from stderr, and I think it did (see the 'stderr:' prefix).  I think the issue is the HTTP 429/return code 1

zmc:
Yeah, #925 has the same effect as this PR. The ticket you're referencing is about the HTTP 429 / rate limiting issues

dmick:
So #925 had problems, and was reverted; this seems safer.  However, as noted, it doesn't resolve http://tracker.ceph.com/issues/16972.

PR #929, opened 2016-08-11
openstack: cache auth tokens

dachary:
http://tracker.ceph.com/issues/16893

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,8 @@ setup(
                       # Babel, and installs 2.3.3, which is forbidden by
                       # python-novaclient 4.0.0
                       'python-novaclient',
-                      'python-openstackclient',
+                      # python-openstackclient == 3.0.0 fails when using a token
+                      'python-openstackclient == 2.2.0',
                       # with openstacklient >= 2.1.0, neutronclient no longer is
                       # a dependency but we need it anyway.
                       'python-neutronclient',
--- a/teuthology/nuke/__init__.py
+++ b/teuthology/nuke/__init__.py
@@ -86,7 +86,7 @@ def stale_openstack_instances(ctx, instances, locked_nodes):
 
 
 def openstack_delete_volume(id):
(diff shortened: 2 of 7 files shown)


teuthology/nuke.py
```diff
@@ -442,16 +442,16 @@ def stale_openstack_instances(ctx, instances, locked_nodes):
 
 
 def openstack_delete_volume(id):
-    sh("openstack volume delete " + id + " || true")
+    OpenStack().run("volume delete " + id + " || true")
```
zmc: Why is this call using an instance where the other one in this module is not?

dachary: My bad: in an earlier version I had run as a static function but it's no longer the case.

smithfarm:
@dachary Rebase?

dachary:
@smithfarm rebased

dmick:
Two of these are "missing -q to the call params", but the last one is "used to be sh(), now is OpenStack.run()" and I'm not immediately sure how to mock that.

dachary:
@dmick which one are missing -q ? None of them should use it now that the run method exists, because it adds --quiet to all.

dmick:
see the test results

dachary:
@dmick oh, right :-) I thought you were refering to parts of the code where -q was missing. Fixed & repushed !

dachary:
<pre>
        def sh(cmd):
            if 'volume show' in cmd:
                return volume_show
    
        with patch.multiple(
                nuke,
                sh=sh,
                openstack_delete_volume=DEFAULT,
                ) as m:
            nuke.stale_openstack_volumes(ctx, volume_list)
>           m['openstack_delete_volume'].assert_called_with(id)
</pre>
the error is because it does not mock the function it should

dmick:
pushed https://github.com/dmick/teuthology/tree/wip-token with a suggested fix for these.  Comments welcome, raiding welcome

dmick:
Filed #942 to extend/replace this

PR #931, opened 2016-08-12
Python 3 compatibility

oprypin:
This pull request makes _teuthology_ compatible with Python 3.x while keeping compatibility with Python 2.x.

The most controversial part of this pull request is the [compat.py](https://github.com/BlaXpirit/teuthology/blob/wip-py3/teuthology/compat.py) module and its uses. The reason that it is needed is simple: throughout the _teuthology_ codebase bytestrings are used for text data. It is especially important to not change the existing behavior on Python 2 to Unicode strings due to assumptions outside of _teuthology_, like _ceph-qa-suite_. That's why on Python 2 the _stringify_ function leaves the bytestrings alone, and on Python 3 decodes them into Unicode strings. Same for `StringIO` - on Python 2 it's still `cStringIO` (byte-based) and on Python 3 it's `io.StringIO` (Unicode-based)

Note that I couldn't properly test these changes because _ceph-qa-suite_ is Python 2.x only at the moment. However, I made sure that unittests pass on both Python 2.x and Python 3.x, and that the number-wise stats for the unittests (passed/error/warning etc) are exactly the same as on the master branch.

Additionally, I ran a few actual tests with _teuthology_, again, in 3 configurations: _master_+_python2_, _wip-py3_+_python2_, _wip-py3_+_python3_. Details and logs [**here**](https://gist.github.com/330fe8639e75d54fba7b7dfb461bb605).
Basically, the simple _interactive_ "test" works well, the _teuthology_ test works the same in all configurations, and the _rados/cephtool_ test fails only on Python 3 because it is not compatible with Python 3.

### DIFF
--- a/teuthology/beanstalk.py
+++ b/teuthology/beanstalk.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import beanstalkc
 import yaml
 import logging
@@ -109,17 +110,17 @@ class JobPrinter(JobProcessor):
         job_priority = job_config['priority']
         job_name = job_config['name']
         job_desc = job_config['description']
-        print 'Job: {i:>4} priority: {pri:>4} {job_name}/{job_id}'.format(
+        print('Job: {i:>4} priority: {pri:>4} {job_name}/{job_id}'.format(
             i=job_index,
             pri=job_priority,
             job_id=job_id,
             job_name=job_name,
-            )
+            ))
         if self.full:
             pprint.pprint(job_config)
         elif job_desc and self.show_desc:
             for desc in job_desc.split():
-                print '\t {desc}'.format(desc=desc)
+                print('\t {desc}'.format(desc=desc))
 
 
 class RunPrinter(JobProcessor):
@@ -131,7 +132,7 @@ class RunPrinter(JobProcessor):
         run = self.jobs[job_id]['job_config']['name']
         if run not in self.runs:
             self.runs.append(run)
-            print run
+            print(run)
 
 
 class JobDeleter(JobProcessor):
@@ -147,10 +148,10 @@ class JobDeleter(JobProcessor):
     def process_job(self, job_id):
         job_config = self.jobs[job_id]['job_config']
         job_name = job_config['name']
-        print 'Deleting {job_name}/{job_id}'.format(
+        print('Deleting {job_name}/{job_id}'.format(
             job_id=job_id,
             job_name=job_name,
-            )
+            ))
         job_obj = self.jobs[job_id].get('job_obj')
         if job_obj:
             job_obj.delete()
@@ -196,7 +197,7 @@ def main(args):
             # it is not needed for pausing tubes
             watch_tube(connection, machine_type)
         if status:
-            print stats_tube(connection, machine_type)
+            print(stats_tube(connection, machine_type))
         elif pause_duration:
             pause_tube(connection, machine_type, pause_duration)
         elif delete:
--- a/teuthology/beanstalk.py
+++ b/teuthology/beanstalk.py
@@ -1,10 +1,14 @@
 from __future__ import print_function
-import beanstalkc
-import yaml
 import logging
 import pprint
 import sys
 from collections import OrderedDict
+
+import yaml
+try:
+    import beanstalkc
+except ImportError:
+    import pystalkd.Beanstalkd as beanstalkc
 
 from .config import config
 from . import report
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -223,7 +223,7 @@ def get_initial_tasks(lock, config, machine_type):
         init_tasks.extend([
             {'pcp': None},
             {'selinux': None},
-            #{'ansible.cephlab': None},
+            {'ansible.cephlab': None},
             {'clock.check': None}
         ])
 
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -35,7 +35,7 @@ def set_up_logging(verbose, archive):
 def install_except_hook():
     def log_exception(exception_class, exception, traceback):
         logging.critical(''.join(format_tb(traceback)))
-        if not exception.message:
+        if not str(exception):
             logging.critical(exception_class.__name__)
             return
         logging.critical('{0}: {1}'.format(
@@ -46,13 +46,13 @@ def install_except_hook():
 
 def write_initial_metadata(archive, config, name, description, owner):
     if archive is not None:
-        with file(os.path.join(archive, 'pid'), 'w') as f:
+        with open(os.path.join(archive, 'pid'), 'w') as f:
             f.write('%d' % os.getpid())
 
-        with file(os.path.join(archive, 'owner'), 'w') as f:
+        with open(os.path.join(archive, 'owner'), 'w') as f:
             f.write(owner + '\n')
 
-        with file(os.path.join(archive, 'orig.config.yaml'), 'w') as f:
+        with open(os.path.join(archive, 'orig.config.yaml'), 'w') as f:
             yaml.safe_dump(config, f, default_flow_style=False)
 
         info = {
@@ -64,7 +64,7 @@ def write_initial_metadata(archive, config, name, description, owner):
         if 'job_id' in config:
             info['job_id'] = config['job_id']
 
-        with file(os.path.join(archive, 'info.yaml'), 'w') as f:
+        with open(os.path.join(archive, 'info.yaml'), 'w') as f:
             yaml.safe_dump(info, f, default_flow_style=False)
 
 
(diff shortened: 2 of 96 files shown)


BlaXpirit:
I dropped the commit that adds *py34* to *tox* so it can pass the check.

There is a problem that the test environment does not have header files for Python 3 installed.

jdurgin:
@BlaXpirit @zmc pointed out that the test script uses './bootstrap install' to get the necessary packages - could you add whatever is needed for py3 there?

BlaXpirit:
@jdurgin Well, it's not as simple as that, because Python 3 development packages are not required to use teuthology with Python 2.

Extending the script to properly support Python 3 separately seems complex: first, it would have to be an additional argument to the script (not entirely clear how to accept separate arguments), then separate packages per version / per OS (especially fun on CentOS).

And in the end the CI job would have to be modified to call it with that argument anyway.

jdurgin:
@BlaXpirit I was thinking of not making it optional - just checking for/installing both py2 and py3 packages, so tox works with both envs by default

BlaXpirit:
@jdurgin, don't forget about users. Python 3 is not available to install by default on CentOS!

jdurgin:
@BlaXpirit users of teuthology are ceph devs - not so worried about extra dependencies, just need to make sure bootstrap supports them, and if needed add additional docs on extra repo setup

Now that we've got py3 ceph packages, ceph devs will have those installed anyway, so it shouldn't be much of an issue

teuthology/beanstalk.py
```diff
@@ -1,11 +1,15 @@
...
 from collections import OrderedDict
 
+import yaml
+try:
+    import beanstalkc
+except ImportError:
+    import pystalkd.Beanstalkd as beanstalkc
+
 from .config import config
 from . import report
```
jdurgin: did you try various teuthology-queue operations? Wondering if there's any more adjustment needed for pystalkd

teuthology/run.py
```diff
@@ -223,7 +223,7 @@ def get_initial_tasks(lock, config, machine_type):
         init_tasks.extend([
             {'pcp': None},
             {'selinux': None},
-            {'ansible.cephlab': None},
+            #{'ansible.cephlab': None},
```
jdurgin: will need to remove this commit before merging

teuthology/suite/matrix.py
```diff
@@ -161,7 +161,7 @@ def _index(self, i, submats):
         off = (i // clen) % cycles
 
         def combine(r, s=frozenset()):
-            if type(r) is frozenset:
+            if isinstance(r, frozenset):
```
jdurgin: nice catch

teuthology/beanstalk.py
BlaXpirit: I have not tried any *teuthology-queue* operations.

teuthology/run.py
BlaXpirit: OK but doesn't it prevent teuthology from working on Python 3?

vasukulkarni:
@rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?

vasukulkarni:
test this please

kshtsk:
> @rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?

@vasukulkarni I actually in the middle of making teuthology code py2/py3 compatible taking code gradually from it, this PR is going to  be outdated as far as I finish the effort and close this PR. In other words, no-one should resolve conflicts within this PR, since the conflicts will increase conflicts.

vasukulkarni:
> > @rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?
> 
> @vasukulkarni I actually in the middle of making teuthology code py2/py3 compatible taking code gradually from it, this PR is going to be outdated as far as I finish the effort and close this PR. In other words, no-one should resolve conflicts within this PR, since the conflicts will increase in amount.

That is good to know, but why have py2 support? Its EOL anyway, but doesn't hurt.

kshtsk:
> > > @rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?
> > 
> > 
> > @vasukulkarni I actually in the middle of making teuthology code py2/py3 compatible taking code gradually from it, this PR is going to be outdated as far as I finish the effort and close this PR. In other words, no-one should resolve conflicts within this PR, since the conflicts will increase in amount.
> 
> That is good to know, but why have py2 support? Its EOL anyway, but doesn't hurt.

Just because we cannot switch everything to py3 at once, we need to keep it working on py2 until all ceph code base be compatible, so the work is staged in following steps:
1) make code compatible py2/py3 (still running infra on py2) (teuthology + ceph release branches)
2) make code runable on pure py3
3) get rid off py2 after stabilization period in infra
4) wipe py2 trace out of the code base

vasukulkarni:
> > > > @rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?
> > > 
> > > 
> > > @vasukulkarni I actually in the middle of making teuthology code py2/py3 compatible taking code gradually from it, this PR is going to be outdated as far as I finish the effort and close this PR. In other words, no-one should resolve conflicts within this PR, since the conflicts will increase in amount.
> > 
> > 
> > That is good to know, but why have py2 support? Its EOL anyway, but doesn't hurt.
> 
> Just because we cannot switch everything to py3 at once, we need to keep it working on py2 until all ceph code base be compatible, so the work is staged in following steps:
> 
> 1. make code compatible py2/py3 (still running infra on py2) (teuthology + ceph release branches)
> 2. make code runable on pure py3
> 3. get rid off py2 after stabilization period in infra
> 4. wipe py2 trace out of the code base

I was talking only about the teuthology code, I think the ceph part would be in ceph repo, but afaik I dont think there is a test code in teuthology.

kshtsk:
> > > > > @rakeshgm @kshtsk @racpatel @zmc there is quite a bit of work here on moving to python3 ? I think one can resolve conflicts, tests and get this merged? WDYT ?
> > > > 
> > > > 
> > > > @vasukulkarni I actually in the middle of making teuthology code py2/py3 compatible taking code gradually from it, this PR is going to be outdated as far as I finish the effort and close this PR. In other words, no-one should resolve conflicts within this PR, since the conflicts will increase in amount.
> > > 
> > > 
> > > That is good to know, but why have py2 support? Its EOL anyway, but doesn't hurt.
> > 
> > 
> > Just because we cannot switch everything to py3 at once, we need to keep it working on py2 until all ceph code base be compatible, so the work is staged in following steps:
> > 
> > 1. make code compatible py2/py3 (still running infra on py2) (teuthology + ceph release branches)
> > 2. make code runable on pure py3
> > 3. get rid off py2 after stabilization period in infra
> > 4. wipe py2 trace out of the code base
> 
> I was talking only about the teuthology code, I think the ceph part would be in ceph repo, but afaik I dont think there is a test code in teuthology.

right the test code is in the ceph repo, but you can't run teuthology on py3 without having ceph  code ported to py3 ;-)

vasukulkarni:
> > I was talking only about the teuthology code, I think the ceph part would be in ceph repo, but afaik I dont think there is a test code in teuthology.
> 
> right the test code is in the ceph repo, but you can't run teuthology on py3 without having ceph code ported to py3 ;-)

Can you tell me the ETA for your PR for python3 support? I might be missing something but ceph repo should be independent work, it should really affect tests in ceph repo. (just the bootstrap and all the parts in teuthology/beanstalkd are python3 ready is what we need in teuthology) and the python2 support is for remote nodes that run work units and is not affected.

kshtsk:
> Can you tell me the ETA for your PR for python3 support? I might be missing something but ceph repo should be independent work, it should really affect tests in ceph repo. (just the bootstrap and all the parts in teuthology/beanstalkd are python3 ready is what we need in teuthology) and the python2 support is for remote nodes that run work units and is not affected.

There is not a single PR, there is a set of PRs which are incorporated into master one-by-one, in order to not break the infrastructure, and minimize teuthology server out of service. Ceph repo can not be independent since qa/tasks are using teuthology modules at source level, and they are not running remotely, test workers are run locally by the worker. Not sure I understand your comment regarding `teuthology/beanstalkd`. Have you tested in on py3 at least once? I never saw it could even run without syntax errors.

kshtsk:
@vasukulkarni not sure I follow your point, would you like to take over the py3 porting effort?

vasukulkarni:
> @vasukulkarni not sure I follow your point, would you like to take over the py3 porting effort?

If you have PR in progress, we dont want to take over, but if its not ready yet then I think there is urgency to get it working now.  I think we can syncup once in Wednesday's meeting so that I can explain why I am thinking ceph repo source files shouldn't be affected and we can fix the teuthology part (based off the work done here) faster.

kshtsk:
So far so good, the time's come and we can close this PR in favor of #1362 since we were able to run successfully ceph tests against master branch on py3 deployed teuthology environment.

PR #932, opened 2016-08-12
[DNM] Add xfs bug to syslog error exceptions

djgalloway:
Fixes: http://tracker.ceph.com/issues/17003

### DIFF
--- a/teuthology/task/internal.py
+++ b/teuthology/task/internal.py
@@ -768,6 +768,8 @@ def syslog(ctx, config):
                     run.Raw('|'),
                     'grep', '-v', 'ceph-create-keys: INFO',
                     run.Raw('|'),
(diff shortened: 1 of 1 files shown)


zmc:
Is there evidence that this is truly something we should ignore?

djgalloway:
http://oss.sgi.com/archives/xfs/2016-06/msg00206.html maybe?

I also found http://tracker.ceph.com/issues/3058

djgalloway:
@jcsp @gregsfortytwo Can you guys weigh in on http://tracker.ceph.com/issues/17003#note-5 please?

gregsfortytwo:
Yeah, we have hit that with our own code before. A filter would need to check for xfs specifically. :(

idryomov:
Please don't do this - see my comment on the tracker.

PR #935, opened 2016-08-16
Nuke improvements, and a bonus refactor

zmc:
... because I can't seem to touch old code without some refactoring.

The main functional improvements are:
* Removing packages between killing processes and rebooting
* Attempting a power cycle if the console is unresponsive
* If the console is unresponsive after a power cycle, attempting to continue anyway by using SSH

This work was prompted by http://tracker.ceph.com/issues/17049


teuthology/orchestra/remote.py
```diff
@@ -479,7 +479,7 @@ def _exec(self, cmd):
 
     def _exit_session(self, child, timeout=None):
         child.send('~.')
-        t = timeout
+        t = timeout or self.timeout
         if not t:
```
dmick: Isn't this if redundant given the added or?

zmc: Good catch; I'll push a commit removing those two lines

teuthology/orchestra/remote.py
```diff
@@ -638,8 +634,7 @@ class VirtualConsole():
     """
     Virtual Console (set from getRemoteConsole)
     """
-    def __init__(self, name, ipmiuser, ipmipass, ipmidomain, logfile=None,
```
dmick: oh...because we have a factory function, we can just elide the unused param.  I see.

teuthology/nuke.py
```diff
@@ -400,19 +400,24 @@ def check_console(hostname):
         host=shortname,
         domain=config['ipmi_domain'])
     log.info('checking console status of %s' % cname)
-    if not console.check_status():
-        # not powered on or can't get IPMI status.  Try to power on
-        console.power_on()
```
dmick: ew.  I didn't know nuke did this.  That's unintuitive to me, and you're not changing it, but ... blech.

dmick:
:thumbsup:  Waiting for tests

dmick:
Just because it's been such an annoying couple of weeks, let's merge this first thing tomorrow

teuthology/nuke.py
zmc: Well, I do change it in a later commit: I add a power cycle to help recover systems who may have hung. If you think that's a bad idea we should discuss

dmick: it's just behavior I wouldn't expect, and would in general prefer to enable with a switch.  but it's more about surprise anyway.  carry on with the Large Hammer Nuke.

PR #937, opened 2016-08-22
Log IPMI console output for bare metal hosts

zmc:


zmc:
Before I merge, I also want to see if I can safely move the task earlier into job execution.

zmc:
@djgalloway I think the new conserver deployment broke this work (and all ipmitool console usage). We should talk tomorrow about how to proceed

teuthology/orchestra/console.py
```diff
@@ -38,9 +38,10 @@ def _pexpect_spawn(self, cmd):
             log.error('Must set ipmi_user, ipmi_password, and ipmi_domain in .teuthology.yaml')  # noqa
         full_command = self._build_command(cmd)
         log.debug('pexpect command: %s', full_command)
-        child = pexpect.spawn(full_command)
```
dmick: commit message: _pexpect_spawn

dmick:
I'm assuming you'll connect with conserver rather than ipmitool?...but yeah, talking good

teuthology/test/task/__init__.py
```diff
@@ -146,6 +146,7 @@ def test_setup_called(self):
             setup=DEFAULT,
             begin=DEFAULT,
             end=DEFAULT,
+            teardown=DEFAULT,
```
dmick: And because I am thick, patching these to DEFAULT just means they are patched to Mocks, so they won't be doing anything, but the assert_called_* still prove they made it?   and you need to patch teardown for completeness, because it's pass anyway, and isn't asserted about here, but both those could change in future?

teuthology/task/console_log.py
```diff
@@ -0,0 +1,88 @@
...
+
+from teuthology.orchestra.cluster import Cluster
+
+from . import Task
+
+log = logging.getLogger(__name__)
+
+
+class ConsoleLog(Task):
+    enabled = True
```
dmick: is this vestigial?

teuthology/task/console_log.py
```diff
@@ -0,0 +1,88 @@
...
+            self.enabled = False
+
+    def filter_hosts(self):
+        super(ConsoleLog, self).filter_hosts()
+        if not hasattr(self.ctx, 'cluster'):
+            return
+        new_cluster = Cluster()
+        for (remote, roles) in self.cluster.remotes.iteritems():
+            if not hasattr(remote.console, 'spawn_sol_log'):
+                print "not PC: %s" % str(remote.console)
```
dmick: I don't understand "not PC"

zmc: Just junk left over from debugging a test; removing

teuthology/task/console_log.py
zmc: Nope, it is what defaults the task to being enabled. ``__init__`` disables it conditionally.

teuthology/test/task/__init__.py
zmc: Exactly

zmc:
Running live: http://pulpito.ceph.com/zack-2016-08-23_01:24:46-krbd-master-testing-basic-smithi/380512/

zmc:
Another job, with the ``console_log`` task moved up: http://pulpito.ceph.com/zack-2016-08-23_03:21:34-krbd-master-testing-basic-smithi/380672/

zmc:
One for vps to make sure it doesn't break them jobs w/o IPMI: http://pulpito.ceph.com/zack-2016-08-23_03:23:00-krbd-master-testing-basic-smithi/380674/

dmick:
LGTM.

PR #939, opened 2016-08-23
Revert "packages.yaml: reflect python-ceph package split"

smithfarm:
This reverts commit 4543fb64cedffc97c816feff7d4a69cd0456ab15.

Fixes: http://tracker.ceph.com/issues/17075

### DIFF
--- a/teuthology/task/packages.yaml
+++ b/teuthology/task/packages.yaml
@@ -32,8 +32,6 @@ ceph:
   - libcephfs1
   - librados2
   - librbd1
-  - python-rados
-  - python-rbd
-  - python-cephfs
+  - python-ceph
   - rbd-fuse
   - ceph-debuginfo


smithfarm:
@vasukulkarni This looks like the simplest fix. I will also update https://github.com/ceph/ceph/pull/9757 by dropping the commit that drops the `python-ceph-compat` package. We'll hold off dropping that package as long as traces of pre-hammer releases remain in ceph-qa-suite.

zmc:
After speaking with @smithfarm I agree we at least need this. I'm not sure, however whether we need/want https://github.com/ceph/ceph-qa-suite/pull/1130 - let's talk there after this is merged

vasukulkarni:
@smithfarm @zmc this will break jewel/rpm,  The right place to fix is to add packages.yaml and use overrides in places required or else the purpose of packages.yaml is not going to be useful.

PR #941, opened 2016-08-25
Support console sessions using conserver

zmc:


teuthology/orchestra/console.py
```diff
@@ -35,6 +35,18 @@ def __init__(self, name, ipmiuser, ipmipass, ipmidomain, logfile=None,
         self.has_credentials = all(
             [self.ipmiuser, self.ipmipass, self.ipmidomain]
         )
+        self.conserver_master = config.conserver_master
+        self.conserver_port = config.conserver_port
+        conserver_client_found = subprocess.Popen(
+            'which console',
```
dmick: Not 100% sure about this.  1, design: is having the client enough to turn this on?  2, mechanism: is which universal, and always an external binary, and should we allow it to happen from any path?

teuthology/task/console_log.py
```diff
@@ -28,7 +28,8 @@ def filter_hosts(self):
             if not hasattr(remote.console, 'spawn_sol_log'):
                 log.debug("%s does not support IPMI; excluding",
                           remote.shortname)
-            elif not remote.console.has_credentials:
+            elif (not remote.console.has_ipmi_credentials and
+                  not remote.console.has_conserver):
                 log.debug("IPMI credentials not found for %s; excluding",
```
dmick: should probably change the msg too

teuthology/orchestra/console.py
```diff
@@ -271,13 +271,15 @@ def spawn_sol_log(self, dest_path):
         pexpect_templ = \
             "import pexpect; " \
             "pexpect.run('{cmd}', logfile=file('{log}', 'w'), timeout=None)"
-        python_cmd = 'python -c "%s"' % pexpect_templ.format(
-            cmd=console_cmd,
-            log=dest_path,
-        )
+        python_cmd = [
+            '/usr/bin/env', 'python', '-c',
```
dmick: 1) interested in details about spawn termination, but 2) why env here as well?  seems separate  (entirely possible I just don't understand)

zmc: 1) Not sure what you mean here
2) So that we find the correct python binary

teuthology/task/console_log.py
zmc: Good point, I'll update that

teuthology/orchestra/console.py
zmc: 1) I'm unsure, actually. I'd love feedback for how to better make this decision
2) Well, it's a binary on f24, centos 7, trusty, and xenial. I haven't yet seen a system without it.

teuthology/orchestra/console.py
dmick: I wanted to understand what caused you to rewrite this and how this helps.

teuthology/orchestra/console.py
dmick: Mech: yeah let's discuss in the mroning.  which: OK, but we should specify absolute path.

zmc:
test this please

zmc:
test this please

zmc:
test this please

PR #942, opened 2016-08-26
openstack: Cache auth tokens, add unit test fixes

dmick:
Loic wrote up #929, but the unit tests were failing.  As ovh is completely broken without it, I made a branch that extends that PR with unit test fixes.


zmc:
I don't see any big issues here. Has anyone tested this against an actual cloud? Should we?

zmc:
On second thought, since OVH is entirely broken without this, let's run with it

zmc:
Ouch http://pulpito.ovh.sepia.ceph.com:8081/zack-2016-08-29_23:08:32-smoke-master---basic-openstack/

PR #943, opened 2016-08-26
Install required packages when downbursting VMs

djgalloway:
http://pulpito.ceph.com/teuthology-2016-08-26_22:40:40-smoke-master-testing-basic-vps/

### DIFF
--- a/teuthology/provision/downburst.py
+++ b/teuthology/provision/downburst.py
@@ -180,6 +180,13 @@ class Downburst(object):
                 ['passwd', '-d', self.user],
             ]
         }
+        # Install git on downbursted VMs to clone upstream linux-firmware.
+        # Issue #17154
+        if 'packages' not in user_info:
+            user_info['packages'] = list()
+        user_info['packages'].extend([
+            'git',
+        ])
         # On CentOS/RHEL/Fedora, write the correct mac address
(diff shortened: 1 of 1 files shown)


teuthology/provision/downburst.py
```diff
@@ -180,12 +180,26 @@ def build_config(self):
...
+            'wget',
+        ])
+        # On CentOS/RHEL/Fedora, write the correct mac address and
+        # install redhab-lsb-core for `lsb_release`
         if os_type in ['centos', 'rhel', 'fedora']:
             user_info['runcmd'].extend([
                 ['sed', '-ie', 's/HWADDR=".*"/HWADDR="%s"/' % mac_address,
                  '/etc/sysconfig/network-scripts/ifcfg-eth0'],
             ])
+            if 'packages' not in user_info:
```
dmick: This test isn't necessary anymore, because the above is unconditional

teuthology/provision/downburst.py
```diff
@@ -180,12 +180,26 @@ def build_config(self):
...
+                user_info['packages'] = list()
+            user_info['packages'].extend([
```
zmc: Should probably just use ``append()`` here

dmick:
test this please (jenkins)

PR #946, opened 2016-08-30
Dont try to remove packages before reboot due to stale locks

vasukulkarni:
Signed-off-by: Vasu Kulkarni vasu@redhat.com

### DIFF
--- a/teuthology/nuke/__init__.py
+++ b/teuthology/nuke/__init__.py
@@ -326,8 +326,6 @@ def nuke_helper(ctx, should_unlock):
     shutdown_daemons(ctx)
     log.info('All daemons killed.')
     kill_valgrind(ctx)
-    # Try to remove packages before reboot
-    remove_installed_packages(ctx)
 
     remotes = ctx.cluster.remotes.keys()
     reboot(ctx, remotes)


zmc:
The change was intentionally added in https://github.com/ceph/teuthology/pull/935/commits/11dd4986bc28e733b4b1de3e97480a3dadb1dd0d

vasukulkarni:
@zmc that is not going to help as seen in various cases where the osd locks are held.

zmc:
I'm not really sure how that's related?

The reason for removing packages before reboot was daemons were starting up after reboot, wedging the machines.

vasukulkarni:
@zmc and now we are seeing machines that are left unlocked and what do you expect? cleanup manually?
Afair I haven't seen stale issue before this order was changed,  

you are talking about kernel issue fix there, for the kernel issue there is no single good solution other than proper reimage.

zmc:
@vasukulkarni I don't understand your comment and I'm not sure how to respond to "what do you expect? cleanup manually?"

No, I do not expect that at all.

You submitted this PR with no explanation that I can see. I referenced the commit where the package removal was added with a justification for that change.

Can you help me understand why the change was problematic? And also, what we should do about the original problem that it intended to solve?

vasukulkarni:
@zmc, sorry I thought you had the context, this is related to last 4 comments from http://tracker.ceph.com/issues/17084 and also stale issues that i am seeing now in octo.

for kernel wedge issue i am really not sure,  i think yuri runs krbd/kcephfs in vps and the owner's have to take responsibility of properly releasing the nodes back in good state when nuke fails.

zmc:
I didn't have the context, no (this is why we mention issues in commits and PRs). I'll repeat here that the later comments in 17084 are a separate issue from the original.

Anyway, if the root cause of the problem in those later comments is that processes that should be killed, aren't getting killed, we should address that. Perhaps ``shutdown_daemons()`` should be more aggressive?

vasukulkarni:
@zmc as per discussion in infra call last week,  shutdown_daemons() is not really a issue here, I finally raised a different tracker for this here: http://tracker.ceph.com/issues/17233, we have to revert back to the original cleanup method and handle commit 11dd498 in a different way and not break this.

zmc:
@vasukulkarni I don't understand your comment, or that bug report. Isn't the root cause of the problem that ceph processes are still running?

vasukulkarni:
@zmc we are seeing this *now* due to the commit 11dd498  which i mentioned above, i understand test failure should be followed by a defect against ceph component, afaik this is not a major issue for ceph since at the end of cleanup(uninstall, kill, etc) a issue seen is fixed by reboot. but if you think *ceph* should be fixed for this issue(in older releases - mainly hammer) then i will just *close* this and someone who wants to do a manually cleanup can do it.

zmc:
As I've said before several times, I do want the teuthology issue to be fixed, but fixed in a way that doesn't break something else. That is why I'm asking about the root cause from a teuthology perspective. Whatever ceph bug causing this isn't what I'm interested in. I'm interested in how we can get nuke to clean up machines in all cases.

This is why I'm asking about ``shutdown_daemons()``, but I'm not really getting an answer.

vasukulkarni:
ok just to be more clearn shutdown_daemons() is not a culprit here, it has issued the kill and the one thats complaining about the locks is 'rm -rf /var/lib/ceph', when you hit that state and issue lsof there i no process that it will list as holding locks, so *reboot* is the only good solution here, or maybe a better solution to fix this and http://tracker.ceph.com/issues/17027 , would be to just issue ipmi shutdown and proceed with cleanup.

PR #947, opened 2016-08-31
Fix logging regression

zmc:
http://tracker.ceph.com/issues/17102

#914 caused a regression in logging remote command output whereby each line wasn't logged immediately as it was received. This PR reverts the regression and adds an integration test so that we can avoid making the same mistake in the future.

Note that integrations tests aren't (and can't currently be) run by Jenkins-initiated testing as they need a live machine to SSH to.

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -237,14 +237,11 @@ def quote(args):
 
 
 def copy_to_log(f, logger, loglevel=logging.INFO):
-    """
-    Interface to older xreadlines api.
-    """
     # Work-around for http://tracker.ceph.com/issues/8313
     if isinstance(f, ChannelFile):
         f._flags += ChannelFile.FLAG_BINARY
 
-    for line in f.readlines():
+    for line in f:
         line = line.rstrip()
         # Second part of work-around for http://tracker.ceph.com/issues/8313
         try:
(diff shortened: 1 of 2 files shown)


dmick:
lgtm.  :thumbsup:

idryomov:
Hmm, looks like it's a bit more involved: StringIO is file-like, but doesn't have xreadlines()...

```
  File "/home/dis/src/teuthology/virtualenv/local/lib/python2.7/site-packages/gevent/greenlet.py", line 534, in run
    result = self._run(*self.args, **self.kwargs)
  File "/home/dis/src/teuthology/teuthology/orchestra/run.py", line 280, in copy_file_to
    copy_to_log(src, logger)
  File "/home/dis/src/teuthology/teuthology/orchestra/run.py", line 247, in copy_to_log
    for line in f.xreadlines():
AttributeError: 'cStringIO.StringO' object has no attribute 'xreadlines'
<Greenlet at 0x7f82c4d0a690: copy_file_to(<paramiko.ChannelFile from <paramiko.Channel 0 (cl, <logging.Logger object at 0x7f82ba700d10>, <cStringIO.StringO object at 0x7f82ba72e068>)> failed with AttributeError
```

dmick:
oh foo.  that's why it changed, right

idryomov:
Can safely/need to fallback to readline() for memory buffers?

zmc:
Whoops. I messed this up. I just pushed a different fix (which passes the same test and all the rest) which actually uses the recommended method for streaming file-like objects.

zmc:
test this please

zmc:
test this please

PR #949, opened 2016-09-02
openstack: use OS_AUTH_TYPE=token_endpoint when possible

dachary:
It requires setting the OS_URL depending on the command and avoids
calling the auth endpoint entirely.

Fixes: http://tracker.ceph.com/issues/16893

### DIFF
--- a/teuthology/provision/openstack.py
+++ b/teuthology/provision/openstack.py
@@ -66,22 +66,33 @@ class ProvisionOpenStack(OpenStack):
             except subprocess.CalledProcessError as e:
                 if 'No volume with a name or ID' not in e.output:
                     raise e
-                self.run("volume create -f json " +
-                         config['openstack'].get('volume-create', '') + " " +
-                         " --property ownedby=" + config.openstack['ip'] +
-                         " --size " + str(volumes['size']) + " " +
-                         volume_name)
+                # do not use OpenStack().run because its
+                # bugous for volume create as of openstackclient 3.2.0
+                # https://bugs.launchpad.net/python-openstackclient/+bug/1619726
+                misc.sh(
+                    "openstack volume create -f json " +
+                    config['openstack'].get('volume-create', '') + " " +
+                    " --property ownedby=" + config.openstack['ip'] +
+                    " --size " + str(volumes['size']) + " " +
+                    volume_name)
             with safe_while(sleep=2, tries=100,
                             action="volume " + volume_name) as proceed:
                 while proceed():
-                    r = self.run("volume show  -f json " + volume_name)
-                    status = self.get_value(json.loads(r), 'status')
-                    if status == 'available':
-                        break
-                    else:
-                        log.info("volume " + volume_name +
-                                 " not available yet")
-            self.run("server add volume " + name + " " + volume_name)
+                    try:
+                        r = OpenStack().run("volume show  -f json " +
+                                            volume_name)
+                        status = self.get_value(json.loads(r), 'status')
+                        if status == 'available':
+                            break
+                        else:
+                            log.info("volume " + volume_name +
+                                     " not available yet")
+                    except subprocess.CalledProcessError:
+                            log.info("volume " + volume_name +
+                                     " not information available yet")
+            # do not use OpenStack().run because its
+            # bugous for volume
+            misc.sh("openstack server add volume " + name + " " + volume_name)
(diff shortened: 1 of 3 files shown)


dachary:
@dmick this should avoid 429 errors on OVH

teuthology/provision/openstack.py
```diff
@@ -111,7 +116,7 @@ def create(self, num, os_type, os_version, arch, resources_hint):
         flavor = self.flavor(resources_hint['machine'],
                              config['openstack'].get('flavor-select-regexp'))
         cmd = ("flock --close --timeout 28800 /tmp/teuthology-server-create.lock" +
-               " openstack server create" +
+               " openstack --quiet server create" +
```
dmick: This seems unconnected?

teuthology/provision/openstack.py
```diff
@@ -127,7 +132,7 @@ def create(self, num, os_type, os_version, arch, resources_hint):
                " --wait " +
                " " + self.basename)
         try:
-            misc.sh(cmd)
+            self.run(cmd, type='compute')
```
dmick: as does this?

teuthology/provision/openstack.py
```diff
@@ -66,7 +66,10 @@ def attach_volumes(self, name, volumes):
             except subprocess.CalledProcessError as e:
                 if 'No volume with a name or ID' not in e.output:
                     raise e
-                self.run("volume create -f json " +
+                # do not use OpenStack().run because its
+                # bugous for volume create as of openstackclient 3.2.0
```
dmick: "it's buggy" or "it's bogus" (I would assume you really mean the former)

teuthology/provision/openstack.py
dachary: what does unconnected mean in this context ?

dmick: The change has nothing to do with the rest of the changes or the symptom

teuthology/provision/openstack.py
```diff
@@ -66,7 +66,10 @@ def attach_volumes(self, name, volumes):
...
+                # https://bugs.launchpad.net/python-openstackclient/+bug/1619726
+                misc.sh("volume create -f json " +
```
dmick: This needs to be prefixed with "openstack", yes?

teuthology/provision/openstack.py
dachary: ah, right

teuthology/provision/openstack.py
dachary: without this change the **cmd** will use password based authentication instead of reusing an existing token

teuthology/provision/openstack.py
dachary: yes, good catch ! I repushed after testing and also added a try-catch to cope with the (rare) case when a volume does not even exist at all after volume create returns.

dmick:
Test failures?

dachary:
@dmick test failures ? which test failures ;-)

dmick:
Thank you, rereviewing

dmick:
I've tried several times to find the actual reference for how Openstack auth works that would allow me to understand these changes, and failed.  Do you have a pointer?

also, can we separate the unconnected changes into a separate commit for clarity?

dachary:
> I've tried several times to find the actual reference for how Openstack auth works that would allow me to understand these changes, and failed. Do you have a pointer?

I don't unfortunately. I browsed the code of https://pypi.python.org/pypi/python-openstackclient and https://github.com/openstack/keystone until I figured out how to do what we need which is reusing tokens.

I think auth token throttling is unique to OVH, reason why documented strategies / implementations to cope with that do not exist. It also explains why some code paths have bugs when using an existing token instead of obtaining a new one. It's not a blocker since we can always fallback to the user/password authentication when it is the case.

I will separate the unconnected changes into a separate commit.

dmick:
> I will separate the unconnected changes into a separate commit.

Do you still plan to do this?  Our ovh runs are still failing every night

dachary:
I did it right away, did I missin anything ?

dmick:
I still only see one commit, so didn't look.  Did you just remove that code altogether?

dachary:
Yes, the --quiet option that you mentionned as unconnected was unnecessary.

PR #950, opened 2016-09-05
orchestra.run: fix generating ConnectionLostError

jcsp:
Previously you could get AttributeError if
a node went away, because it tried to call
getpeername after get_transport() had
returned None.


zmc:
Thanks!

PR #954, opened 2016-09-20
Set ctx.cluster even when a job has no roles

zmc:
This is to allow for teuthology to run self-tests that don't use
resources.


zmc:
Example job: http://pulpito.ceph.com/zack-2016-09-20_10:20:49-teuthology-master-distro-basic-smithi/425954/

zmc:
test this please

PR #956, opened 2016-09-21
Set ctx.cluster even when a job has no roles

zmc:
This is a re-work of #954 without the "break ``teuthology-nuke``" feature.

I've already tested nuke and I just scheduled:
✅ http://pulpito.ceph.com/zack-2016-09-21_10:06:35-teuthology-master-distro-basic-smithi/


dmick:
Dnm because waiting for test?

zmc:
Oh. Yes. Changing now

PR #958, opened 2016-09-26
ceph.conf: set min rweight min bytes per osd = 10

liewegas:
This lets the reweight-by-utilization command work even after mons are
restarted (ceph_manager is currently setting this via injectargs, but that
is not persistent).


tchaikov:
better than https://github.com/ceph/ceph-qa-suite/pull/1128!

PR #959, opened 2016-09-28
openstack: use get_provider() instead of self.provider

dachary:
Using self.provider directly will fail if set_provider() has never been
called. Use get_provider() instead because it will call set_provider()
if necessary.

The tests are adapted accordingly: some of them assumed set_provider()
could be skipped because the data member was accessed directly.

### DIFF
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -211,8 +211,21 @@ class OpenStack(object):
     token_expires = None
     token_cache_duration = 3600
 
+    @staticmethod
+    def clear_token_class():
+        OpenStack.token = None
+        OpenStack.token_expires = None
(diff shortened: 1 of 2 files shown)


dachary:
@dmick change is small: most of the diff is about fixing tests. We should be good now :-)

dmick:
Test failure, please investigate

dachary:
@dmick I will, thanks for the ping !

dmick:
@dachary any news?

dachary:
@dmick there was the week-end and it was great, really. Then today I got busy with other things (I'm happy about the refactor of jerasure SIMD that was merged). I should have time to work on this tomorrow. Thanks for asking :-)

dachary:
I don't have time to fix this, someone else will have to work on it.

PR #960, opened 2016-09-28
teuthology-nuke: lazy-unmount osd mount points

zmc:


zmc:
(this hasn't been tested yet, hence the DNM)

PR #963, opened 2016-10-06
Add rerun capability to teuthology-suite

zmc:
This isn't quite done yet but is ready for testing. Feedback is welcome!

Things that should be added:
- [x] A docs section in the "intro for testers" doc ?
- [x] Docs should mention that the implementation schedules a new run and filters it
- [x] Docs should mention that the new run doesn't inherit anything from the original, and flags must be used to set things like branch and machine type

Update: I think this is ready now.

### DIFF
--- a/docs/intro_testers.rst
+++ b/docs/intro_testers.rst
@@ -64,3 +64,11 @@ get an email when the test run completes.
 <https://github.com/ceph/pulpito/>`__ that will display the current status of
 each job. The Sepia lab's pulpito instance is `here
 <http://pulpito.ceph.com/>`__.
+
+There may be times when, after scheduling a run containing a large number of
+jobs, that you want to reschedule only those jobs which have failed or died for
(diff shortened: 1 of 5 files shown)


docs/intro_testers.rst
```diff
@@ -64,3 +64,10 @@ get an email when the test run completes.
 <https://github.com/ceph/pulpito/>`__ that will display the current status of
 each job. The Sepia lab's pulpito instance is `here
 <http://pulpito.ceph.com/>`__.
+
+There may be times when, after scheduling a run containing a large number of
+jobs, that you want to reschedule only those jobs which have failed or died for
+some other reason. For that use-case, `teuthology-suite` has a `--rerun` flag,
```
yuriw: I'd say `--rerun` (or `-r`) and  `--rerun-statuses` (or `-R`)

PR #964, opened 2016-10-08
orchestra.connect: do not override user/host from ~/.ssh/config

dmick:
I don't know why this was ever the case, but ignoring user@host
from the targets list just seems like a bad idea, and causes
problems, because tasks/ansible/etc. assume that the user to use
is ubuntu.  If that's not what .ssh/config says, there's just
going to be trouble...but there's no reason to pay attention
to .ssh/config there; if you really want to play tricks, edit
your target list accordingly.

Fixes: http://tracker.ceph.com/issues/17125

### DIFF
--- a/teuthology/orchestra/connection.py
+++ b/teuthology/orchestra/connection.py
@@ -79,26 +79,18 @@ def connect(user_at_host, host_key=None, keep_alive=False, timeout=60,
         username=user,
         timeout=timeout
     )
-    if key_filename:
-        connect_args['key_filename'] = key_filename
 
     ssh_config_path = os.path.expanduser("~/.ssh/config")
     if os.path.exists(ssh_config_path):
         ssh_config = paramiko.SSHConfig()
         ssh_config.parse(open(ssh_config_path))
         opts = ssh_config.lookup(host)
-        opts_to_args = {
-            'host': 'hostname',
-            'user': 'username'
-        }
-        if not key_filename:
-            opts_to_args['identityfile'] = 'key_filename'
-        for opt_name, arg_name in opts_to_args.items():
-            if opt_name in opts:
-                value = opts[opt_name]
-                if arg_name == 'key_filename' and '~' in value:
-                    value = os.path.expanduser(value)
-                connect_args[arg_name] = value
+        if not key_filename and 'identityfile' in opts:
+            key_filename = opts['identityfile']
+
+    if key_filename:
+        key_filename = [os.path.expanduser(f) for f in key_filename]
+        connect_args['key_filename'] = key_filename
 
     log.debug(connect_args)
 


djgalloway:
I just tried to use `teuthology-kill` (and thus `teuthology-nuke`) using this branch and bad things happened.

```
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:2016-10-10 14:53:43,017.017 ERROR:teuthology.nuke:Could not nuke {'smithi078.front.sepia.ceph.com': 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQD6xd+indcBVYVA2VHa+KwsdWBZ9LcH/eX7gmOu/uArXmX03lHarbz2otPqLF4sWbflJXWA8XPFjMv/3LFHE6L+aHJe9ZwFpt+91/EMDjSWd9pp4IJiFpZUAX/0ULSZFsSVK0Wz6iZK6x55BdNUaJNtRAqMwKTeXWicMvlbxhyNrbVL4WjC4cRJhTYIIBTg9XvRemt+UiIGJNNxUqMCIpqfAjHY/LlAHL0pSFc+jrp10ReB30kc/EEOjXiysU16kVqcNmbaS13yXhTr5wHFLn0EJ58KlrzhoIrYZOTLxVgDcE0ZC3Fhs0VB99EUyzj8JVMde5A1olhnT5Yuc+1/nGed'}
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:Traceback (most recent call last):
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/teuthology/nuke/__init__.py", line 281, in nuke_one
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:    nuke_helper(ctx, should_unlock)
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/teuthology/nuke/__init__.py", line 319, in nuke_helper
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:    connect(ctx, None)
2016-10-10 14:53:43,018.018 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/teuthology/task/internal.py", line 346, in connect
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:    rem.connect()
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/teuthology/orchestra/remote.py", line 59, in connect
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:    self.ssh = connection.connect(**args)
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/teuthology/orchestra/connection.py", line 92, in connect
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:    connect_args['key_filename'] = os.path.expanduser(key_filename)
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:  File "/home/dgalloway/src/teuthology/virtualenv/lib/python2.7/posixpath.py", line 254, in expanduser
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:    if not path.startswith('~'):
2016-10-10 14:53:43,019.019 INFO:teuthology.kill:AttributeError: 'list' object has no attribute 'startswith'
```

dmick:
agh.  one can have multiple identity filenames, in which case paramiko returns a list.  WIll fix.

dmick:
I think the old code also did not handle this properly, but did not raise an exception (it would look for '~' in the item, and if it were a list, that would always fail, so the passed list would be unmodified, and presumably then fail).  I'd like to know what you had in your .ssh/config for the host you tested, and how the auth ultimately worked (because paramiko will continue to search .ssh for keys anyway).

zmc:
When we (eventually and finally) stop using the ``ubuntu`` user everywhere, will this PR make that harder?

dmick:
@zmc it shouldn't; in that case the lock db ought to have <user> instead of 'ubuntu', and thus the list-targets etc. should have <user>@<host>

djgalloway:
Works for me now.

dmick:
Repushed/squashed.  @zmc, can you review plz

dmick:
ping @zmc

zmc:
@dmick Sorry, I totally failed to see your pings here. Thanks for the fix!

PR #965, opened 2016-10-11
Don't leak console-logging processes when a job is killed

zmc:
This PR first adds a framework for installing and removing signal handlers, allowing small tasks to be queued for execution when teuthology receives a given signal, or set of signals.

We then tell the ``console_log`` task to use this feature to ensure that the processes it spawns to log serial console output are killed when teuthology gets a ``SIGTERM``.

### DIFF
--- a/teuthology/exit.py
+++ b/teuthology/exit.py
@@ -1,0 +1,78 @@
+import logging
+import os
+import signal
+
+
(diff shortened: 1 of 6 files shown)


zmc:
I think this is ready to go!

PR #971, opened 2016-10-20
teuthology: update nuke for ceph-mgr

jcsp:


jcsp:
This goes with https://github.com/ceph/ceph-qa-suite/pull/1086, but doesn't require it (if mgr is not installed we'll not care during nuke)

PR #972, opened 2016-10-27
Pull in libcephfs2 instead of libcephfs1

jtlayton:
Companion pull request for the ceph one here:

https://github.com/ceph/ceph/pull/11647

...and the ceph-qa-suite one here:

https://github.com/ceph/ceph-qa-suite/pull/1223

Mostly just ripple effects of the package rename.

### DIFF
--- a/teuthology/task/packages.yaml
+++ b/teuthology/task/packages.yaml
@@ -8,7 +8,7 @@ ceph:
   - ceph-test
   - radosgw
   - python-ceph
-  - libcephfs1
+  - libcephfs2
   - libcephfs-java
   - libcephfs-jni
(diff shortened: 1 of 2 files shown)


teuthology/task/packages.yaml
```diff
@@ -8,7 +8,7 @@ ceph:
   - ceph-test
   - radosgw
   - python-ceph
-  - libcephfs1
```
jtlayton: Thanks, I'll reinstate it there.

jtlayton:
Yeah, looks like this PR isn't necessary. I think we can drop it for now.

PR #973, opened 2016-11-02
Change pull_directory to use a remote tar streaming to the local extract

dmick:
Avoids remote and local temp files, and subsequent read/write thrashing of local dir.  Since some of those directories are multiple tens or hundreds of gigabytes, the extra file load is important, and this should save time as well just by skipping the extra I/O thrashing (which the cephfs particularly hates)

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -19,7 +19,6 @@ import urlparse
 import yaml
 import json
 import re
-import tempfile
 import pprint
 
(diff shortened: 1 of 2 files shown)


dmick:
http://pulpito.ceph.com/dmick-2016-11-02_11:56:32-rados-master---basic-mira/ is one.  I can do more of a different stripe if you think it will help

PR #975, opened 2016-11-04
install: apply exclude_packages globally

zmc:
Previously, we were only applying the setting for upgrades


zmc:
http://pulpito.ceph.com/zack-2016-11-04_14:30:40-upgrade:jewel-x-master---basic-smithi/517676/

dmick:
I really want that code that calls install() to use "private_config = config; private_config.setdefault('project', 'ceph')... " but that can be later

PR #979, opened 2016-11-10
teuthology: update nuke for libcephfs2

tchaikov:
try to remove libcephfs2 as well.


tchaikov:
otherwise the libcephfs2 will be lingering around in the test node forever.

PR #980, opened 2016-11-10
teuthology/nuke: update for python-rgw

jcsp:


zmc:
Looks like this needs a rebase

PR #982, opened 2016-11-10
swift: added "--cluster" to rgw-admin command for multisite support

alimaredia:
This needs testing but I wanted to get it out so I don't forget about it

Signed-off-by: Ali Maredia <amaredia@redhat.com>


alimaredia:
This should be good to merge and seems to be a barrier to testing the rest of the multi-site changes

PR #983, opened 2016-11-14
kernel: update get_sha1_from_pkg_name() regex

idryomov:
ceph-build kernel packages will be named -ceph-g${SHA1:0:12}.  As the
first step towards using ceph-build, change the local package regex.
With this, I'm able to download post ceph/ceph-build#540 packages and
install them as a local deb/rpm.

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -1020,16 +1020,15 @@ def get_latest_image_version_deb(remote, ostype):
 
 def get_sha1_from_pkg_name(path):
     """
-    Get commit hash (min 7 max 40 chars) from (rpm or deb) package name.
-    Sample basenames of "make deb-pkg" and "make rpm-pkg" packages
-        linux-image-3.10.0-ceph-rhdeb-00050-g687d1a5f0083_3.10.0-ceph-rhdeb-00050-g687d1a5f0083-6_amd64.deb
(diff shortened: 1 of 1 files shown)


teuthology/task/kernel.py
```diff
@@ -1020,16 +1020,15 @@ def get_latest_image_version_deb(remote, ostype):
...
+    Get commit hash (min 12 max 40 chars) from (rpm or deb) package name.
+    Example package names ("make bindeb-pkg" and "make binrpm-pkg"):
+        linux-image-4.9.0-rc4-ceph-g156db39ecfbd_4.9.0-rc4-ceph-g156db39ecfbd-1_amd64.deb
+        kernel-4.9.0_rc4_ceph_g156db39ecfbd-2.x86_64.rpm
 
     :param path: (rpm or deb) package path (only basename is used)
     """
     basename = os.path.basename(path)
-    match = re.search('\d+[-_]g([0-9a-f]{7,40})', basename)
+    match = re.search('[-_]ceph[-_]g([0-9a-f]{12,40})', basename)
```
dmick: dear god that is so ugly

idryomov: @dmick Well, the alternative is extracting it from the package metadata.  It's more code and would still require that packages are built in a specific manner, so that sha1 gets embedded.  If you have a better solution, I'm all ears...

PR #986, opened 2016-11-17
valgrind: suppression for md_config_t string leak

cbodley:
Adds a valgrind suppression for a very strange memory leak only seen in radosgw under teuthology. When it reproduces, it always complains about the same 4 std::strings under md_config_t, with sizes 28, 51, 61, and 63.

I've been trying for months to track down the root cause, without success. I wasn't able to locate an offending commit via git bisect. I haven't seen it reproduce on packages built by shaman, so I suspect it's some artifact of the gitbuilders. We really want to get the rgw suite green so that kraken backports can be verified, so we're adding valgrind suppression as a last resort.

Fixes: http://tracker.ceph.com/issues/17924

### DIFF
--- a/teuthology/task/install/valgrind.supp
+++ b/teuthology/task/install/valgrind.supp
@@ -370,3 +370,14 @@
 	fun:_dl_init
 	...
 }
+{
+	strange leak of std::string memory from md_config_t seen in radosgw
+	Memcheck:Leak
+	...
+	fun:_ZNSs4_Rep9_S_createEmmRKSaIcE
+	fun:_ZNSs12_S_constructIPKcEEPcT_S3_RKSaIcESt20forward_iterator_tag
+	...
+	fun:_ZN11md_config_tC1Ev
+	fun:_ZN11CephContextC1Eji
+	...
+}


cbodley:
scheduled at http://pulpito.ceph.com/cbodley-2016-11-17_11:46:57-rgw-wip-17792---basic-mira

cbodley:
marked DNM, trying another approach suggested by @liewegas

liewegas:
<pre>
../src/rgw/rgw_rados.cc:    int ret = cct->_conf->set_val("rgw_zonegroup", cct->_conf->rgw_region, true, false);
</pre> looks like a likely culprit

cbodley:
@liewegas I tried moving this unsafe `set_val()` into rgw_main.cc before `common_init_finish()`, but the valgrind failures persist - see https://github.com/ceph/ceph/pull/12051

cbodley:
@liewegas a fix for that unsafe change to `rgw_zonegroup` was merged in https://github.com/ceph/ceph/pull/12075, but the leaks persist. would you consider merging this suppression?

cc @mattbenjamin @yehudasa

liewegas:
I would like to properly fix this :(

PR #989, opened 2016-11-30
Pre-support for colocated suite

zmc:
We want to be able to move the contents of `ceph-qa-suite.git` into `ceph.git`. This PR adds the functionality needed to test the remaining changes that are required for that support.

@liewegas: This is built around the user only having to supply a single argument to `teuthology-suite`: `-C` or `--colocated-suite <relpath>`; in your case you'd use `-C qa` to point inside the subdir.

### DIFF
--- a/teuthology/repo_utils.py
+++ b/teuthology/repo_utils.py
@@ -267,14 +267,21 @@ def fetch_repo(url, branch, bootstrap=None, lock=True):
     return dest_path
 
 
-def fetch_qa_suite(branch, lock=True):
+def fetch_qa_suite(branch, lock=True, colocated_suite=False):
     """
     Make sure ceph-qa-suite is checked out.
 
-    :param branch: The branch to fetch
-    :returns:      The destination path
+    :param branch:           The branch to fetch
+    :param lock:             See fetch_repo()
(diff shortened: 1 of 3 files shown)


liewegas:
This still allows for a --suite-branch to be specified, right? Because I 
think that will still be important in many cases.  E.g., I already have a 
build i want to test, but I'm fiddling with the tests.  In that case, I'd 
probably actually want to specify the --suite-repo as 
github.com/liewegas/ceph.git and avoid pushing to the ceph[-ci].git 
entirely and triggering new builds.

The approach I had in mind was:

 - add --suite-repo (defaults to github.com/ceph/ceph-qa-suite.git, later 
to ceph[-ci].git)
 - add --suite-repo-path (default to '' now, 'qa' later)
 - make --suite-branch default to whatever -c is (i think it does that 
now?)

?

zmc:
@liewegas it would still support different branches, yeah. If you use `-c master` and `-C qa --suite-branch wip-foo` it would use a `ceph` repo with `wip-foo` checked out for the suite bits.

The one thing it wouldn't support is if you for some reason want to test `ceph.git` with suite bits in e.g. `ceph-ci.git`

zmc:
Going to take another approach

PR #990, opened 2016-11-30
Allow using an alternate suite repo

zmc:
... and also optionally look for tasks and suite definitions in a subdirectory of that repo

Related:
http://pad.ceph.com/p/qa-suite-merge
https://paste.fedoraproject.org/493541/04623891/

### DIFF
--- a/teuthology/repo_utils.py
+++ b/teuthology/repo_utils.py
@@ -240,12 +240,11 @@ def fetch_repo(url, branch, bootstrap=None, lock=True):
     :param branch:     The branch we want
     :returns:          The destination path
     """
-    # 'url/to/project.git' -> 'project'
-    name = url.split('/')[-1].split('.git')[0]
     src_base_path = config.src_base_path
     if not os.path.exists(src_base_path):
         os.mkdir(src_base_path)
-    dest_path = os.path.join(src_base_path, '%s_%s' % (name, branch))
+    dirname = '%s_%s' % (url_to_dirname(url), branch)
+    dest_path = os.path.join(src_base_path, dirname)
     # only let one worker create/update the checkout at a time
     lock_path = dest_path.rstrip('/') + '.lock'
     with FileLock(lock_path, noop=not lock):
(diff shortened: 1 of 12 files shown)


zmc:
Note to self: repos on-disk need to be named to avoid collisions

teuthology/suite/run.py
```diff
@@ -189,30 +194,42 @@ def choose_teuthology_branch(self):
...
+            return 'ceph-qa-suite'
+
     def choose_suite_branch(self):
+        suite_repo_name = self.suite_repo_name
         suite_branch = self.args.suite_branch
         ceph_branch = self.args.ceph_branch
         if suite_branch and suite_branch != 'master':
-            if not util.git_branch_exists('ceph-qa-suite', suite_branch):
-                exc = BranchNotFoundError(suite_branch, 'ceph-qa-suite.git')
+            if not util.git_branch_exists(suite_repo_name, suite_branch):
```
liewegas: Does this assume that it's a repo under github.com/ceph?

zmc: Ah, it will but it doesn't have to be that way of course. I'll work on that in the morning

zmc:
@liewegas I *believe* this does what we want now!

zmc:
Testing new worker code with new suite features:
http://pulpito.ceph.com/zack-2016-12-02_09:49:30-smoke-master---basic-mira/596203/

Testing new worker code without new suite features:
http://pulpito.ceph.com/zack-2016-12-02_09:50:46-smoke-master---basic-mira/596205/

Both are doing what I would expect.

zmc:
I'm going to push some unit-test-only changes

dmick:
Do you want to wait for those unit test changes?

zmc:
@dmick I pushed them Friday morning

dmick:
so confused; gh doesn't show any new commits after your comment.  sigh.

zmc:
It was a force push on top of an existing commit

teuthology/repo_utils.py
```diff
@@ -97,20 +97,25 @@ def enforce_repo_state(repo_url, dest_path, branch, remove_on_error=True):
         raise
 
 
-def clone_repo(repo_url, dest_path, branch):
+def clone_repo(repo_url, dest_path, branch, shallow=True):
```
zmc: Are you sure? I couldn't find any.

dmick: It's all inside repo_utils, but it gets used for teuthology and ceph-qa-suite both.  The other one I saw was in the cephfs tests in ceph-qa-suite, but I see now that's a locally-defined routine of the same name.

Mostly I was raising the question to make certain you'd doublechecked all the call paths.  I can't think of why teuthology would care that it's depth 1, but it's a change.

zmc: So yeah, there's one caller of `clone_repo` - it's the one that needs the changes here.

I can't imagine why a shallow clone would disturb anything else teuthology does. It's eternally possible that I've missed something, though :)

dmick: yeah.  and nothing appears to use fetch_branch but tests.  shrug.  It's likely fine.

liewegas:
a shallow clone also seems safe to me.  +1

PR #992, opened 2016-12-02
kernel: fix reboot into the new kernel on xenial

idryomov:
shutdown is a symlink to systemctl on xenial, and on "shutdown -r now"
network gets shut down before ssh has a chance to close gracefully.
There have been multiple complaints about this and some distantly
related fixes have made it into recent systemds [1].

This doesn't seem to happen on centos, probably because of different
systemd versions or sshd configuration, so this band-aid is for ubuntu
only.  "systemctl --no-block reboot" is nicer, but not available on
trusty.

[1] https://github.com/systemd/systemd/issues/2390

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -596,10 +596,18 @@ def install_and_reboot(ctx, config):
                 'rm',
                 remote_pkg_path(role_remote),
                 run.Raw('&&'),
+                # work around a systemd issue, where network gets shut down
+                # before ssh can close its session
+                run.Raw('('),
+                'sleep',
+                '1',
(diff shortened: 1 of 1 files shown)


idryomov:
Before (dead jobs):
http://pulpito.ceph.com/teuthology-2016-11-27_10:15:02-krbd-jewel-testing-basic-smithi/
http://pulpito.ceph.com/teuthology-2016-11-30_10:15:01-krbd-jewel-testing-basic-smithi/

After (just 3 jobs, but all xenial):
http://pulpito.ceph.com/dis-2016-12-02_15:34:17-krbd:unmap-jewel-testing-basic-smithi/

PR #994, opened 2016-12-06
Default use_shaman to True

zmc:


zmc:
I've updated the Sepia cronjobs to disable Shaman when scheduling the Samba suite, since it's not ready yet.

liewegas:
yay!

PR #998, opened 2016-12-14
Reflect the move from ceph-qa-suite.git -> ceph.git

zmc:


liewegas:
lgtm.  the usage args thing is magic!

PR #1000, opened 2016-12-15
ShamanProject._tag_to_sha1(): retry w/ ceph.git

zmc:
For upgrade tests that are otherwise using ceph-ci.git, we need to also
look in ceph.git to lookup released tags.

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -889,17 +889,35 @@ class ShamanProject(GitbuilderProject):
         """
         Shaman doesn't know about tags. Use git ls-remote to query the remote
         repo in order to map tags to their sha1 value.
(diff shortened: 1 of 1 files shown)


teuthology/packaging.py
```diff
@@ -889,17 +889,36 @@ def _tag_to_sha1(self):
         """
         Shaman doesn't know about tags. Use git ls-remote to query the remote
         repo in order to map tags to their sha1 value.
-        """
+
+        This method will also retry against ceph.git if the original request
+        uses ceph-ci.git and fails.
+        """
+        def get_sha1(url):
+            print url
```
liewegas: drop this line?

PR #1001, opened 2016-12-15
platform.linux_distribution() is deprecated; stop using it

smithfarm:
Fixes: http://tracker.ceph.com/issues/18141

### DIFF
--- a/teuthology/orchestra/opsys.py
+++ b/teuthology/orchestra/opsys.py
@@ -3,8 +3,13 @@ import re
 DISTRO_CODENAME_MAP = {
     "ubuntu": {
         "18.04": "bionic",
+        "17.10": "artful",
+        "17.04": "zesty",
+        "16.10": "yakkety",
         "16.04": "xenial",
+        "15.10": "wily",
         "15.04": "vivid",
+        "14.10": "utopic",
         "14.04": "trusty",
         "13.10": "saucy",
         "12.04": "precise",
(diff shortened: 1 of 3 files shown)


smithfarm:
@zmc Ping

smithfarm:
platform.linux_distribution() is deprecated since python 3.5, and will be removed in python 3.7 (see https://docs.python.org/3.6/library/platform.html#platform.linux_distribution).

smithfarm:
The implementation in Python 2.7 returns `(,,)` on newer distros like openSUSE Tumbleweed (which dropped the old `/etc/SuSE-release` now that `/etc/os-release` is the new standard). I think this `(,,)` return value happens in newer Fedoras as well, but I haven't tried it.

smithfarm:
@zmc When you say "all OSes" which OSes do you mean? Are there any that don't have `/etc/os-release` ?

ktdreyer:
FYI [mock](https://github.com/rpm-software-management/mock) recently switched to the [distro](https://github.com/nir0s/distro) lib, now that `platform.dist` and `platform.linux_distribution` are going away in py37.

smithfarm:
@ktdreyer Do we need an elaborate python module when a glance at `/etc/os-release` gives us the information we need?

ktdreyer:
Teuthology does a lot already - it was just a suggestion :)

smithfarm:
In https://github.com/ceph/ceph/pull/23111 I have unit tests that cover "all" OSes. I will add a commit to this PR implementing these unit tests in teuthology so we can be confident that `platform.linux_distribution` is no longer needed.

smithfarm:
Added unit tests and support for new distros. Pinging again for review.

smithfarm:
@zmc Ping

PR #1003, opened 2016-12-20
spawn_sol_log: use sys.executable to find bin/python

dmick:
An activated virtualenv has PATH set to find Python with /usr/bin/env.
A binary run from virtualenv/bin doesn't set PATH.  Use sys.executable
to handle both invocation methods.

Fixes: http://tracker.ceph.com/issues/17986


dmick:
Ran a trivial test on a smithi and verified that the console log was collected.  pexpect is only present in my teuthology virtualenv.

PR #1006, opened 2017-01-10
Remove high level of debug logging which is not required

vasukulkarni:
High levels of debug messages are not required for ceph-deploy tests, also its causing remoto to hang during get of clent.admin key


vasukulkarni:
just to add more info, without this we are going to hit timeout's in remoto as seen here http://qa-proxy.ceph.com/teuthology/vasu-2017-01-10_03:37:51-ceph-deploy-kraken---basic-vps/703474/teuthology.log  and also the other part being the logging is just way to much to meaningfully debug ceph-deploy outputs which is what we are more interested in the ceph-deploy test.

PR #1007, opened 2017-01-14
Add redhat internal tasks that setup repo

vasukulkarni:
setup repo based on config items, either use
cdn or compose sytle urls or plain repo's.

Still testing with other change,

### DIFF
--- a/teuthology/task/internal/redhat.py
+++ b/teuthology/task/internal/redhat.py
@@ -1,0 +1,143 @@
+"""
+Internal tasks  for redhat downstream builds
+"""
+import contextlib
+import logging
+import requests
+from tempfile import NamedTemporaryFile
+from teuthology.parallel import parallel
+from teuthology.orchestra import run
+from teuthology.task.install.redhat import set_deb_repo
+
+log = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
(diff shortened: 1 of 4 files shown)


zmc:
Curious - why not take advantage of the `downstream_setup` role that Andrew created ages ago? Wasn't it created for exactly this purpose?

GregMeno:
@vasukulkarni when do you think this will be out of [DNM] ?

vasukulkarni:
I have this already working, need to cleanup commits, setting up meeting with zack, ken to get most of issues sorted out and merged.

teuthology/task/internal/redhat.py
```diff
@@ -0,0 +1,143 @@
...
+@contextlib.contextmanager
+def setup_cdn_repo(ctx, config):
+    """
+     setup repo if set_cdn_repo exists in config
+     redhat:
+      set_cdn_repo:
+        rhbuild: 2.0 or 1.3.2 or 1.3.3
+    """
+    # do import of tasks here since the qa task path should be set here
+    if ctx.config.get('redhat').get('set-cdn-repo', None):
```
zmc: If `redhat` is undefined, you'll still get an `AttributeError` here, so the first `get()` won't help much

PR #1009, opened 2017-01-16
Avoid a race condition with job archive creation

zmc:
This will hopefully fix http://tracker.ceph.com/issues/18482


zmc:
cf07393 is just a bonus I discovered while testing. It is almost certainly contributing to the issue, though.

PR #1010, opened 2017-01-17
Dont set qa-suite repo during worker startup

vasukulkarni:
There is no need to initialize qa-suite repo during worker startup,
moreover it fails with default values trying to checkout 'master'
branch from ceph-ci.

### DIFF
--- a/teuthology/worker.py
+++ b/teuthology/worker.py
@@ -93,7 +93,6 @@ def main(ctx):
 
     if teuth_config.teuthology_path is None:
         fetch_teuthology('master')
-    fetch_qa_suite('master')
 
     keep_running = True
     while keep_running:


zmc:
Can I see more information on the bug this wants to fix please?

vasukulkarni:
sorry here's the traceback: https://paste.fedoraproject.org/531660/93521614/raw/

zmc:
We set this in the workers' `teuthology.yaml` in sepia - this may be what you need:
```
ceph_git_base_url: git://git.ceph.com/git/
ceph_git_url: git://git.ceph.com/ceph.git
ceph_qa_suite_git_url: git://git.ceph.com/ceph.git
```

PR #1013, opened 2017-01-19
Change to use task 'clock' instead of 'clock.check'

dmick:
See commit comments, but, clock will force a sync with ntpdate, which is probably the better path.

Tested with a local teuthology run and a small suite excerpt (look for "Syncing clocks"):

http://pulpito.ceph.com/dmick-2017-01-18_13:39:43-rados-kraken---basic-mira/

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -234,7 +234,7 @@ def get_initial_tasks(lock, config, machine_type):
             {'pcp': None},
             {'selinux': None},
             {'ansible.cephlab': None},
-            {'clock.check': None}
+            {'clock': None}
         ])
 
     return init_tasks
--- a/teuthology/task/clock.py
+++ b/teuthology/task/clock.py
@@ -54,7 +54,7 @@ def task(ctx, config):
             'PATH=/usr/bin:/usr/sbin',
             'ntpdc', '-p',
         ])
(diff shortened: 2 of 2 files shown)


dmick:
Ping, more jobs broke because of clock sync:  
(04:45:43 PM) jcsp: something's probably wrong with NTP on one or more smithi nodes, a slew of jobs just failed with mon clock skew
(04:45:51 PM) jcsp: http://pulpito.ceph.com/jspray-2017-01-26_23:25:23-fs-wip-jcsp-testing-20170126-distro-basic-smithi/

dmick:
@zmc points out that the service name differs on CentOS (ntpd).  In changing/testing, I find that ntpdc apparently doesn't work on xenial.

I hate my life.

dmick:
ntpdc is deprecated,  replacing with ntpq.  now if I can figure out why ntpd refuses to start on my test node...

dmick:
ntpd fails on xenial because of Debian bugs.  See https://github.com/ceph/ceph-cm-ansible/pull/305 for a workaround (and reference to the issue)

I think this PR is ok once we merge https://github.com/ceph/ceph-cm-ansible/pull/305

dmick:
@zmc I think this is ready to go

PR #1017, opened 2017-01-24
Use manhole to provide a way to debug hung jobs

zmc:
https://pypi.python.org/pypi/manhole


zmc:
I have a few jobs running with this commit now and it doesn't seem to break anything.

PR #1023, opened 2017-02-02
nuke: If normal reboot fails, power cycle

zmc:
I'm pretty surprised that we weren't doing this already.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -444,6 +444,12 @@ class Remote(object):
             self._console = getRemoteConsole(self.name)
         return self._console
 
+    @property
+    def uptime(self):
+        proc = self.run(args=['cat', '/proc/uptime'], stdout=StringIO())
+        proc.wait()
+        return float(proc.stdout.read().split(' ')[0])
+
     def __del__(self):
         if self.ssh is not None:
             self.ssh.close()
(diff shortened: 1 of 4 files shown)


idryomov:
Could you please hold on on merging this until I try a few things?

teuthology/orchestra/remote.py
```diff
@@ -444,6 +444,12 @@ def console(self):
             self._console = getRemoteConsole(self.name)
         return self._console
 
+    @property
+    def uptime(self):
+        proc = self.run(args=['cat', '/proc/uptime'], stdout=StringIO())
+        proc.wait()
+        return float(proc.stdout.read().split(' ')[0])
+
```
idryomov: The whole "connect and then power cycle from the outside" seems wrong.  I've posted PR #1024 that should hopefully supersede this.

dmick:
We decided against this in favor of @idryomov's solution, right?  Shall we close this if so?

idryomov:
Yeah, I think we agreed on https://github.com/ceph/teuthology/pull/1024#issuecomment-277349956.  I have a couple of items on my TODO, but I haven't seen any new reports from @yuriw.

dmick:
@zmc, do you agree this is dead?

zmc:
I think so.

PR #1024, opened 2017-02-03
nuke: work around a reboot -n trouble

idryomov:


idryomov:
@zmc I've tested this on @yuriw's t-nuke --stale pile.  Some had their rootfs filled up with stray OSD logs and required manual intervention, but that has nothing to do with this change.

Right now the -n branch is taken only if stale kernel client state is detected (so k* runs only), but I plan to try to extend it to cover some generic kernel issues as well (such as soft lockups on old distro kernels during rados runs).  A power cycle fall back is definitely valuable, but in order to cover the rest it needs to work from the outside and not try to ssh in, so I say let's merge this today and work on a better version of PR #1023 next week (or see for how long we can get by without a fall back, just as an exercise).

zmc:
@idryomov That sounds good. We can follow up with other PRs next week. Thanks!

PR #1025, opened 2017-02-07
console: force existing connections into spy mode if !readonly

idryomov:
If someone watching the console didn't think of using "console -s", we
end up power cycling the node in an attempt to get the login prompt.
This is futile -- if the watcher is still there after the node comes
back up, our connection will get dropped to spy mode again.

Use -f to temporarily force existing connections into spy mode when we
attach to save a power cycle.


djgalloway:
I wondered if/when this would become a necessity  +1 from me.  Wish I had mentioned it sooner.

PR #1028, opened 2017-02-09
openstack: get it working again

dachary:
This is a large patch that could be divided into smaller parts. However,
some of these fixes lingered for months and it seems useful to get all
of it together for the record.


dachary:
this is the result of collecting fixes from various sources, putting them together and making it so the integration test work. It is not intented as a pull request to be merged, although it could. It is more likely to be useful to someone willing to get teuthology-openstack working at this point in time.

dmick:
O_o

kshtsk:
Closed the PR in favor of #1326

PR #1030, opened 2017-02-16
suite: use "suite_hash" as the default sha1 for workunit

tchaikov:
as "workunits" reside in ceph/qa/workunits, it's more intuitive to
respect suite-branch option when cloning workunits.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -345,6 +345,9 @@ def main(args):
     ceph_repo = config.get('repo')
     if ceph_repo:
         teuth_config.ceph_git_url = ceph_repo
+    suite_repo = config.get('suite_repo')
+    if suite_repo:
+        teuth_config.ceph_qa_suite_git_url = suite_repo
 
     config["tasks"] = validate_tasks(config)
 
--- a/teuthology/suite/placeholder.py
+++ b/teuthology/suite/placeholder.py
@@ -89,7 +89,7 @@ dict_templ = {
             }
         },
         'workunit': {
-            'sha1': Placeholder('ceph_hash'),
+            'sha1': Placeholder('suite_hash'),
         }
     },
     'repo': Placeholder('ceph_repo'),


dmick:
This affects all of teuthology and all tasks, though, not just workunit; I'm not sure that's what we'd want, or how to tell what might break

idryomov:
@tchaikov I think we need one more change here: https://github.com/ceph/teuthology/commits/wip-teuth-config-suite-repo.  Otherwise it attempts to clone from git://git.ceph.com/ceph.git.

Do you want to pull into your branch or should I open a separate PR?

tchaikov:
@idryomov cherry-picked and repushed. this PR is a bug fix now!

vasukulkarni:
shouldn't the workunit match whatever ceph version is under test? as specified by -c flag, on the other hand I have been requesting to make workunits more generic so that they dont fail if not tested with right ceph version and skip tests that are not valid.

idryomov:
@vasukulkarni It will match the ceph version under test, if you don't specify --suite-repo / --suite-branch.
All this does is allows you to change that, should you want to.

vasukulkarni:
@idryomov bit confusing, the placeholder change will overwrite all the workunits config, where will it select ceph version when no option is specified?

idryomov:
@vasukulkarni suite_hash defaults to ceph_hash, so if you don't specify --suite-repo / --suite-branch you'll get ceph_hash in there -- the current behavior IOW.

dmick:
It makes sense to me.  I'd also like to see the commandline argument from teuthology, --suite-path, be able to set the path for workunits, but that's yet another PR

vasukulkarni:
I think we definitely want to test this for 1 or 2 days with this, although it makes sense to me that it should match suite-repo, I am thinking there will be cases where we want to use ceph version even when we specify the separte suite branch for some of the qa wip branches during testing.

idryomov:
As mentioned in https://github.com/ceph/ceph/pull/13495, I tested this with http://pulpito.ceph.com/dis-2017-02-17_15:43:00-krbd-luminous-testing-basic-smithi/.

I don't see why, when running tests from branch A against branch B, you would want to use workunit scripts (i.e. a small subset of tests) from branch B.  If you really want to do that, just pull those scripts into A.

My use case is trying out new workunit scripts with existing (i.e. already built) (named) ceph branches.  For example, I paired a workunit from my ceph fork with luminous binaries from a week ago in that run.

vasukulkarni:
@idryomov that is my use case as well,  but now since we are changing the behavior one has to  be careful when making wip-* branches which doesn't change workunits but work on other scripts(eg install script changes etc) 

probably we can make that as an additional cli option to enforce this option and keep default as is? 

Again for both the use cases we have overrides option to override the workunit branch option, but what I am asking here is we should keep the default as is, since that is the most common use case.

zmc:
This looks good to me - any objections?

vasukulkarni:
@zmc sure feel free to merge

PR #1036, opened 2017-03-01
misc.wait_until_{healthy,osds_up}: support timeout param

tchaikov:
so we are able to specify the timeout for them, when valgrind is
enabled, OSDs takes more time to boot and active,

Refs: http://tracker.ceph.com/issues/19113

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -879,12 +879,12 @@ def get_scratch_devices(remote):
     return retval
 
 
-def wait_until_healthy(ctx, remote, ceph_cluster='ceph'):
+def wait_until_healthy(ctx, remote, ceph_cluster='ceph', timeout=900):
     """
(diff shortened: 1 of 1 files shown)


smithfarm:
@kefu Is the valgrind delay related to boost being in a submodule (statically linked?) now?

tchaikov:
@smithfarm i am not sure. static linkage only prevents valgrind, for example,  from replacing some malloc related function, so it will be faster. but in our case, we still links against glibc dynamically. so it's not relevant, i guess. but it's a good direction to investigate. i should try run valgrind with osd locally with/without libboost from system.

tchaikov:
since https://github.com/ceph/ceph/pull/13748 is merge, i am closing this. maybe it will be useful in the future? who knows.

PR #1037, opened 2017-03-01
cloud.openstack: Also retry on BaseHTTPError

zmc:
We attach volumes immediately after creating them; sometimes they are
still momentarily in the 'creating' state, causing the attach call to
throw a BaseHTTPError. When that happens, simply retry the request
instead of failing node creation, starting the entire cycle all over
again.


dmick:
That is a truly fascinating exception name

PR #1040, opened 2017-03-06
provision.downburst: Tweak destroy return code

zmc:
Specifically, if the instance doesn't even exist, consider the destroy
op to have succeeded.


zmc:
@dmick oh, that would have been fair too

PR #1041, opened 2017-03-06
Invoke downburst with logging

dmick:


teuthology/provision/__init__.py
```diff
@@ -6,11 +6,18 @@
...
 import downburst
 import openstack
+import os
 
 
 log = logging.getLogger(__name__)
 
 
+def _logfile(ctx, shortname):
+    if (ctx.config['archive_path']):
```
zmc: Is `config['archive_path']` always present? Would it be safer to:
```
if ctx.config.get('archive_path', None) is not None:
```
?

dmick: yep, good catch, fixing

PR #1045, opened 2017-03-10
Fix a couple bugs re: remote daemon execution

zmc:


dmick:
Should probably wait to merge until the lab is back, so we can catch immediate problems

PR #1046, opened 2017-03-13
run.py: Provide --os-{type,version} values to get_initial_tasks

smithfarm:
We want to be able to check the final values of os_type and os_version
in get_initial_tasks, so move the command-line override up to before
the get_initial_tasks() method call.


smithfarm:
So this is ready to merge?

gregsfortytwo:
Sure, looks fine to me. @zmc, any objections?

smithfarm:
@zmc Is this ready to merge?

PR #1047, opened 2017-03-13
orchestra/daemon: separate register_daemon from add_daemon

liewegas:
The former registers without starting; the latter does both (as before).

### DIFF
--- a/teuthology/orchestra/daemon.py
+++ b/teuthology/orchestra/daemon.py
@@ -150,7 +150,25 @@ class DaemonGroup(object):
     def add_daemon(self, remote, type_, id_, *args, **kwargs):
         """
         Add a daemon.  If there already is a daemon for this id_ and role, stop
-        that daemon and.  Restart the damon once the new value is set.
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/daemon.py
```diff
@@ -160,6 +160,24 @@ def add_daemon(self, remote, type_, id_, *args, **kwargs):
...
         # we can only get optional args from unused kwargs entries
+        self.register_daemon(remote, type_, id_, *args, **kwargs)
+        cluster = kwargs.pop('cluster', 'ceph')
+        role = cluster + '.' + type_
+        self.daemons[role][id_].restart()
+
+    def register_daemon(self, remote, type_, id_, *args, **kwargs):
+        """
+        Add a daemon.  If there already is a daemon for this id_ and role, stop
+        that daemon and.
```
zmc: "stop that daemon and." ?

liewegas:
@zmc fixed comment. this is working fine in my test runs.. ok to merge?

PR #1048, opened 2017-03-16
setup.py: disallow requests 2.13.0

dmick:
python-cinderclient and/or keystoneauth1 apparently don't want
requests 2.13.0, so make it illegal for now


dmick:
https://github.com/openstack/requirements/commit/08b589c5ad0f0f49d8d5880f3a703cfae43b0a39 seems to be why/how this was blacklisted.

PR #1049, opened 2017-03-17
misc: add flexibility to delete_ and move_file

jan--f:
Add check argument to delete_file to provide the ability to ignore a
failed delete (say if the file to delete does not exist)
Add preserve_perms argument to move_file to make the preservation of
permissions optional. This makes move_file usable if the target
file does not yet exist.
Existing invocations are unchanged.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>
(cherry picked from commit f9aef0126313a24b791bc16918eb47d113ee6139)

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -550,27 +550,31 @@ def copy_file(from_remote, from_path, to_remote, to_path=None):
     ])
 
 
-def move_file(remote, from_path, to_path, sudo=False):
+def move_file(remote, from_path, to_path, sudo=False, preserve_perms=True):
     """
     Move a file from one path to another on a remote site
 
-    The file needs to be stat'ed first, to make sure we
-    maintain the same permissions
+    If preserve_perms is true, the contents of the destination file (to_path,
+    which must already exist in this case) are replaced with the contents of the
+    source file (from_path) and the permissions of to_path are preserved. If
+    preserve_perms is false, to_path does not need to exist, and is simply
+    clobbered if it does.
     """
-    args = []
-    if sudo:
-        args.append('sudo')
-    args.extend([
-        'stat',
-        '-c',
-        '\"%a\"',
-        to_path
-    ])
-    proc = remote.run(
-        args=args,
-        stdout=StringIO(),
-    )
-    perms = proc.stdout.getvalue().rstrip().strip('\"')
+    if preserve_perms:
+        args = []
+        if sudo:
+            args.append('sudo')
+        args.extend([
+            'stat',
+            '-c',
+            '\"%a\"',
+            to_path
+        ])
+        proc = remote.run(
+            args=args,
+            stdout=StringIO(),
+        )
+        perms = proc.stdout.getvalue().rstrip().strip('\"')
 
     args = []
     if sudo:
@@ -586,24 +590,26 @@ def move_file(remote, from_path, to_path, sudo=False):
         stdout=StringIO(),
     )
 
-    # reset the file back to the original permissions
-    args = []
-    if sudo:
-        args.append('sudo')
-    args.extend([
-        'chmod',
-        perms,
-        to_path,
-    ])
-    proc = remote.run(
-        args=args,
-        stdout=StringIO(),
-    )
+    if preserve_perms:
+        # reset the file back to the original permissions
+        args = []
+        if sudo:
+            args.append('sudo')
+        args.extend([
+            'chmod',
+            perms,
+            to_path,
+        ])
+        proc = remote.run(
+            args=args,
+            stdout=StringIO(),
+        )
 
 
 def delete_file(remote, path, sudo=False, force=False):
     """
-    rm a file on a remote site.
+    rm a file on a remote site. Use force=True if the call should succeed even
+    if the file is absent or rm path would otherwise fail.
     """
     args = []
     if sudo:


dmick:
a couple of minor issues so far: 1) try not to introduce new pep8 issues (I see a couple of instances of if(condition))  2) in the commit message, you typo "preserver_parms" instead of "preserve_parms"

dmick:
ok to test

jan--f:
@dmick thx...fixed.

smithfarm:
@jan--f: Can you drop the "cherry-picked from" line, too?

jan--f:
@smithfarm dropped.

teuthology/misc.py
```diff
@@ -586,22 +590,23 @@ def move_file(remote, from_path, to_path, sudo=False):
...
+        if sudo:
+            args.append('sudo')
+        args.extend([
+            'chmod',
+            perms,
+            to_path,
+        ])
+        proc = remote.run(
+            args=args,
+            stdout=StringIO(),
```
dmick: but you're not consuming stdout, so why set it up to a StringIO() object?

teuthology/misc.py
```diff
@@ -586,22 +590,23 @@ def move_file(remote, from_path, to_path, sudo=False):
...
+        )
 
 
-def delete_file(remote, path, sudo=False, force=False):
+def delete_file(remote, path, sudo=False, force=False, check=True):
```
dmick: Why do we need this given that we have force, which ignores "file not found" errors as well?

teuthology/misc.py
```diff
@@ -550,27 +550,31 @@ def copy_file(from_remote, from_path, to_remote, to_path=None):
...
+    clobbered if it does.
+    """
+    if preserve_perms:
+        args = []
+        if sudo:
+            args.append('sudo')
+        args.extend([
+            'stat',
+            '-c',
+            '\"%a\"',
```
dmick: Are the backslashes for the further shell escape?  This is what "run.Raw()" is for, but I also don't think the quotes are necessary:

>>> r.run(args=['echo', '%a'])
2017-03-20 09:13:46,118.118 INFO:teuthology.orchestra.run.mira001:Running: 'echo %a'
2017-03-20 09:13:46,302.302 INFO:teuthology.orchestra.run.mira001.stdout:%a

Edit: I see you were just moving code.  Still.

teuthology/misc.py
```diff
@@ -550,27 +550,31 @@ def copy_file(from_remote, from_path, to_remote, to_path=None):
...
+            to_path
+        ])
+        proc = remote.run(
+            args=args,
+            stdout=StringIO(),
+        )
+        perms = proc.stdout.getvalue().rstrip().strip('\"')
```
dmick: ...which means you wouldn't need this stripping here too:

```
>>> r.run(args=['stat', '-c', '%a', '/tmp'])
2017-03-20 09:15:16,281.281 INFO:teuthology.orchestra.run.mira001:Running: 'stat -c %a /tmp'
2017-03-20 09:15:16,466.466 INFO:teuthology.orchestra.run.mira001.stdout:1777

```

teuthology/misc.py
```diff
@@ -550,27 +550,31 @@ def copy_file(from_remote, from_path, to_remote, to_path=None):
     ])
 
 
-def move_file(remote, from_path, to_path, sudo=False):
+def move_file(remote, from_path, to_path, sudo=False, preserve_perms=True):
```
dmick: So the description says the desire is to make this workable even if the destination doesn't exist, but, the method could simply ignore the 'no perms to update' without requiring the caller to request it.  The issue is, however, what permissions for the destination file ought to be used (whatever the source has, or some default, or?...)

I think I'd be happier if the change here were to add a permission for the case of the file not existing (as an optional kw parameter, still).  If to_path doesn't exist, then use perm if supplied, otherwise default to the perm of from_path.  What do you think?

teuthology/misc.py
jan--f: True, I did not consider using -f. Will update.

teuthology/misc.py
jan--f: Yeah I just indented existing code. I see your points, but shouldn't this be addressed in a separate PR?

teuthology/misc.py
jan--f: I think it is better this way tbh. ```move_file``` already does quite a bit more then its name suggests. Starting to now also set permissions (but only if ```to_path``` doesn't exist) seems to add complexity beyond this functions scope.
Personally I'd expect it to behave like ```mv```, so preserve permissions of the source file. I added the option to keep things compatible with existing code.

teuthology/misc.py
dmick: yep.

teuthology/misc.py
dmick: I guess that's the part that rankles a bit; it's changing the intent of the original definition, however misguided, while leaving the default behavior, but not going the step further to add the capability of choosing the actual permission.  But I can see that it's certainly a matter of opinion which way is more in keeping with expectations.  I withdraw the suggestion.

dmick:
Sorry I let this drop @jan--f.  @zmc, if you're ok with it I'm ok.

PR #1051, opened 2017-03-21
DNM: Multi-OS locking

zmc:
I've been working on this for the last few days and it's close - but it needs more testing before it's merged.


teuthology/run.py
```diff
@@ -179,26 +180,89 @@ def validate_tasks(config):
...
-        init_tasks.append({'internal.lock_machines': (
-            len(config['roles']), machine_type)})
+
+    target_conflict_msg = 'You cannot specify targets in a config file when' \
+        'using the --lock option'
+    if lock and ('roles' in config or 'nodes' in config):
+        assert 'targets' not in config, target_conflict_msg
+    if lock and 'roles' in config:
+        if 'nodes' in config:
+            log.warn(
```
vasukulkarni: I think we should exit here, It would be a user error to have both and fix them to use only one format

zmc: I'd be open to that

kshtsk:
I would like to have a documentation update of how this feature looks from user point of view, 'node' format and use cases.

smithfarm:
If someone took over where @zmc left off and landed this feature, that would be a huge accomplishment IMO. This feature is sorely missing in teuthology.

kshtsk:
@vasukulkarni @zmc I'm looking into this PR, did you have any suite for testing this already?

batrick:
https://tracker.ceph.com/issues/43612#note-2

(agree with you @kshtsk )

PR #1055, opened 2017-03-30
suite: use "suite_branch" for workunit if avaiable

tchaikov:
"git clone" allows us to clone a single branch, but not a sha1, so
prefer "branch" if it's available.


dillaman:
FYI -- this is the reason why the upgrade tests started to fail (since the upgrade tests specify specific branches to pull from and this new global "branch" override was stomping on it)

tchaikov:
@dillaman thanks for tracking down the root cause.

PR #1056, opened 2017-03-30
Provide a way for jobs to detect daemon failure

zmc:


teuthology/orchestra/daemon.py
```diff
@@ -43,7 +44,10 @@ def stop(self, timeout=300):
             return
         self.proc.stdin.close()
         self.log.debug('waiting for process to exit')
-        run.wait([self.proc], timeout=timeout)
+        try:
+            run.wait([self.proc], timeout=timeout)
+        except CommandFailedError:
```
dmick: does CommandFailedError result from a failed start that we're only now noticing?

PR #1058, opened 2017-03-31
Fix run-playbook errors

wusui:
The setuptools and ansible_ver fields need to be properly quoted in the remote run.
The ansible operation that is remotely run needs to be broken up so that 'ansible-ceph -vv...'
is interpreted properly.
The workaround recommended in https://access.redhat.com/solutions/2943461 has been implemented.

Fixes: 19425

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -336,7 +336,7 @@ class CephAnsible(Task):
         ceph_installer.run(args=('cat', 'ceph-ansible/site.yml'))
         ceph_installer.run(args=('cat', 'ceph-ansible/group_vars/all'))
         str_args = ' '.join(args)
-        ceph_installer.run(args=[
+        next_args=[
             run.Raw('cd ~/ceph-ansible'),
             run.Raw(';'),
             'virtualenv',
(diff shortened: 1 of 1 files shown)


vasukulkarni:
This shouldn't be merged,  AFAIK there are no issues with recent runs in pulpito.

vasukulkarni:
The title is misleading, there are no errors in ceph-ansilble test runs, If you are looking to add journal disk you can also add them to group_vars/all in test yaml and ensure the test node has sufficient number for osd's (use 1 journal and 2 for data etc)

wusui:
I was originally trying to solve http://tracker.ceph.com/issues/18528.  While doing so, I executed run-playbook in ceph_ansible.py.  That run failed and ended up creating a file named =11.3.  So it looked like the setuptools >=11.3 part of the code run remotely was interpreted as a shell redirect instead of being passed to the remote operation.   I fixed this as part of 19425.  Things still did not work because the ansible_vers variable also was being mis-interpretted.  So  I changed this to behave the same as the setuptools part. 

After that, it appeared that the  'ansible-playbook -vv ...' command was not running.  I talked to zack and it seems to me that since this was part of a list of args, it should be broken up into ['ansible-playbook', '-vv' ...] parts so that the 'ansible-playbook -vv...' string is not searched for as one command.  Once I made this change, the task still coredumped and I saw log output that read:

[vpm013.front.sepia.ceph.com]: FAILED! => {"changed": false, "failed": true, "msg": "please choose an osd scenario"}

So looking around I found the following:
https://bugzilla.redhat.com/show_bug.cgi?id=1354701
and
https://access.redhat.com/solutions/2943461

The suggestions on that second web page involved the changing of the journal_collocation value.   I did not like it but I tried those changes and they allowed that whole run operation to complete.   After that, I got to the point where I could start to work on 18528.  So I think that the solution suggested is a little dicey but since it helped me out I decided to present it for review.

It is probably the case that the change suggested by the solutions web page is not what we want and there is a configuration problem with the yaml file that I am using.   However, I think that the other two issues are real problems. (1. the format problems with the setuptools parameter and the ansible_vers variables, and 2. the ansible-playbook operation not being parameterized into multiple parts).

kshtsk:
Hi @wusui, @vasukulkarni do we still need this PR? If so, how can we reproduce the issue this patch is addressing?

If it is not required may we can close the PR?
If it is still actual maybe it is better to use multiline script using with remote.run or remote.sh of ceph_installer for better readability?

kshtsk:
@wusui looks like a more than 2 years old stale branch, maybe we can close it cleanup branch?

kshtsk:
Looks like the issue is fixed by someother commits, and this change is outdated, closing.

PR #1059, opened 2017-04-05
If we're going to fail, fail sooner

zmc:
The change to `misc.wait_until_osds_up()` is the follow-up to #1056, #1045 and #1033.

The change to `parallel` fixes a very longstanding bug that prevented jobs from failing properly. If one of the parallel greenlets encountered an exception, it did not interrupt any others. In certain common situations, e.g. the `rados` task running `ceph_test_rados`, it meant the job kept running until it hit the global timeout and was killed - without any useful error message.

An example of this in action:
[This upgrade run using teuthology master](http://pulpito.ceph.com/zack-2017-03-23_14:08:11-upgrade:hammer-jewel-x-kraken---basic-ovh/) timed out.
[This run](http://pulpito.ceph.com/zack-2017-04-05_12:19:17-upgrade:hammer-jewel-x-kraken---basic-ovh/), which is otherwise identical, failed very quickly and with a much more useful error message: ``Command failed on ovh011 with status 1: 'sudo [...] ceph-osd -f --cluster ceph -i 0'``

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -911,6 +911,9 @@ def wait_until_osds_up(ctx, cluster, remote, ceph_cluster='ceph'):
     testdir = get_testdir(ctx)
     with safe_while(sleep=6, tries=50) as proceed:
         while proceed():
+            daemons = ctx.daemons.iter_daemons_of_role('osd', ceph_cluster)
+            for daemon in daemons:
+                daemon.check_status()
             r = remote.run(
                 args=[
                     'adjust-ulimits',
--- a/teuthology/parallel.py
+++ b/teuthology/parallel.py
@@ -1,14 +1,17 @@
 import logging
 import sys
 
+import gevent
 import gevent.pool
 import gevent.queue
 
 log = logging.getLogger(__name__)
 
+
 class ExceptionHolder(object):
     def __init__(self, exc_info):
         self.exc_info = exc_info
+
 
 def capture_traceback(func, *args, **kwargs):
     """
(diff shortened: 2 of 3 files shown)


teuthology/parallel.py
```diff
@@ -73,20 +78,13 @@ def __enter__(self):
...
-        try:
-            # raises if any greenlets exited with an exception
-            for result in self:
-                log.debug('result is %s', repr(result))
-                pass
-        except Exception:
-            # Emit message here because traceback gets stomped when we re-raise
-            log.exception("Exception in parallel execution")
-            raise
+        # raises if any greenlets exited with an exception
```
dmick: worth noting that it also blocks until completion, IMO, so readers of __exit__ understand that this is essentially a join().

Also, I think this implies that if there are N tasks, and 1..N-1 are still running, but N has raised, we will not notice the exception until N is visited (last)?  I'm not saying that matters, but it's a thing to be aware of if true.

teuthology/misc.py
```diff
@@ -911,6 +911,9 @@ def wait_until_osds_up(ctx, cluster, remote, ceph_cluster='ceph'):
     testdir = get_testdir(ctx)
     with safe_while(sleep=6, tries=50) as proceed:
         while proceed():
+            daemons = ctx.daemons.iter_daemons_of_role('osd', ceph_cluster)
```
dmick: maybe we also want to do something similar in wait_until_healthy, perhaps with more daemon types (don't remember if non-osds are also started without waiting)?  Can be a followon, but seems similarly useful

zmc: Possibly - but that would warrant more testing IMO

dmick: yeah, hence 'followon'

teuthology/parallel.py
dmick: I was misreading this as greenlet.get(), not queue.get().  The way the queuing works, any reason for greenlet termination is available immediately.  So I rescind my request for a comment too; either you understand queues or you don't, and I didn't, but now I do

PR #1060, opened 2017-04-11
Add option to use sudo to check ceph health when required

vasukulkarni:
http://pulpito.ceph.com/vasu-2017-04-11_21:01:29-smoke-master---basic-vps/, Failed xenial test is a ceph issue

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -885,22 +885,24 @@ def get_scratch_devices(remote):
     return retval
 
 
-def wait_until_healthy(ctx, remote, ceph_cluster='ceph'):
+def wait_until_healthy(ctx, remote, ceph_cluster='ceph', use_sudo=False):
     """
     Wait until a Ceph cluster is healthy. Give up after 15min.
     """
     testdir = get_testdir(ctx)
+    # when cluster is setup using ceph-deploy or ansible
+    # access to admin key is readonly for ceph user
+    cmd = ['ceph', '--cluster', ceph_cluster, 'health']
+    if use_sudo:
+        cmd.insert(0, 'sudo')
+    args = ['adjust-ulimits',
+            'ceph-coverage',
+            '{tdir}/archive/coverage'.format(tdir=testdir)]
+    args.extend(cmd)
     with safe_while(tries=(900 / 6), action="wait_until_healthy") as proceed:
         while proceed():
             r = remote.run(
-                args=[
-                    'adjust-ulimits',
-                    'ceph-coverage',
-                    '{tdir}/archive/coverage'.format(tdir=testdir),
-                    'ceph',
-                    '--cluster', ceph_cluster,
-                    'health',
-                ],
+                args=args,
                 stdout=StringIO(),
                 logger=log.getChild('health'),
             )


teuthology/misc.py
```diff
@@ -885,18 +885,24 @@ def get_scratch_devices(remote):
...
+    if use_sudo:
+        sudo = 'sudo'
     with safe_while(tries=(900 / 6), action="wait_until_healthy") as proceed:
         while proceed():
             r = remote.run(
                 args=[
                     'adjust-ulimits',
                     'ceph-coverage',
                     '{tdir}/archive/coverage'.format(tdir=testdir),
+                    '{sudo}'.format(sudo=sudo),
```
tchaikov: i don't think this works if `sudo` is `None`. instead maybe we can rename the parameter of `use_sudo` to `sudo` and concat a `cmd` list to `args`, like:
```py
cmd = ['ceph', '--cluster', ceph_cluster, 'health']
if sudo:
  cmd.insert(0, 'sudo')
with safe_while(...):
  while proceed():
    r = remote.run(
      args=[....] + cmd,
      stdout=StringIO(),
      logger=log.getChild('health')
    )
```

tchaikov:
@vasukulkarni could you elaborate a little bit on why `systemd` tests need `sudo`, while others don't?

vasukulkarni:
@tchaikov when we use ceph-deploy or ceph-ansible the permission on admin key is rw for ceph with no access to others, but in ceph.py we relax this for easy testing and set it to '-rw-r--r--.' so that any user can run ceph admin commands.

teuthology/misc.py
vasukulkarni: @tchaikov thanks, I can also try to add ' {sudo}' to make it a string instead of None, that can be simple change?

zmc: I like the look of @tchaikov's suggestion

vasukulkarni: @zmc sure, I will just update with the above snippet thanks

teuthology/misc.py
```diff
@@ -885,18 +885,24 @@ def get_scratch_devices(remote):
...
+    if use_sudo:
+        sudo = 'sudo'
     with safe_while(tries=(900 / 6), action="wait_until_healthy") as proceed:
         while proceed():
             r = remote.run(
                 args=[
                     'adjust-ulimits',
                     'ceph-coverage',
                     '{tdir}/archive/coverage'.format(tdir=testdir),
+                    ' {sudo}'.format(sudo=sudo),
```
tchaikov: please note `' {sudo}'.format(sudo=None)` evaluates to "` None`".

tchaikov:
@vasukulkarni thanks the explanation, got it! =)

dmick:
Is this the only place teuthology issues a ceph command that might need sudo?  I'm wondering whether  this might be common and that we maybe should fix at a lower level.

jdurgin:
many tasks and workunits use the admin key and assume it's readable by the user running the tests - it's much simpler to change its permissions to allow this than to fix every user of the key

dmick:
it also occurs to me that changing the key permissions means we don't mask other possible permission problems (i.e. no longer running as root)

vasukulkarni:
@tchaikov @zmc review comments has been fixed, can one of you guys merge this?

PR #1062, opened 2017-04-17
ceph.conf: adjust "mon reweight min *" for mgr also

tchaikov:
Fixes: http://tracker.ceph.com/issues/19631


tchaikov:
being tested at http://pulpito.ceph.com/kchai-2017-04-17_07:04:05-rados-master---basic-mira/

PR #1069, opened 2017-04-27
rgw: add nose dependency for rgw_multisite_tests task

cbodley:
nose is a dependency for https://github.com/ceph/ceph/pull/14688, which builds a qa suite around the existing rgw multisite tests in `src/test/rgw/test_multi.py`. those tests need to be invoked by the teuthology task itself, because they run `radosgw-admin` commands on different nodes, start/stop daemons, etc

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
                       'httplib2',
                       'paramiko',
                       'pexpect',
+                      'nose', # for qa/tasks/rgw_multisite_tests.py',
                       'requests != 2.13.0',
                       'raven',
                       'web.py',


vasukulkarni:
LGTM

setup.py
```diff
@@ -61,6 +61,7 @@
                       'configobj',
                       'six >= 1.9', # python-openstackclient won't work properly with less
                       'httplib2',
+                      'nose',
```
zmc: I don't *love* the idea of adding a dependency for just one task, but I don't have a better solution yet. In the meantime could you add a comment explaining what this new dep is for? I don't want the reason to get buried in commit history.

cbodley:
thanks @zmc, added a comment. i agree that it's not ideal, but the only alternative i could think of was rewriting all of the existing tests for `unittest` (someday, maybe..)

alfredodeza:
Actually, py.test has support for nosetests :) So this shouldn't even be necessary? Also, please keep in mind that even though nosetests is used in the suites, it is *no longer maintained*, and that has been the case for a while.

cbodley:
hi @alfredodeza, i'm not familiar with py.test, but that sounds promising. are there any examples you could point me at?

vasukulkarni:
there are quite a few known issues running nosetests using py.test, https://docs.pytest.org/en/latest/nose.html,  Better to run them with nose than with py.test

alfredodeza:
@cbodley I think that if you just try `pytest` instead of `nose` as the executable, things should just work. 

Certainly, like @vasukulkarni points out, there is no 100% support for everything nose has (e.g. classes that don't inherit from `unittest.TestCase` will not work), but worth give it a shot to help transition the tests to a supported test runner.

cbodley:
> @cbodley I think that if you just try pytest instead of nose as the executable, things should just work.

i'm using the [nose api](http://nose.readthedocs.io/en/latest/api/core.html) to run the tests within the [rgw_multisite_tests task](https://github.com/ceph/ceph/pull/14688/files#diff-af5a9932c326616bdd541d55c2b5d103R62), so i'd have to do something similar with pytest (it looks like there's a `pytest.main()` for that)

> e.g. classes that don't inherit from unittest.TestCase will not work

that's a problem, then. the tests in `src/test/rgw/rgw_multi/tests.py` are free functions

cbodley:
>> e.g. classes that don't inherit from unittest.TestCase will not work
> 
> that's a problem, then. the tests in src/test/rgw/rgw_multi/tests.py are free functions

maybe not. running those tests on the command line with `pytest`, it seems to find the test functions just fine. i'll give `pytest.main()` a try in the rgw_multisite_tests task

cbodley:
@alfredodeza i'm having trouble getting `pytest.main()` to do what i need here - my main requirement is that it runs tests in an already-imported `rgw_multi.tests` module, because the rgw_multisite_tests task needs to set up the test environment. it does this and calls `rgw_multi.tests.init_multi(realm, user)`, which sets some global variables that are used by all of the tests

i tried using `pytest.main(['--pyargs', tests.__name__])` to do this, but it appears to load a separate module, and the tests fail because those global variables are `None`

cbodley:
i don't see pytest listed as a dependency either - is that coming from somewhere else?

testing my [updated suite](https://github.com/ceph/ceph/pull/14688) on [ovh](http://pulpito.ceph.com/cbodley-2017-05-04_14:25:40-rgw-wip-19834---basic-ovh/), i got `Could not find task 'rgw_multisite_tests'`, which i'm guessing was due to the `import pytest`

zmc:
@cbodley Ah, we don't list pytest as a dependency since it's only been used for unit tests so far. If you have things working with pytest, we can definitely add it as a dependency.

cbodley:
updated pr, now adds `pytest` dependency

cbodley:
@zmc does this look ok? would you prefer a similar comment for the `pytest` dependency?

cbodley:
i changed the dependency back to nose. i was able to get it 'working' with pytest, but i couldn't get pytest to print its results to the teuthology log. the nose api gives more control over logging, so i decided to stick with that

alfredodeza:
py.test "captures" stderr and stdout, and spits them out if there are any failures. If you wish to disable that capturing (done so that test runner output is clean) then there is a flag for that: `-s`

cbodley:
@alfredodeza i tried passing each of `-s`, `--capture=no`, and `--capture=sys` without success

cbodley:
with help from @alfredodeza, i did make some progressing in getting pytest output into teuthology.log

but in further testing, i realized that the tests themselves depend on `nose` (for `nose.tools`, `nose.plugins.attrib`, `nose.plugins.skip`), so i don't think there's a way to work around that

can we merge this as is?

cbodley:
ping @zmc

cbodley:
ping @zmc

vasukulkarni:
@zmc is on pto this week, If you have run the tests with this change and post the results I can merge.

cbodley:
thanks @vasukulkarni, here's the last successful run of the rgw:multisite suite that uses nose: http://pulpito.ceph.com/cbodley-2017-05-16_10:46:23-rgw:multisite-master---basic-smithi/

the jobs fail with valgrind issues, but the tests themselves pass:

> INFO:tasks.rgw_multisite_tests:----------------------------------------------------------------------
INFO:tasks.rgw_multisite_tests:Ran 11 tests in 921.748s
INFO:tasks.rgw_multisite_tests:
INFO:tasks.rgw_multisite_tests:OK

PR #1070, opened 2017-04-27
[tasks/ansible] Add support for mgr role

vasukulkarni:

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -26,6 +26,11 @@ class CephAnsible(Task):
             roles=['ceph-mon'],
         ),
         dict(
+            hosts='mgrs',
+            become=True,
+            roles=['ceph-mgr'],
+        ),
+        dict(
             hosts='osds',
             become=True,
             roles=['ceph-osd'],
@@ -152,6 +157,7 @@ class CephAnsible(Task):
     def generate_hosts_file(self):
         groups_to_roles = dict(
             mons='mon',
+            mgrs='mgr',
             mdss='mds',
             osds='osd',
             rgws='rgw',
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -18,8 +18,14 @@ log = logging.getLogger(__name__)
 
 class CephAnsible(Task):
     name = 'ceph_ansible'
-
+     # TODO: Eventually, we should drop _default_playbook in favor of site.yml.sample from ceph-ansible.git
     _default_playbook = [
+        # Run gather facts at the top as required by mgr-role 
+        dict(
+            hosts='all',
+            become=True,
+            tasks=[],
+        ),
         dict(
             hosts='mons',
             become=True,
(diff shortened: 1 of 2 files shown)


vasukulkarni:
tests are passing we need at add gather facts to playbook as well: http://pulpito.ceph.com/vasu-2017-06-20_21:18:49-smoke-master---basic-vps/1306490/

vasukulkarni:
The following PR https://github.com/ceph/ceph/pull/14624 depends on this.

teuthology/task/ceph_ansible.py
```diff
@@ -21,11 +21,21 @@ class CephAnsible(Task):
 
     _default_playbook = [
         dict(
+            hosts='all',
+            become=True,
+            tasks=[],
+        ),
```
zmc: What's this for?

vasukulkarni: the default playbook we are defining in the task has to run gatherfacts or else the master branch wont work, ideally we have use the site.yml.sample that comes with ceph-ansible instead of using the default playbook and it needs more testing across other branches, until that is done we should use this to unbreak master runs, more background at : https://github.com/ceph/ceph-ansible/issues/1498

andrewschoen: I think a simple comment here explaining the need for ansible to connect to all nodes at once and gather facts would be helpful.

vasukulkarni:
@andrewschoen updated

teuthology/task/ceph_ansible.py
```diff
@@ -20,12 +20,25 @@ class CephAnsible(Task):
     name = 'ceph_ansible'
 
     _default_playbook = [
+        # run gather facts at the top as required by mgr-role and eventually
+        # stop using _default_playbook and instead use site.yml.sample
+        # from ceph-ansbile.git
```
andrewschoen: s/ceph-ansbile.git/ceph-ansible.git

vasukulkarni: http://i.imgur.com/MjZEvxT.gif

teuthology/task/ceph_ansible.py
```diff
@@ -20,12 +20,25 @@ class CephAnsible(Task):
     name = 'ceph_ansible'
 
     _default_playbook = [
+        # run gather facts at the top as required by mgr-role and eventually
+        # stop using _default_playbook and instead use site.yml.sample
+        # from ceph-ansible.git
```
zmc: I'm confused about the above: we're modifying `_default_playbook` to gather facts first... but what about `site.yml.sample`?

zmc: Suggestion for this line:
```
# First run a play with all hosts and no tasks, just to get ansible to gather facts before the mgr role is run.
```

teuthology/task/ceph_ansible.py
```diff
@@ -20,12 +20,25 @@ class CephAnsible(Task):
     name = 'ceph_ansible'
```
zmc: Suggestion for this line:
```
# TODO: Eventually, we should drop _default_playbook in favor of site.yml.sample from ceph-ansible.git
```

PR #1072, opened 2017-05-24
New QA path is <ceph_git_url>/qa ,so we need add this, otherwise the …

RonnyChen:
…tasks will not be imported

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -89,6 +89,7 @@ def fetch_tasks_if_needed(job_config):
     suite_branch = job_config.get('suite_branch', ceph_branch)
     suite_path = os.path.normpath(os.path.join(
         fetch_qa_suite(suite_branch),
+        'qa',
         job_config.get('suite_relpath', ''),
     ))
     sys.path.insert(1, suite_path)


RonnyChen:
@zmc

vasukulkarni:
qa path has to be specified with --suite-relpath option, maybe that option can be defaulted to 'qa' if not specified.

RonnyChen:
@vasukulkarni thanks for your comment~
Well, we can specified the suite_path in various  ways:
1. write “suite_path” in yaml
2. config <ceph_qa_suite_git_url>  in "~/.teuthology.yaml" or  "/etc/teuthology.yaml"
3. teuthology-suite --suite-relpath xxx ...
 
I think all this options will use function fetch_tasks_if_needed in run.py
if specified the path ,it will use 
``` python
            sys.path.insert(1, suite_path)
```

but if not specified the path in yaml or --suite-relpath   it will inovke fetch_qa_suite(suite_branch) --> get_ceph_qa_suite_git_url() in config.py
``` python
    def get_ceph_qa_suite_git_url(self):
        return (self.ceph_qa_suite_git_url or
                self.get_ceph_git_url())
```

the problem is if someone not config ceph_qa_suite_git_url in "~/.teuthology.yaml" or  "/etc/teuthology.yaml" it will use <ceph_git_url> 

so I think we need add the "qa" and maybe we need remove the <ceph_qa_suite_git_url> in _defaults beause ceph-qa-suite is DEPRECATED

========

Another problem is also in config.py
``` python
def _get_config_path():
    system_config_path = '/etc/teuthology.yaml'
    if not os.path.exists(TeuthologyConfig.yaml_path) and \
            os.path.exists(system_config_path):
        return system_config_path
    return TeuthologyConfig.yaml_path

```
if someone use the teuthology first, like http://docs.ceph.com/teuthology/docs/detailed_test_config.html
there isn't write the "~/.teuthology.yaml" or  "/etc/teuthology.yaml" 

In this situation, some parameter in "_default" will not work
for example: 
if I config the  <test_user> in "_default", it will not use in get_test_user() in misc.py beacause the function read the config from file.
so I think we need dump the _default to file 
https://github.com/ceph/teuthology/pull/1074/files
or we have to write the  config  file manually every time

Maybe there are many aspects I did not take into account, thanks for read

kshtsk:
@RonnyChen do you still need this or we can close this PR?

kshtsk:
The PR is more than two years old and looks like the patch is doing nothing... Closing it.
@RonnyChen thanks for submitting it, if you still have a time to work on this, please, create a ticket by tracker.ceph.com in teuthology project and describe your problem there, and feel free to create another pull request. Thank you very much.

PR #1073, opened 2017-05-24
fix grammatical error 

RonnyChen:
@zmc 

### DIFF
--- a/teuthology/orchestra/cluster.py
+++ b/teuthology/orchestra/cluster.py
@@ -17,7 +17,7 @@ class Cluster(object):
         """
         self.remotes = {}
         if remotes is not None:
-            for remote, roles in remotes:
+            for remote, roles in remotes.iteritems():
                 self.add(remote, roles)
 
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/cluster.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, remotes=None):
         """
         self.remotes = {}
         if remotes is not None:
-            for remote, roles in remotes:
+            for remote, roles in remotes.iteritems():
```
joscollin: I don't think you are fixing a grammatical error or typo here. Please provide a clear description, title and Commit message.

tchaikov: agreed. remotes is a `list` not a `dict` here. please see `teuthology/orchestra/test/test_cluster.py` for examples. if i am wrong, and this is indeed a syntax error, could you add a test case for it?

RonnyChen:
I‘m sorry about I used the wrong type

PR #1074, opened 2017-05-24
Dump '_defaults' to file, otherwise some functions in misc.py will use None

RonnyChen:
@zmc 


kshtsk:
looks like outdated commit? please reopen PR if it is still actual

PR #1076, opened 2017-06-06
whitelist denials from chronyc

vasukulkarni:

### DIFF
--- a/teuthology/task/selinux.py
+++ b/teuthology/task/selinux.py
@@ -107,6 +107,8 @@ class SELinux(Task):
         #  https://bugzilla.redhat.com/show_bug.cgi?id=1289274
         # tracker for chronyd/cephtest issue:
         #  http://tracker.ceph.com/issues/14244
+        # AVC denial about dac_read_search chronyc
+        #  https://bugzilla.redhat.com/show_bug.cgi?id=1449108
         known_denials = [
             'comm="dmidecode"',
             'chronyd.service',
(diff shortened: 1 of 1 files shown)


b-ranto:
I guess the idea here is to get rid of all the dac_read_search denials but this won't be the case (not all of them are caused by the given comm, there are some for postfix, etc). The upstream kernel bug is still being worked on/debugged and I personally believe that this should be fixed in the kernel itself. They changed the behaviour and started requiring CAP_DAC_READ_SEARCH for something they only required CAP_DAC_OVERRIDE before.

vasukulkarni:
@b-ranto yes you are right,  i looked at the bz which Ilya provided and it seems to be coming from single source but looking at the teuthology logs we see it from few others too. We will need some temporary solution to ignore dac_read_search or else its going to be too much noise for many suites.

b-ranto:
In that case, I would suggest ignoring CAP_DAC_READ_SEARCH, especially if there is no "ceph" somewhere in the AVC denial string.

vasukulkarni:
we could check only for ceph denials in that case, I have seen the labels all start with ceph_* so we will not miss anything and we could add more to blacklist,  proposing https://github.com/ceph/teuthology/pull/1077 will need more inputs from dev though.

b-ranto:
@vasukulkarni We don't want to check for ceph denials only, we could miss some cases when the files are mislabelled.

vasukulkarni:
@b-ranto If the dir/files are mislabelled then we will definitely catch them due to ceph selinux policy(there will be denials), what am i missing? for other ceph cases we ignore them and shouldn't cause problems in permissive mode.

b-ranto:
@vasukulkarni but the denials do not have to necessarily contain 'ceph' I believe, the paths are oftentimes shortened and/or scrambled, there is also radosgw command and the commands are sometimes scrambled as well.

PR #1077, opened 2017-06-06
[tasks/selinux]: check denials for ceph only, use blacklist instead of whitelist

vasukulkarni:
Instead of https://github.com/ceph/teuthology/pull/1076


vasukulkarni:
Closing

PR #1078, opened 2017-06-07
scripts/suite.py: use distro kernel as default for all tests

vasukulkarni:
Avoids using existing kernel on test nodes and always installs distro kernel if no option provided

Fixes: http://tracker.ceph.com/issues/20203


zmc:
@liewegas, we could use your input on this

liewegas:
:+1:

PR #1089, opened 2017-06-29
When bootstrapping teuthology, avoid deadlocks

zmc:
Use Popen.communicate() instead of Popen.wait()


dmick:
ooooh, do you think it was overrunning the pipe buffer?

djgalloway:
@dmick check this out: https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

zmc:
@dmick it was certainly a possibility here

djgalloway:
So far so good.  Last night and yesterday, it was to the point where I'd kill the hung pip process, ~100 jobs would start then pip would get hung up again.

PR #1090, opened 2017-07-03
[DNM] upgrade check: upgrade to same version is a noop

smithfarm:
We actually had a test that was "upgrading" from one version to the very same
version. This commit adds a check for that case.

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -335,6 +335,11 @@ def upgrade_remote_to_config(ctx, config):
 
     return result
 
+def _upgrade_is_noop(installed_version, upgrade_version):
+    assert installed_version, "installed_version is empty"
+    assert upgrade_version, "upgrade_version is empty"
+    return LooseVersion(installed_version) == LooseVersion(upgrade_version)
+
 def _upgrade_is_downgrade(installed_version, upgrade_version):
     assert installed_version, "installed_version is empty"
     assert upgrade_version, "upgrade_version is empty"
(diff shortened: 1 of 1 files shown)


smithfarm:
I thought this would address the following:

```
2017-07-03T16:12:30.234 INFO:teuthology.task.install:Ceph rpm upgrade from 10.2.7-430.g53a3be7.el7 to 10.2.7-430.g53a3be7
```

But now I see there's a `.el7` on the end of the first version, which is missing in the second, and that causes the second one to be "less" than the first.

Not sure if we can reasonably check for that, though.

smithfarm:
Maybe just strip off any `.el\d+` at the end of the version number, before doing any comparisons?

smithfarm:
@gregsfortytwo In general our upgrade testing seems like it could be improved. Your take on the comments above?

smithfarm:
@gregsfortytwo This fix is based on two hypotheses:

Hypothesis #1: The code in `upgrade_common` only runs when the `install.upgrade` task is specified in the test yaml. 

Hypothesis #2: `install.upgrade` is currently used only to upgrade packages from the Ceph monolith.

Now, if someone constructs test yaml that ends up instructing `install.upgrade` to either (a) downgrade Ceph packages or (b) keep them at the same version, then that would indicate the test is not doing what was intended.

The idea here is to help discover semantic issues in the upgrade tests - i.e. cases when `install.upgrade` was intended to upgrade Ceph, but in reality that's not happening.

I already landed a patch that fails the test if `install.upgrade` ends up **downgrading**, but I would like (and maintain that it makes sense) to fail the test if it ends up installing the same version on top of the existing packages, as well.

Can you think of a case when `install.upgrade` would be intentionally used to just reinstall the same version of the Ceph packages?

smithfarm:
Marking DNM until I can figure out what can be done (if anything) about https://github.com/ceph/teuthology/pull/1090#issuecomment-312738217

zmc:
Ugh, I'm sorry. I merged without refreshing this browser tab, which I opened this morning...

PR #1091, opened 2017-07-06
[ansible-doc-fix]: correct the doc to fix the overrides name

vasukulkarni:

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -358,8 +358,9 @@ class CephLab(Ansible):
     __doc__ = """
     A very simple subclass of Ansible that defaults to:
 
-    - ansible:
+    - ansible.cephlab:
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@zmc ping ?

teuthology/task/ansible.py
```diff
@@ -358,8 +358,9 @@ class CephLab(Ansible):
     __doc__ = """
     A very simple subclass of Ansible that defaults to:
 
-    - ansible:
+    - ansible.cephlab:
         repo: {git_base}ceph-cm-ansible.git
+        branch: mybranch
```
zmc: This isn't the default value :)

vasukulkarni: I was just trying to give an example but do you want to default it to master?

PR #1092, opened 2017-07-10
docs: use kernel version distro instead of jewel in the example of ru…

bellaalleb:
…n a suite

When try the example of run a suite in README, an error would accur,
complaining that "Kernel branch 'jewel' not found". Use "-k distro" instead.

Signed-off-by: Jing Li <lijing@gohighsec.com>

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -103,7 +103,7 @@ where:
 
 For example, consider::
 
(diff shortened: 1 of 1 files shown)


kshtsk:
ok to test

gregsfortytwo:
retest this please

PR #1094, opened 2017-07-19
packaging: pass --force-yes to allow for unsigned deb packages

idryomov:
rbd_fio task has been switched to {install,remove}_package().  This
commit mirrors ceph.git commit d14f2da437a3 ("task/rbd_fio: allow for
unsigned packages"):

"Similar to what the teuthology install.py task does, add --force-yes to
the apt-get install so that unsigned packages are successfully
installed. It is needed when the buildpackages task is used to create
packages on the fly."

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -74,6 +74,7 @@ def install_package(package, remote):
                   '-E',
                   'apt-get',
                   '-y',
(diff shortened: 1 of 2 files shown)


idryomov:
rbd_fio PR: https://github.com/ceph/ceph/pull/16428

PR #1099, opened 2017-08-03
[teuthology/provision]: Increase number of drives for vps to 4

vasukulkarni:
Related to:  https://github.com/ceph/ceph/pull/16799

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>

PR #1100, opened 2017-08-15
downburst: Decrease disk size to 100GB

djgalloway:
The first VPS on every VPSHOST runs off the VPSHOST root drive.  Because
of this, the root drive is filling up due to the addition of another OSD
drive: https://github.com/ceph/teuthology/pull/1099

This not only results in nagios alerts but could also result in needless
job failures.  OSD drives/partitions are already only 100G on smithi so
this should be fine.

### DIFF
--- a/teuthology/provision/downburst.py
+++ b/teuthology/provision/downburst.py
@@ -170,7 +170,7 @@ class Downburst(object):
             'distro': os_type,
             'distroversion': self.os_version,
             'additional-disks': 4,
-            'additional-disks-size': '200G',
+            'additional-disks-size': '100G',
             'arch': 'x86_64',
         }
         fqdn = self.name.split('@')[1]


dmick:
Why does adding another OSD increase the load on root?  /var/log/<new-osd>?

dmick:
(and why did we add another drive anyway?)

djgalloway:
> Why does adding another OSD increase the load on root? /var/log/?
`/srv/libvirtpool/vpm$FIRSTVPS` is on the root drive instead of a mounted 1TB spinner

> (and why did we add another drive anyway?)
See linked PRs in https://github.com/ceph/teuthology/pull/1099 discussion.

dmick:
> /srv/libvirtpool/vpm$FIRSTVPS is on the root drive instead of a mounted 1TB spinner

yes, but it was when we had only 3 drives too, right?

djgalloway:
> yes, but it was when we had only 3 drives too, right?

Right but now we have the multiple centos and ubuntu cloud images, the running OS image, 4x 200GB OSD drives on top of the the root filesystem.  It's probably fine for now as the nagios alerts typically hover around 95-96% utilization on '/' but next time logrotate fails or whatever, that VPS will cause a job failure.

dmick:
The obvious thing I was missing was that *each vps* now has more drives, and that includes the one on the actual root, which was already at capacity.

PR #1103, opened 2017-08-23
requries: fix install requries

LianneWang:
Fixes: http://tracker.ceph.com/issues/21074
Signed-off-by: Lianne Wang <liyan.wang@xtaotech.com>


kshtsk:
Cannot reproduce.

PR #1108, opened 2017-09-01
DNM - Moved qa/machine_types and qa/nightlies from ceph tree

yuriw:
Signed-off-by: Yuri Weinstein <yweinste@redhat.com>

### DIFF
--- a/nightlies/cron_wrapper
+++ b/nightlies/cron_wrapper
@@ -1,0 +1,53 @@
+#!/bin/bash
+# /nightlies/cron_wrapper.sh
+
+# check for no argument case and stop
+if [ -z $1 ]; then
+  echo "need argument"
+  exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


yuriw:
We have these scripts in `ceph` tree (`ceph/qa/nightlies/` and `ceph/qa/machine_types/`) and used to keep check out directory of `ceph-qa-suites` on the teuthology box in order to use with nightlies.  I think it'd be easier if we move them, and keep pulling latest for teuthology.

yuriw:
per chat with @zmc and @djgalloway the location for scripts suggested `https://github.com/ceph/ceph-cm-ansible/tree/master/roles/teuthology`

yuriw:
replaced by https://github.com/ceph/ceph-cm-ansible/pull/332

PR #1111, opened 2017-09-17
ceph.conf: set new pg max per osd limit

liewegas:
This is replacing the old one, but we need to keep the old one for
upgrades etc.


liewegas:
needed for https://github.com/ceph/ceph/pull/17427

PR #1112, opened 2017-09-19
task/ceph-ansible: copy purge-cluster to top-level directory for it to work

vasukulkarni:
set ANSIBLE_LIBRARY environment variable for purge cluster playbook to work properly.

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -185,11 +185,20 @@ class CephAnsible(Task):
             'ANSIBLE_STDOUT_CALLBACK=debug',
             'ansible-playbook', '-vv',
             '-e', 'ireallymeanit=yes',
-            '-i', 'inven.yml', 'infrastructure-playbooks/purge-cluster.yml'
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@andrewschoen @leseb suggested either to set the env or move the purge-cluster.yml one level above, I find this more better.

vasukulkarni:
Doesn't like inside venv 'ERROR! Unexpected Exception: 'module' object has no attribute 'SSL_ST_INIT''

http://qa-proxy.ceph.com/teuthology/vasu-2017-09-19_23:53:18-ceph-ansible-master-distro-basic-ovh/1650200/teuthology.log

andrewschoen:
@vasukulkarni If this works I guess it's fine, but my contention is that our CI does not test this method nor do our docs mention it. They both take the approach of moving the playbook up one level.

vasukulkarni:
@andrewschoen anyway that didn't work on Ubuntu,  I think I will just move the playbook one folder up.

vasukulkarni:
@andrewschoen done!

PR #1114, opened 2017-09-27
[DNM] packaging: Pin *.ceph.com deb repos higher than Ubuntu repos

djgalloway:
This will ensure apt prioritizes dev repos higher when performing
dependency resolution.


djgalloway:
We do this in ceph-cm-ansible

PR #1119, opened 2017-10-09
downburst: Explicitly type our user-data

zmc:
http://tracker.ceph.com/issues/21727

Tested on:
- CentOS 7.4
- CentOS 7.3
- Ubuntu 16.04
- Ubuntu 14.04


djgalloway:
See http://tracker.ceph.com/issues/21727

PR #1121, opened 2017-10-20
Add a cephmetrics task

zmc:


zmc:
http://pulpito.ceph.com/zack-2017-10-26_16:17:45-cephmetrics-master-distro-basic-vps/

vasukulkarni:
overall LGTM. thanks for fixing the log collection, which will help when we run it on other than mon.a role for smoke tests

PR #1122, opened 2017-10-26
[orchestra/systemd]: check status when daemon is restarted

vasukulkarni:
Also check status when daemon is restarted, any failed daemon restart will raise an error early on.


vasukulkarni:
@zmc ping

PR #1132, opened 2017-12-05
Fix cloning of ceph-ansible.

wusui:
Make sure we are in the right directory when we clone ceph-ansible.
Ceph-ansible clones into ~, and the ansible playbook is run from ~/ceph_ansible


wusui:
I do not think that this change needs to be made.

PR #1134, opened 2017-12-05
PhysicalConsole fixes for FOG

zmc:


teuthology/orchestra/console.py
```diff
@@ -234,9 +228,10 @@ def power_on(self):
...
             if r == 0:
                 break
-        if not self.check_power('on'):
+        if self.check_power('on'):
+            log.info('Power on for {s} completed'.format(s=self.shortname))
+        else:
             log.error('Failed to power on {s}'.format(s=self.shortname))
-        log.info('Power on for {s} completed'.format(s=self.shortname))
 
     def power_off(self):
```
djgalloway: Should probably fix this function too (along with any others that need it)

PR #1136, opened 2017-12-06
nuke: Power off FOG machines instead of reimaging

zmc:
Now that FOG is in master, we don't have to reimage on nuke. Let's shut
them down - the next job will bring them up again during the reimage
process.


zmc:
test this please

PR #1139, opened 2017-12-12
valgrind: added suppression for cython constants

dillaman:
Signed-off-by: Jason Dillaman <dillaman@redhat.com>

### DIFF
--- a/teuthology/task/install/valgrind.supp
+++ b/teuthology/task/install/valgrind.supp
@@ -445,3 +445,39 @@
     ...
     fun:PyObject_Call
 }
+
+{
+   rados cython constants
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:PyObject_Malloc
+   fun:PyCode_New
+   fun:__Pyx_InitCachedConstants
+   fun:initrados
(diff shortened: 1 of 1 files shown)


dillaman:
Fixes false positives in test runs: http://qa-proxy.ceph.com/teuthology/teuthology-2017-12-08_02:01:03-rbd-master-distro-basic-smithi/1943792/teuthology.log

teuthology/task/install/valgrind.supp
```diff
@@ -445,3 +445,41 @@
     ...
     fun:PyObject_Call
 }
+
+# rados cython constants
+{
+   <insert_a_suppression_name_here>
```
trociny: @dillaman probably insert a suppression name here as suggested or remove this line (and below)?

teuthology/task/install/valgrind.supp
```diff
@@ -445,3 +445,41 @@
     ...
     fun:PyObject_Call
 }
+
+# rados cython constants
```
trociny: could be used as a suppression name

dillaman:
@trociny addressed

PR #1140, opened 2017-12-15
Implement installer.0 role.

wusui:
This allows one to place the ceph ansible installer node on another machine.
It defaults back to the lowest monitor if there is no installer.0 role.


zmc:
@wusui has this been tested yet?

wusui:
This has been tested running with installer.0 on its own machine, with installer.0 on the mon.a machine, installer.0 on a machine where mon.a was not running, and without installer.0 (so it defaults to mon.a).

yuriw:
@wusui can you add a link to tests ?

PR #1141, opened 2017-12-21
Fix installer.0 bug.

wusui:
Insure that mon node is used for rbd_pools.


wusui:
This fixes installer.0 when rbd_pool is created and installer.0 has no monitors on the same node.

yuriw:
tests http://pulpito.ceph.com/teuthology-2017-12-22_14:55:03-ceph-ansible-master-distro-basic-vps/

PR #1142, opened 2018-01-05
task/exec: Add option to retry for exec task

vasukulkarni:
retry and sleep_for_retry options can be used with exec task without
needing to use lengthy shell based while loops for similar
task(with chances of infinite loop). Also adds timeout option 
for commands that can be used to return within specified time.

### DIFF
--- a/teuthology/task/exec.py
+++ b/teuthology/task/exec.py
@@ -4,6 +4,7 @@ Exececute custom commands
 import logging
 
 from teuthology import misc as teuthology
+from teuthology.contextutil import safe_while
 
 log = logging.getLogger(__name__)
 
@@ -15,6 +16,9 @@ def task(ctx, config):
         - ceph:
         - kclient: [client.a]
         - exec:
+            retry: 10
+            sleep_for_retry: 5 (in seconds) # defauts to 1 second
+            timeout: 100 (in seconds)
             client.a:
               - "echo 'module libceph +p' > /sys/kernel/debug/dynamic_debug/control"
               - "echo 'module ceph +p' > /sys/kernel/debug/dynamic_debug/control"
@@ -23,7 +27,7 @@ def task(ctx, config):
     It stops and fails with the first command that does not return on success. It means
     that if the first command fails, the second won't run at all.
 
-    To avoid confusion it is recommended to explicitly enclose the commands in 
+    To avoid confusion it is recommended to explicitly enclose the commands in
     double quotes. For instance if the command is false (without double quotes) it will
     be interpreted as a boolean by the YAML parser.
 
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@rakeshgm when you get a chance, please review this!

teuthology/task/exec.py
```diff
@@ -35,6 +39,11 @@ def task(ctx, config):
 
     testdir = teuthology.get_testdir(ctx)
 
+    timeout = config.get('timeout', None)
```
smithfarm: maybe 0 instead of None ?

teuthology/task/exec.py
```diff
@@ -45,12 +54,30 @@ def task(ctx, config):
...
+                                action="exec_with_retry") as proceed:
+                    while proceed():
+                        proc = remote.run(
+                                  args=[
+                                      'sudo',
+                                      'TESTDIR={tdir}'.format(tdir=testdir),
+                                      'bash',
+                                      '-c',
+                                      c],
+                                  timeout=timeout,
```
smithfarm: what happens when timeout is None here? perhaps better for timeout to default to 0

teuthology/task/exec.py
```diff
@@ -45,12 +54,30 @@ def task(ctx, config):
...
+                            break
+            else:
+                remote.run(
+                    args=[
+                        'sudo',
+                        'TESTDIR={tdir}'.format(tdir=testdir),
+                        'bash',
+                        '-c',
+                        c],
+                    timeout=timeout
```
smithfarm: what happens when timeout is None here? perhaps better for timeout to default to 0

vasukulkarni: I think the paramiko timeout of 0 means infinite wait, I will check that and come back here, Also I think we are defaulting to some high value when its None, I have to check on both.

teuthology/task/exec.py
kshtsk: The timeout=0 is equivalent to timeout=None in this case, and just disables socket timeout for command execution.

kshtsk:
@vasukulkarni ping, any plans to finish this effort?

vasukulkarni:
@rakeshgm can you take over this one? ping me on chat

teuthology/task/exec.py
kshtsk: I think you're trying to use 'timeout' incorrectly here and above, because this parameter means 'socket timeout' (or read/write operation timeout on ssh channel), but what you expect to have is a command execution timeout, this functionality is missing for remote.run.

kshtsk:
still no progress here

vasukulkarni:
> still no progress here

We have new fixes, will ask @rakeshgm to wrap up this one.

kshtsk:
@rakeshgm @vasukulkarni I don't see comments are addressed, are you still interested in this, or we can close it?

rakeshgm:
@kshtsk I think we need this. but I am slightly busy. will take some time to work on this. sorry for the delay

kshtsk:
I think we can close this, and if someone needs it we can reopen?

smithfarm:
> I think we can close this, and if someone needs it we can reopen?

Reasonable if you consider how many years it has been languishing. Just because a PR is closed doesn't mean it can't be reopened later...

PR #1143, opened 2018-01-10
teuthology/task/install/valgrind.supp: add suppression for dlopen()

tchaikov:
the analysis in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=700899
also applies to ceph-common.

Fixes: http://tracker.ceph.com/issues/22438

### DIFF
--- a/teuthology/task/install/valgrind.supp
+++ b/teuthology/task/install/valgrind.supp
@@ -481,3 +481,13 @@
    fun:PyEval_CallObjectWithKeywords
    fun:PyEval_EvalFrameEx
 }
+
+{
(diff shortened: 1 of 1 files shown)


tchaikov:
@jdurgin you asked me why valgrind reports "still reachable" the other day in the standup. after revisiting this issue, i believe that it's a false alarm. and `/usr/lib/valgrind/debian.supp` shipped by debian/ubuntu actually has following suppression rule:
```
{
  dlopen() with -lpthread bug#700899
  Memcheck:Leak
  fun:calloc
  fun:_dlerror_run
  fun:dlopen@@GLIBC_2.2.5
}
```

PR #1145, opened 2018-01-18
install: extend python-related valgrind rules

jcsp:
These were set for kind 'possible' which was failing
to suppress Leak_DefinitelyLost errors.

Also add several suppressions related to module
initialization.


jcsp:
Tested here (http://pulpito.ceph.com/jspray-2018-01-15_14:36:50-fs:verify-wip-mgr-valgrind-distro-basic-smithi/)

It's all still fails but the valgrind complaints targeted by this change are gone.

PR #1147, opened 2018-02-14
Enable FOG reimage by default for all machine types

vasukulkarni:
Can be enabled for multi queue without any additional bookkeeping.

the downside: you can't control from teuthology.yaml what
machines can be enabled for FOG but that is fine.


kshtsk:
Do we still need this? If not let us close this.

PR #1153, opened 2018-03-15
Disable running cm-ansible by default

vasukulkarni:
Disable running cm-ansible by default, FOG images have already run cm-ansible code and it would be redundant to run them again.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -221,9 +221,13 @@ def get_initial_tasks(lock, config, machine_type):
         init_tasks.extend([
             {'pcp': None},
             {'selinux': None},
-            {'ansible.cephlab': None},
             {'clock': None}
         ])
+    # dont run cm-ansible by default unless requested in config
+    # nodes are reimaged by FOG and the images provided
+    # by FOG have already gone through cm-ansible run
+    if config.get('run-cm-ansible', False):
+        init_tasks.extend([{'ansible.cephlab': None}])
 
     if 'redhat' in config:
         init_tasks.extend([


djgalloway:
I don't think this is the right approach.  Some stuff in ceph-cm-ansible should still be run on every job.

e.g., Adding users, updating pubkeys, zapping disks, setting up LVM

zmc:
Agreed; if anything we should use specific tags and not skip ceph-cm-ansible entirely.

djgalloway:
This can be better handled with https://github.com/ceph/teuthology/pull/1164

kshtsk:
> This can be better handled with #1164

so what should be done to skip all everything from roles, so tautology even does not make any attempt to login to the nodes?

djgalloway:
@kshtsk You're welcome to file a PR to enable this functionality explicitly.  This PR was closed because the premise is incorrect.  In our upstream and downstream labs, ceph-cm-ansible should not be disabled by default.  There are some tasks that must always run or teuthology jobs will fail.

As teuthology stands today, you can try `skip_tags: all` to skip all tasks/roles.

kshtsk:
> @kshtsk You're welcome to file a PR to enable this functionality explicitly. This PR was closed because the premise is incorrect. In our upstream and downstream labs, ceph-cm-ansible should not be disabled by default. There are some tasks that must always run or teuthology jobs will fail.
> 
> As teuthology stands today, you can try `skip_tags: all` to skip all tasks/roles.

Thanks.

PR #1154, opened 2018-03-15
task.kernel: Only show latest kernel when running rpm -q kernel

djgalloway:
Fixes: http://tracker.ceph.com/issues/23381

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -766,7 +766,7 @@ def need_to_install_distro(remote):
     if current in newest or current.replace('-', '_') in newest:
         return False
     log.info(
-        'Not newest distro kernel. Curent: {cur} Expected: {new}'.format(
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@djgalloway this task for distro option needs to be revisited, we have an existing task for redhat, probably it can use be used for all rpm systems https://github.com/ceph/teuthology/blob/master/teuthology/task/kernel.py#L459 but centos will still fail..

The kernel tasks are still run with custom kernel as many other upstream kernels are behind so ...

djgalloway:
Does that task get run before or after ceph-cm-ansible?

tchaikov:
retest this please.

vasukulkarni:
@zmc @djgalloway was this tested with kernel task to see if its working as expected? the issue we have in tracker is due to run on a distro kernel instead of test kernel and doesn't fully address the problem.

PR #1157, opened 2018-03-20
Implement $ option to randomly choose yamls.

wusui:
This implements tracker #23208

### DIFF
--- a/teuthology/suite/build_matrix.py
+++ b/teuthology/suite/build_matrix.py
@@ -34,6 +34,9 @@ def build_matrix(path, subset=None):
     for each item in the directory, and then do a product to generate
     a result list with all combinations (A Product).
 
+    For a directory with a magic '$' file, we generate a list of all
+    items that we will randomly choose from.
+
     The final description (after recursion) for each item will look
     like a relative path.  If there was a % product, that path
     component will appear as a file with braces listing the selection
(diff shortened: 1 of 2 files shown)


wusui:
The previous change randomly picked distro yamls but ran for each distro anyway.  The problem was that the build_matrix changes were not sufficient to prevent this.  This new code implements a PickRandom class in matrix.py which effectively models the Concat code, but instead on Concatenating all the yamls together, it randomly picks one instead.  This seems to have fixed things when running the following test:

./virtualenv/bin/teuthology-suite  -vv --ceph-repo https://github.com/ceph/ceph.git --suite-repo https://github.com/ceph/ceph.git --dry-run -c $CEPH_BRANCH -m ovh -s smoke -k distro -p 10000 -e $CEPH_QA_MAIL -t wusui-23208 --suite-branch wip-23208A

wip-2308A is a smoke test suite with $ entries in distro directories.  grepping for os_type in the above output shows 27 runs (instead of the 50+ previously) and both ubuntu and Centos are checked.

Yuri -- can you test this please?

yuriw:
@wusui @zmc we probably need to update docs for this change.  Where is it?

wusui:
I'm looking around for docs right now.  All I have seen is the convolution
description in the ceph doc.  I think that there also has to be some
documentation for teuthology somewhere

yuriw:
@wusui from @dmick => 
https://github.com/ceph/teuthology has a link, to http://docs.ceph.com/teuthology/docs/

yuriw:
@wusui @zmc 
I tried run `smoke` with "--distro ubuntu --distro-version 16.04" => http://pulpito.ceph.com/yuriw-2018-03-21_23:38:37-smoke-luminous-distro-basic-ovh/  and expected to see all jobs on ubuntu.  We need to clarify correct behavior.

wusui:
I have added suite tests for this change and pushed to wip-23208

zmc:
@wusui I don't see your changes

smithfarm:
Changes should be pushed to wusui-23208 branch (not wip-23208) for them to appear in this PR

PR #1163, opened 2018-04-05
Implement $ at symlink directory level.

wusui:
This change implements the 'choose a random yaml file' behavior for directory names and is not keyed on having a $ file in the directory.  The following is an example of how it should be used:

foo is a directory containing distro definitions -- yaml files that define os_type and os_version.  The files in foo are:
          %
          ubuntu.yaml
          centos.yaml
          rhel.yaml

A suite that links distros to foo will run a test for ubuntu, centos, and rhel.

A suite that links distros$ to foo will run a test, randomly selecting from ubuntu, centos and rhel.

This would allow a suite of n combinations of tests to run roughly n/3 tests on each distro when distro$ is used as the name of the symbolic link.  When distro is the name of the symbolic link 3*n tests will be run (once for each yaml file).

### DIFF
--- a/docs/intro_testers.rst
+++ b/docs/intro_testers.rst
@@ -10,7 +10,7 @@ We assume here that you have access to an operational test lab; if not, ask
 your local admin for access!
 
 If you're here to test upstream Ceph, start `here
-<http://ceph.github.io/sepia/adding_users/>`__.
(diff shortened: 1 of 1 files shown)


wusui:
I have modified the unit tests to run for either $ or directory_name$

wusui:
test me please

PR #1166, opened 2018-04-11
placeholder: update whitelist for osd slow op wrn

batrick:
Caused by: https://github.com/ceph/ceph/pull/20660 (ea97c120d2173f2fc70d979d57a7edb2a6c5da5e)

Fixes: https://tracker.ceph.com/issues/23662

### DIFF
--- a/teuthology/suite/placeholder.py
+++ b/teuthology/suite/placeholder.py
@@ -70,7 +70,7 @@ dict_templ = {
                     'debug osd': 25
                 }
             },
-            'log-whitelist': ['slow request'],
+            'log-whitelist': ['slow request', '\(SLOW_OPS\)'],
             'sha1': Placeholder('ceph_hash'),
         },
         'ceph-deploy': {


teuthology/suite/placeholder.py
```diff
@@ -70,7 +70,7 @@ def _substitute(input_dict, values_dict):
                     'debug osd': 25
                 }
             },
-            'log-whitelist': ['slow request'],
+            'log-whitelist': ['\(SLOW_OPS\)'],
```
jdurgin: we'll need to leave the original variant here too, since teuthology master is used for all ceph branches

batrick: good point, I'll add it back in.

batrick:
Before: http://pulpito.ceph.com/pdonnell-2018-04-11_21:31:18-multimds-wip-multimds-regression2-testing-basic-smithi/

After: http://pulpito.ceph.com/pdonnell-2018-04-11_23:34:55-multimds-wip-multimds-regression2-testing-basic-smithi/

PR #1167, opened 2018-04-16
teuthology/lock: ignore none 200 jobs in node_job_is_active()

tchaikov:
there is chance that we will have 404 when accessing a job's URI.


vasukulkarni:
@tchaikov 404 should cause a traceback with requests.get()?

tchaikov:
> 404 should cause a traceback with requests.get()

@vasukulkarni no, that's what `raise_for_status()` is for. see http://docs.python-requests.org/en/master/user/quickstart/#json-response-content

tchaikov:
@zmc ping?

PR #1168, opened 2018-04-21
placeholder: whitelist MDS_ALL_DOWN by default

tchaikov:
because, in ceph/qa/tasks/ceph.py, we start mon, mgr, osd, and then mds.
there is a time window where there is no mds around, but mgr is checking
mdsmap for MDS_ALL_DOWN errors. there is no way to disable this check in
this time window. so we just whitelist MDS_ALL_DOWN here.

### DIFF
--- a/teuthology/suite/placeholder.py
+++ b/teuthology/suite/placeholder.py
@@ -70,7 +70,10 @@ dict_templ = {
                     'debug osd': 25
                 }
             },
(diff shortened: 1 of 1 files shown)


tchaikov:
this should addresses most of the failures in http://pulpito.ceph.com/kchai-2018-04-21_11:36:06-rados-wip-kefu-testing-2018-04-21-1717-distro-basic-mira/

PR #1171, opened 2018-04-30
[DNM]: setup daemongroup to use systemd by default when cluster is setup using ceph-ansible

vasukulkarni:
Need to test if this is all we need or register them as well.


kshtsk:
The PR has conflicts, @vasukulkarni do you still need this or we can close it?

vasukulkarni:
Closing because we need to handle this in better way with ceph-adm.

PR #1173, opened 2018-05-16
ansible.py: Write "foo" to ~/.vault_pass.txt instead of touching

djgalloway:
ansible-playbook will not run with an empty vault password file.  It
will run if the password file has something in it even if it doesn't get
used.

Multiple ansible vault identities won't be fixed until at least 2.5 (https://github.com/ansible/ansible/issues/38146) so this is probably our safest bet for now.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -416,12 +416,13 @@ class CephLab(Ansible):
         super(CephLab, self).__init__(ctx, config)
 
     def begin(self):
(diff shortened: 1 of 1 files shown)


zmc:
test this please

djgalloway:
Does the `pass` need to remain after `f.write('foo')`?

zmc:
The `pass` isn't needed, no

PR #1175, opened 2018-05-18
Unpin libvirt-python

zmc:
http://tracker.ceph.com/issues/24168


zmc:
@dmick Jenkins tested it on trusty, and I retested on trusty manually (as well as MacOS, because that's ever-so-relevant ;)

PR #1176, opened 2018-05-22
ceph-ansible: add hard-coded notario dependency

cbodley:
ceph-ansible has some new dependencies in requirements.txt which need to be installed to avoid the error:
`[ERROR]: The python-notario library is missing. Please install it to continue.`

Fixes: http://tracker.ceph.com/issues/24230

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -446,6 +446,7 @@ class CephAnsible(Task):
             'pip',
             'install',
             run.Raw('setuptools>=11.3'),
+            run.Raw('notario>=0.0.13'), # FIXME: use requirements.txt
             run.Raw(ansible_ver),
             run.Raw(';'),
             run.Raw(str_args)


alfredodeza:
This is not going to work nicely unfortunately. We tried implementing the same for ceph-volume tests, and failed immediately on both CentOS7 and Xenial and all test scenarios with https://github.com/ceph/ceph-ansible/issues/2629

We are now just hard-coding the dependencies in https://github.com/ceph/ceph/pull/22116

cbodley:
thanks for review @alfredodeza. i'm not sure exactly what you mean there in `# XXX Ideally we should be able to consume the requirements for ceph-ansible directly, but the master branch doesn't pin dependencies so we can't guarantee to work correctly` - i see the same `notario>=0.0.13` in requirements.txt there

was ceph-ansible always a python project, or are these python dependencies a recent addition? what will it take to fix ceph-ansible in a way that upstream changes there don't keep breaking these different users and their hard-coded dependencies?

alfredodeza:
I probably consider `notario` more stable because it doesn't move at the pace of ceph-ansible or Ansible itself, and it has never broken backwards compatibility.

ceph-ansible is not a Python project, but it relies on Python libraries, the main one being Ansible, and now it has added `notario` for validation (implemented as a Python plugin in ceph-ansible).

There is currently no way to avoid this problem, the only thing teuthology can do is to pin these dependencies and wait to see if https://github.com/ceph/ceph-ansible/issues/2631 gets addressed so that we can benefit directly from whatever ceph-ansible wants to pin to

vasukulkarni:
@zmc, unit test failure for - > pcp.py                     322 ERROR    Downloading graphs failed!

alfredodeza:
@vasu I don't know what builds you are referring to

vasukulkarni:
@alfredodeza ceph-ansible rpm and debian builds like ceph builds.

vasukulkarni:
jenkins test this please

vasukulkarni:
Jenkins retest this please

cbodley:
thanks folks!

ktdreyer:
@vasukulkarni we have ceph-ansible RPMs directly from master, at https://shaman.ceph.com/repos/ceph-ansible/master/

python-notario is available as an RPM from the CentOS Storage SIG, https://cbs.centos.org/koji/buildinfo?buildID=22798

ktdreyer:
python-notario-0.0.14 at https://cbs.centos.org/koji/buildinfo?buildID=23004

PR #1177, opened 2018-05-31
[DNM]: Add misc cleanup options before we unlock the system

vasukulkarni:
we are running to subscription issues as we are running out of entitlements(verified by David), unregistering before unlock should keep the count down.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -22,6 +22,7 @@ import re
 import pprint
 
 from teuthology import safepath
+from teuthology.parallel import parallel
 from teuthology.exceptions import (CommandCrashedError, CommandFailedError,
                                    ConnectionLostError)
 from .orchestra import run
@@ -1275,6 +1276,17 @@ def get_distro(ctx):
 
     # if os_type is None, return the default of ubuntu
     return os_type or "ubuntu"
+
+
+def cleanup(ctx):
+    unregister_rhel_systems(ctx)
+
+
+def unregister_rhel_systems(ctx):
+    with parallel():
+        for remote in ctx.cluster.remotes.iterkeys():
+            if remote.os.name == 'rhel':
+                remote.run(args=['sudo', 'subscription-manager', 'unregister'])
 
 
 def get_distro_version(ctx):
--- a/teuthology/task/internal/lock_machines.py
+++ b/teuthology/task/internal/lock_machines.py
@@ -150,6 +150,7 @@ def lock_machines(ctx, config):
     finally:
         # If both unlock_on_failure and nuke-on-error are set, don't unlock now
         # because we're just going to nuke (and unlock) later.
+        misc.cleanup(ctx)
         unlock_on_failure = (
             ctx.config.get('unlock_on_failure', False)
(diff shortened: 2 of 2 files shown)


vasukulkarni:
Testing this out in Octo

vasukulkarni:
Works fine http://pulpito.ceph.redhat.com/vasu-2018-05-31_13:47:02-rgw:multifs-jewel-distro-basic-multi/

zmc:
This seems like something that ought to go in `nuke`

vasukulkarni:
@zmc we dont run nuke with FOG, I intentionally kept it out of nuke.

zmc:
@vasukulkarni we do; we just bail early. https://github.com/ceph/teuthology/blob/master/teuthology/nuke/__init__.py#L306-L309

vasukulkarni:
@zmc i disagree here, I am not sure why it went there, nuke is something that has to be deprecated, adding more functions there would just delay that.

zmc:
@vasukulkarni It went there because nuke still gets run, and it would be a waste of time to run all the nuke ops on a system that is going to be wiped.

A couple issues with this PR as it stands:
  - It needs to respect `unlock_on_failure`
  - It shouldn't add functions to `misc`; just keep them in the same module
  - We will still leak subscriptions when jobs are killed by the timeout, or if anything other than a clean pass/fail happens. This is why we still have nuke.

djgalloway:
I thought `nuke` only gets run if the job fails?

Despite my initial recommendation, I don't think `-lock` is the best place.  It should happen before the end of the job like whenever this [1] function gets called.

[1] https://github.com/ceph/teuthology/blob/master/teuthology/task/install/__init__.py#L75-L103

zmc:
@djgalloway right; even putting this in nuke wouldn't be enough...

vasukulkarni:
@zmc  can you show me a failed job or any job that is running nuke of failure?  95% of the code I see in nuke is a dead code that never gets run for scheduled jobs, it may have some use for manual mode but thats outside the scope.

a) How is unlock_on_failure linked with unregistering a rhel system?
b) I dont get why this function cant go in misc? 
c) if we are nuking on teuthology-kill then its thats a different issue which we should fix.

vasukulkarni:
@djgalloway please show me a logs where you see nuke is running, probably that code should be revisited?

vasukulkarni:
@djgalloway yeah right with FOG in picture we have to revisit many of the unwinding functions, I think its waste of testbed time to do such cleanups(heavy weight functions like uninstalling packages or even small cleanup's), It was valid for pre-FOG scenario's but now I question the use of contextlib functions, do we really need a individual test function cleanup's? 

The cleanup I have added is during the *lock* unwinding, It can be in any of the unwinding but not install since we use ceph-ansible install method as well or better yet is to have a non contextlib function that runs at the end of test suite by default.

vasukulkarni:
@djgalloway I am going to close this, since I am not happy to put this in nuke and also I think we can do unregister in cm-ansible itself before we do register ? probably the best place, and I will gather profile from timing.yaml how much time we are spending in cleanup and take up a task to end those.

PR #1180, opened 2018-06-06
Fix fetch_binaries_for_coredumps

kshtsk:
Addresses error message:

  AttributeError: 'tuple' object has no attribute 'split'

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -304,7 +304,7 @@ def fetch_binaries_for_coredumps(path, remote):
             dump_path = os.path.join(coredump_path, dump)
             dump_info = subprocess.Popen(['file', dump_path],
                                          stdout=subprocess.PIPE)
-            dump_out = dump_info.communicate()
+            dump_out = dump_info.communicate()[0]
 
             # Parse file output to get program, Example output:
             # 1422917770.7450.core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, \


kshtsk:
This issue we have in SUSE when teuthology finds a crash. 

Found similar tickets on redmine
https://tracker.ceph.com/issues/20538
https://tracker.ceph.com/issues/18304
https://tracker.ceph.com/issues/16493

gregsfortytwo:
LGTM; I guess we can get both stdout and stderr from that function and weren't handling error returns, huh?

dmick:
given that dump_err is unused, it would be equivalent to assign dump_out to ....communicate()[0] (or add an if() to dump dump_err if there are contents).  I don't have strong feelings either way.

kshtsk:
@dmick will log.warning(dump_err) or log.error(dump_err) be a good usage for it?

gregsfortytwo:
I searched the repo to see what we do elsewhere:

> ~/src/teuthology [master]$ git grep communicate\(
teuthology/coverage.py:        output, _ = proc.communicate()
teuthology/coverage.py:        output, _ = proc.communicate()
teuthology/kill.py:    stdout, stderr = proc.communicate()
teuthology/openstack/test/openstack-integration.py:        (stdoutdata, stderrdata) = self.worker.communicate()
teuthology/provision/downburst.py:        out, err = proc.communicate()
teuthology/provision/downburst.py:        out, err = proc.communicate()
teuthology/repo_utils.py:        out, err = boot_proc.communicate()
teuthology/task/internal/__init__.py:            dump_out = dump_info.communicate()
teuthology/task/internal/vm_setup.py:                    _, err = p2.communicate()

If you look at those results more carefully (eg with "-A 15"), in some cases we optionally dump and in others we just ignore the stderr result. So I'd say it's not a concern either way, but a thought for similar changes in future.

dmick:
I'm good with log.error()

kshtsk:
@dmick seems like overall code does not care about stderr, the subprocess.Popen does not assign any special value to it, that means communicate() will never return anything in second element of tuple rather than None. And handling it does not make any sense. I was not going to rewrite the function fetch_binaries_for_coredumps so it does everything correctly, and improve the logic, just addressing type error.

teuthology/task/internal/__init__.py
```diff
@@ -304,7 +304,7 @@ def fetch_binaries_for_coredumps(path, remote):
             dump_path = os.path.join(coredump_path, dump)
             dump_info = subprocess.Popen(['file', dump_path],
                                          stdout=subprocess.PIPE)
-            dump_out = dump_info.communicate()
+            dump_out, _ = dump_info.communicate()
```
zmc: I think that if we know we're not going to use stderr, we should do:
`dump_out = dump_info.communicate()[0]`

zmc:
ok to test

teuthology/task/internal/__init__.py
kshtsk: Done

dmick:
I don't understand @kshtsk's last comment; if the command generated stderr, stderr will not be None.  Of course the command has to generate output there, but it's certainly possible (imagine an attempt to run a command on a file which doesn't allow access, just as one example).  The point is, it's an unexpected case which causes the error output, and by suppressing/ignoring it, if something *does* go wrong with the command, there will be no evidence to the teuthology user that it did.

The justification for ignoring any error is probably "this command almost never fails", but in the case where it does, it's annoying to find that the program that issued the command just refuses to tell the user why the command failed.  Given how easy it is to log the error, I don't see how ignoring is better in any respect.

kshtsk:
@dmick ok, I'll elaborate on this, if you look closer the code:
```
dump_info = subprocess.Popen(['file', dump_path],
                             stdout=subprocess.PIPE)
```
means that only stdout will be piped and will be returned by _communicate()_ method as first element of tuple, the second element will be always None in this circumstance. If there an error occurs, then _Popen_ will raise an exception and print error messages to _stderr_. If we want to process _stderr_, i.e. make _communicate()_ return it as second value of the tuple, we need to add `stderr=subprocess.PIPE` as a parameter to _Popen_ invocation, and in this case it will not raise an exception.

dmick:
I see what you mean now.   A slight correction: if there is stderr output with stderr==None, the output comes to the parent's stderr, however that might be processed; AFAIK there's no exception. 

```
>>> import subprocess
>>> pr=subprocess.Popen(['/bin/ls', '/does.not.exist'], stdout=s.PIPE)
>>> /bin/ls: cannot access '/does.not.exist': No such file or directory
pr.communicate()
```
('', None)


I probably would have set stderr=subprocess.PIPE and logged the error, if any, but yes, it would have been a more invasive change.

batrick:
Fixes: https://tracker.ceph.com/issues/20538

PR #1185, opened 2018-06-22
build_matrix: ignore hidden files

batrick:
Idea is to support position independent suites so copies do not break symlinks. e.g.: 

https://github.com/batrick/ceph/commit/73f636db1b2616c0acccc2b2356549d1e03600c3

### DIFF
--- a/teuthology/suite/build_matrix.py
+++ b/teuthology/suite/build_matrix.py
@@ -78,6 +78,8 @@ def _get_matrix(path, subset=None):
 
 
 def _build_matrix(path, mincyclicity=0, item=''):
+    if os.path.basename(path)[0] == '.':
+        return None
     if not os.path.exists(path):
         raise IOError('%s does not exist (abs %s)' % (path, os.path.abspath(path)))
     if os.path.isfile(path):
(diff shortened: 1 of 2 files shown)


liewegas:
I like it!

batrick:
retest this please

vasukulkarni:
@batrick sorry I am having hard time understanding this, I see you are ignoring anything that starts with '.' but then there are yaml's that start with .qa and has whitelist.yaml, how is it being used there?

batrick:
> but then there are yaml's that start with .qa and has whitelist.yaml, how is it being used there?

The key point is we want teuthology to ignore `.qa`. It is only used as a helper symlink. We are using `.qa` as a reference to the top-level `qa` directory. The links work no matter where the suite is copied to because each refers to the parent directory's `.qa` directory.

batrick:
@zmc done (I think!).

zmc:
Thanks @batrick!

vasukulkarni:
@batrick thanks for the explanation, I think I see the logic of links working, but the last part I am missing is when '.' is ignored during matrix build how is that the top level folder with '.qa' included in matrix?  If I am understanding correctly the top level should not start with '.'?

batrick:
> how is that the top level folder with '.qa' included in matrix?

The folder's name doesn't begin with `.qa` so it won't be filtered. TBH, I don't really understand your question.

vasukulkarni:
well, looking at the new qa PR  explains it much better now, so to answer my own question, the files that are shared across sub folders will use .qa and since they dont start with  '.' (eg: centos_latest.yaml) are included in matrix, the links are top down rather than bottom up.

PR #1187, opened 2018-07-02
Restart dead workers

kshtsk:
This patch allows to restart dead workers separately not halting the rest of the teuthology components,
and what is more importantly the beanstalkd service. That makes it possible to extend the number of workers as well. Either of pulpito and paddles can be restarted alone too.


zmc:
ok to test

PR #1188, opened 2018-07-02
Fix worker can't figure out owner of a job

kshtsk:
The patch fixes uncaught exception RuntimeError:

  I could not figure out the owner of the requested job.
  Please pass --owner <owner>.

Worker dies with unhandled exception in run_with_watchdog
if it can't figure out owner of a job, which it tries
to kill when job runs longer then given limit of time.


zmc:
ok to test

PR #1189, opened 2018-07-09
results,schedule,woker: persist --seed and --subset in results.log

tchaikov:
to create a repeatable test suite, in addition to `--seed <SEED>`, we also
need to pass the same `--subset <SUBSET>` to teuthology-suite when
rerunning the failed tests. but it would be handy if teuthology-suite
could remember these settings and recall them when `--rerun <RUN>`.

in this change, we repurpose the last job sending the email to report
the test result to note down the subset and seed used for scheduling the
test suite. these variables are stored in results.log at this moment.


yuriw:
@tchaikov ideally if we store  `--seed` value in results.log, then we maybe not even need to specify it when rerunning.

E.g. we provide run name as `RERUN` and thus can parse `--seed` from results.log

Can't we ?

tchaikov:
@yuriw please see https://github.com/ceph/teuthology/pull/1189/files#diff-9874d141940b077c6e46ba0503dae76fR112  and https://github.com/ceph/teuthology/pull/1189/files#diff-b78ebb036cf91aca8c804c64147861faR391

tchaikov:
@zmc ping?

PR #1192, opened 2018-07-20
install: make extra_packages distro-agnostic (in a limited way) and support extra_system_packages option

smithfarm:
Fixes: http://tracker.ceph.com/issues/25020
Fixes: http://tracker.ceph.com/issues/25026

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -227,13 +227,17 @@ def install(ctx, config):
 
     package_list = get_package_list(ctx, config)
     debs = package_list['deb']
-    rpm = package_list['rpm']
+    rpms = package_list['rpm']
 
     # pull any additional packages out of config
-    extra_pkgs = config.get('extra_packages')
+    extra_pkgs = config.get('extra_packages', [])
     log.info('extra packages: {packages}'.format(packages=extra_pkgs))
-    debs += extra_pkgs
-    rpm += extra_pkgs
+    if isinstance(extra_pkgs, dict):
+        debs += extra_pkgs.get('deb', [])
+        rpms += extra_pkgs.get('rpm', [])
+    else:
+        debs += extra_pkgs
+        rpms += extra_pkgs
 
     # When extras is in the config we want to purposely not install ceph.
     # This is typically used on jobs that use ceph-deploy to install ceph
@@ -245,8 +249,8 @@ def install(ctx, config):
         debs = ['ceph-test', 'ceph-fuse',
                 'librados2', 'librbd1',
                 'python-ceph']
-        rpm = ['ceph-fuse', 'librbd1', 'librados2', 'ceph-test', 'python-ceph']
-    package_list = dict(deb=debs, rpm=rpm)
+        rpms = ['ceph-fuse', 'librbd1', 'librados2', 'ceph-test', 'python-ceph']
+    package_list = dict(deb=debs, rpm=rpms)
     install_packages(ctx, package_list, config)
     try:
         yield
@@ -351,7 +355,6 @@ def upgrade_common(ctx, config, deploy_style):
     remotes = upgrade_remote_to_config(ctx, config)
     project = config.get('project', 'ceph')
 
-    # FIXME: extra_pkgs is not distro-agnostic
     extra_pkgs = config.get('extra_packages', [])
     log.info('extra packages: {packages}'.format(packages=extra_pkgs))
 
@@ -362,8 +365,10 @@ def upgrade_common(ctx, config, deploy_style):
         pkgs = get_package_list(ctx, config)[system_type]
         log.info("Upgrading {proj} {system_type} packages: {pkgs}".format(
             proj=project, system_type=system_type, pkgs=', '.join(pkgs)))
-        # FIXME: again, make extra_pkgs distro-agnostic
-        pkgs += extra_pkgs
+        if isinstance(extra_pkgs, dict):
+            pkgs += extra_pkgs.get(system_type, [])
+        else:
+            pkgs += extra_pkgs
 
         installed_version = packaging.get_package_version(remote, 'ceph-common')
         upgrade_version = get_upgrade_version(ctx, node, remote)
@@ -462,6 +467,10 @@ def task(ctx, config):
         project: samba
         branch: foo
         extra_packages: ['samba']
+    - install:
+        extra_packages:
+           deb: ['librados-dev', 'libradosstriper-dev']
+           rpm: ['librados-devel', 'libradosstriper-devel']
     - install:
         rhbuild: 1.3.0
         playbook: downstream_setup.yml
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -198,15 +198,17 @@ def get_package_list(ctx, config):
         debs = filter(lambda p: not p.endswith('-dbg'), debs)
         rpms = filter(lambda p: not p.endswith('-debuginfo'), rpms)
 
-    excluded_packages = config.get('exclude_packages', list())
-    if excluded_packages:
+    def exclude(pkgs, exclude_list):
+        return list(set(pkgs).difference(set(exclude_list)))
+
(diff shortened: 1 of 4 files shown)


teuthology/task/install/__init__.py
```diff
@@ -362,8 +365,10 @@ def upgrade_common(ctx, config, deploy_style):
         pkgs = get_package_list(ctx, config)[system_type]
         log.info("Upgrading {proj} {system_type} packages: {pkgs}".format(
             proj=project, system_type=system_type, pkgs=', '.join(pkgs)))
-        # FIXME: again, make extra_pkgs distro-agnostic
-        pkgs += extra_pkgs
+        if isinstance(extra_pkgs, dict):
+            pkgs += extra_pkgs.get(system_type)
```
gregsfortytwo: Why are we using ```system_type``` here but explicit ```'deb'``` or ```'rpm'``` strings above?

smithfarm:
@gregsfortytwo Based on my understanding of the current state of this code, I cannot see how `qa/suites/samba/install/install.yaml` could work. The latest run of the samba suite I could find is:

http://pulpito.ceph.com/teuthology-2017-05-15_03:50:04-samba-master---basic-smithi/

All jobs failed. And samba was removed from the nightlies around that time.

I suppose another way to approach this would be to change the `apt-get install` command in `task/install/deb.py` so it no longer forces a particular package version. That would make the second commit in this PR much simpler. On the other hand, since all Ceph packages are built together it makes some sense to force the package version on Ceph packages.

Anyway, once this PR lands we could fix the samba suite :-)

smithfarm:
@gregsfortytwo I could not find any instance of the `ganesha` in `qa/`

extra_packages **is** being used to install:

* rbd-nbd
* rbd-mirror

But these are part of the Ceph packaging monolith.

teuthology/task/install/__init__.py
smithfarm: Because here (in upgrade_common) we're looping over the remotes, and inside the loop we are already getting deb/rpm via `teuthology.get_system_type(remote)`. Whereas above, in the `install` function, we are not yet to the point of looping over remotes.

teuthology/task/install/__init__.py
```diff
@@ -232,8 +232,12 @@ def install(ctx, config):
     # pull any additional packages out of config
     extra_pkgs = config.get('extra_packages')
     log.info('extra packages: {packages}'.format(packages=extra_pkgs))
-    debs += extra_pkgs
-    rpm += extra_pkgs
+    if isinstance(extra_pkgs, dict):
+        debs += extra_pkgs.get('deb')
```
jdurgin: perhaps .get(type, []) instead so that missing one type doesn't result in an exception trying to add None to a list

teuthology/task/install/__init__.py
```diff
@@ -462,6 +467,13 @@ def task(ctx, config):
         project: samba
         branch: foo
         extra_packages: ['samba']
+    - install:
```
jdurgin: config format looks good

gregsfortytwo:
Hmm, maybe the extra_packages got more broken then.

Though now that you mention it like that I do remember we had to pick one OS to do the samba testing on (when it worked) because of the package names. It's really the version strings that I'm surprised about. *shrug*

gregsfortytwo:
Anyway, this all seems fine, I was just confused by it apparently not working at all.
Reviewed-by: Greg Farnum <gfarnum@redhat.com>

smithfarm:
I will do some ad-hoc testing.

teuthology/task/install/__init__.py
zmc: Similar to Josh's comment, we should probably do `get(system_type, [])`

teuthology/task/install/deb.py
```diff
@@ -57,15 +64,20 @@ def _update_package_list_and_install(ctx, remote, debs, config):
     builder.install_repo()
 
     remote.run(args=['sudo', 'apt-get', 'update'], check_status=False)
-    remote.run(
-        args=[
+    install_cmd = [
```
zmc: This seems like a good opportunity to start using this here: https://github.com/ceph/teuthology/blob/master/teuthology/packaging.py#L64

smithfarm: @zmc Switching to that will change the way the install task currently behaves in one key respect. See https://github.com/ceph/teuthology/blob/master/teuthology/task/install/deb.py#L67

Due to this `'%s=%s' % (d, version) for d in debs`, the install task fails the test if apt cannot find the exact version that teuthology thinks is supposed to be installed.

Are you OK with that?

(I'm OK with it, because the equivalent RPM code does not behave in the same way.)

smithfarm:
Another thing that occurred to me is that this patchset doesn't fix `exclude_packages` - in other words, we'll need to be able to give a `{'deb':..., 'rpm': ...}` dictionary to that as well.

teuthology/task/install/deb.py
zmc: Hmm, I'm actually not sure what to think now.

I'm leaning towards *not* making that change, if only to reduce the risk of regressions. Perhaps later on we implement and test what I suggested above...

smithfarm:
OK, so the consensus seems to be that this PR needs the following for merge:

1. do `get(system_type, [])`
2. fix `exclude_packages` so it also optionally takes a deb/rpm dict

smithfarm:
@zmc @gregsfortytwo Done. Please re-review.

smithfarm:
retest this please

smithfarm:
The Jenkins tox failure appears to be reproducible on master.

tchaikov:
retest this please.

tchaikov:
the issue on master was fixed by #1199

smithfarm:
Thanks, @tchaikov, for unblocking the unit tests. Now I see there are some that will need tweaking as part of this PR.

batrick:
Noting that this PR would be nice to have for: https://github.com/ceph/ceph/pull/23551#issuecomment-412579492

teuthology/task/install/__init__.py
```diff
@@ -198,15 +198,17 @@ def get_package_list(ctx, config):
...
         rpms = filter(lambda p: not p.endswith('-debuginfo'), rpms)
 
-    excluded_packages = config.get('exclude_packages', list())
-    if excluded_packages:
-        log.debug("Excluding packages: {}".format(excluded_packages))
-
-        def exclude(pkgs):
-            return list(set(pkgs).difference(set(excluded_packages)))
+    def exclude(pkgs, exclude_list):
+        return list(set(pkgs).difference(set(exclude_list)))
```
tchaikov: might want to update 

https://github.com/ceph/teuthology/blob/da24833932b65ee8d660a6b9452a9cc33fee1c91/teuthology/test/task/test_install.py#L43-L45

accordingly, as the order of elements in `list(set(["foo", "bar"]))` is determined by how python hashes string.

smithfarm:
Tests tweaked.

smithfarm:
@batrick Thanks. I found the problem and opened https://github.com/ceph/teuthology/pull/1202 to fix it. BTW the "compiled" yaml of your test looks weird:

```
  tasks:
  - install:
      extra_packages:
      - python3-cephfs
  - ceph: null
  - ceph-fuse: null
  - install:
      extra_system_packages:
        deb:
        - bison
        - flex
        rpm:
        - bison
        - flex
```

(i.e. install task is run twice, once before deployment and again after)

But it *should* work, and hopefully will, once #1202 is merged.

smithfarm:
@batrick Since the package names are the same for rpm and deb, your yaml could look like this:

```
tasks:
- install:
    extra_packages:
    - python3-cephfs
    extra_system_packages:
    - bison
    - flex
```

batrick:
@smithfarm I was hoping to have bison/flex installed only when running the kernel_untar_build task which is why there's a second install task.

smithfarm:
@batrick Got it - it should start working once #1202 is merged. You didn't make any mistake that I can see.

batrick:
It appears to work. Thanks @smithfarm !

PR #1194, opened 2018-07-24
Add no-validate-refs option to teuthology-suite

adamwg:
Allow users to skip validation of SHA1s passed to -S when submitting
jobs via teuthology-suite. This is useful when testing against
non-GitHub repos, since teuthology can't validate refs for a plain git
repository.

Signed-off-by: Adam Wolfe Gordon <awg@digitalocean.com>

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -71,6 +71,9 @@ Standard arguments:
                               <suite_branch> to be ignored for scheduling
                               purposes, but it will still be used for test
                               running.
+  --validate-sha1 <bool>
+                              Validate that git SHA1s passed to -S exist.
+                              [default: true]
 
 Scheduler arguments:
   --owner <owner>             Job owner
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -6,6 +6,7 @@ import logging
 import os
 import random
 import time
+from distutils.util import strtobool
 
 import teuthology
 from ..config import config, YamlConfig
@@ -56,6 +57,8 @@ def process_args(args):
             value = expand_short_repo_name(
                 value,
                 config.get_ceph_qa_suite_git_url())
+        elif key in ('validate_sha1'):
+            value = strtobool(value)
         conf[key] = value
     return conf
 
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -151,7 +151,9 @@ class Run(object):
         repo_name = self.ceph_repo_name
 
         if self.args.ceph_sha1:
-            ceph_hash = util.git_validate_sha1(repo_name, self.args.ceph_sha1)
+            ceph_hash = self.args.ceph_sha1
+            if self.args.validate_sha1:
+                ceph_hash = util.git_validate_sha1(repo_name, ceph_hash)
             if not ceph_hash:
                 exc = CommitNotFoundError(
                     self.args.ceph_sha1,


adamwg:
I'm not sold on the naming of this option - alternative suggestions more than welcome.

teuthology/suite/__init__.py
```diff
@@ -28,6 +28,7 @@ def process_args(args):
         'flavor': 'kernel_flavor',
         '<config_yaml>': 'base_yaml_paths',
         'filter': 'filter_in',
+        'no_validate_refs': 'validate_refs',
```
vasukulkarni: for clarity we dont really have to rename the arg and avoid "value = not value" negation, the if branch could check use the negation check.

adamwg: Indeed, my thought was just that `if not no_validate_refs:` looks weird, so better to invert here, since there doesn't appear to be a convenient way to make an optional boolean parameter true by default.

teuthology/suite/__init__.py
```diff
@@ -48,6 +49,9 @@ def process_args(args):
                 value = []
             else:
                 value = [x.strip() for x in value.split(',')]
+        elif key == 'validate_refs':
+            value = not value # Since this is re-mapped from no-valididate-refs
```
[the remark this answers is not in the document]
adamwg: Thanks, updated the wording as suggested.

teuthology/suite/__init__.py
vasukulkarni: if no_validate_refs is False: can be used, and there is also else: if we not looks weird, double negation and changing the arg is confusing for code readability later on.

scripts/suite.py
```diff
@@ -71,6 +71,7 @@
                               <suite_branch> to be ignored for scheduling
                               purposes, but it will still be used for test
                               running.
+  --no-validate-refs          Don't validate that git SHA1s passed to -S exist.
```
smithfarm: Could you change this to `--validate-refs` with `default: True`? Then you would avoid the weird "double-negative" later on.

kshtsk: Could you please rename `--no-validate-refs` to `--validate-sha1` defaults: 0, because:
1) Options should not have negative notation.
2) This does not produces double-negative coding.
3) sha1 is better since this option is directly related to SHA1 options.
It is better to put the description for this key next to:
```-S <sha1>, --sha1 <sha1>    The ceph sha1 to run against (overrides -c)```.

Comment (adamwg) [reply scripts/suite.py]: I've tried changing this to `--validate-refs` with the default as `True`, but then i can't figure out how to make it False at runtime. My option definition looks like this:

```
  --validate-refs             Validate that git SHA1s passed to -S exist.
                              [default: True]
```

but I've tried a few ways to set it and none of them work:

```console
$ teuthology-suite -c luminous -S abcd1234 --suite rbd:basic -d ubuntu -D 14.04 -m ubuntu --suite-branch luminous --validate-refs=False /home/teuthology/overrides.yml
--validate-refs must not have an argument
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
$ teuthology-suite -c luminous -S abcd1234 --suite rbd:basic -d ubuntu -D 14.04 -m ubuntu --suite-branch luminous --no-validate-refs /home/teuthology/overrides.yml
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
$ teuthology-suite -c luminous -S abcd1234 --suite rbd:basic -d ubuntu -D 14.04 -m ubuntu --suite-branch luminous --without-validate-refs /home/teuthology/overrides.yml
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
```

Any thoughts on how this is supposed to work?

kshtsk: If you look carefully you can see that '=' is not used anywhere with arguments.

kshtsk: @adamwg I still would like to ask you to rename argument to `--validate_sha1` to avoid future confusion and recall easily the arguments.

adamwg: @kshtsk Not sure you're understanding the problem I'm having. If I remove the `=` I still have the same issue. Updated the code to this:

```
  --validate-sha1 <validate_sha1>
                              Validate that git SHA1s passed to -S exist.
                              [default: 0]
```

but neither of the following makes the option false:

```console
$ teuthology-suite -c luminous -S abcd1234 --suite rbd:basic -d ubuntu -D 14.04 -m ubuntu --suite-branch luminous --validate-sha1 False /home/teuthology/overrides.yml
--validate-refs must not have an argument
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
$ teuthology-suite -c luminous -S abcd1234 --suite rbd:basic -d ubuntu -D 14.04 -m ubuntu --suite-branch luminous --validate-sha1 0 /home/teuthology/overrides.yml
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --machine-type <type> --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
```

It seems like a bug in this silly option parsing library that it can't create a bool flag with true as a default. Unless someone has a solution, I think the only way to go is with my original solution, naming the flag `no-validate-sha1`.

kshtsk: @adamwg It is hard to say what the bug you have not seeing the changes. Seems like you changed option to `--validate-sha1` but still trying to use somewhere `--validate-refs`.
See example, it works fine:
```
import docopt
import sys

doc = """
usage: t.py --help
       t.py [options]

Test arguments.

arguments:
  -b <bln>, --bool <bln> 
                                Boolean value [default: True]
"""

def main(argv=sys.argv[1:]):
    args = docopt.docopt(doc, argv=argv)
    print("Arguments: %s" % args)

if __name__ == "__main__":
    main()
```
1) Default value should is True for ```python t.py```:
```
Arguments: {'--bool': 'True',
 '--help': False}
```
2) Value is changed to False for ```python t.py -b False```
```
Arguments: {'--bool': 'False',
 '--help': False}
```
And, maybe, your problem is that you do not provide **mandatory** arguments:  `--machine-type`, `--ceph` and `--suite` (or `--rerun`).

kshtsk: > I've tried changing this to `--validate-refs` with the default as `True`, but then i can't figure out how to make it False at runtime. My option definition looks like this:
> 
> ```
>   --validate-refs             Validate that git SHA1s passed to -S exist.
>                               [default: True]
> ```
> but I've tried a few ways to set it and none of them work:
> .
> .
> .
> --validate-refs must not have an argument
Looks like you have left this `--validate-refs` declaration somewhere, which is syntactically  incorrect, because you didn't provide a placeholder for value in angle brackets but declaring the default value.

adamwg: @kshtsk In your example, the `--bool` argument value is not a bool, it's a string. If you do something like `if bool` you'll find it's always truey (unless you set it to an empty string, but that's a totally unintuitive way to make a boolean option false).

That did help me figure out the issue, though - it's always treating the argument as a string, so we need to convert it manually. See the latest update to this PR for a working solution.

teuthology/suite/__init__.py
```diff
@@ -48,6 +49,8 @@ def process_args(args):
                 value = []
             else:
                 value = [x.strip() for x in value.split(',')]
+        elif key == 'validate_sha1':
```
kshtsk: Suggestion to change the line to:
```
elif key in ('validate_sha1'):
```

scripts/suite.py
kshtsk: very true, if the default value is not repeatable it is interpreted as a string

teuthology/suite/__init__.py
adamwg: Sure - I followed the examples above where there was only a single key, but makes sense to use the set operation in case there are other bool options like this in the future. Done.

adamwg:
Rebased onto master and resolved merge conflicts. Hoping this is ready to merge.

vasukulkarni:
@djgalloway @zmc can one of you add the retest comment here, thanks

djgalloway:
jenkins test

djgalloway:
jenkins test

djgalloway:
jenkins test

vasukulkarni:
If there are no further comments, then I will merge on monday.

PR #1203, opened 2018-08-31
Fix log remote dir with dual clusters

smanjara:
Fixes failure to log second cluster's ceph logs. Also fixes the teuthology failures due to single remote dir.


vasukulkarni:
test failure is a 'rh' branch issue, I will fix that up and will cherrypick this, thanks!

PR #1204, opened 2018-09-10
 worker: do not pass --timeout to first-in-suite job

tchaikov:
this change fixes the --rerun feature again.

### DIFF
--- a/teuthology/worker.py
+++ b/teuthology/worker.py
@@ -193,23 +193,26 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
     if job_config.get('first_in_suite') or job_config.get('last_in_suite'):
         if teuth_config.results_server:
             report.try_delete_jobs(job_config['name'], job_config['job_id'])
-        log.info('Generating memo/results for %s', job_config['name'])
         args = [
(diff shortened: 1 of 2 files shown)


teuthology/worker.py
```diff
@@ -193,23 +193,26 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
             job_config['name'],
         ]
-        if job_config.get('email'):
-            args.extend(['--email', job_config['email']])
-        if job_config.get('seed'):
-            args.extend(['--seed', job_config['seed']])
-        if job_config.get('subset'):
-            args.extend(['--subset', job_config['subset']])
+        if job_config.get('first_in_suite'):
+            log.info('Generating memo for %s', job_config['name'])
```
zmc: I realize this line isn't new, but why is the word "memo" used here?

tchaikov: @zmc the because the seed and subset noted by the "first-in-suite" job will be read back by `--rerun` command so it is able to use these parameters when creating the test matrix, see also 6e58583a0a5768ce219eb4fa397b21d21c126733. that's why i use "memo" for this job.

PR #1205, opened 2018-09-12
worker,result: create archive_dir for the first job and mark --{seed,subset} optional

tchaikov:


tchaikov:
@zmc i am jumping the gun and merging it. but the change won't kick in unless we restart the workers.

PR #1207, opened 2018-09-15
teuthology/task: improve documentation of "print" internal task

smithfarm:
The documentation is taken from the commit message that first
introduced this task.

### DIFF
--- a/teuthology/task/print.py
+++ b/teuthology/task/print.py
@@ -1,5 +1,17 @@
 """
 Print task
+
+A task that logs whatever is given to it as an argument. Can be used
+like any other task (under sequential, etc...).j
+
+For example, the following would cause the strings "String" and "Another
+string" to appear in the teuthology.log before and after the chef task
+runs, respectively.
+
+tasks:
+- print: "String"
+- chef: null
+- print: "Another String"
 """
 
 import logging


teuthology/task/print.py
```diff
@@ -1,5 +1,16 @@
 """
 Print task
+
+A task that simply prints the output that is given to it as
+an argument. Can be used like any other task (under sequential,
+etc...)
```
dmick: If we're going to add this, how about making it crystal clear that it actually logs the output.  "Print" means "terminal" to me, and of course that makes no sense for scheduled jobs, but one might think this is not useful for scheduled jobs (and it is).  In fact, I'm not sure why it's not called "log"

smithfarm:
Clarification needed here, I see. 

This PR is not introducing the `print` task - it already exists and is used a lot in the upgrade suites. For example, in master, I get:

```
smithfarm@wilbur:~/src/ceph/smithfarm/ceph/qa/suites/upgrade> ag print | wc --lines
82
```

Many more instances in stable branches, of course.

This PR is just adding a comment. I think `s/print/log/g` is beyond the scope of this PR.

smithfarm:
@gregsfortytwo @dmick OK, pushed a new version. Please take a look.

dmick:
wasn't suggesting was was new or that the name choice was yours, but, just an "as a matter of fact" to the subject of clearly documenting.  names are a part of clarity.  I'm happy to make it a separate PR, although that runs the risk of it never being done, of course, and while we're here...

but as you wish.

dmick:
that comment clarifies it well.

PR #1208, opened 2018-09-15
teuthology/task: implement dump_ctx internal task

smithfarm:
This task is supposed to pretty-print the ctx and config parameters that are
passed to each task.

The intended use is didactic: teaching myself and others how to use teuthology. I plan to add this to the existing "dummy" task:

https://github.com/ceph/ceph/tree/master/qa/suites/dummy

### DIFF
--- a/teuthology/task/dump_ctx.py
+++ b/teuthology/task/dump_ctx.py
@@ -1,0 +1,19 @@
+import logging
+import pprint
+
+log = logging.getLogger(__name__)
+pp = pprint.PrettyPrinter(indent=4)
+
+def _pprint_me(thing, prefix):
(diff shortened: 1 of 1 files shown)


vasukulkarni:
Is this not enabled by default? How does the sample output look?

gregsfortytwo:
Hmmm, the config parameters are per-task so this isn't sufficient to print what any given task will receive. Are you planning to somehow embed this within each task? Or do you need to go back and look at the way the per-task configs get generated in the source?

gregsfortytwo:
I'm confused about how you'd use it as a didactic tool. I guess you plan to swap this test out for something else and see what it prints, then go back to the original one?

PR #1210, opened 2018-09-15
suite: Implement --sleep-before-teardown option

smithfarm:
With this option, all jobs in the run sleep for the given number of seconds as
the very last thing they do, regardless of success or failure.

Use case: interactive debugging


gregsfortytwo:
Has this been through anything to test it yet?

smithfarm:
> Has this been through anything to test it yet?

I have tested it (with passing and failing jobs) in our downstream teuthology-openstack environment, yes.

PR #1211, opened 2018-09-17
teuthology/task/install/valgrind.supp: add suppression for DPDK

tchaikov:

### DIFF
--- a/teuthology/task/install/valgrind.supp
+++ b/teuthology/task/install/valgrind.supp
@@ -523,3 +523,13 @@
   fun:_dlerror_run
   fun:dlopen@@GLIBC_2.2.5
 }
+
+{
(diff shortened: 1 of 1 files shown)


tchaikov:
see

```/a/kchai-2018-09-17_04:41:34-rados-wip-kefu-testing-2018-09-16-2359-distro-basic-mira/3032089/remote/mira084/log/valgrind/mon.b.log.gz```

tchaikov:
the valgrind warning is introduced by https://github.com/ceph/ceph/pull/23650/commits/1a6877dc048321226cdb764778c6ed3971f36566 . and the warning is expected. i am wondering if we should just disable `WITH_SPDK=ON`...

jdurgin:
it'd be nice if we could continue enabling spdk, just to ensure we don't break building with it

PR #1214, opened 2018-09-26
Added haproxy and keepalived config

smanjara:


teuthology/task/ceph_ansible.py
```diff
@@ -448,6 +451,82 @@ def run_rh_playbook(self):
         self.fix_keyring_permission()
         self.wait_for_ceph_health()
 
+    def run_haproxy(self):
+
+        """
+        task:
+            ceph-ansible:
+                haproxy: true
```
vasukulkarni: can you add role definition that is needed like haproxy.i and haproxy.1

smanjara: I have already added in 'groups to roles' dict and the deployment was successful. Is it required elsewhere?

smanjara: Updated the doc string

vasukulkarni:
Closing this in favor of #1232

PR #1216, opened 2018-10-01
run: do not block on greenlets after command exits

batrick:
The stdout/stderr greenlets will not necessarily exit when the command does if
child processes are stuck in an uninterruptible sleep. For example, the fsx.sh
workunit spawns fsx processes that may be left behind in the D state after
/bin/timeout kills fsx.sh. These are connected to the stdout/stderr pipes which
prevent the greenlets from exiting normally.

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -139,10 +139,16 @@ class RemoteProcess(object):
 
         :returns: self.returncode
         """
-        for greenlet in self.greenlets:
-            greenlet.get()
 
         status = self._get_exitstatus()
+        if status != 0:
+            log.debug("got remote process result: {}".format(status))
+        for greenlet in self.greenlets:
+            try:
+                greenlet.get(block=True,timeout=60)
+            except gevent.Timeout:
+                log.debug("timed out waiting; will kill: {}".format(greenlet))
+                greenlet.kill(block=False)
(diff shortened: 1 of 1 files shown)


batrick:
Example: /ceph/teuthology-archive/pdonnell-2018-10-01_03:19:12-multimds-wip-pdonnell-testing-20181001.011252-distro-basic-smithi/3090388/teuthology.log

batrick:
I don't see why the build is broken or how it could be caused by this PR.

teuthology/orchestra/run.py
```diff
@@ -139,10 +139,16 @@ def wait(self):
...
         """
-        for greenlet in self.greenlets:
-            greenlet.get()
 
         status = self._get_exitstatus()
+        if status != 0:
+            log.debug("got remote process result: {}".format(status))
+        for greenlet in self.greenlets:
+            try:
+                s = greenlet.get(block=True,timeout=60)
```
tchaikov:
```
teuthology/orchestra/run.py:148:17: F841 local variable 's' is assigned to but never used
```

batrick: Ah, thanks @tchaikov

vasukulkarni:
I think its a flake8 issue,  line 148: s initialized but never used

PR #1217, opened 2018-10-06
Add osd_disk_info to ctx so that any task can access it like ctx.osd_…

shylesh:
…disk_info

Signed-off-by: shylesh kumar <shylesh.mohan@gmail.com>


vasukulkarni:
thanks @shylesh , I have cherrypicked, couple more flake8 issue not from your pr.

PR #1218, opened 2018-10-10
Add custom repos before installing rpm packages

kshtsk:
It is a set of patches from suse/teuthology related to packaging and --test-repo feature.

The idea of this feature is to bypass buildpackages phase and let teuthology install ceph packages from the specified test repos in 'install' task. For openstack it speeds up bootstrapping a lot, first of all it does not spend tons time for building ceph packages, and does not need to allocate automatically builder and packages servers in cloud services which needs extra cost.

Signed-off-by: Kyr Shatskyy <kshatskyy@suse.com>


kshtsk:
retest this please

teuthology/orchestra/opsys.py
```diff
@@ -85,7 +86,7 @@ def __init__(self, name=None, version=None, codename=None):
     @staticmethod
     def _version_to_codename(name, version):
         for (_version, codename) in DISTRO_CODENAME_MAP[name].iteritems():
-            if version == _version or version.split('.')[0] == _version:
+            if str(version) == _version or str(version).split('.')[0] == _version:
```
gregsfortytwo: Why the change to stringify variables here? Aren’t they already strings?

teuthology/packaging.py
```diff
@@ -641,7 +641,7 @@ def _get_version(self):
         if not version:
             version = DEFAULT_OS_VERSION.get(self.os_type)
 
-        return version
+        return str(version)
```
gregsfortytwo: And here.

teuthology/run.py
```diff
@@ -168,10 +168,15 @@ def validate_tasks(config):
 
 
 def get_initial_tasks(lock, config, machine_type):
-    init_tasks = [
-        {'internal.check_packages': None},
-        {'internal.buildpackages_prep': None},
-    ]
+    init_tasks = []
+    overrides = config.get('overrides')
+    having_repos = ('install' in config and 'repos' in config.get('install')) \
```
gregsfortytwo: I don’t think I understand what’s happening here. I presume you want to bypass these internal tasks if not in the upstream lab, but the conditions don’t make a lot of sense to me.

teuthology/orchestra/opsys.py
kshtsk: > Why the change to stringify variables here? Aren’t they already strings?

At the time of patch applied, the version looks like 'X.Y' and it appeared here as floating, so... this code made sure that there is a string.

teuthology/packaging.py
kshtsk: Same as above.

teuthology/run.py
kshtsk: Yes, the idea we don't want to build ceph packages ourself if we provided test repos.

PR #1222, opened 2018-10-15
tests: use dict() instead Mock() to test config

kshtsk:
Addresses failure in verify_package_version:
    TypeError: argument of type 'Mock' is not iterable

### DIFF
--- a/teuthology/test/task/test_install.py
+++ b/teuthology/test/task/test_install.py
@@ -78,7 +78,8 @@ class TestInstall(object):
         gb.project = "ceph"
         m_gitbuilder_project.return_value = gb
         m_get_package_version.return_value = "0.89.0"
-        install.verify_package_version(Mock(), Mock(), Mock())
+        config = dict()
+        install.verify_package_version(Mock(), config, Mock())
 
     @patch("teuthology.task.install._get_builder_project")
     @patch("teuthology.task.install.packaging.get_package_version")
@@ -89,9 +90,7 @@ class TestInstall(object):
         gb.project = "ceph"
         m_gitbuilder_project.return_value = gb
         m_get_package_version.return_value = "0.89.1"
-        config = Mock()
-        # when it looks for config.get('extras') it won't find it
-        config.get.return_value = False
+        config = dict()
         with pytest.raises(RuntimeError):
             install.verify_package_version(Mock(), config, Mock())
 
@@ -105,8 +104,8 @@ class TestInstall(object):
         m_gitbuilder_project.return_value = gb
         # ceph isn't installed because ceph-deploy would install it
         m_get_package_version.return_value = None
-        config = Mock()
-        config.extras = True
+        config = dict()
+        config['extras'] = True
         install.verify_package_version(Mock(), config, Mock())
 
     def test_get_flavor_default(self):


kshtsk:
Using Mock() object in the test makes it impossible to check key presence dict() objects using `key in dictobj` syntax [see https://docs.python.org/3/library/stdtypes.html#typesmapping] somewhere in teuthology code, in particular, verify_package_version.

gregsfortytwo:
I presume a Mock() object is used so that in future it can actually be mocked out properly — is replacing it with a straight dict really appropriate here, or is there some Mock functionality we should implement?
</Greg knowing nothing about proper Python>

kshtsk:
@gregsfortytwo Probably I don't understand what should be changed in mock so it can support following syntax:
```
 'key' in config
```
if config parameter of **verify_package_version** method is a dictionary, or it is not?

teuthology/test/task/test_install.py
```diff
@@ -105,8 +104,8 @@ def test_skip_when_using_ceph_deploy(self, m_get_package_version,
         m_gitbuilder_project.return_value = gb
         # ceph isn't installed because ceph-deploy would install it
         m_get_package_version.return_value = None
-        config = Mock()
-        config.extras = True
+        config = dict()
+        config['extras'] = True
```
tchaikov: could also put
```suggest
config = Mock()
config.get.return_value = True
```

teuthology/test/task/test_install.py
```diff
@@ -78,7 +78,8 @@ def test_verify_ceph_version_success(self, m_get_package_version,
         gb.project = "ceph"
         m_gitbuilder_project.return_value = gb
         m_get_package_version.return_value = "0.89.0"
-        install.verify_package_version(Mock(), Mock(), Mock())
+        config = dict()
```
tchaikov: we could just put
```suggestion
config = Mock()
config.get.return_value = False
```

kshtsk:
@tchaikov check this code:
https://github.com/ceph/teuthology/pull/1218/commits/4f304d5d66cd85076a02ff3166cecadd4e236d53#diff-e97be647744c3039ca9a3e42a4a652e2R44

teuthology/test/task/test_install.py
```diff
@@ -89,9 +90,7 @@ def test_verify_ceph_version_failed(self, m_get_package_version,
         gb.project = "ceph"
         m_gitbuilder_project.return_value = gb
         m_get_package_version.return_value = "0.89.1"
-        config = Mock()
-        # when it looks for config.get('extras') it won't find it
-        config.get.return_value = False
+        config = dict()
```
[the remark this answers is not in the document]
kshtsk: It will fail with `TypeError: argument of type 'Mock' is not iterable` if to check key presence in target method like this:
```
if 'extras' in config
```
instead of
```
if config.get("extras")
```

teuthology/test/task/test_install.py
kshtsk: See previous comment.

kshtsk:
@zmc You can find trace example in this job:
https://jenkins.ceph.com/job/teuthology-pull-requests/1318/consoleFull

kshtsk:
@zmc the tox issue I mentioned on meeting is not related to this PR.

tchaikov:
as replied before, i am not sure the dict is better than Mock in terms of the testing if we 
FWIW, you can also mock the dict using:
```py
# use side_effect if you are after more flexibility
config = Mock(get=Mock(return_value=True), 
              __contains__=Mock(return_value=True))
```

kshtsk:
> as replied before, i am not sure the dict is better than Mock in terms of the testing if we do want to mock `config` with a minimal stub.
> FWIW, you can also mock the dict using:
> 
> ```python
> # use side_effect if you are after more flexibility
> config = Mock(get=Mock(return_value=True), 
>               __contains__=Mock(return_value=True))
> ```

Thanks, but what is the benefit of using Mock for config instead of dict in these particular test cases?

PR #1223, opened 2018-10-15
Increase log level to 'info' of run config

kshtsk:
Make run config visible in the logs even if -v option is not passed.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -332,7 +332,7 @@ def main(args):
         assert lock, \
             'the --block option is only supported with the --lock option'
 
-    log.debug(
+    log.info(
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
Is this not already dumped elsewhere? eg the archive dir gets the job's config dumped at a few different stages of processing. Is there some concern about it not getting handled correctly that can't be diagnosed from that?

kshtsk:
@gregsfortytwo it is dump already elsewhere in separate file but when you look at some nonamed 'teuthology.log' it is not clear which run it corresponds too, it is useful for quick analysis so you don't switch context. I don't insist on this patch, but this is usability improvement.

smithfarm:
This seems reasonable to me.

PR #1224, opened 2018-10-19
valgrind: suppress __libc_csu_init leaks

liewegas:


gregsfortytwo:
Still plenty of valgrind failures, but is this particular suppression working @cbodley?

cbodley:
> Still plenty of valgrind failures, but is this particular suppression working @cbodley?

yes thanks! the issues in that run are in rgw

PR #1230, opened 2018-10-25
task/install: run install commands in a concatenated string

tchaikov:

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -90,8 +90,9 @@ class RemoteProcess(object):
         prefix = "Running:"
         if self.label:
             prefix = "Running ({label}):".format(label=self.label)
-        log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
-            cmd=self.command, prefix=prefix))
+        log.getChild(self.hostname).info(prefix)
+        for line in self.command.split('\n'):
+            log.getChild(self.hostname).info('> %s' % line)
 
         if hasattr(self, 'timeout'):
             (self._stdin_buf, self._stdout_buf, self._stderr_buf) = \
--- a/teuthology/task/install/rpm.py
+++ b/teuthology/task/install/rpm.py
@@ -1,6 +1,5 @@
 import logging
-
-from itertools import chain
+import os.path
 
 from teuthology.config import config as teuth_config
 from teuthology.orchestra import run
@@ -148,48 +147,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
         _yum_set_check_obsoletes(remote)
 
     if dist_release in ['opensuse', 'sle']:
-        remote.run(
-            args=[
-                'sudo', 'zypper', 'clean', '-a',
-            ])
+        remote.run(args='sudo zypper clean -a')
     else:
-        remote.run(
-            args=[
-                'sudo', 'yum', 'clean', 'all',
-            ])
+        remote.run(args='sudo yum clean all')
 
     ldir = _get_local_dir(config, remote)
 
     if dist_release in ['opensuse', 'sle']:
-        pkg_remove_cmd = 'sudo zypper -n remove --capability {pkg}'
-        pkg_install_cmd = 'sudo zypper -n --no-gpg-checks install --capability --no-recommends {pkg}'
+        remove_cmd = 'sudo zypper -n remove --capability'
+        install_cmd = 'sudo zypper -n --no-gpg-checks install --capability --no-recommends'
     else:
-        pkg_remove_cmd = 'sudo yum -y remove {pkg}'
-        pkg_install_cmd = 'sudo yum -y install {pkg}'
+        remove_cmd = 'sudo yum -y remove'
+        install_cmd = 'sudo yum -y install'
 
     for cpack in rpm:
-        pkg = None
         if ldir:
-            pkg = "{ldir}/{cpack}".format(
-                ldir=ldir,
-                cpack=cpack,
-            )
-            remote.run(
-                args=chain(
-                    ['if', 'test', '-e', run.Raw(pkg), run.Raw(';'), 'then'],
-                    pkg_remove_cmd.format(pkg=pkg).split(), [run.Raw(';')],
-                    pkg_install_cmd.format(pkg=pkg).split(), [run.Raw(';')],
-                    ['fi'])
-            )
-        if pkg is None:
-            remote.run(args=pkg_install_cmd.format(pkg=cpack).split())
+            remote.run(args='''
+              if test -e {pkg} ; then
+                {remove_cmd} {pkg} ;
+                {install_cmd} {pkg} ;
+              else
+                {install_cmd} {cpack} ;
+              fi
+            '''.format(remove_cmd=remove_cmd,
+                       install_cmd=install_cmd,
+                       pkg=os.path.join(ldir, cpack),
+                       cpack=cpack))
         else:
-            remote.run(
-                args=chain(
-                    ['if', 'test', run.Raw('!'), '-e', run.Raw(pkg), run.Raw(';'), 'then'],
-                    pkg_install_cmd.format(pkg=cpack).split(), [run.Raw(';')],
-                    ['fi'])
-            )
+            remote.run(args='''
+              {install_cmd} {cpack}
+            '''.format(install_cmd=install_cmd,
+                       cpack=cpack))
 
 
 def _yum_fix_repo_priority(remote, project, uri):


teuthology/task/install/rpm.py
```diff
@@ -148,48 +146,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
-                'sudo', 'yum', 'clean', 'all',
-            ])
+        remote.run(args='sudo yum clean all')
 
     ldir = _get_local_dir(config, remote)
 
     if dist_release in ['opensuse', 'sle']:
-        pkg_remove_cmd = 'sudo zypper -n remove --capability {pkg}'
-        pkg_install_cmd = 'sudo zypper -n --no-gpg-checks install --capability --no-recommends {pkg}'
+        remove_cmd = 'sudo zypper -n remove --capability {pkg}'
```
kshtsk: Here and below, do we really need this {pkg} template here?

kshtsk: Oh... I see, actually, the code is great.

teuthology/task/install/rpm.py
```diff
@@ -148,48 +146,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
     else:
-        pkg_remove_cmd = 'sudo yum -y remove {pkg}'
-        pkg_install_cmd = 'sudo yum -y install {pkg}'
+        remove_cmd = 'sudo yum -y remove {pkg}'
+        install_cmd = 'sudo yum -y install {pkg}'
 
     for cpack in rpm:
-        pkg = None
         if ldir:
             pkg = "{ldir}/{cpack}".format(
```
kshtsk: I would actually get rid of _pkg_ variable. See below.

teuthology/task/install/rpm.py
tchaikov: we need to handle two cases

1. local package where `{ldir}/{cpack}` is the first choice, and `{cpack}` is an alternative, 
1. and remote package, where `{ldir}` is `None` .

teuthology/task/install/rpm.py
```diff
@@ -148,48 +146,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
+            remote.run(args='''
+              if test -e {pkg} ; then
+                {remove_cmd_pkg} ;
+                {install_cmd_pkg} ;
+              else
+                {install_cmd_cpack} ;
+              fi
+            '''.format(remove_cmd_pkg=remove_cmd.format(pkg=pkg),
+                       install_cmd_pkg=install_cmd.format(pkg=pkg),
+                       install_cmd_cpack=install_cmd.format(pkg=cpack)))
```
kshtsk: What about this?
```
remote.run(args='''
    if test -e {ldir}/{pkg} ; then
        {remove_cmd} {ldir}/{pkg}
        {install_cmd} {ldir}/{pkg}
    else
        {install_cmd} {pkg}
    fi
'''.format(ldir=ldir, pkg=cpack,
        install_cmd=install_cmd,
        remove_cmd=remove_cmd))
```

tchaikov: @kshtsk i revised the change based on your proposal.

teuthology/task/install/rpm.py
```diff
@@ -148,48 +147,34 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
+            remote.run(args='''
+              if test -e {pkg} ; then
+                {remove_cmd} {pkg} ;
+                {install_cmd} {pkg} ;
+              else
+                {install_cmd} {cpack} ;
+              fi
+            '''.format(remove_cmd=remove_cmd,
+                       install_cmd=install_cmd,
+                       pkg=os.path.join(ldir, pkg),
```
kshtsk: You meant _os.path.join(ldir, **cpack**)_?

tchaikov: d'oh! fixed and repushed.

teuthology/task/install/rpm.py
```diff
@@ -148,48 +147,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
-            remote.run(
-                args=chain(
-                    ['if', 'test', run.Raw('!'), '-e', run.Raw(pkg), run.Raw(';'), 'then'],
-                    pkg_install_cmd.format(pkg=cpack).split(), [run.Raw(';')],
-                    ['fi'])
-            )
+            remote.run(args='''
+              {install_cmd} {pkg}
+            '''.format(install_cmd=install_cmd,
+                       pkg=cpack))
```
tchaikov: yeah, it's just for better symmetry..

tchaikov:
@kshtsk i think it's should be very close to the optimal form =)

teuthology/task/install/rpm.py
kshtsk: For symmetry it would be `cpack=cpack`?

tchaikov: @kshtsk fixed and repushed.

teuthology/task/install/rpm.py
```diff
@@ -148,48 +147,37 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
...
+              if test -e {pkg} ; then
+                {remove_cmd} {pkg} ;
+                {install_cmd} {pkg} ;
+              else
+                {install_cmd} {cpack} ;
+              fi
+            '''.format(remove_cmd=remove_cmd,
+                       install_cmd=install_cmd,
+                       pkg=os.path.join(ldir, cpack),
+                       cpack=cpack)
```
kshtsk: Missing parenthesis for _run_ method.

tchaikov: @kshtsk fixed and repushed.

smithfarm:
While this looks great in the code, it looks less great in the teuthology.log...

tchaikov:
@smithfarm could you be more specific? are you testing this change? what’s the teuthology.log look like?

smithfarm:
@tchaikov I tested similar code before. If I'm not mistaken, a string specified like this:

```
       """
            my
            pretty
            string
       """
```

will get logged (by `Remote.run()` via `logging`) like this:

`\n            my\n            pretty\n            string\n        `

so accompanying change to `Remote.run()` might be needed.

kshtsk:
@smithfarm the ';' longliners does not look much better either.
IMHO source code readability is more important than logs in this particular case.
I would suggest to fix logger as separate patch, like:
```
diff --git a/teuthology/orchestra/run.py b/teuthology/orchestra/run.py
index cccb8d38..2f48d080 100644
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -84,8 +84,9 @@ class RemoteProcess(object):
         prefix = "Running:"
         if self.label:
             prefix = "Running ({label}):".format(label=self.label)
-        log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
-            cmd=self.command, prefix=prefix))
+        for l in self.command.split('\n'):
+            log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
+                cmd=l, prefix=prefix))
 
         if hasattr(self, 'timeout'):
             (self._stdin_buf, self._stdout_buf, self._stderr_buf) = \
```

smithfarm:
> IMHO source code readability is more important than logs in this particular case.

I respectfully disagree. I spend *way* more time reading teuthology logs than I do this code.

smithfarm:
Anyway, let's wait for @zmc to weigh in.

@zmc Do you pass commands to `RemoteProcess` as strings anywhere? I've only ever seen them passed as lists.

Passing them as lists seems to have the following advantages:

1. list args can take advantage of quoting https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/run.py#L62-L65 (string args are taken literally)
2. list args look "normal" in the teuthology.log (thanks to their being passed through the `quote()` function, while string args get logged as-is (multiline strings end up being unreadable in the log)

Are there any other pros/cons?

kshtsk:
```
>>> R('xiii.local').run(args="""
...     x=my
...     y=pretty
...     z=string
...     for i in $x $y $z ; do
...             echo -n "$i "
...     done""")
2018-10-25 21:13:05,890.890 INFO:teuthology.orchestra.run.xiii:Running:
2018-10-25 21:13:05,890.890 INFO:teuthology.orchestra.run.xiii:true
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:Running:
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:	x=my
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:	y=pretty
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:	z=string
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:	for i in $x $y $z ; do
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:		echo -n "$i "
2018-10-25 21:13:05,944.944 INFO:teuthology.orchestra.run.xiii:	done
2018-10-25 21:13:05,987.987 INFO:teuthology.orchestra.run.xiii.stdout:my pretty string
RemoteProcess(client=<paramiko.client.SSHClient object at 0x1073b0650>, args='\n\tx=my\n\ty=pretty\n\tz=string\n\tfor i in $x $y $z ; do\n\t\techo -n "$i "\n\tdone', check_status=True, hostname='xiii')
```

kshtsk:
> Anyway, let's wait for @zmc to weigh in.
> 
Sure.
> @zmc Do you pass commands to `RemoteProcess` as strings anywhere? I've only ever seen them passed as lists.
> 
> Passing them as lists seems to have the following advantages:

The `run` supports already both variants, why should we use only one? If so, lets delete confusing code. I'm okay with both variants.

> 
> 1. list args can take advantage of quoting https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/run.py#L62-L65 (string args are taken literally)

For string args quoting is not required.

> 2. list args look "normal" in the teuthology.log (thanks to their being passed through the `quote()` function), while string args get logged as-is (multiline strings end up being unreadable in the log)
See my comment with fix log. Code looks much more readable than list of super-quoted elements.
> 
> Are there any other pros/cons?

Quoting disadvantages:
1. Cryptic, not readable, difficult to maintain. Looks like a garbage for long.
2. Makes you build crazy list joining manipulations.

Script as string advantages:
1. Human readable commands in the source code. That's saves you from mistakes.
2. You can copy&paste code from you bash script, easily and quickly integrate into your python code.
3. You can copy&paste command from the log and run locally for checking it in your environment.
4. Still able to copy python code into your program.

gregsfortytwo:
I missed the issue diagnosis that led to this PR and so the argument we're having over strings has left me really confused...in general I'd like us to match the other code, which I think is reasonably consistent about doing
```"I", "am", "a", "weird", "set", "of", "arguments"```
but I presume there was some specific issue with following our normal conventions in this bit of code?

**Comment by kshtsk** [comment]
@gregsfortytwo The starting point was this patch https://github.com/ceph/teuthology/pull/1229 where I suggesting improving code quality and readability based on my personal impression after reading teuthology code base.
The anamnesis for the strings vs. lists I venture to say the initial implementation of remote or node communication where using 'exec' like or process which has been executed 'ssh' with command as splitted list. Probably later, someone integrated **paramiko** which is using ssh protocol directly escaping command line ssh interface, which gave a possibility to normally run scripts via ssh _execute_ method. So far so good there left a lot of code using the old method and were not refactored due to compatibility or lack of the time, or will, or other more reasonable point. It's my _assumption only_ and I can be 100% wrong here. However the real history can be learned from the code.

**Comment by tchaikov** [comment]
@gregsfortytwo there are more occurrence of passing a whole command or even a script as `arg` to `RemoteClient.run()` in teuthology, for instance, see 

https://github.com/ceph/teuthology/blob/93566516cf4a6635cac3a9b3325f0ce200b3a157/teuthology/provision/fog.py#L290-L291 

i don't think we should drop the support of running `string`. as it's a convenient way to pass command to remote test node. actually, i think we should prefer `"A long time ago in a galaxy"` to `["A", "long", "time", "ago", "in", "a", "galaxy"]` under most circumstances.

**Comment by tchaikov** [comment]
changelog

- revise the commit message to explain the rationale of the change
- add a commit to flatten the multiple-line command line to a single in `RemoteProcess.execute()`.

File: teuthology/orchestra/run.py [hunk 7]
```diff
@@ -91,7 +91,7 @@ def execute(self):
         if self.label:
             prefix = "Running ({label}):".format(label=self.label)
         log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
-            cmd=self.command, prefix=prefix))
+            cmd=self.command.strip().replace('\n', ' '), prefix=prefix))
```
kshtsk: I thinks replacing the newlines has two disadvantages, this breaks bash syntax, it is not that readable as multiline.
So I vote for logging multiline commands as multiline logs commands.
From the other side, it is not necessary to prefix each line with 'Running (label) prefix' and it should be only logged once before the script.

tchaikov: @kshtsk agreed. i incorporated proposed change.

teuthology/orchestra/run.py
```diff
@@ -90,8 +90,9 @@ def execute(self):
         prefix = "Running:"
         if self.label:
             prefix = "Running ({label}):".format(label=self.label)
-        log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
-            cmd=self.command, prefix=prefix))
+        log.getChild(self.hostname).info(prefix)
+        for line in self.command.split('\n'):
+            log.getChild(self.hostname).info(line)
```
kshtsk: Nice, this is what I meant. However, since there is used _logger info_, and not _print()_ used it is probably better to prefix each line of command with specially character to give the user an easy way to recognize a script, for example, '> '.

tchaikov: @kshtsk updated and repushed.

teuthology/orchestra/run.py
```diff
@@ -90,8 +90,10 @@ def execute(self):
         prefix = "Running:"
         if self.label:
             prefix = "Running ({label}):".format(label=self.label)
-        log.getChild(self.hostname).info(u"{prefix} {cmd!r}".format(
-            cmd=self.command, prefix=prefix))
+        log.getChild(self.hostname).info(prefix)
+        for line in self.command.split('\n'):
+            log.getChild(self.hostname).info('{:>{indent}} {line}'.format(
+                '>', indent=len(prefix), line=line))
```
kshtsk: @tchaikov Thanks for the commit :-) please excuse me, but looks like that long indentation is overkill, especially when one would like to use _label_, check it out, the code:
```python
from teuthology.orchestra.remote import Remote as R
r=R('127.0.0.1')
r.run(label="I am not a weird set of arguments any more",
      args="""\
                for i in one two ; do
                    echo "$(( 1 + a++ )) is $i"
                done""")
```
will produce:
```
2018-10-26 21:21:08,740.740 INFO:teuthology.orchestra.run.127:Running (I am not a weird set of arguments any more):
2018-10-26 21:21:08,740.740 INFO:teuthology.orchestra.run.127:                                                    >                 for i in one two ; do
2018-10-26 21:21:08,740.740 INFO:teuthology.orchestra.run.127:                                                    >                     echo "$(( 1 + a++ )) is $i"
2018-10-26 21:21:08,740.740 INFO:teuthology.orchestra.run.127:                                                    >                 done
2018-10-26 21:21:08,785.785 INFO:teuthology.orchestra.run.127.stdout:1 is one
2018-10-26 21:21:08,785.785 INFO:teuthology.orchestra.run.127.stdout:2 is two
```
However if you use just `> ` prefix, it looks better, for example, this change:
```diff
diff --git a/teuthology/orchestra/run.py b/teuthology/orchestra/run.py
index 2f2d89c6..49266b89 100644
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -92,8 +92,7 @@ class RemoteProcess(object):
             prefix = "Running ({label}):".format(label=self.label)
         log.getChild(self.hostname).info(prefix)
         for line in self.command.split('\n'):
-            log.getChild(self.hostname).info('{:>{indent}} {line}'.format(
-                '>', indent=len(prefix), line=line))
+            log.getChild(self.hostname).info('> %s' % line)
 
         if hasattr(self, 'timeout'):
             (self._stdin_buf, self._stdout_buf, self._stderr_buf) = \
```
will give you out for the same script:
```
2018-10-26 21:25:12,926.926 INFO:teuthology.orchestra.run.127:Running (I am not a weird set of arguments any more):
2018-10-26 21:25:12,926.926 INFO:teuthology.orchestra.run.127:>                 for i in one two ; do
2018-10-26 21:25:12,926.926 INFO:teuthology.orchestra.run.127:>                     echo "$(( 1 + a++ )) is $i"
2018-10-26 21:25:12,926.926 INFO:teuthology.orchestra.run.127:>                 done
2018-10-26 21:25:12,970.970 INFO:teuthology.orchestra.run.127.stdout:1 is one
2018-10-26 21:25:12,970.970 INFO:teuthology.orchestra.run.127.stdout:2 is two
```
Thanks for patience, and patches.

tchaikov: @kshtsk updated and repushed.

tchaikov:
@zmc ping?

tchaikov:
@zmc ping?

tchaikov:
i am very sorry! i merged this chance by accident with the github web UI. thought it was "comment" button, not the "merge" button.

PR #1233, opened 2018-10-30
orchestra: add remote.sh commands analogous to misc.sh

kshtsk:
Adds a remote.sh similar to misc.sh, in fact a shortcut for remote.run,
but return output instead of proc
Example:
```
    my_name = Remote('127.0.0.1').sh('whoami')
```
Adds a remote.sh_file run a script as file on a remote with or without sudo
Example 1: Run python script with shebang
```
    Remote('127.0.0.1').sh_file("#!/usr/bin/env python3\n"
                                "import sys\n"
                                "print(sys.version_info)")
```
Example 2: Run script as root
```
    Remote('user@host.domain').sh_file("whoami", sudo=True,
                                       label="who-am-i-for-the-real")
```
Example 3: Run script as other user
```
    Remote('user@host.domain').sh_file("whoami", sudo='nobody', )
```
Example 4: Run local file script remotely with label
```
with open('/path/my/very/nice/script') as script_file:
   Remote('user@host.domain').sh_file(script_file, label='my-nice-script', )
```
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.de>

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -218,6 +218,47 @@ class Remote(object):
         data = proc.stdout.getvalue()
         return data
 
+    def sh(self, script, **kwargs):
+        """
+        Shortcut for run method.
+
+        Usage:
+            my_name = remote.sh('whoami')
+            remote_date = remote.sh('date')
+        """
+        if 'stdout' not in kwargs:
+            kwargs['stdout'] = StringIO()
+        if 'args' not in kwargs:
+            kwargs['args'] = script
+        proc=self.run(**kwargs)
+        return proc.stdout.getvalue()
+
+
+    def sh_file(self, script, label="script", sudo=False, **kwargs):
+        """
+        Run shell script after copying its contents to a remote file
+
+        :param script:  string with script text, or file object
+        :param sudo:    run command with sudo if True,
+                        run as user name if string value (defaults to False)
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/remote.py
```diff
@@ -218,6 +218,47 @@ def mktemp(self):
...
+        """
+        Run shell script after copying its contents to a remote file
+
+        :param script:  string with script text, or file object
+        :param sudo:    run command with sudo if True,
+                        run as user name if string value (defaults to False)
+        :param label:   string value which will be part of file name
+        Returns: stdout
+        """
+        ftempl = '/tmp/teuthology-remote-$(date +%Y%m%d%H%M%S)-{}-XXXX'\
```
gregsfortytwo: Why the XXXX at the end there?
(Rest of this looks good to me!)

tchaikov: quote from MKTEMP(1)

> SYNOPSIS
>        mktemp [OPTION]... [TEMPLATE]
> 
> DESCRIPTION
>        Create  a  temporary  file or directory, safely, and print its name.  TEMPLATE must contain at least 3 consecutive 'X's in last component.  If TEMPLATE is not specified, use tmp.XXXXXXXXXX, and --tmpdir is implied.

PR #1234, opened 2018-11-07
 skip calling greenlet exception handler

batrick:
We print the error to teuthology log anyway. This avoids inexplicable EIO
    errors printed by the greenlet exception errors.
    
    Fixes: http://tracker.ceph.com/issues/36731

### DIFF
--- a/docs/_static/worker_start.sh
+++ b/docs/_static/worker_start.sh
@@ -24,8 +24,8 @@ function start_all {
 }
  
 function main {
-    echo "$@"
-    if [[ -z "$@" ]]
+    printf '%s\n' "$*"
+    if [[ -z "$*" ]]
     then
         start_all
     elif [ ! -z "$2" ] && [ "$2" -gt "0" ]
@@ -37,4 +37,4 @@ function main {
     fi
 }
  
-main $@
+main "$@"
--- a/teuthology/__init__.py
+++ b/teuthology/__init__.py
@@ -100,10 +100,12 @@ def patch_gevent_hub_error_handler():
     Hub._origin_handle_error = Hub.handle_error
 
     def custom_handle_error(self, context, type, value, tb):
-        if not issubclass(type, Hub.SYSTEM_ERROR + Hub.NOT_ERROR):
+        if context is None or issubclass(type, Hub.SYSTEM_ERROR):
+            self.handle_system_error(type, value)
+        elif issubclass(type, Hub.NOT_ERROR):
+            pass
+        else:
             log.error("Uncaught exception (Hub)", exc_info=(type, value, tb))
-
-        self._origin_handle_error(context, type, value, tb)
 
     Hub.handle_error = custom_handle_error
 


teuthology/__init__.py
```diff
@@ -103,7 +103,9 @@ def custom_handle_error(self, context, type, value, tb):
         if not issubclass(type, Hub.SYSTEM_ERROR + Hub.NOT_ERROR):
             log.error("Uncaught exception (Hub)", exc_info=(type, value, tb))
 
-        self._origin_handle_error(context, type, value, tb)
+        # Do not call greenlet exception handler, it encounters an EIO error
```
tchaikov: i think the reason why we have i/o error is that: teuthworker is running without terminal attached to it. put in other way, when a background process not properly daemonized tries to write to stdout/stderr, i/o error is expected.

we can either fix it by 

- closing its stdin,  stdout and stderr or 
- redirecting them to `/dev/null`. or, 
- just don't try to read/write to them.

because gevent not only prints the backtrace but also takes some extra measures to fix things when handling `hub.SYSTEM_ERROR`. see  https://github.com/gevent/gevent/blob/a8c1f1712bd2010cf28eb2f42774be273ce7dd4d/src/gevent/hub.py#L454 

i'd suggest put something like

```py
if context is None or issubclass(type, Hub.SYSTEM_ERROR):
  self.handle_system_error(type, value)
elif issubclass(type, Hub.NOT_ERROR):
  pass
else:
  log.error("Uncaught exception (Hub)", exc_info=(type, value, tb))
```

batrick: Thanks, that makes sense Kefu. I've added your suggested code and also modified worker_start.sh to redirect to /dev/null. What do you think?

docs/_static/worker_start.sh
```diff
@@ -9,7 +9,7 @@ function start_workers_for_tube {
     echo "Starting $2 workers for $1"
     for i in `seq 1 $2`
     do
-        teuthology-worker -v --archive-dir $ARCHIVE --tube $1 --log-dir $WORKER_LOGS &
+        teuthology-worker -v --archive-dir $ARCHIVE --tube $1 --log-dir $WORKER_LOGS > dev/null 2>&1 &
```
gregsfortytwo: Shouldn't we write it...somewhere? Rather than nowhere?

tchaikov: yeah, ideally we should write it to somewhere. but

1. question `#1`:  what is "it". we have checked 3 different cases, namely:

   1. the `Hub.NOT_ERROR` case, where an exception is thrown when gevent stops or exits. they are bypassed.
   1. the `Hub.SYSTEM_ERROR` case, where the exception is pretty much only gevent is care understand/about. they are sent to https://github.com/gevent/gevent/blob/a8c1f1712bd2010cf28eb2f42774be273ce7dd4d/src/gevent/hub.py#L454 . i assume they will not be printed to stderr/stdout. but if they will, i guess it's something *unexpected*. and an i/o error helps us to notice these kinds of error in a way that not very "developer friendly".
   1. the other exceptions. they are printed to `log`.

2. question '#2': where is somewhere?  
as @batrick put in 2dbc296e98a18a1dc9e903944801b026bdb215ba, we already print the exception that we are interested in to the log file. so this is the best we can have for a  process running in background?

tl;dr: i think this is the best we can have.

gregsfortytwo: Okay, I must be missing how the exceptions can get printed in those cases, then?

tchaikov: yeah, i think they are directed to `teuthology.log` with Patrick's change. =)

batrick:
I'm planning to run test this with a cephfs run sometime this week.

tchaikov:
@batrick i am afraid that we cannot test teuthworker changes without deploying them in sepia =(. one way to test the change it is probably 

1. to deploy the changed copy somewhere 
1. start another teuthworker process using the new code 
1. schedule a teuthology run in hope the new worker gets picked

but it still needs the privilege to launch process on behalf of "teuthworker" in lab.

teuthology/__init__.py
vasukulkarni: Doesn't this also mean somewhere in the test there is print statement instead of log.info() inside the gevent block? If so one can remove those print and use log.info() to fix.

batrick: The print is in the gevent code. Not feasible to fix.

batrick:
I think let's just merge and deploy the changes then since risk is really low IMO. What do you think @tchaikov ?

docs/_static/worker_start.sh
kshtsk: @tchaikov 
a) You've missed 'root' slash for '/dev/null' device.
b) `somewhere`, for example, $WORKER_LOGS/worker-$1.$i.txt, often stderr contains some errors which does not get into python logger, and sometimes it is easier to find logs by constant filenames, in opposite to names based on random process id.

batrick:
@tchaikov @kshtsk oops, fixed!

kshtsk:
@batrick what about syntax errors from python interpreter, you will never see the reason why worker script can't be started.

tchaikov:
@batrick @kshtsk how about dropping 51855e1 ? and hope for the best? and probably close or redirect stdout/stderr in worker script later in a follow-up PR?

batrick:
> @batrick @kshtsk how about dropping 51855e1 ? and hope for the best? and probably close or redirect stdout/stderr in worker script later in a follow-up PR?

Okay, sounds good to me. PR updated.

batrick:
Works as advertised to prevent the IOErrors, although there are still some other issues with timeouts that need figured out:

/ceph/teuthology-archive/pdonnell-2018-11-21_02:23:40-multimds-wip-pdonnell-testing-20181120.235909-distro-basic-smithi/3275521/teuthology.log

PR #1236, opened 2018-11-13
ceph_ansible: Install netaddr

djgalloway:
See https://github.com/ceph/ceph-ansible/pull/3199


yuriw:
http://pulpito.ceph.com/yuriw-2018-11-13_21:01:17-ceph-ansible-luminous-distro-basic-ovh/

PR #1237, opened 2018-11-21
opsys: Add RHEL8

djgalloway:


smithfarm:
@djgalloway Maybe add it for centos as well (just a couple lines down)?

smithfarm:
> There isn't a CentOS 8 yet :-)

But there will be :-)

PR #1238, opened 2018-11-22
run_tasks.py: log the right ImportError message

smithfarm:
It turns out it's possible for a file qa/tasks/foo.py to exist,
yet importing it still raises an ImportError because it references a
non-existent symbol.
    
In this case, teuthology was clobbering the real ImportError with its
own bogus text.

Fixes: http://tracker.ceph.com/issues/37370

### DIFF
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -23,12 +23,9 @@ def get_task(name):
 
     # First look for the tasks's module inside teuthology
     module = _import('teuthology.task', module_name, task_name)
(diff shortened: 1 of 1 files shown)


smithfarm:
test this please

gregsfortytwo:
jenkins test this please

...I don't think I'm actually on the admin list, gonna need to get that fixed...

gregsfortytwo:
Oh hey, I am at least whitelisted!

zmc:
Thanks!

PR #1243, opened 2018-12-10
task/install: keep the order of pkgs to be installed

tchaikov:
before this change, the package list to be installed are ordered by
the hash function used by set().

after this change, the order of packages are not changed. so we can
ensure the order of package installation is the identical to that
of package list in yaml.


tchaikov:
see also https://github.com/ceph/ceph/pull/25463

PR #1244, opened 2018-12-11
task/install: add "downgrade_packages" option for "install" task

tchaikov:
see https://github.com/ceph/ceph/pull/25463 for a use case of this change.

see http://pulpito.ceph.com/kchai-2018-12-11_05:52:36-rados-wip-kefu2-testing-2018-12-11-0943-distro-basic-smithi/ for the addressed failures.

### DIFF
--- a/teuthology/task/install/rpm.py
+++ b/teuthology/task/install/rpm.py
@@ -1,6 +1,8 @@
 import logging
 import os.path
 
+from distutils.version import LooseVersion
+
 from teuthology.config import config as teuth_config
 from teuthology.orchestra import run
 from teuthology import packaging
@@ -99,6 +101,38 @@ def _zypper_removerepo(remote, repo_list):
             'sudo', 'zypper', '-n', 'removerepo', repo['name'],
         ])
 
+
+def _downgrade_packages(ctx, remote, pkgs, pkg_version, config):
+    """
+    Downgrade packages listed by 'downgrade_packages'
+
+    Downgrade specified packages to given version. The list of packages
+    downgrade is provided by 'downgrade_packages' as a property of "install"
+    task.
+
+    :param ctx: the argparse.Namespace object
+    :param remote: the teuthology.orchestra.remote.Remote object
+    :param pkgs: list of package names to install
+    :param pkg_version: the version to which all packages will be downgraded
+    :param config: the config dict
+    :return: list of package names from 'pkgs' which are not yet
+             installed/downgraded
+    """
+    downgrade_pkgs = config.get('downgrade_packages', [])
+    if not downgrade_pkgs:
+        return pkgs
+    # assuming we are going to downgrade packages with the same version
+    first_pkg = downgrade_pkgs[0]
+    installed_version = packaging.get_package_version(remote, first_pkg)
+    assert installed_version, "failed to get version of {}".format(first_pkg)
+    assert LooseVersion(installed_version) < LooseVersion(pkg_version)
+    # to compose package name like "librados2-0.94.10-87.g116a558.el7"
+    pkgs_opt = ['-'.join([pkg, pkg_version]) for pkg in downgrade_pkgs]
+    downgrade_cmd = 'sudo yum -y downgrade {}'.format(' '.join(pkgs_opt))
+    remote.run(args=downgrade_cmd.format(pkgs=pkgs_opt))
+    return [pkg for pkg in pkgs if pkg not in downgrade_pkgs]
+
+
 def _update_package_list_and_install(ctx, remote, rpm, config):
     """
     Installs the repository for the relevant branch, then installs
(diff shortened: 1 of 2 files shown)


teuthology/task/install/rpm.py
```diff
@@ -99,6 +101,23 @@ def _zypper_removerepo(remote, repo_list):
             'sudo', 'zypper', '-n', 'removerepo', repo['name'],
         ])
 
+
+def _downgrade_packages(ctx, remote, pkgs, pkg_version, config):
```
susebot: Add inline descriptions for the function, what should it do and what returns.

teuthology/task/install/rpm.py
```diff
@@ -99,6 +101,23 @@ def _zypper_removerepo(remote, repo_list):
...
+    downgrade_pkgs = config.get('downgrade_packages', [])
+    if not downgrade_pkgs:
+        return pkgs
+    # assuming we are going to downgrade packages with the same version
+    first_pkg = downgrade_pkgs[0]
```
susebot: What if downgrade_pkgs is empty list?

teuthology/task/install/rpm.py
```diff
@@ -99,6 +101,23 @@ def _zypper_removerepo(remote, repo_list):
...
+    installed_version = packaging.get_package_version(remote, first_pkg)
```
susebot: Why only the first element is checked? What if the condition is different for first and the rest elements?

teuthology/task/install/rpm.py
```diff
@@ -99,6 +101,23 @@ def _zypper_removerepo(remote, repo_list):
...
+    assert installed_version, "failed to get version of {}".format(first_pkg)
+    assert LooseVersion(installed_version) < LooseVersion(pkg_version)
+    # to compose package name like "librados2-0.94.10-87.g116a558.el7"
+    pkgs_opt = ['-'.join([pkg, pkg_version]) for pkg in downgrade_pkgs]
+    downgrade_cmd = 'sudo yum -y downgrade {}'.format(' '.join(pkgs_opt))
+    remote.run(args=downgrade_cmd.format(pkgs=pkgs_opt))
+    return [pkg for pkg in pkgs if pkg not in downgrade_pkgs]
```
susebot: Do I understand correctly the method returns the list of packages which has not been downgraded?

teuthology/task/install/rpm.py
tchaikov: in that case, this function will return before it reaches this line. see https://github.com/ceph/teuthology/pull/1244#discussion-diff-240600067R108

teuthology/task/install/rpm.py
tchaikov: please see the comment. it's only for downgrade the packages of the same version.

teuthology/task/install/rpm.py
kshtsk: Aha.

teuthology/task/install/rpm.py
kshtsk: I don't understand, if the downgrade_pkgs is a list of packages supposed to be downgraded, why only the first packages from the list is checked for the version?

kshtsk: Are all packages from pkgs supposed to be the same version?

kshtsk: So this function is only for usage with ceph packages which have unique version convention/

tchaikov: @kshtsk this fix is designed to address the needs of https://github.com/ceph/ceph/pull/25463 . where librados2 and librbd1 are of the *same* version. and they are supposed to be downgrade in the *single* yum command. otherwise, yum will refuse to downgrade the package if we just pass either of them to `yum downgrade` separately.

tchaikov: > So this function is only for usage with ceph packages which have unique version convention/

or for any project which creates multiple packages.

susebot: oh... it is homogeneous packages downgrade... I see

teuthology/task/install/rpm.py
```diff
@@ -99,6 +101,35 @@ def _zypper_removerepo(remote, repo_list):
             'sudo', 'zypper', '-n', 'removerepo', repo['name'],
         ])
 
+
+def _downgrade_packages(ctx, remote, pkgs, pkg_version, config):
+    """
+    Downgrade packages listed by 'downgrade_packages'
```
kshtsk: Maybe add 
```
Downgrade packages which are supposed to have the same version and are provided by 'downgrade_packages' in teuthology config file. Do nothing if package list empty.
```

tchaikov: as the suggested change is not quite accurate, i adapted it and revised the doc string.

tchaikov:
changelog

- fix typos and syntax errors in doc string.

PR #1245, opened 2018-12-21
task/install: populate "downgrade_packages" to nested_config

tchaikov:
- populate "downgrade_packages" to `nested_config` in `task()`.
  otherwise, the "downgrade_packages" is invisible to the `install()`
  function.
- also print logging message before downgrading packages.
- cleanup rpm._downgrade_packages(), no need to format `downgrade_cmd`.
  it's already formated.


tchaikov:
@neha-ojha unlike the tasks reside in `ceph/qa`, the built-in tasks in teuthology are native in teuthology worker, and we cannot test it until deploy them. put it in other words, we need to install them into `/home/teuthworker/src/git.ceph.com_git_teuthology_master` for testing them.

PR #1248, opened 2019-01-15
teuthology-describe-tests: add --summary command

kshtsk:
Add --summary argument to teuthology-describe-tests for estimating run capacity, printing descriptions, dumping matrix,  yaml fragments.

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -122,6 +122,11 @@ Scheduler arguments:
   --filter-all KEYWORDS       Only run jobs whose description contains each one
                               of the keywords in the comma separated keyword
                               string specified.
+  -F, --filter-fragments <bool>
+                              Check yaml fragments too if job description
+                              does not match the filters provided with
+                              options --filter, --filter-out, and --filter-all.
+                              [default: false]
   --archive-upload RSYNC_DEST Rsync destination to upload archives.
   --archive-upload-url URL    Public facing URL where archives are uploaded.
   --throttle SLEEP            When scheduling, wait SLEEP seconds between jobs.
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -75,7 +75,8 @@ Standard arguments:
                               assembling jobs from yaml fragments. This causes
                               <suite_branch> to be ignored for scheduling
                               purposes, but it will still be used for test
-                              running.
+                              running. The <suite_dir> must have `qa/suite`
+                              sub-directory.
   --validate-sha1 <bool>
                               Validate that git SHA1s passed to -S exist.
                               [default: true]
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -78,7 +78,7 @@ def process_args(args):
             value = expand_short_repo_name(
                 value,
                 config.get_ceph_qa_suite_git_url())
-        elif key in ('validate_sha1'):
+        elif key in ('validate_sha1', 'filter_fragments'):
             value = strtobool(value)
         conf[key] = value
     return conf
--- a/scripts/describe_tests.py
+++ b/scripts/describe_tests.py
@@ -37,12 +37,23 @@ optional arguments:
 options only for describing combinations represented by a suite:
   -c, --combinations                 Describe test combinations rather than
                                      individual yaml fragments
-  --filter <keywords>                Only list jobs whose filenames contain at
-                                     least one of the keywords in the comma
-                                     separated keyword string specified.
-  --filter-out <keywords>            Do not list jobs whose filenames contain
+  -s, --summary                      Print summary
+  --filter <keywords>                Only list tests whose description contains
+                                     at least one of the keywords in the comma
+                                     separated keyword string specified
+  --filter-out <keywords>            Do not list tests whose description contains
                                      any of the keywords in the comma separated
-                                     keyword string specified.
+                                     keyword string specified
+  --filter-all <keywords>            Only list tests whose description contains
+                                     each of the keywords in the comma separated
+                                     keyword string specified
+  -F, --filter-fragments             Check fragments additionaly to descriptions
+                                     using keywords specified with 'filter',
+                                     'filter-out' and 'filter-all' options.
+  -D, --print-description            Print job descriptions for the suite,
+                                     used only in combination with 'summary'
+  -L, --print-fragments              Print file list inovolved for each facet,
+                                     used only in combination with 'summary'
   -l <jobs>, --limit <jobs>          List at most this many jobs
                                      [default: 0]
   --subset <index/outof>             Instead of listing the entire
@@ -54,6 +65,11 @@ options only for describing combinations represented by a suite:
                                      2/<outof> ... <outof>-1/<outof>
                                      will list all jobs in the
                                      suite (many more than once).
+  -S <seed>, --seed <seed>           Used for pseudo-random tests generation
+                                     involving facet whose path ends with '$'
+                                     operator, where negative value used for
+                                     a random seed
+                                     [default: -1]
 """
 
 
--- a/teuthology/describe_tests.py
+++ b/teuthology/describe_tests.py
@@ -7,9 +7,13 @@ import os
 import sys
 import yaml
 
-from teuthology.exceptions import ParseError
-from teuthology.suite.build_matrix import build_matrix, combine_path
+import random
+from distutils.util import strtobool
 
+from teuthology.exceptions import ParseError
+from teuthology.suite.build_matrix import \
+        build_matrix, generate_combinations, _get_matrix
+from teuthology.suite import util
 
 def main(args):
     try:
@@ -20,27 +24,60 @@ def main(args):
 
 def describe_tests(args):
     suite_dir = os.path.abspath(args["<suite_dir>"])
-    fields = args["--fields"].split(',')
-    include_facet = args['--show-facet'] == 'yes'
     output_format = args['--format']
 
+    conf=dict()
+    rename_args = {
+        'filter': 'filter_in',
+    }
+    for (key, value) in args.items():
+        key = key.lstrip('--').replace('-', '_')
+        key = rename_args.get(key) or key
+        if key in ('filter_all', 'filter_in', 'filter_out', 'fields'):
+            if not value:
+                value = []
+            else:
+                value = [_ for _ in
+                            (x.strip() for x in value.split(',')) if _]
+        elif key in ('limit'):
+            value = int(value)
+        elif key in ('seed'):
+            value = int(value)
+            if value < 0:
+                value = None
+        elif key == 'subset' and value is not None:
+            # take input string '2/3' and turn into (2, 3)
+            value = tuple(map(int, value.split('/')))
+        elif key in ('show_facet'):
+            value = strtobool(value)
+        conf[key] = value
+
     if args['--combinations']:
-        limit = int(args['--limit'])
-        filter_in = None
-        if args['--filter']:
-            filter_in = [f.strip() for f in args['--filter'].split(',')]
-        filter_out = None
-        if args['--filter-out']:
-            filter_out = [f.strip() for f in args['--filter-out'].split(',')]
-        subset = None
-        if args['--subset']:
-            subset = map(int, args['--subset'].split('/'))
-        headers, rows = get_combinations(suite_dir, fields, subset,
-                                         limit, filter_in,
-                                         filter_out, include_facet)
+        headers, rows = get_combinations(suite_dir,
+                                         limit=conf['limit'],
+                                         seed=conf['seed'],
+                                         subset=conf['subset'],
+                                         fields=conf['fields'],
+                                         filter_in=conf['filter_in'],
+                                         filter_out=conf['filter_out'],
+                                         filter_all=conf['filter_all'],
+                                         filter_fragments=conf['filter_fragments'],
+                                         include_facet=conf['show_facet'])
         hrule = ALL
+    elif args['--summary']:
+        output_summary(suite_dir,
+                       limit=conf['limit'],
+                       seed=conf['seed'],
+                       subset=conf['subset'],
+                       show_desc=conf['print_description'],
+                       show_frag=conf['print_fragments'],
+                       filter_in=conf['filter_in'],
+                       filter_out=conf['filter_out'],
+                       filter_all=conf['filter_all'],
+                       filter_fragments=conf['filter_fragments'])
+        exit(0)
     else:
-        headers, rows = describe_suite(suite_dir, fields, include_facet,
+        headers, rows = describe_suite(suite_dir, conf['fields'], conf['show_facet'],
                                        output_format)
         hrule = FRAME
 
@@ -69,9 +106,60 @@ def output_results(headers, rows, output_format, hrule):
         print(table)
 
 
-def get_combinations(suite_dir, fields, subset,
-                     limit, filter_in, filter_out,
-                     include_facet):
+def output_summary(path, limit=0,
+                         seed=None,
+                         subset=None,
+                         show_desc=True,
+                         show_frag=False,
+                         show_matrix=False,
+                         filter_in=None,
+                         filter_out=None,
+                         filter_all=None,
+                         filter_fragments=True):
+    """
+    Prints number of all facets for a given suite for inspection,
+    taking into accout such options like --subset, --filter,
+    --filter-out and --filter-all. Optionally dumps matrix objects,
+    yaml files which is used for generating combinations.
+    """
+
+    random.seed(seed)
+    mat, first, matlimit = _get_matrix(path, subset)
+    configs = generate_combinations(path, mat, first, matlimit)
+    print("# {} (not filtered) {}".format(len(configs), path))
+    count = 0
+    suite = os.path.basename(path)
+    config_list = util.filter_configs(configs,
+                                      suite_name=suite,
+                                      filter_in=filter_in,
+                                      filter_out=filter_out,
+                                      filter_all=filter_all,
+                                      filter_fragments=filter_fragments)
+    if show_desc or show_frag:
+        for c in config_list:
+            if limit and count >= limit:
+                break
+            count += 1
+            print("  {}".format(c[0]))
+            if show_frag:
+                for path in c[1]:
+                    print("    {}".format(util.strip_fragment_path(path)))
+    else:
+        count=sum(1 for _ in config_list)
+    if show_matrix:
+       print(mat.tostr(1))
+    print("  {} (total filtered)".format(count))
+
+def get_combinations(suite_dir,
+                     limit=0,
+                     seed=None,
+                     subset=None,
+                     fields=[],
+                     filter_in=None,
+                     filter_out=None,
+                     filter_all=None,
+                     filter_fragments=False,
+                     include_facet=True):
     """
     Describes the combinations of a suite, optionally limiting
     or filtering output based on the given parameters. Includes
@@ -80,8 +168,8 @@ def get_combinations(suite_dir, fields, subset,
     Returns a tuple of (headers, rows) where both elements are lists
     of strings.
     """
-    configs = [(combine_path(suite_dir, item[0]), item[1]) for item in
-               build_matrix(suite_dir, subset)]
+    suite = os.path.basename(suite_dir)
+    configs = build_matrix(suite_dir, subset, seed)
 
     num_listed = 0
     rows = []
@@ -90,15 +178,15 @@ def get_combinations(suite_dir, fields, subset,
     dirs = {}
     max_dir_depth = 0
 
+    configs = util.filter_configs(configs,
+                                  suite_name=suite,
+                                  filter_in=filter_in,
+                                  filter_out=filter_out,
+                                  filter_all=filter_all,
+                                  filter_fragments=filter_fragments)
     for _, fragment_paths in configs:
         if limit > 0 and num_listed >= limit:
             break
-        if filter_in and not any([f in path for f in filter_in
-                                  for path in fragment_paths]):
-            continue
-        if filter_out and any([f in path for f in filter_out
-                               for path in fragment_paths]):
-            continue
 
         fragment_fields = [extract_info(path, fields)
                            for path in fragment_paths]
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -21,7 +21,7 @@ from teuthology.orchestra.opsys import OS
 from teuthology.repo_utils import build_git_url
 
 from teuthology.suite import util
-from teuthology.suite.build_matrix import combine_path, build_matrix
+from teuthology.suite.build_matrix import build_matrix
 from teuthology.suite.placeholder import substitute_placeholders, dict_templ
 
 log = logging.getLogger(__name__)
@@ -363,34 +363,11 @@ class Run(object):
         jobs_to_schedule = []
         jobs_missing_packages = []
         for description, fragment_paths in configs:
-            base_frag_paths = [
-                util.strip_fragment_path(x) for x in fragment_paths
-            ]
             if limit > 0 and len(jobs_to_schedule) >= limit:
                 log.info(
                     'Stopped after {limit} jobs due to --limit={limit}'.format(
                         limit=limit))
                 break
-            # Break apart the filter parameter (one string) into comma
-            # separated components to be used in searches.
-            def matches(f):
-                if f in description:
-                    return True
-                if any(f in path for path in base_frag_paths):
-                    return True
-                return False
-            filter_all = self.args.filter_all
-            if filter_all:
-                if not all(matches(f) for f in filter_all):
-                    continue
-            filter_in = self.args.filter_in
-            if filter_in:
-                if not any(matches(f) for f in filter_in):
-                    continue
-            filter_out = self.args.filter_out
-            if filter_out:
-                if any(matches(f) for f in filter_out):
-                    continue
 
             raw_yaml = '\n'.join([open(a, 'r').read() for a in fragment_paths])
 
@@ -487,6 +464,7 @@ class Run(object):
                 log.info("pause between jobs : --throttle " + str(throttle))
                 time.sleep(int(throttle))
 
+
     def schedule_suite(self):
         """
         Schedule the suite-run. Returns the number of jobs scheduled.
@@ -501,10 +479,9 @@ class Run(object):
             self.base_config.suite.replace(':', '/'),
         ))
         log.debug('Suite %s in %s' % (suite_name, suite_path))
-        configs = [
-            (combine_path(suite_name, item[0]), item[1]) for item in
-            build_matrix(suite_path, subset=self.args.subset, seed=self.args.seed)
-        ]
+        configs = build_matrix(suite_path,
+                               subset=self.args.subset,
+                               seed=self.args.seed)
         log.info('Suite %s in %s generated %d jobs (not yet filtered)' % (
             suite_name, suite_path, len(configs)))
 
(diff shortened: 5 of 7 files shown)


kshtsk:
jenkins retest this please

jdurgin:
this is similar to teuthology-describe-tests - would it make sense to add options for this kind of output there instead of a new command?

kshtsk:
> this is similar to teuthology-describe-tests - would it make sense to add options for this kind of output there instead of a new command?

I don't know what does this tools supposed to, I've just checked it, does not work for me with errors like:
```
IOError: [Errno 40] Too many levels of symbolic links: '/mnt/terra/ceph/suse/qa/suites/rgw/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/.qa/deepsea/distros/opensuse_42.3.yaml'
```

kshtsk:
@jdurgin do you mean to add '--calc' option same as '--combinations'?

kshtsk:
@jdurgin from the descriptions of describe-tests:
```
Describe the contents of a qa suite by reading 'meta' elements from
yaml files in the suite.

The 'meta' element should contain a list with a dictionary
of key/value pairs for entries, i.e.:
```
Looks like the purpose of those scripts is different.

scripts/calc.py
```diff
@@ -0,0 +1,23 @@
+"""
+usage: teuthology-calc [-h] [options] [<suite_path>...]
+
+Print suite config description, matrix dump, file list
```
vasukulkarni: What is the purpose of this cli? what is it useful for?  Feels like part of it is same as teuthology-describe?

teuthology/calc.py
```diff
@@ -0,0 +1,31 @@
+from teuthology.suite.build_matrix import build_matrix
+from teuthology.suite.build_matrix import _get_matrix
+
+def  main(args):
```
vasukulkarni: Needs doc string

gregsfortytwo:
Actually @kshtsk, can you explain how this is different from just doing a teuthology-suite dry run?

And I would like to strengthen Josh's suggestion that whatever we're after be implemented as a teuthology-suite subcommand.

kshtsk:
@gregsfortytwo @jdurgin 
I actually was  not aware of teuthology-describe-tests when I started to code this tool,  probably I just didn't figure out how to use it in first place because it just does not work for me because of infinity loop bug, and from the description I didn't understand it does something useful for me.
So my original plan was to put this functionallity to teuthology-suite, but since it DOES checks out corresponding repos from github and requires some config file intrusion for changing the repo, it is not doing the thing I wanted. Also teuthology-suite prints a lot of logging garbage and doing not-necessary actions. The purpose of the _calc_ is to get the "picture" what would be executed by teuthology-suite but in more quick and flexible way, this is why providing a path to the _suite_ is much better than just branch, because in order to see what you get in a suite does not require any git push operation and can easily proceed in local directory. Probably I was not much specific in the help of the tool. So, what does it do:
1) It shows the number of jobs it will run
2) Show the exact same "description" which is used in pulpito per each job.
3) Can show all the files which are used for the given suite the same way teuthology does, not trying to go throw the directory structure itself.
4) Can show inter Matrix representation for the given suite, like product, subset cycle index, sum, etc.

scripts/calc.py
kshtsk: @vasukulkarni the teuthology-describe does not do the same.

teuthology/calc.py
kshtsk: @vasukulkarni Not sure I understand, the docstring is in the script/calc.py already

scripts/calc.py
kshtsk: @vasukulkarni from the descriptions of describe-tests:
```
Describe the contents of a qa suite by reading 'meta' elements from
yaml files in the suite.

The 'meta' element should contain a list with a dictionary
of key/value pairs for entries, i.e.:
```
The calc has nothing to do with 'meta' elements, it just shows the teuthology forensic data, exactly what is parsed by teuthology.suite.build_matrix.

kshtsk:
@vasukulkarni @jdurgin @gregsfortytwo Moved the code to _teuthology-suite_ tool.

kshtsk:
jenkins retest this please

kshtsk:
jenkins retest this please

kshtsk:
I have no idea what make the jenkins fail on this PR.

teuthology/calc.py
kshtsk: The code moved to teuthology/suite.py

teuthology/suite/__init__.py
```diff
@@ -86,11 +88,55 @@ def expand_short_repo_name(name, orig):
...
+    incl_keys=[]
+    excl_keys=[]
+    if args['--filter']:
+        incl_keys = args['--filter'].split(',')
+        log.debug("Using include filter: {}".format(incl_keys))
+    if args['--filter-out']:
+        excl_keys = args['--filter-out'].split(',')
+        log.debug("Using exclude filter: {}".format(excl_keys))
+
+    _seed=int(args['--seed'])
```
gregsfortytwo: Shouldn't we be using some default value in case the seed just wasn't set? Setting it in the "normal" path is definitely weird but also a little more complicated than we're doing here...

kshtsk: @gregsfortytwo the default value for seed is -1, if the default value is given there is code which generates a random seed value, which is not what we want sine the seed will be different each time. So I just set to None default value.

kshtsk: However, I tried to use run --calc with different --seed value and it looks like nothing is changed, though it does not generate different Cycle sequence with combination --subset, it is not critical right now.

gregsfortytwo: Do we actually go through that default path when running ```--calc```?
Granted I don't really understand how the seed is used here but it looks like it's only interpreted in the rerun path, so since we're shortcutting everything except for ```process_args``` we can't make any assumptions what it looks like.

scripts/suite.py
```diff
@@ -29,6 +30,16 @@
   <config_yaml>               Optional extra job yaml to include
   -s <suite>, --suite <suite>
                               The suite to schedule
+  -C <suite_path>, --calc <suite_path>
+                              Path to a suite directory for calculation,
+                              see also --desc, --list, --matrix, as well as 
+                              options --subset, --filter and --filter-out
+  -P, --desc                  Print job descriptions for the suite,
```
vasukulkarni: I used the cli to understand better, -P and -L are similar to --dry-run, are you saying dry-run doesnt work with local dirs and hence this change?  Is it worth fixing dry-run to handle this, my concern is two cli's doing almost similar job.

kshtsk: @vasukulkarni From my understanding --dry-run is to go throw the run, like run but do not schedule a job, correspondingly it generates a lot of noise in the logs, the idea of calc is to drop the noise and focus on suite analysis only.

vasukulkarni: I probably think you are little bit misunderstanding the -dry-run option using same seed

Here is what I tried(dry-run also takes --suite-dir option so it can work with local dirs)

a) ./teuthology-suite --dry-run -s fs -c master  --suite-dir ~/ceph/ceph/  -m vps --seed 3550
b)./teuthology-suite --calc ~/ceph/ceph/qa/suites/fs -P --seed 3550

-L would just print the number of tests and in both cases is easy to figure out
-P has less noise but the yaml snippets are exactly the same, but when the number grows bit large its upto user to make sense, but should be possible to fix that noise better?

-M is a new option in this change.

kshtsk: @vasukulkarni I was able to run --dry-run without extra teuthology configured file only on working freshly deployed system. And what I get is not what I expect to have:
1) For me it is not good thing that just for investigation of a suite I need to add noisy arguments like machine type, branch repos, as a user I want to have simple way, and not afraid that if I miss for example one of --ceph, --suite, or --machine (have no idea why I need to enter all those parameters to examine a suite). Probably for someone else it useable.

2) Noisy output, where each facet printed in one line as a debug log, it is not a convenient way of representing data and I don't like there idea to keep in memory the best way to parse to understand whole this:
```
2019-01-23 01:05:25,672.672 INFO:teuthology.suite.util:/home/ubuntu/teuthology/virtualenv/bin/teuthology-schedule --name ubuntu-2019-01-23_01:05:21-rbd-ses6---basic-openstack --num 1 --worker openstack --dry-run --priority 1000 --description 'rbd/cli_v1/{base/install.yaml clusters/{fixed-1.yaml openstack.yaml} features/format-1.yaml msgr-failures/few.yaml objectstore/bluestore-bitmap.yaml pool/small-cache-pool.yaml supported-random-distro$/rhel_latest.yaml workloads/rbd_cli_import_export.yaml}' -- /tmp/schedule_suite_cUtuwe src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/base/install.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/clusters/fixed-1.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/clusters/openstack.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/features/format-1.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/msgr-failures/few.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/objectstore/bluestore-bitmap.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/pool/small-cache-pool.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/supported-random-distro$/rhel_latest.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/workloads/rbd_cli_import_export.yaml
2019-01-23 01:05:25,672.672 INFO:teuthology.suite.run:Scheduling rbd/mirror/{base/install.yaml cluster/{2-node.yaml openstack.yaml} msgr-failures/few.yaml objectstore/bluestore-comp.yaml supported-random-distro$/centos_latest.yaml users/mirror.yaml workloads/rbd-mirror-workunit-config-key.yaml}
```
Like I need to add always '2>&1 | grep -v teuthology-schedule ... | sed ...' or 'grep Scheduling' it complicates everything so much.
3) It does extra steps, which I don't like to be executed anyway while working on local copy, like, each try to run teuthology-suite:

```
2019-01-23 01:22:10,102.102 INFO:teuthology.suite.run:ceph sha1: 237cb653cce794c517680eb53da8045214f4e7b1
2019-01-23 01:22:10,103.103 INFO:teuthology.suite.run:skipping ceph package verification
2019-01-23 01:22:10,103.103 INFO:teuthology.suite.run:teuthology branch: master
2019-01-23 01:22:11,492.492 INFO:teuthology.suite.run:ceph branch: ses6 237cb653cce794c517680eb53da8045214f4e7b1
```
and it is in the best case, when you have correctly setup the client node with all configs.

vasukulkarni: > @vasukulkarni I was able to run --dry-run without extra teuthology configured file only on working freshly deployed system. And what I get is not what I expect to have:
> 
> 1. For me it is not good thing that just for investigation of a suite I need to add noisy arguments like machine type, branch repos, as a user I want to have simple way, and not afraid that if I miss for example one of --ceph, --suite, or --machine (have no idea why I need to enter all those parameters to examine a suite). Probably for someone else it useable.
What I am saying is since the output yamls are exactly same with few additional info(which I really dont consider as noise, since the user is expected few details when working with teuthology) it might be better we standardize on same options rather than new sub options, probably add --quiet option, I am not saying we use crazy cli options to filter out things we dont want, it should be easy to fix this in the suite function and silence the logs we dont need. It *will* be helpful for everybody if we fix the standard options everyone is already using and simplify things if they are doing same things.

> 2. Noisy output, where each facet printed in one line as a debug log, it is not a convenient way of representing data and I don't like there idea to keep in memory the best way to parse to understand whole this:
> 
> ```
> 2019-01-23 01:05:25,672.672 INFO:teuthology.suite.util:/home/ubuntu/teuthology/virtualenv/bin/teuthology-schedule --name ubuntu-2019-01-23_01:05:21-rbd-ses6---basic-openstack --num 1 --worker openstack --dry-run --priority 1000 --description 'rbd/cli_v1/{base/install.yaml clusters/{fixed-1.yaml openstack.yaml} features/format-1.yaml msgr-failures/few.yaml objectstore/bluestore-bitmap.yaml pool/small-cache-pool.yaml supported-random-distro$/rhel_latest.yaml workloads/rbd_cli_import_export.yaml}' -- /tmp/schedule_suite_cUtuwe src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/base/install.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/clusters/fixed-1.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/clusters/openstack.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/features/format-1.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/msgr-failures/few.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/objectstore/bluestore-bitmap.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/pool/small-cache-pool.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/supported-random-distro$/rhel_latest.yaml src/github.com_suse_ceph_ses6/qa/suites/rbd/cli_v1/workloads/rbd_cli_import_export.yaml
> 2019-01-23 01:05:25,672.672 INFO:teuthology.suite.run:Scheduling rbd/mirror/{base/install.yaml cluster/{2-node.yaml openstack.yaml} msgr-failures/few.yaml objectstore/bluestore-comp.yaml supported-random-distro$/centos_latest.yaml users/mirror.yaml workloads/rbd-mirror-workunit-config-key.yaml}
> ```
> Like I need to add always '2>&1 | grep -v teuthology-schedule ... | sed ...' or 'grep Scheduling' it complicates everything so much.
> 3) It does extra steps, which I don't like to be executed anyway while working on local copy, like, each try to run teuthology-suite:
> 
> ```
> 2019-01-23 01:22:10,102.102 INFO:teuthology.suite.run:ceph sha1: 237cb653cce794c517680eb53da8045214f4e7b1
> 2019-01-23 01:22:10,103.103 INFO:teuthology.suite.run:skipping ceph package verification
> 2019-01-23 01:22:10,103.103 INFO:teuthology.suite.run:teuthology branch: master
> 2019-01-23 01:22:11,492.492 INFO:teuthology.suite.run:ceph branch: ses6 237cb653cce794c517680eb53da8045214f4e7b1
> ```
> and it is in the best case, when you have correctly setup the client node with all configs.

kshtsk: @vasukulkarni 
I see, and want to admit that, it would be great to have some kind of standardization, the only think I don't like is to use `--dry-run` command for just listing a suite, just because the purpose of dry-run is different, as a confirmation of it, --dry-run does not prints out the suites in pretty format it dumps the command logs (info or debug), like dump `teuthology-schedule` command to stderr. And it is just coincidence that all those yaml files have got into debug dump. In other words using dry-run for calc purpose looks like workaround. I would agree that instead `--calc` we would rather use `--suite-dir` however I would prefer it to point to the directory of a particular suite, instead of pointing to the root of some repo with hardcoded file structure. Summarizing using debug like output with all those timestamps and noisy decorations  just for investigating suites looks very weird and clumsy.

vasukulkarni: I didnt' get what you meant by pointing at root dir, I can point at suite level and it works for me. 

If we want to keep this outside `--dry-run`, then how about using a better name instead of --calc (like --analyze) ?

and instead of calculation use analysis in description below,  Also -a option is available.

```
  -C <suite_path>, --calc <suite_path>
                              Path to a suite directory for calculation,
                              see also --desc, --list, --matrix, as well as 
                              options --subset, --filter and --filter-out
```

vasukulkarni: Also -P and -L should be default option.

kshtsk: @vasukulkarni The way I use this command is to evaluate the capacity of a suite, so, the default behaviour just to look at the number of cases for the suite, so I don't need grep out all the cases. And after playing with filters, the is a sense to print out the description (-P option). Btw, this is why I used 'calc' keyword, short for calculate, however I don't mind to rename it to something else, but I would to leave a possibility to enter it multiple times, since it is useful sometimes. I also prefer short command name over the longer, like 'teuthology-describe-tests', As for -L options it will also be overkill because the list of yaml is the same for many facets. I don't like it be enable by default, and by the way to disable will require more complicated command line combination either: '-L false' or 'grep -V ...'

kshtsk: > I didnt' get what you meant by pointing at root dir, I can point at suite level and it works for me.
I mean 'root' git repository dir, see example:
1) You have to provide '--ceph', '--suite', '--machine' (why should I?):
```
$ teuthology-suite --dry-run --suite-dir ~/src/github.com_SUSE_ceph_ses5/qa/suites/rbd/basic
usage: teuthology-suite --help
       teuthology-suite [-v | -vv ] --ceph <ceph> --suite <suite> [options] [<config_yaml>...]
       teuthology-suite [-v | -vv ] --ceph <ceph> --rerun <name>  [options] [<config_yaml>...]
```
2) You have to provide suite name explicitly as separate argument, and even after that it looks for specific sub-directory of '--suite-dir', in particular 'qa/suites':
```

$ teuthology-suite --dry-run --suite-dir ~/src/github.com_SUSE_ceph_ses5/qa/suites/rbd/basic --ceph ses5 --machine openstack --suite rbd/basic
2019-01-29 14:30:22,731.731 INFO:teuthology.suite.run:ceph sha1: b10c702661a31c8563b3421d6d664de93a0cb0e2
2019-01-29 14:30:22,732.732 INFO:teuthology.suite.run:skipping ceph package verification
2019-01-29 14:30:22,733.733 INFO:teuthology.suite.run:teuthology branch: master
2019-01-29 14:30:24,452.452 INFO:teuthology.suite.run:ceph branch: ses5 b10c702661a31c8563b3421d6d664de93a0cb0e2
  File "/home/ubuntu/teuthology/virtualenv/bin/teuthology-suite", line 11, in <module>
    load_entry_point('teuthology', 'console_scripts', 'teuthology-suite')()
  File "/home/ubuntu/teuthology/scripts/suite.py", line 142, in main
    return teuthology.suite.main(args)
  File "/home/ubuntu/teuthology/teuthology/suite/__init__.py", line 88, in main
    run.prepare_and_schedule()
  File "/home/ubuntu/teuthology/teuthology/suite/run.py", line 311, in prepare_and_schedule
    num_jobs = self.schedule_suite()
  File "/home/ubuntu/teuthology/teuthology/suite/run.py", line 478, in schedule_suite
    build_matrix(suite_path, subset=self.args.subset)
  File "/home/ubuntu/teuthology/teuthology/suite/build_matrix.py", line 50, in build_matrix
    mat, first, matlimit = _get_matrix(path, subset)
  File "/home/ubuntu/teuthology/teuthology/suite/build_matrix.py", line 68, in _get_matrix
    mat = _build_matrix(path)
  File "/home/ubuntu/teuthology/teuthology/suite/build_matrix.py", line 77, in _build_matrix
    raise IOError('%s does not exist (abs %s)' % (path, os.path.abspath(path)))
IOError: /home/ubuntu/src/github.com_SUSE_ceph_ses5/qa/suites/rbd/basic/qa/suites/rbd/basic does not exist (abs /home/ubuntu/src/github.com_SUSE_ceph_ses5/qa/suites/rbd/basic/qa/suites/rbd/basic)
```

kshtsk: > @kshtsk I dont have much to say here except that it can have a more helpful name for folks who can print help and know better , 'calc' seems like generic name. 
I'm not against other convenient name, but would like to keep the logic.
> Also when I tried this command with default and -L it just prints the number of tests as shown below, so I strongly feel it should set those as defaults, Is there something I am missing here on usage?
Yes, you missed this:
```
-L, --list                  Print file list inovolved for building suite,
                               used only with -C and --desc arguments
```
So, if -P not provided, -L does not make sense, however, we can change it to 'if -L is used, -P is treated as true.'

kshtsk: @varadakari I've addressed not working '-L' if '-P'.

> how about using a better name instead of --calc (like --analyze) ?

do you think it is really good name? :-) maybe `--analyse`  is better, or support both `--analyze` and `--analyse` as interchangeable arguments?

kshtsk:
jenkins retest this please

kshtsk:
@gregsfortytwo any idea what happens on jenkins?

kshtsk:
@gregsfortytwo Probably something wrong with jenkins environment or master branch is broken. 
The only suspect thing I can see in the log is:
```flake8 runtests: commands[0] | flake8 --select=F,E9 teuthology scripts
  /home/jenkins-build/build/workspace/teuthology-pull-requests$ /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/flake8/bin/flake8 --select=F,E9 teuthology scripts 
teuthology/suite/run.py:139:16: F632 use ==/!= to compare str, bytes, and int literals
teuthology/suite/test/test_build_matrix.py:624:16: F632 use ==/!= to compare str, bytes, and int literals
teuthology/suite/test/test_build_matrix.py:626:16: F632 use ==/!= to compare str, bytes, and int literals
teuthology/suite/test/test_build_matrix.py:629:16: F632 use ==/!= to compare str, bytes, and int literals
teuthology/suite/test/test_build_matrix.py:631:16: F632 use ==/!= to compare str, bytes, and int literals
teuthology/task/lockfile.py:228:12: F632 use ==/!= to compare str, bytes, and int literals
ERROR: InvocationError for command '/home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/flake8/bin/flake8 --select=F,E9 teuthology scripts' (exited with code 1)```

But other PR has the same error message, for example:
https://jenkins.ceph.com/job/teuthology-pull-requests/1411/

kshtsk:
jenkins retest this please

kshtsk:
So the jenkins failure are not related to the PR.

kshtsk:
@jdurgin @gregsfortytwo @vasukulkarni are we good with this PR?

gregsfortytwo:
Going back to my last review, looks like the seed thing is still messed up. And a test to validate it against dry-run output would still be good!

gregsfortytwo:
Anything automated will do, but yeah I would probably start with some grep'ing, sort'ing, and diff'ing.

kshtsk:
> Anything automated will do, but yeah I would probably start with some grep'ing, sort'ing, and diff'ing.

It will be equivalent to fix --dry-run so it can be executed without special production environment, because as I said earlier you can just run teuthology-suite --dry-run without special environment. This would be an overkill.

kshtsk:
@gregsfortytwo There is no sense in making test comparing `--dry-run` and `--inspect` in sense of seed, because using of seed is different by design for both, by default `--dry-run` should generate random seed for randomized suite generation, for the latter it should not be random.
We could try and run both for different same seed values. However it will not be unit test, it is system test because it requires specific test environment, if the `--inspect` can be run just from source code directory, the `--dry-run` requires more fixes, not related to this PR. Since this PR is not changing any functionality, I would like to merge it, and suggest to create another PR for testing `--dry-run`, anyways I do not see a benefit in this test.

gregsfortytwo:
Trying to match up this output with that of dry-run isn't to test dry-run, it's to validate that --inspect is doing what we expect. It would be VERY BAD if somebody ran inspect and somehow got different results than were to come out of the normal suite execution engine!
That doesn't mean we need to execute a test on every install, but we can make it part of the teuthology self-tests that run in Jenkins: do the setup necessary to invoke dry-run, grab its output, and compare it to the output of --inspect.

(I think the seed is a red herring; it only applies when you're doing a subset with the magic $ (random) input. So for a full suite run it won't apply? Or if it does, we can force the same seed value in for both.)

kshtsk:
> Trying to match up this output with that of dry-run isn't to test dry-run, it's to validate that --inspect is doing what we expect. It would be VERY BAD if somebody ran inspect and somehow got different results than were to come out of the normal suite execution engine!

But we will need to run both.
Well the `inspect` is using same API as `dry-run`, correspondingly it should generate same facet set. I ran manually a comparison test on suites from ceph repo. But if you want to run it for each commit on each PR it will require checkout whole ceph repo each time. Or we will need to have some test hierarchy of empty files, and we are not safe that 'dry-run' won't fail while trying to detect some git repo stuff by this path. 

I also don't seed  a sense in this test, because later one can decide to remove test description from the log of dry-run.

If we really want to test --inspect, not 'dry-run' there is no sense to compare it with `dry-run` at all, because `dry-run` also requires to be tested that it is doing what we expect, and we are not testing this now.

So in ideal world we need to have test subtree in teuthology source tree to play role as test vectors for inspect. Btw, the matrix tests are using dummy in memory fs, which can not be used on cli level.


> That doesn't mean we need to execute a test on every install, but we can make it part of the teuthology self-tests that run in Jenkins: do the setup necessary to invoke dry-run, grab its output, and compare it to the output of --inspect.
> 
> (I think the seed is a red herring; it only applies when you're doing a subset with the magic $ (random) input. So for a full suite run it won't apply? Or if it does, we can force the same seed value in for both.)

Right, the seed make sense only when 'subset' option is used, otherwise it is just ignored.

gregsfortytwo:
> > Trying to match up this output with that of dry-run isn't to test dry-run, it's to validate that --inspect is doing what we expect. It would be VERY BAD if somebody ran inspect and somehow got different results than were to come out of the normal suite execution engine!
> 
> But we will need to run both.
> Well the `inspect` is using same API as `dry-run`, correspondingly it should generate same facet set.

Hmm, the implementation is mostly in build_matrix, but I'm worried about some subtle difference creeping in as part of the filtering since that's duplicated. I guess if that were factored out and the same code used we wouldn't need an extra test here.

zmc:
I just spoke to @kshtsk and did a little bit of digging re: how to test this. What I'm thinking the right thing to do would be to take a look at the `build_matrix` tests - specifically things similar to [this](https://github.com/ceph/teuthology/blob/master/teuthology/suite/test/test_build_matrix.py#L61-L81). The use of `mock` to create a fake 'suite' has been effective, and it seems like it would work fell for this feature.

I'd also argue that it doesn't need to e.g. verify all possible flag combinations, and that testing that the proper args are passed to `build_matrix`, trusting that those tests do their job.

gregsfortytwo:
Yeah that would be fine; like I said it's the re-implementation of filtering that concerns me at this point.

zmc:
Oh, sure - good point. I agree that it'd be worth attempting to reuse that code.

kshtsk:
@jdurgin @gregsfortytwo 
The functionality moved to teuthology-describe-tests, the `teuthology-suite --inspect` reincarnated as `teuthology-describe-tests --summary`. The code is now using same execution path as teuthology.suite.run Run.collect_jobs for filters. Also teuthology-describe-tests fixed to use --filter --filter-out and --filter-all in same manner as in teuthology-suite, as well as --seed and new option --filter-fragments.

kshtsk:
Also I suggest to rename `teuthology-describe-tests` to shorter name `teuthology-describe`, what do you think about it?

kshtsk:
@susebot run deploy

jdurgin:
> Also I suggest to rename `teuthology-describe-tests` to shorter name `teuthology-describe`, what do you think about it?

sounds good to me

PR #1249, opened 2019-01-18
Cluster name support removed 

smanjara:
ceph-ansible task yaml will use the name 'cluster' in rgw multisite test suite, purely to differentiate between role names provided to the clusters in the yaml. But it will not be using the cluster names during configuration.


smanjara:
Reorganizing the commits in PR.

PR #1251, opened 2019-01-21
Fix matrix dump tostr() method

kshtsk:
Addresses issue while trying to dump matrix object:
```
  AttributeError: Base instance has no attribute '__getitem__'
```
Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.de>

### DIFF
--- a/teuthology/suite/matrix.py
+++ b/teuthology/suite/matrix.py
@@ -196,7 +196,7 @@ class Concat(Matrix):
 
     def tostr(self, depth):
         ret = '\t'*depth + "Concat({item}):\n".format(item=self.item)
-        return ret + ''.join([i[1].tostr(depth+1) for i in self.submats])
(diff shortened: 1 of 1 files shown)


teuthology/suite/matrix.py
```diff
@@ -196,7 +196,7 @@ def index(self, i):
 
     def tostr(self, depth):
         ret = '\t'*depth + "Concat({item}):\n".format(item=self.item)
-        return ret + ''.join([i[1].tostr(depth+1) for i in self.submats])
+        return ret + ''.join([i.tostr(depth+1) for i in self.submats])
```
tchaikov: i think `PickRandom` suffers from the same problem. could you help fix it as well?

kshtsk: Done

PR #1252, opened 2019-01-24
New cli teuthology-reimage-fog to reimage FOG provisioned nodes

vasukulkarni:
add new cli teuthology-reimage-fog to reimage FOG provisioned nodes without actually
locking the nodes.

### DIFF
--- a/scripts/reimage_fog.py
+++ b/scripts/reimage_fog.py
@@ -1,0 +1,22 @@
+import docopt
+
+from teuthology.misc import reimage_fog
+import sys
+
+doc = """
+usage: teuthology-reimage-fog -h
+       teuthology-reimage-fog --nodes node1,node2 --os-type distro --os-version version
+
+Reimage nodes using FOG without locking the nodes
+
+standard arguments:
+  -h, --help                           Show this help message and exit
+  --nodes node1,node2                  List of nodes to reimage
+  --os-type <os-type>                  Distribution type eg: rhel, ubuntu
+  --os-version <os-version>            OS version eg: 7.6, 16.04 etc
+"""
+
+
+def main(argv=sys.argv[1:]):
+    args = docopt.docopt(doc, argv=argv)
+    reimage_fog(args)
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -20,6 +20,8 @@ import yaml
 import json
 import re
 import pprint
+import parallel
+
 
 from netaddr.strategy.ipv4 import valid_str as _is_ipv4
 from netaddr.strategy.ipv6 import valid_str as _is_ipv6
@@ -1371,6 +1373,21 @@ def is_in_dict(searchkey, searchval, d):
         return True
     else:
         return searchval == val
+
+
+def reimage_fog(args):
+    """
+    Reimage FOG nodes with options specified
+    """
+    machines = args['--nodes']
+    nodes = machines.rstrip(',').split(',')
+    ctx = argparse.Namespace()
+    ctx.os_type = args['--os-type']
+    ctx.os_version = args['--os-version']
+    from teuthology.provision import reimage
+    with parallel() as p:
+        for node in nodes:
+            p.spawn(reimage, ctx, node)
 
 
 def sh(command, log_limit=1024):
--- a/tox.ini
+++ b/tox.ini
@@ -30,6 +30,7 @@ commands=py.test --cov=teuthology --cov-report=term -v {posargs:teuthology/test/
 basepython=python2.7
 
 [testenv:flake8]
+ignore=F632
 install_command = pip install --upgrade {opts} {packages}
 deps=
   flake8
--- a/setup.py
+++ b/setup.py
@@ -115,6 +115,7 @@ setup(
             'teuthology-queue = scripts.queue:main',
             'teuthology-prune-logs = scripts.prune_logs:main',
             'teuthology-describe-tests = scripts.describe_tests:main',
+            'teuthology-reimage-fog = scripts.reimage_fog:main'
             ],
         },
 


vasukulkarni:
@djgalloway @alfredodeza: I am getting a python3 environment for tests, Did anything changes in test nodes request for Jenkins jobs?

djgalloway:
> reimage FOG provisioned nodes without actually locking the nodes

Why would you want to do this?  Won't this be more confusing than manually locking nodes and not knowing whether they're actually being used?

vasukulkarni:
> > reimage FOG provisioned nodes without actually locking the nodes
> 
> Why would you want to do this? Won't this be more confusing than manually locking nodes and not knowing whether they're actually being used?

@djgalloway  This is specifically for 2 use cases  a) when the user has the systems locked but wants to reimage them again same or different OS  without going through unlock/update status down/lock again b) In octo we are having trouble with mero/callypso with timeouts and panic,  so I am fixing this where we will just reimage things once and jobs can use the same OS.

djgalloway:
jenkins retest this

kshtsk:
@vasukulkarni @djgalloway one of my PR has a fix for current jenkins failures #1255 or #1254 check it out.

teuthology/misc.py
```diff
@@ -1373,6 +1374,21 @@ def is_in_dict(searchkey, searchval, d):
...
+def reimage_fog(args):
+    """
+    Reimage FOG nodes with options specified
+    """
+    machines = args['--nodes']
+    nodes = machines.rstrip(',').split(',')
+    ctx = argparse.Namespace()
+    ctx.os_type = args['--os-type']
+    ctx.os_version = args['--os-version']
+    from teuthology.provision import reimage
```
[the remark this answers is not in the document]
kshtsk: @vasukulkarni can this be resolved in other way than putting import inside methods?

tox.ini
```diff
@@ -30,6 +30,7 @@ commands=py.test --cov=teuthology --cov-report=term -v {posargs:teuthology/test/
 basepython=python2.7
 
 [testenv:flake8]
+ignore=F632
```
kshtsk: I would not use this, but fix the python3 related issues one by one, especially when it is declared that 2.7 will be deprecate in less than a year.

vasukulkarni: @kshtsk can you separate the commit into its own PR? the subject would be clear then, Also, teuthology is not ready for python3 yet, all the tests use python2 environment, I see this as error in flake8 thats doing py3 flake tests in py2 environment, when we actually run in py3 environment it would make sense to look at all the errors.

kshtsk: > @kshtsk can you separate the commit into its own PR? the subject would be clear then, Also, teuthology is not ready for python3 yet, all the tests use python2 environment, I see this as error in flake8 thats doing py3 flake tests in py2 environment, when we actually run in py3 environment it would make sense to look at all the errors.

"is" should not be used for python2.7 as well:
```
$ python2.7
Python 2.7.15 (default, Jan 18 2019, 02:15:11) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a=257
>>> b=257
>>> a==b
True
>>> a is b
False
>>> 

```

vasukulkarni:
> It's perfectly sensible to want to reimage your own locked nodes, but as with all the stuff teuthology does, we should not add functionality that touches nodes without the lock status. If the current user owns the lock it can proceed; otherwise it should fail. We can add the standard "--owner" flag (or whatever it is) if we need to do this from a different account.

Sure I fogot about the owner option, I will fix this.

tox.ini
kshtsk: flake8 related commit should be dropped

scripts/reimage_fog.py
```diff
@@ -0,0 +1,22 @@
+import docopt
+
+from teuthology.misc import reimage_fog
+import sys
+
+doc = """
+usage: teuthology-reimage-fog -h
+       teuthology-reimage-fog --nodes node1,node2 --os-type distro --os-version version
```
kshtsk: It is more usable to provide node names without a key, so bash expansion can be used, like in teuthology-lock:
```
teuthology-reimage-fog --os-type distro --os-version X.Y node{1,2}
```

kshtsk:
> It's perfectly sensible to want to reimage your own locked nodes, but as with all the stuff teuthology does, we should not add functionality that touches nodes without the lock status. If the current user owns the lock it can proceed; otherwise it should fail. We can add the standard "--owner" flag (or whatever it is) if we need to do this from a different account.

I agree with Greg that the lock should be always for machines which are registered in paddles, using --owner flag does not help much because some other jobs you've scheduled earlier can jump in the middle of reimaging. And vice versa. Although the --owner flag is not supper secure.

teuthology/misc.py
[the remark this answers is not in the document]
kshtsk: We should avoid circular imports.
I would suggest reimage_fog to put to a separate module like others corresponding to scripts or just put it directly to to script, since you're importing teuthology.misc there?

kshtsk:
@vasukulkarni are you going to work on this?

kshtsk:
another idea, we better rename this cli to `teuthology-reimage` and allow it to respect teuthology config to use the provisioning method machine has

ryg-:
> another idea, we better rename this cli to `teuthology-reimage` and allow it to respect teuthology config to use the provisioning method machine has
I agree with that, I don't see FOG specific functionality in it, pelagos should be compatible with it without code changes, so it bill be good if names be generic

kshtsk:
Closing this in favor of #1526 since it will be merged soon.

PR #1257, opened 2019-02-11
test_repo_utils: fix shared temporary directory

kshtsk:
Signed-off-by: Kyr <kyrylo.shatskyy@suse.com>

### DIFF
--- a/teuthology/test/test_repo_utils.py
+++ b/teuthology/test/test_repo_utils.py
@@ -4,6 +4,7 @@ import os.path
 from pytest import raises, mark
 import shutil
 import subprocess
+import tempfile
 
 from ..exceptions import BranchNotFoundError
 from .. import repo_utils
@@ -12,20 +13,21 @@ repo_utils.log.setLevel(logging.WARNING)
 
 
 class TestRepoUtils(object):
-    src_path = '/tmp/empty_src'
-    # online_repo_url = 'https://github.com/ceph/teuthology.git'
-    # online_repo_url = 'git://ceph.newdream.net/git/teuthology.git'
-    online_repo_url = 'https://github.com/ceph/empty.git'
-    offline_repo_url = 'file://' + src_path
-    repo_url = None
-    dest_path = '/tmp/empty_dest'
 
     @classmethod
     def setup_class(cls):
+        cls.temp_path = tempfile.mkdtemp(prefix='test_repo-')
+        cls.dest_path = cls.temp_path + '/empty_dest'
+        cls.src_path = cls.temp_path + '/empty_src'
+
         if 'TEST_ONLINE' in os.environ:
-            cls.repo_url = cls.online_repo_url
+            cls.repo_url = 'https://github.com/ceph/empty.git'
         else:
-            cls.repo_url = cls.offline_repo_url
+            cls.repo_url = 'file://' + cls.src_path
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(cls.temp_path)
 
     def setup_method(self, method):
         assert not os.path.exists(self.dest_path)
@@ -67,7 +69,7 @@ class TestRepoUtils(object):
         assert not os.path.exists(self.dest_path)
 
     def test_fetch_no_repo(self):
-        fake_dest_path = '/tmp/not_a_repo'
+        fake_dest_path = self.temp_path + '/not_a_repo'
         assert not os.path.exists(fake_dest_path)
         with raises(OSError):
             repo_utils.fetch(fake_dest_path)
@@ -79,7 +81,7 @@ class TestRepoUtils(object):
         assert os.path.exists(self.dest_path)
 
     def test_fetch_branch_no_repo(self):
-        fake_dest_path = '/tmp/not_a_repo'
+        fake_dest_path = self.temp_path + '/not_a_repo'
         assert not os.path.exists(fake_dest_path)
         with raises(OSError):
             repo_utils.fetch_branch(fake_dest_path, 'master')


teuthology/test/test_repo_utils.py
```diff
@@ -12,13 +13,14 @@
 
 
 class TestRepoUtils(object):
-    src_path = '/tmp/empty_src'
+    temp_path = tempfile.mkdtemp(prefix='test_repo', dir='/tmp')
```
tchaikov: nit, i think it'd be better to allow user to customize the default directory by specifying environmental variables like `TMPDIR`, `TEMP` or `TMP`. and `/tmp` is the first directory used by `tempfile.mkdtemp()` tries to create a tempfile on UNIX-like OSes.

tchaikov: also, `temp_path` is created when this `class` is evaluated. but you `rm -rf` it whenever a test method is executed.

if this directory is supposed to be shared among the test methods, we should create it in `setup_class()` and nuke it in `teardown_class()`. and, yeah, initialize `dest_path` in `setup_method()`.

kshtsk: > nit, i think it'd be better to allow user to customize the default directory by specifying environmental variables like `TMPDIR`, `TEMP` or `TMP`. and `/tmp` is the first directory used by `tempfile.mkdtemp()` tries to create a tempfile on UNIX-like OSes.
This change is for resolving test failures in multiuser environment. I don't understand what exactly you want, which of the variable above you want to use? Do you want to use 'dir=sys.getenv('TEMP', '/tmp')`? Could you provide code example for it.

kshtsk: > also, `temp_path` is created when this `class` is evaluated. but you `rm -rf` it whenever a test method is executed.
> 
So, you mean that *teardown_method* should be renamed to `teardown_class`?

> if this directory is supposed to be shared among the test methods, we should create it in `setup_class()` and nuke it in `teardown_class()`. and, yeah, initialize `dest_path` in `setup_method()`.

We should create a temporary directory in the same place where we setup all those path variables, they are in constructor, do you like to move them to setup_class method?

tchaikov: @kshtsk see https://github.com/enthought/Python-2.7.3/blob/master/Lib/tempfile.py#L154 . i'd suggest remove `dir` parameter.

tchaikov: > So, you mean that *teardown_method* should be renamed to `teardown_class`?

no. `self.dest_path` is created by tests, for instance, see https://github.com/ceph/teuthology/pull/1257/files#diff-3bee1a3b7d5220a2093743ab3f0004f9R127 . so we should at least remove that file in `teardown_method()`. but i think the better approach is probably to remove the whole directory in `teardown_method()`. and create / initialize `temp_path` in `setup_method()`, unless we want to reuse the parent directory. but i cannot think of a reason to do so.

kshtsk: > @kshtsk see https://github.com/enthought/Python-2.7.3/blob/master/Lib/tempfile.py#L154 . i'd suggest remove `dir` parameter.

Agree.

kshtsk: @tchaikov Changes pushed, could you please review again.

teuthology/test/test_repo_utils.py
```diff
@@ -12,20 +13,23 @@
...
-    # online_repo_url = 'https://github.com/ceph/teuthology.git'
-    # online_repo_url = 'git://ceph.newdream.net/git/teuthology.git'
-    online_repo_url = 'https://github.com/ceph/empty.git'
-    offline_repo_url = 'file://' + src_path
-    repo_url = None
-    dest_path = '/tmp/empty_dest'
 
     @classmethod
-    def setup_class(cls):
+    def setup_class(self):
```
tchaikov: why s/cls/self/ ? i feel `self` in this context is a kinda misleading.

kshtsk: > why s/cls/self/ ? i feel `self` in this context is a kinda misleading.
There is no difference which is name of the first argument from one side, from other side since all test_ methods have 'self', it is better to be consistent and use the same naming.

tchaikov: please note the `self` here is not an instance of `TestRepoUtils`. and `setup_class()` is not a test method.

teuthology/test/test_repo_utils.py
```diff
@@ -12,20 +13,23 @@
...
+
         if 'TEST_ONLINE' in os.environ:
-            cls.repo_url = cls.online_repo_url
+            self.repo_url = self.online_repo_url
         else:
-            cls.repo_url = cls.offline_repo_url
+            self.repo_url = self.offline_repo_url
+
+    @classmethod
+    def teardown_class(self):
```
tchaikov: please s/self/cls/

teuthology/test/test_repo_utils.py
kshtsk: > please note the `self` here is not an instance of `TestRepoUtils`. and `setup_class()` is not a test method.
What is it if not the test class instance?

kshtsk: @tchaikov I addressed you request in the recent push.

teuthology/test/test_repo_utils.py
kshtsk: Addressed.

teuthology/test/test_repo_utils.py
tchaikov: probably you could take a look at https://docs.python.org/3/library/functions.html#classmethod ? and see also, https://github.com/pytest-dev/pytest/blob/5ca81596bbbb35da0b8b742afaf7ed2ef075ccda/src/_pytest/python.py#L732 , i don't have enough bandwidth looking into the details. but the gist is that it is very likely to be an instance of the class. and we cannot assume that it is an instance of `TestRepoUtils`.

kshtsk: > @tchaikov I addressed you request in the recent push.

@tchaikov ^

teuthology/test/test_repo_utils.py
```diff
@@ -12,21 +13,24 @@
...
-    offline_repo_url = 'file://' + src_path
-    repo_url = None
-    dest_path = '/tmp/empty_dest'
 
     @classmethod
     def setup_class(cls):
+        cls.temp_path = tempfile.mkdtemp(prefix='test_repo-')
+        cls.dest_path = cls.temp_path + '/empty_dest'
+        cls.src_path = cls.temp_path + '/empty_src'
+        cls.offline_repo_url = 'file://' + cls.src_path
```
tchaikov: nit, no need to assign these variable as class variables. they could just be local variables.

kshtsk: done, review

PR #1258, opened 2019-02-11
openstack: fix all volumes nuke

kshtsk:
Do not allow teuthology-nuke to delete all volumes on openstack.

There is a bug on some openstack clients that --name option
does not work correctly for 'openstack volume list' in contrast
to 'openstack server list' and lists all available volumes
ignoring '--name' option at all.

### DIFF
--- a/teuthology/nuke/__init__.py
+++ b/teuthology/nuke/__init__.py
@@ -34,6 +34,17 @@ from ..task.internal import check_lock, add_remotes, connect
 log = logging.getLogger(__name__)
 
 
+def openstack_volume_id(volume):
+    return (volume.get('ID') or volume['id'])
+
+
+def openstack_volume_name(volume):
+    return (volume.get('Display Name') or
+            volume.get('display_name') or
+            volume.get('Name') or
+            volume.get('name') or "")
+
+
 def stale_openstack(ctx):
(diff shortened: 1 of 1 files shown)


teuthology/nuke/__init__.py
```diff
@@ -156,11 +166,12 @@ def openstack_remove_again():
     xargs --no-run-if-empty --max-args 1 -P20 openstack server delete --wait
     true
     """)
-    sh("""
-    openstack volume list --name REMOVE-ME --column ID --format value |
-    xargs --no-run-if-empty --max-args 1 -P20 openstack volume delete
-    true
-    """)
+    volumes = OpenStack().run("volume list -f json --long")
```
tchaikov: i assume that `volumes` will the output of

```
openstack --quiet volume list -f json --long
```

so it will be a unicode string, right? can we iterate through a unicode string and invoke the `v.get()` on each of its element? because, i think `v` is a `str` presenting an element in `volumes`.

kshtsk: @tchaikov I've updated the code a little bit. It actually works in similar way to OpenStack().list_volumes()

tchaikov: yeah. now it does.

PR #1259, opened 2019-02-11
python-cov: update python mysql dependency

kshtsk:
Use more recent version of mysql python module

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,7 +47,7 @@ markupsafe==1.0           # via jinja2
 monotonic==1.3            # via oslo.utils
 more-itertools==4.3.0     # via pytest
 msgpack-python==0.4.8     # via oslo.serialization
-mysql-python==1.2.3
+mysqlclient==1.4.2
 ndg-httpsclient==0.4.2
 netaddr==0.7.19           # via oslo.config, oslo.utils, python-neutronclient
 netifaces==0.10.5         # via oslo.utils
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
                       'tox',
                       'gevent',
                       # For teuthology-coverage
-                      'MySQL-python == 1.2.3',
+                      'mysqlclient == 1.4.2',
                       'PyYAML',
                       'argparse >= 1.2.1',
                       'beanstalkc >= 0.2.0',


gregsfortytwo:
Looks fine to me, but did you check that this version is available everywhere? Or is that not a problem with the pip stuff?

kshtsk:
> Looks fine to me, but did you check that this version is available everywhere? Or is that not a problem with the pip stuff?

Actually I want this package because I have no problem with pip stuff for it on any system I have, I've tested it with purest py27 target on opensuse 42.3 and opensuse 15.0, as well as Mojave. If you can test it on redhat or debian, it would be great.

vasukulkarni:
> > Looks fine to me, but did you check that this version is available everywhere? Or is that not a problem with the pip stuff?
> 
> Actually I want this package because I have no problem with pip stuff for it on any system I have, I've tested it with purest py27 target on opensuse 42.3 and opensuse 15.0, as well as Mojave. If you can test it on redhat or debian, it would be great.

We have teuthology smoke suite that can be run in sepia lab, why are we not testing these changes there?

kshtsk:
> > > Looks fine to me, but did you check that this version is available everywhere? Or is that not a problem with the pip stuff?
> > 
> > 
> > Actually I want this package because I have no problem with pip stuff for it on any system I have, I've tested it with purest py27 target on opensuse 42.3 and opensuse 15.0, as well as Mojave. If you can test it on redhat or debian, it would be great.
> 
> We have teuthology smoke suite that can be run in sepia lab, why are we not testing these changes there?

1) I'm not sure I have access to the lab and I don't know how to test anything there?
2) The mysql python is not used for in teuthology, and only used with coverage pytest pluging afaik, correspondingly the Jenkins already tested this tested, see the status for this PR.

vasukulkarni:
> > > > Looks fine to me, but did you check that this version is available everywhere? Or is that not a problem with the pip stuff?
> > > 
> > > 
> > > Actually I want this package because I have no problem with pip stuff for it on any system I have, I've tested it with purest py27 target on opensuse 42.3 and opensuse 15.0, as well as Mojave. If you can test it on redhat or debian, it would be great.
> > 
> > 
> > We have teuthology smoke suite that can be run in sepia lab, why are we not testing these changes there?
> 
> 1. I'm not sure I have access to the lab and I don't know how to test anything there?
I am sorry that shouldn't be the reason for not testing the PR, sepia lab access is well documented here: https://ceph.github.io/sepia/adding_users/ and I am sure you know how to use the cli.

> 2. The mysql python is not used for in teuthology, and only used with coverage pytest pluging afaik, correspondingly the Jenkins already tested this tested, see the status for this PR.
Its used in coverage: https://github.com/ceph/teuthology/blob/master/teuthology/coverage.py#L6 , regardless we should be doing testing in sepia lab to ensure its not causing new regression.

kshtsk:
> Its used in coverage: https://github.com/ceph/teuthology/blob/master/teuthology/coverage.py#L6 , regardless we should be doing testing in sepia lab to ensure its not causing new regression.

Do you have an idea how to test this in the lab? Like, remove mysql-python and install mysqclient?

kshtsk:
@vasukulkarni 
> Its used in coverage: https://github.com/ceph/teuthology/blob/master/teuthology/coverage.py#L6 , regardless we should be doing testing in sepia lab to ensure its not causing new regression.

since this change only related to deployment of teuthology environment I don't understand how this patch can be tested in sepia, for example, it will not remove stale python-mysql package and not install the new one from requirements. If I am wrong could you please provide command which will test the 'coverage.py'?

vasukulkarni:
> @vasukulkarni
> 
> > Its used in coverage: https://github.com/ceph/teuthology/blob/master/teuthology/coverage.py#L6 , regardless we should be doing testing in sepia lab to ensure its not causing new regression.
> 
> since this change only related to deployment of teuthology environment I don't understand how this patch can be tested in sepia, for example, it will not remove stale python-mysql package and not install the new one from requirements.txt. If I am wrong could you please provide command which will test the 'coverage.py'?

The current coverage script is stale since its not been in use for a while, If you are confident that it wont break for later use then we can merge, but we shouldn't break it completely due to new pip module replacement.

PR #1263, opened 2019-02-28
openstack: take defaults from environment

kshtsk:
This patch introduces TEUTH_ based environment variables
which can be used to avoid bulky commands while overriding
default values for some of the teuthology arguments.
Make --ceph argument optional for teuthology-suite,
and defaults to 'master'.

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -2,12 +2,13 @@ import docopt
 import sys
 
 import teuthology.suite
+from teuthology.suite import override_arg_defaults as defaults
 from teuthology.config import config
 
 doc = """
 usage: teuthology-suite --help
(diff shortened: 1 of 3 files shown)


scripts/suite.py
```diff
@@ -49,7 +51,7 @@
                               [default: basic]
   -t <branch>, --teuthology-branch <branch>
                               The teuthology branch to run against.
-                              [default: master]
+                              [default: {default_teuthology_branch}]
```
gregsfortytwo: Maybe instead of silently changing the default in the output, we should note that it's being overridden by an environment variable?

kshtsk: > Maybe instead of silently changing the default in the output, we should note that it's being overridden by an environment variable?

this looks like an overkill for this feature, because it will require more complicated code as for me, and what will be the benefit? from other side, I think it is a noise

kshtsk:
@gregsfortytwo ping

PR #1264, opened 2019-03-04
ansible.py: catch yaml.YAMLError when yaml.safe_load()

tchaikov:
yaml.safe_load() also throws yaml.scanner.ScannerError. so use
yaml.YAMLError to cover them all. see
https://pyyaml.org/wiki/PyYAMLDocumentation .


tchaikov:
see http://pulpito.ceph.com/dzafman-2019-03-01_09:55:16-rados-wip-zafman-testing-luminous2-distro-basic-smithi/3654188/

PR #1267, opened 2019-03-15
task/kernel: fix Depends: parsing for dependencies with "|"

dmick:
also make the "Checking kernel version...want" logmsg include quotes
for spotting problems more quickly

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -172,7 +172,7 @@ def need_to_install(ctx, role, version):
                     or a sha1.
     """
     ret = True
-    log.info('Checking kernel version of {role}, want {ver}...'.format(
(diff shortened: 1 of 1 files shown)


dmick:
This should fix the failures on Ubuntu 16.04 noticed today

dmick:
Tested with
```
interactive-on-error: true
nuke-on-error: false
kernel:
   sha1: distro
roles:
  - [client.0]

tasks:
  - interactive:
```

yuriw:
@dmick @liewegas suspect this PR caused http://tracker.ceph.com/issues/38802

dmick:
No, this PR should have *fixed* 38802, but you have to pick up the new code

PR #1269, opened 2019-03-21
openshift ansible installation

shylesh:
Signed-off-by: shylesh kumar <shylesh.mohan@gmail.com>

### DIFF
--- a/teuthology/task/openshift_ansible.py
+++ b/teuthology/task/openshift_ansible.py
@@ -1,0 +1,446 @@
+import json
+import os
+import re
+import logging
+import yaml
+import time
+import errno
+import socket
+from cStringIO import StringIO
+
+from . import Task
+from tempfile import NamedTemporaryFile
+from ..config import config as teuth_config
+from ..misc import get_scratch_devices
+from teuthology import contextutil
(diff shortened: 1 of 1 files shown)


teuthology/task/openshift_ansible.py
```diff
@@ -0,0 +1,446 @@
...
+import yaml
+import time
+import errno
+import socket
+from cStringIO import StringIO
+
+from . import Task
+from tempfile import NamedTemporaryFile
+from ..config import config as teuth_config
+from ..misc import get_scratch_devices
```
kshtsk: please use full path to the module, i.e.:
```
from teuthology.config
from teuthology.misc
```
```

File: teuthology/task/openshift_ansible.py [hunk 2]
```diff
@@ -0,0 +1,446 @@
...
+from teuthology import contextutil
+from teuthology.orchestra import run
+from teuthology.orchestra.daemon import DaemonGroup
+from teuthology.task.install import ship_utilities
+from teuthology import misc
+from teuthology import misc as teuthology
```
kshtsk: this is really strange, can you use 'misc' namespace?

PR #1272, opened 2019-03-28
add ssh_interface property

kshtsk:


teuthology/provision/cloud/openstack.py
```diff
@@ -74,6 +74,12 @@ def _get_driver_args(self):
             driver_args['ex_force_base_url'] = self._auth_token.endpoint
         return driver_args
 
+    @property
+    def ssh_interface(self):
+        if not hasattr(self, '_ssh_interface'):
+            self._ssh_interface = self.conf.get('ssh_interface', 'public_ips')
```
zmc: Can you explain how this works?

kshtsk: Basic explanation is that different OpenStack setups can assign IP addresses to different kind of networks. For example, OVH only provides one network which is assigned as public for our use, same as I believe for you, but in general OpenStack can be configured to have several networks, which are not exposed to the world, they are "magically" assigned by libcloud as "private_ips". I would also add an option to allow to select which ip should be used instead the first (zero index) in case there several networks configured, but not in this commit, since we do not use such complex configuration in teuthology yet. Depending which ssh_interface you point, either public_ips, or private_ips, corresponding set of addresses one get from libcloud when user polling it.

PR #1274, opened 2019-03-29
valgrind.supp: move the whitelist file to /qa dir of ceph.git

rzarzynski:
This change drops valgrind.supp whitelist file from teuthology repo and alters the install task to acquire it from `/qa` dir of ceph-qa-suite repository.
It's expected the commit will have [sibling in ceph.git moving the proper valgrind.supp there](https://github.com/ceph/ceph-ci/commit/c9195f48b9bdb18e57a1c616fd9c52c82ea2d816).

The rationale behind the entire change is duplication of these files between two repos, their inconsistencies and confusion.


teuthology/task/install/util.py
```diff
@@ -67,8 +67,9 @@ def ship_utilities(ctx, config):
     filenames = []
 
     log.info('Shipping valgrind.supp...')
+    assert 'suite_path' in ctx.config
     with file(
-        os.path.join(os.path.dirname(__file__), 'valgrind.supp'),
+        os.path.join(ctx.config['suite_path'], 'valgrind.supp'),
```
batrick: I think this would cause problems with teuthology runs on older versions of Ceph?

rzarzynski: Indeed, that could happen. However, the whitelist file is now ported to older Ceph branches as well.

liewegas: I think the only think we might add here is a try/catch or whatever so that we do not fail hard if the valgrind.supp file isn't present in the ceph repo for some reason

rzarzynski: Added try/catch + log message to not fail silently.
There are first greens in http://pulpito.front.sepia.ceph.com/rzarzynski-2019-04-24_14:40:46-rados-wip-sage-testing-2019-04-23-1419-distro-basic-smithi/.

PR #1275, opened 2019-04-03
update default OS versions

liewegas:
centos/rhel 7.4 -> 7.5
ubuntu 16.04 -> 18.04


djgalloway:
We have 7.6 images for RHEL and CentOS

kshtsk:
I don't know why unit tests are hardcoded to the default os version, but this patch requires to fix teuthology/test/test_get_distro_version.py

PR #1277, opened 2019-04-12
[DNM] Revert kernel default

kshtsk:
This is reverting 'add default' for kernel argument, which makes it is impossible to disable kernel task.

By default we do not want any kernel operations and since there no-functionality provided for disabling kernel task, it is better to revert this changes.

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -45,7 +45,6 @@ Standard arguments:
   -k <kernel>, --kernel <kernel>
                               The kernel branch to run against; if not
                               supplied, the installed kernel is unchanged
-                              [default: distro]
   -f <flavor>, --flavor <flavor>
                               The kernel flavor to run against: ('basic',
                               'gcov', 'notcmalloc')
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -43,7 +43,8 @@ Standard arguments:
                               either --ceph or --sha1, backtracking
                               up to <newest> commits [default: 0]
   -k <kernel>, --kernel <kernel>
-                              The kernel branch to run against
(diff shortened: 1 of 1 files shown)


kshtsk:
@vasukulkarni this breaks our downstream ci testing, since we do not use kernel operations.

vasukulkarni:
@kshtsk I think specifying default as distro actually updates the kernel distro for other non kernel tests which probably is needed, you can check the logs to ensure it doesn't break that.

kshtsk:
> @kshtsk I think specifying default as distro actually updates the kernel distro for other non kernel tests which probably is needed, you can check the logs to ensure it doesn't break that.

Yes this is what we do not want for sure, if you need 'distro' by default better add -k distro to your CI. The default tasks of teuthology-suite should be as light as possible.

vasukulkarni:
> > @kshtsk I think specifying default as distro actually updates the kernel distro for other non kernel tests which probably is needed, you can check the logs to ensure it doesn't break that.
> 
> Yes this is what we do not want for sure, if you need 'distro' by default better add -k distro to your CI. The default tasks of teuthology-suite should be as light as possible.

Essentially it boils down to who has to add the option in their run and who gets the default :-) , Right now folks who are invoking are assuming distro is default and their kernel is not going to be updated.

zmc:
Would it be sensible to add a `none` or similar value that `-k` can take, to allow skipping the task? It sounds like some coordination will be needed before that behavior can be the default, and this could buy some time to enable @kshtsk's workflow while not disrupting others'.

kshtsk:
This patch can be replaced with #1299

PR #1279, opened 2019-04-16
suse related fixes

kshtsk:


kshtsk:
@zmc could you please review

teuthology/misc.py
```diff
@@ -654,6 +654,7 @@ def delete_file(remote, path, sudo=False, force=False):
     remote.run(
         args=args,
         stdout=StringIO(),
+        check_status = check
```
zmc: nitpick, but please drop the spaces around the `=`

teuthology/misc.py
```diff
@@ -531,9 +531,8 @@ def write_file(remote, path, data):
     """
     remote.run(
         args=[
-            'python',
-            '-c',
-            'import shutil, sys; shutil.copyfileobj(sys.stdin, file(sys.argv[1], "wb"))',
+            'cat',
+            run.Raw('>'),
```
zmc: Has this been tested pretty well? It looks sane to me (I don't remember why the python approach was used in the first place), but we've been bit by silly things in the past

kshtsk: We used this test code "for ages" on opensuse/sle, I don't 'cat' is different in redhat and debian, but literally we didn't run it in the lab.

teuthology/misc.py
kshtsk: I can add cosmetic fix as separate commit.

kshtsk:
> Looks great overall. In addition to comments I've added, it looks like some of these commits could be squashed - could you take a look at that?

I would avoid squashing, since, it will complicate cherry-picking for me. But if you insist, I can do that.

kshtsk:
@zmc I think I addressed your change requests.

PR #1282, opened 2019-04-16
initial_tasks: make it possible to disable ansible

kshtsk:

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -226,9 +226,13 @@ def get_initial_tasks(lock, config, machine_type):
         init_tasks.extend([
             {'pcp': None},
             {'selinux': None},
-            {'ansible.cephlab': None},
-            {'clock': None}
         ])
+    if config.get('ceph_cm_ansible', True):
+        init_tasks.append({'ansible.cephlab': None})
+
+    # clock_sync_task: 'clock' or 'clock.check'
+    clock_sync_task = config.get('clock_sync_task', 'clock')
+    init_tasks.append({clock_sync_task: None})
 
     if 'redhat' in config:
         init_tasks.extend([


kshtsk:
> This will make it impossible to run the core suites on openSUSE machines in the Sepia lab. . .

@smithfarm come again

smithfarm:
Adding DNM for now (will change `ceph_cm: [STRING]` to `ceph_cm_ansible: [BOOLEAN]` which will hopefully be less confusing)

kshtsk:
@smithfarm could you please review once again

teuthology/run.py
```diff
@@ -226,9 +226,13 @@ def get_initial_tasks(lock, config, machine_type):
...
             {'selinux': None},
-            {'ansible.cephlab': None},
-            {'clock': None}
         ])
+    if config.get('ceph_cm_ansible', True):
+        init_tasks.append({'ansible.cephlab': None})
+
+    # sync_clock: 'clock' or 'clock.check'
+    sync_clock = config.get('sync_clock', 'clock')
+    init_tasks.append({sync_clock: None})
```
smithfarm: Could the config option be called `sync_clock_task` instead of `sync_clock`?

kshtsk: @smithfarm easy, check the next commit out

teuthology/run.py
```diff
@@ -226,9 +226,13 @@ def get_initial_tasks(lock, config, machine_type):
...
             {'selinux': None},
-            {'ansible.cephlab': None},
-            {'clock': None}
         ])
+    if config.get('ceph_cm_ansible', True):
+        init_tasks.append({'ansible.cephlab': None})
+
+    # clock_sync_task: 'clock' or 'clock.check'
+    clock_sync_task = config.get('clock_sync_task', 'clock')
+    init_tasks.append({clock_sync_task: None})
```
zmc: Is this an intentional change? I don't see it mentioned anywhere

kshtsk: this is a collateral change and does not alter the behavior

PR #1289, opened 2019-05-02
imports: make imports absolute

kshtsk:


kshtsk:
@zmc I have no idea, but I have "Merging is blocked" for this PR, nevertheless it has one approve already.

PR #1291, opened 2019-05-08
orchestra/run: add omit_sudo arg used by vstart_runner

batrick:
Introduced-by: ceph/ceph.git:3e0a1361f78452d79c283f70b5dc2b865d62cb61
Fixes: https://tracker.ceph.com/issues/39554

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -377,6 +377,8 @@ def run(
     label=None,
     timeout=None,
     cwd=None,
+    # omit_sudo is used by vstart_runner.py
+    omit_sudo=False
 ):
     """
     Run a command remotely.  If any of 'args' contains shell metacharacters


batrick:
I'll run this through testing with my next batch of PRs.

rishabh-d-dave:
> > @batrick
> > `-t`? I am attaching a log that contains the output of the steps I use to test this patch.
> > [test-pr-1291.log](https://github.com/ceph/teuthology/files/3157547/test-pr-1291.log)
> 
> That's a different problem. Is there a tracker for it?

Nope. Do you hit it too? This might be an error specific to my machine...

rishabh-d-dave:
> > @batrick The exact error due to which [the traceback I see](https://tracker.ceph.com/issues/39554#note-1) occurs is that command's arglist contains `'cd', '|/usr/lib/systemd'`. First, '|/usr/lib/systemd' is a **not** a path and second, even if we remove the vertical bar from the supposed path `/usr/lib/systemd` is only an executable.
> 
> That `cd` command looks wrong. Please create a tracker for it.

Done - http://tracker.ceph.com/issues/39651.

teuthology/orchestra/run.py
```diff
@@ -377,6 +377,7 @@ def run(
     label=None,
     timeout=None,
     cwd=None,
+    omit_sudo=False
```
gregsfortytwo: Please add a comment on why we need this since it's not actually used within the full framework right now.

batrick: done, thanks

PR #1292, opened 2019-05-28
[WIP] add rabbitmq-server for end2end pubsub tests

yuvalif:
- added pika python lib for doing the testing on the client side
- added rabbitmq-server installation. this part is currently *not working*. not sure if the root cause is related to the rabbitmq-server installation, since there are no log files. for example:
  - http://pulpito.front.sepia.ceph.com/yuvalif-2019-06-06_08:31:46-rgw:multisite-add_http_endpoint_tests-distro-basic-smithi/
 - manual installation using: yum, apt, and zypper was verified on: centos7, ubuntu16 and opensuse15.1

This PR is a needed for following test PR: https://github.com/ceph/ceph/pull/28263 to execute the amqp end2end tests

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -16,7 +16,7 @@ case "$(uname -s)" in
 Linux)
     case "$(lsb_release --id --short)" in
     Ubuntu|Debian)
-        for package in qemu-utils python-dev libssl-dev python-pip python-virtualenv libev-dev libvirt-dev libmysqlclient-dev libffi-dev libyaml-dev ; do
+        for package in qemu-utils python-dev libssl-dev python-pip python-virtualenv libev-dev libvirt-dev libmysqlclient-dev libffi-dev libyaml-dev rabbitmq-server; do
             if [ "$(dpkg --status -- $package|sed -n 's/^Status: //p')" != "install ok installed" ]; then
                 # add a space after old values
                 missing="${missing:+$missing }$package"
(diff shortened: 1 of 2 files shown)


yuvalif:
@djgalloway I'm in the process of adding notification to RGW (see https://github.com/ceph/ceph/pull/28971).
Unlike in the case of pubsub, the only tests that are possible with notification are "push" tests.
Allowing rabbitmq installation on the test nodes would enable these tests, as currently the feature is not tesable at all in teuthology.
Note, that once merged, using "notifications" would be recommended over fetching events from pubsub.
Would appreciate your help with this one!

kshtsk:
The first commit 784fec9 and the last one d19794d can be squashed if the latter is really wished.

kshtsk:
I'm just wondering why this test requires any dependencies on the teuthology server side?
I just don't know why I need all this on my teuthology server in my development box.
From other side, why not try and design a test so it could make use of some role and install there all the required stuff and run the tests and later grab results?

yuvalif:
> I'm just wondering why this test requires any dependencies on the teuthology server side?
> I just don't know why I need all this on my teuthology server in my development box.
> From other side, why not try and design a test so it could make use of some role and install there all the required stuff and run the tests and later grab results?

Agree. Even the HTTP tests that does not require any software are problematic when they require connectivity from RGW towards the servers executing the tests.
OTOH, this is much easier when running standalone test on you local dev machine (not in teuthology), as no other role is needed.
This would require different test code for local vs. teuthology execution of the test.
Will cancel this PR.

PR #1297, opened 2019-07-03
orchestra/run: add a method to write file

rishabh-d-dave:
The method writes given data to a temporary file and then uses
put_file() to move the file to required remote machine on the given
path.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -401,6 +401,13 @@ class Remote(object):
             self.remove(path)
         return local_path
 
+    def write_file(self, path, data, sudo=False):
+        """
+        Write a file on a remote machine.
+        """
+        args = 'sudo tee' if sudo else 'tee'
+        self.run('%s %s' % (args, path), stdin=data)
+
     def get_tar(self, path, to_path, sudo=False):
         """
         Tar a remote directory and copy it locally


rishabh-d-dave:
In [Ceph PR#28692](https://github.com/ceph/ceph/pull/28692/), I moved `sudo_write_file()` from test_volume_client.py to vstart_runner.py since it was generic enough and could be used by other testsuites too. However, a change like that can't pass through teuthology testing since a method with same name and functionality would be required in teuthology codebase too.

teuthology/orchestra/remote.py
```diff
@@ -401,6 +401,17 @@ def get_file(self, path, sudo=False, dest_dir='/tmp'):
...
 
+    def write_file(self, path, data):
+        """
+        Write a file on a remote machine.
+
+        :param path: Path on the remote being written to.
+        :param data: Data to be written.
+        """
+        tempfilename = tempfile.mkstemp()[1]
+        self.run(['tee', tempfilename], stdin=data)
```
kshtsk: Using tee process for dumping local files is a bad python coding style.

teuthology/orchestra/remote.py
```diff
@@ -401,6 +401,17 @@ def get_file(self, path, sudo=False, dest_dir='/tmp'):
...
+        self.put_file(tempfilename, path)
```
kshtsk: It is better to use _putfo_ method of sftp client if you wish to write file over sftp protocol.

teuthology/orchestra/remote.py
```diff
@@ -401,6 +401,17 @@ def get_file(self, path, sudo=False, dest_dir='/tmp'):
             self.remove(path)
         return local_path
 
+    def write_file(self, path, data):
```
kshtsk: I think it would be more usable have the same style as 'put_file' and 'get_file', i.e. give the name `put_data` or `get_data` correspondingly.

teuthology/orchestra/remote.py
kshtsk: If you are teeing the data on a remote why not to do it to the temporary destination, it is better, to do using `path` directly in the first place and by the way it is easy to add sudo support here as well, like:
```
. . .  (path, data, sudo=False):
  tee = 'sudo tee' if sudo else 'tee'
  self.run(args='%s %s' % (tee, path), stdin=data)
```

rishabh-d-dave: > If you are teeing the data on a remote why not to do it to the temporary destination, it is better, to do using path directly in the first place

`tempfilename` already holds a path for temporary destination.

teuthology/orchestra/remote.py
rishabh-d-dave: Sorry, you are unclear to me here. I can't see `get_data` and `put_data` in [`get_file`](https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/remote.py#L364) and [`put_file`](https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/remote.py#L354).

teuthology/orchestra/remote.py
rishabh-d-dave: Why not `put_file()`? I thought it did the same thing.

I can't find `putfo`/`putfoo` method. Are you talking about [`sftp.put()`](https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/remote.py#L310)?

teuthology/orchestra/remote.py
rishabh-d-dave: > and by the way it is easy to add sudo support here as well, like:

Done.

teuthology/orchestra/remote.py
kshtsk: @rishabh-d-dave I was talking about "put file object" method _sftp.putfo()_ in opposite to _sftp.put()_, for details look `pydoc paramiko.SFTPClient.putfo`.

teuthology/orchestra/remote.py
kshtsk: @rishabh-d-dave I meant create new methods `get_data` and `put_data`, which can accept string, byte or file object as source argument instead of local file path as in `write_file` method.

teuthology/orchestra/remote.py
```diff
@@ -401,6 +401,15 @@ def get_file(self, path, sudo=False, dest_dir='/tmp'):
             self.remove(path)
         return local_path
 
+    def write_file(self, path, data, sudo=False):
+        """
+        Write a file on a remote machine.
+        """
+        args = 'sudo tee' if sudo else 'tee'
+        tempfilename = tempfile.mkstemp()[1]
+        self.run('%s %s' % (args, tempfilename), stdin=data)
```
kshtsk: You do not need `tempfile` here, since it `self.run` is already executed on a remote host and you can give it a destination file path.

teuthology/orchestra/remote.py
kshtsk: @rishabh-d-dave and you're trying to use both method of delivering file to a remote host, you have to chose one of: running a tee on a remote host within paramiko.SSHClient.exec_command in the backend, or using paramiko.SFTPClient.putfo in a backend. And as I said it does not require any temporary file creation.

teuthology/orchestra/remote.py
rishabh-d-dave: Right. I've made the change, write_file does `sudo tee path`/`tee path` now.

teuthology/orchestra/remote.py
rishabh-d-dave: >  and you're trying to use both method of delivering file to a remote host, you have to chose one of: running a tee on a remote host within paramiko.SSHClient.exec_command in the backend, or using paramiko.SFTPClient.putfo in a backend. And as I said it does not require any temporary file creation.

Yep, I get it now.

teuthology/orchestra/remote.py
rishabh-d-dave: > @rishabh-d-dave I meant create new methods `get_data` and `put_data`, which can accept string, byte or file object as source argument instead of local file path as in `write_file` method. I hope this helps.

Ok, this helps but I am only partially clear. Is it necessary to do it on this PR? I only wanted to get `write_file` merged so that I can test [this PR](https://github.com/ceph/ceph/pull/28692/) on teuthology.

kshtsk: @rishabh-d-dave The common approach is to test the code before merge, so one should avoid merging before testing the code. My suggesting to give a name get/put to keep API style, if it is easy to do, why not to have it in a PR. Giving better names earlier, can minimize refactoring in future. Having 'write_file and put_file' makes it a confusion for coder which method to use, until discover whole method signature. Having pairs 'put_file and put_data' makes it clear at least for me, that in first case I'm going to use a filepath as a source, and in seconds case it is python data as a source. Hope it makes clear.

teuthology/orchestra/remote.py
```diff
@@ -401,6 +401,13 @@ def get_file(self, path, sudo=False, dest_dir='/tmp'):
             self.remove(path)
         return local_path
 
+    def write_file(self, path, data, sudo=False):
+        """
+        Write a file on a remote machine.
+        """
+        args = 'sudo tee' if sudo else 'tee'
+        self.run('%s %s' % (args, path), stdin=data)
```
kshtsk: This code will fail since args should be provided to run as 'args' parameter.
Here is an example of correct usage:
```
>>> from teuthology.orchestra.remote import Remote as r
>>> from StringIO import StringIO as s
>>> localhost=r('localhost')
>>> localhost.run(args="cat", stdin=s("Hello World!"))
2019-07-08 15:51:40,969.969 INFO:teuthology.orchestra.remote:Trying to reconnect to host
2019-07-08 15:51:41,159.159 INFO:teuthology.orchestra.run.localhost:Running:
2019-07-08 15:51:41,159.159 INFO:teuthology.orchestra.run.localhost:> true
2019-07-08 15:51:41,311.311 INFO:teuthology.orchestra.run.localhost:Running:
2019-07-08 15:51:41,311.311 INFO:teuthology.orchestra.run.localhost:> cat
2019-07-08 15:51:41,364.364 INFO:teuthology.orchestra.run.localhost.stdout:Hello World!
RemoteProcess(client=<paramiko.client.SSHClient object at 0x7fa841bffe50>, args='cat', check_status=True, hostname='localhost')
>>> 
```

kshtsk:
> This method already exists:
> 
> https://github.com/ceph/teuthology/blob/52d8af779b377ace2b4f4d7d7353259a13ec194c/teuthology/misc.py#L524
> 
> It's also used in several places in qa/ already.

@batrick That's brilliant :-D But I thought @rishabh-d-dave wanted 'remote' object context method.
Btw, if someone still wants it, the misc method could just reuse the code of 'remote'.

rishabh-d-dave:
> This method already exists:
> 
> https://github.com/ceph/teuthology/blob/52d8af779b377ace2b4f4d7d7353259a13ec194c/teuthology/misc.py#L524
> 
> It's also used in several places in qa/ already.

Closing this PR since this method works for me.

PR #1299, opened 2019-07-10
suite/run: add support 'none' for kernel branch parameter

kshtsk:
This is can be useful when one wants to disable kernel hash

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -119,8 +119,9 @@ class Run(object):
         # Put together a stanza specifying the kernel hash
         if self.args.kernel_branch == 'distro':
             kernel_hash = 'distro'
-        # Skip the stanza if no -k given
-        elif self.args.kernel_branch is None:
+        # Skip the stanza if '-k none' is given
+        elif self.args.kernel_branch is None or \
+             self.args.kernel_branch.lower() == 'none':
             kernel_hash = None
         else:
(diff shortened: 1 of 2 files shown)


teuthology/suite/run.py
```diff
@@ -120,7 +120,7 @@ def choose_kernel(self):
         if self.args.kernel_branch == 'distro':
             kernel_hash = 'distro'
         # Skip the stanza if no -k given
```
smithfarm: Either update this comment or drop it entirely?

teuthology/suite/run.py
```diff
@@ -119,8 +119,9 @@ def choose_kernel(self):
         # Put together a stanza specifying the kernel hash
         if self.args.kernel_branch == 'distro':
             kernel_hash = 'distro'
-        # Skip the stanza if no -k given
-        elif self.args.kernel_branch is None:
+        # Skip the stanza if '-k none' is given
+        elif self.args.kernel_branch is None or \
+             self.args.kernel_branch.lower() == 'none':
```
idryomov: Nit: I don't think converting to lower case is necessary -- we don't do that for `distro`.

kshtsk: Thanks for the comment. Sometimes people may wish to provide None as a in python, it is a special type constant, that's why I would prefer to match both 'None' and 'none'.

PR #1300, opened 2019-07-10
install: drop purge_data and related code

smithfarm:
purge_data was used to remove any Ceph daemon metadata remaining after the 
teardown phase of the install task runs. In particular, the ceph task was 
known to leave OSD metadata behind in the OSD directories under /var/lib/ceph.

While the install task causes the ceph-osd package to be installed and creates
directories under /var/lib/ceph, it is the ceph task that actually
populates these directories. Thus, their "unpopulation" should be
implemented in the ceph task, and not here in the install task.

(If the directories were empty after the ceph task teardown phase, they would be
removed when the install task removes the ceph-osd package from the system. But 
the package management tool refuses to remove them when they contain files not 
owned by a package.)

Not running purge_data during the teardown phase of the install task makes it
easier to implement alternative deployment tasks (i.e. tests that deploy Ceph
clusters, but do not call ceph.py).

In the final analysis, however, this code is no longer needed now that all test
nodes (virtual and bare metal) are reimaged between runs.

This commit does not move purge_data into the ceph task because the author felt
that the ceph task should take more care to clean up after itself. For example,
nowadays there are ceph CLI commands for removing OSDs.

Aside: this commit also drops two lines of dead code that flake8 was 
complaining about.

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -8,7 +8,6 @@ import yaml
 from teuthology import misc as teuthology
 from teuthology import contextutil, packaging
 from teuthology.parallel import parallel
-from teuthology.orchestra import run
 from teuthology.task import ansible
 
 from distutils.version import LooseVersion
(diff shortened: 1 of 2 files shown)


kshtsk:
@smithfarm flake8 failed:
```
teuthology/task/install/__init__.py:11:1: F401 'teuthology.orchestra.run' imported but unused
teuthology/task/install/rpm.py:358:5: F841 local variable 'proj' is assigned to but never used
```
These are required to be resolved.

smithfarm:
@kshtsk The second one I understand, but I don't see how my patch caused the first one ("teuthology/task/install/__init__.py:11:1: F401 'teuthology.orchestra.run' imported but unused").

kshtsk:
> @kshtsk The second one I understand, but I don't see how my patch caused the first one ("teuthology/task/install/**init**.py:11:1: F401 'teuthology.orchestra.run' imported but unused").

@smithfarm JFYI you've remove `purge_data` which has been using _run.Raw_.

gregsfortytwo:
Hmm. I'd been hoping at some point to make the test development workflow where you
1) lock a set of nodes
2) run your test directly (not through the queue) against your nodes
3) make changes
4) run your test directly against your nodes again

would behave. (It doesn't work right now because you can't nuke nodes without reimaging any more.) This moves us away from that. :/ Not necessarily a blocker, but it makes me sad.

kshtsk:
So, are we good to merge it?

kshtsk:
> As @gregsfortytwo said, for manual workflow(testing scripts in local environment) it might still be useful(if deployed via ceph.py) , I am curious you are removing the functions but its called somewhere ?

I can't find the usage from anywhere in ceph/ceph repo and in ceph/teuthology rather than that.

kshtsk:
I'm going to squash and merge this PR if you guys don't mind.

smithfarm:
@gregsfortytwo @vasukulkarni I would argue there are better ways of purging (i.e. destroying) OSDs nowadays than this. Also, since the `install` task does not create OSDs, it should not be in the business of purging/destroying them.

In other words, this code is not only misplaced - it's also outdated. So I don't see any problem with removing it from where it clearly should not be, without reimplementing it in the `ceph` task (since "reimplementing" is not just a matter of copy/paste).

PR #1303, opened 2019-08-22
setup.py: move myqslclient/coverage to optional deps

epuertat:
Use [extras_require feature](https://github.com/rhcs-dashboard/teuthology/archive/wip-minimal-reqs.zip#egg=teuthology\[coverage\]) from setuptools:

Coverage is now an extra, and can be automatically installed by:
- `pip install teuthology[coverage]`
- `pip install git+https://github.com/ceph/teuthology.git#egg=teuthology\[coverage\]`
- `pip install git+https://github.com/ceph/teuthology/archive/wip-minimal-reqs.zip#egg=teuthology\[coverage\]`

In requirements.txt the proper syntax is:
`teuthology[coverage]==<version>`

Signed-off-by: Ernesto Puerta <epuertat@redhat.com>


kshtsk:
Why this is required?
> BTW, motivation for this change comes from ceph/ceph#29504 (comment), where we have to install teuthology deps to run dashboard tests.

kshtsk:
retest this please

epuertat:
> Why this is required?
> 
> > BTW, motivation for this change comes from [ceph/ceph#29504](https://github.com/ceph/ceph/pull/29504) (comment), where we have to install teuthology deps to run dashboard tests.

I kind of assume that by quoting things that way you already got the answer you were looking for. But just to shed more light into this: that `mysqlclient` dependency, which is not strictly required to run teuthology (only for coverage measurement), cannot be fully satisfied by pip. It needs Python & MySQL devel packages installed from the distro package manager.

tchaikov:
@kshtsk i think this makes sense. as mysqlclient is not a necessarily a runtime or even build-time dependency in all cases. so probably we can just make it optional.

kshtsk:
> @kshtsk i think this makes sense. as mysqlclient is not a necessarily a runtime or even build-time dependency in all cases. so probably we can just make it optional.

I like this idea.

@gregsfortytwo  @zmc do you know the cases we where teuthology-coverage is used?

epuertat:
> Since coverage may require there should bootstrap script fixed as well to use extra.

But `bootstrap` script is using `requirements.txt` (which still includes `mysqlclient`) for the dependencies, and as I haven't touched that, it should work fine.

kshtsk:
> > Since coverage may require there should bootstrap script fixed as well to use extra.
> 
> But `bootstrap` script is using `requirements.txt` (which still includes `mysqlclient`) for the dependencies, and as I haven't touched that, it should work fine.

bootstrap uses `./virtualenv/bin/python setup.py develop`

epuertat:
> > > Since coverage may require there should bootstrap script fixed as well to use extra.
> > 
> > 
> > But `bootstrap` script is using `requirements.txt` (which still includes `mysqlclient`) for the dependencies, and as I haven't touched that, it should work fine.
> 
> bootstrap uses `./virtualenv/bin/python setup.py develop`

Which does nothing, regarding dependency install. And a few lines before:
https://github.com/ceph/teuthology/blob/42ebe0d580e9669c38aaeabf768cc1fef31de16f/bootstrap#L139-L140

And inside that:
https://github.com/ceph/teuthology/blob/42ebe0d580e9669c38aaeabf768cc1fef31de16f/requirements.txt#L58

kshtsk:
> > > > Since coverage may require there should bootstrap script fixed as well to use extra.
> > > 
> > > 
> > > But `bootstrap` script is using `requirements.txt` (which still includes `mysqlclient`) for the dependencies, and as I haven't touched that, it should work fine.
> > 
> > 
> > bootstrap uses `./virtualenv/bin/python setup.py develop`
> 
> Which does nothing, regarding dependency install. And a few lines before:
> 
> https://github.com/ceph/teuthology/blob/42ebe0d580e9669c38aaeabf768cc1fef31de16f/bootstrap#L139-L140
> 
> And inside that:
> 
> https://github.com/ceph/teuthology/blob/42ebe0d580e9669c38aaeabf768cc1fef31de16f/requirements.txt#L58

Actually it does, look carefully, it creates binary scripts.

epuertat:
I tried to [refactor](https://github.com/ceph/teuthology/compare/5f33567c3ce6fddbcc1acfba627187236c9bcc0c..6abb043e68af3160e0b60c38697343eb5d8a8b74) `bootstrap/tox.ini` to use `pip install -e .\[coverage]`, but it required more and more changes and caused complex dependency issues (as this requires `sitepackages` too).

Reverted to the original proposal.

kshtsk:
going to run deployment test

epuertat:
Thanks, @kshtsk !

PR #1309, opened 2019-09-18
Minor bugfixes for teuthology

jtlayton:
Just a couple of small bugfixes for running teuthology commands on a modern setup. Avoid symlink loops in teuthology-describe-tests and force the usage of python2 in the bootstrap script.

### DIFF
--- a/teuthology/describe_tests.py
+++ b/teuthology/describe_tests.py
@@ -243,6 +243,9 @@ def tree_with_info(cur_dir, fields, include_facet, prefix, rows,
     has_yamls = any([x.endswith('.yaml') for x in files])
     facet = os.path.basename(cur_dir) if has_yamls else ''
     for i, f in enumerate(files):
+        # skip any hidden files
+        if f.startswith('.'):
+            continue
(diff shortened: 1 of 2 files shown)


teuthology/describe_tests.py
```diff
@@ -243,6 +243,8 @@ def tree_with_info(cur_dir, fields, include_facet, prefix, rows,
     has_yamls = any([x.endswith('.yaml') for x in files])
     facet = os.path.basename(cur_dir) if has_yamls else ''
     for i, f in enumerate(files):
+        if f == ".qa":
```
kshtsk: This is not a loop detection code, though it makes more sense to use `f.startswith('.')` to ignore "hidden" files, which is supposed to be a case if I understand correctly the idea of having '.qa' links.

jtlayton: Good idea. Done.

jtlayton:
Clean up patch description

smithfarm:
@susebot run deploy

PR #1310, opened 2019-09-27
openstack: encode instance name with the full IP

kshtsk:
Using only the last two numbers is problematic when mixing multiple
OpenStack clusters.

Conflicts:
	teuthology/provision.py

### DIFF
--- a/teuthology/describe_tests.py
+++ b/teuthology/describe_tests.py
@@ -198,7 +198,7 @@ def extract_info(file_name, fields):
         return empty_result
 
     with open(file_name, 'r') as f:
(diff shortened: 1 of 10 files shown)


kshtsk:
@zmc since Loic removed his account, trying to give this patch another try replacing #798

kshtsk:
@zmc I don't believe you guys are still using this DNS feature for your setups since it is only related to legacy openstack framework which only used in SUSE :-)

kshtsk:
@susebot run deploy

kshtsk:
@gregsfortytwo thanks

PR #1311, opened 2019-09-27
worker: Archive job config for further reference

kshtsk:
When worker runs teuthology it provides job config as temporary file
which will be deleted after job finished.
This patch makes it possible to save the config in the job log archive
directory and is useful when debugging the job and for the reference.

Signed-off-by: Kyr <kyrylo.shatskyy@suse.com>


kshtsk:
@susebot run deploy

kshtsk:
It appears that this file is stored by teuthology.run to `orig.config.yaml` file in job archive already. Closing this ticket then.

PR #1313, opened 2019-10-02
python3: use open() function instead file() class

kshtsk:
Since python3 does not have file() we should stop using it.


teuthology/suite/test/test_run_.py
```diff
@@ -232,9 +233,10 @@ def test_successful_schedule(
         m_build_matrix.return_value = build_matrix_output
         frag1_read_output = 'field1: val1'
         frag2_read_output = 'field2: val2'
-        m_file.side_effect = [
+        m_open.side_effect = [
             StringIO(frag1_read_output),
             StringIO(frag2_read_output),
+            contextlib.closing(StringIO())
```
gregsfortytwo: What's this do?

kshtsk: in short it makes the test pass

gregsfortytwo: https://docs.python.org/3/library/contextlib.html#contextlib.closing

so it's for automatically closing resources opened with the ContextManager, which makes sense, but I don't get how passing it a StringIO() works. I guess it closes all opened StringIO objects somehow? This level of python-foo is beyond me and google isn't helping. :(

kshtsk: The side_effect array is used to replace output of each call, in this particular case it is open() method. The call of this method is found three times in this particular case:
Two of them are called via this route:
```
test_successful_schedule -> schedule_suite -> collect_jobs
```
where two fragments are read in this code:
```
    raw_yaml = '\n'.join([open(a, 'r').read() for a in fragment_paths])
```
And the third one is via this route:
```
test_successful_schedule -> schedule_suite
```
in the block:
```
        with open(base_yaml_path, 'w+b') as base_yaml:
            base_yaml.write(str(self.base_config))
```
We've been using a `file()` class before, which is not a context type and has no `__exit__` method. But after we switched to `open()` function, which return an context object with `__exit__` method, then on exit from `with ...` block it is invoked (for the file() python interpreter just didn't call it. This why we need to add extra side_effect with value of an object with `__exit__` method additionally declared, like based on contextlib.closing() class. The open() function also returns an object with `close()` method which is used as well, this is why we need to use something like StringIO() class, etc.

I hope this helps.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/config.py
```diff
@@ -37,7 +37,7 @@ def load(self, conf=None):
                 self._conf = yaml.safe_load(conf)
             return
         if os.path.exists(self.yaml_path):
-            self._conf = yaml.safe_load(file(self.yaml_path))
+            self._conf = yaml.safe_load(open(self.yaml_path))
```
tchaikov: probably it's not in the scope of this change, could you add a commit to
```py
with open(self.yaml_path) as f:
  self._conf = yaml.safe_load(f)
```
?

kshtsk: done

kshtsk:
@susebot run deploy

PR #1315, opened 2019-10-07
use python3 compatible print

kshtsk:


kshtsk:
@susebot run deploy

teuthology/openstack/setup-openstack.sh
```diff
@@ -347,7 +347,7 @@ function setup_dnsmasq() {
 
 function subnet_names_and_ips() {
     local subnet=$1
-    python -c 'import netaddr; print "\n".join([str(i) for i in netaddr.IPNetwork("'$subnet'")])' |
+    python -c 'import netaddr; print("\n".join([str(i) for i in netaddr.IPNetwork("'$subnet'")]))' |
```
tchaikov: can we assume python version is greater or equal to 2.6 here?

teuthology/describe_tests.py
```diff
@@ -207,9 +207,9 @@ def extract_info(file_name, fields):
     if not (isinstance(meta, list) and
             len(meta) == 1 and
             isinstance(meta[0], dict)):
-        print 'Error in meta format in', file_name
-        print 'Meta must be a list containing exactly one dict.'
-        print 'Meta is:', meta
+        print('Error in meta format in %s' % file_name)
```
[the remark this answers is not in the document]
kshtsk: > i don't really understand why we need to `from __future__ import print_function` somewhere, but we can dispense with it in other places? so we have different assumptions on python versions in these two cases?

future print requires for, for `file=std.err` parameter and `end=''`. In other cases it is not required rather than you want to have same output for `print a, b` and `print(a,b)` which is different.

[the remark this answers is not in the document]
kshtsk: > nit, you could also put

I could, but I prefer this.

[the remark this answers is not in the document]
kshtsk:
```
> python2 -c "print 'a'"
a
> python2 -c "print('a')"
a
> python3 -c "print('a')"
a
> python2 -c "print 'a', 'b'"
a b
> python2 -c "print('a', 'b')"
('a', 'b')
> python2 -c "from __future__ import print_function; print('a', 'b')"
a b
> python3 -c "print('a', 'b')"
a b
```

teuthology/openstack/setup-openstack.sh
tchaikov: > Python 2.6 is not supported anymore

thanks. i didn't realize this. just to make sure we are on the same page. where did you get this message? by checking the "classifiers" section in `setup.py` or by checking `tox.ini` ?

> Why are you asking?

because python2.5 does not support `print()` at all.

kshtsk: do we have systems with python2.5 where teuthology is used?

kshtsk: Regarding python2.5 
https://www.python.org/downloads/release/python-255/
> Release Date: Jan. 31, 2010
> This is a source-only release that only includes security fixes. The last full bug-fix release of Python 2.5 was Python 2.5.4. User are encouraged to upgrade to the latest release of Python 2.7 (which is 2.7.2 at this point).

It was almost 10 years ago. :-D

kshtsk: Regarding python2.6
https://www.python.org/downloads/release/python-269/
> Python 2.6.9
> Release Date: Oct. 29, 2013
> With the 2.6.9 release, and five years after its first release, the Python 2.6 series is now officially retired. All official maintenance for Python 2.6, including security patches, has ended. For ongoing maintenance releases, please see the Python 2.7 series. The NEWS file lists every change in each alpha, beta, release candidate, and final release of Python 2.6.

They said 'retired' 6 years ago.

tchaikov: > do we have systems with python2.5 where teuthology is used?

that's my question.

smithfarm: If there are such systems (which I doubt), the question is: do we have any reason to support them?

kshtsk:
```kyr@teuthology:~$ python2 --version
Python 2.7.12
```
This is what we have on teuthology server in sepia.

PR #1316, opened 2019-10-08
Wip 42211 remove os before provisioning

ryg-:
https://tracker.ceph.com/issues/42211
Remove os_type and os_version before provisioning for reflecting actual node status.
It helps to quickly understand what is node status, especially for manual work or provisioning
failed


ryg-:
made decision to merge with 1312

PR #1317, opened 2019-10-08
suite/matrix: use isinstance for python3 compatibility

kshtsk:


kshtsk:
@susebot run deploy

PR #1318, opened 2019-10-08
misc: use remote.sh instead of remote.run

kshtsk:
To ease porting and compatibility in order to avoid
usage of cStringIO.StringIO it is better to use
remote.sh method which supposed to return captured
value as a string.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -511,11 +511,7 @@ def create_simple_monmap(ctx, remote, conf, path=None,
         path
     ])
 
-    r = remote.run(
-        args=args,
-        stdout=StringIO()
-    )
-    monmap_output = r.stdout.getvalue()
+    monmap_output = remote.sh(args)
     fsid = re.search("generated fsid (.+)$",
                      monmap_output, re.MULTILINE).group(1)
     return fsid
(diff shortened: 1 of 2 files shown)


kshtsk:
@susebot run deploy

gregsfortytwo:
The changes look equivalent to me but I don't think I understand your comment about avoiding cstringIO.StringIO. (I am very not up on the py3 transition requirements.)
Why does this help?

kshtsk:
> The changes look equivalent to me but I don't think I understand your comment about avoiding cstringIO.StringIO. (I am very not up on the py3 transition requirements.)
> Why does this help?

The cStringIO module is absent in python3 as well as StringIO, more over, the io.StringIO is not the same in python3 and requires much more effort in wrapping things up with encoding/deconing method, and since we already have a good method `remote.sh` which does not refer any of it and just return a string at the end of refactoring it will be enough to make this only method to be compatible with both versions. As side effect the code looks much simpler than before.

kshtsk:
> @kshtsk i assume that your will have a PR fixing `Remote.sh()`, am i correct?

Yes... that is a plan, but at first need to resolve the easiest part

teuthology/misc.py
```diff
@@ -630,10 +619,7 @@ def move_file(remote, from_path, to_path, sudo=False, preserve_perms=True):
             perms,
             to_path,
         ])
-        proc = remote.run(
-            args=args,
-            stdout=StringIO(),
-        )
+        remote.sh(args)
```
smithfarm: Did you mean `proc = remote.sh(args)` here?

smithfarm: ah, I see, `proc` is not used for anything

PR #1319, opened 2019-10-08
misc: use BytesIO instead StringIO for python3

kshtsk:
For python3 compatibility it is preferable to use BytesIO
over StringIO.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -21,6 +21,8 @@ import json
 import re
 import pprint
 
+from io import BytesIO
+
 from netaddr.strategy.ipv4 import valid_str as _is_ipv4
 from netaddr.strategy.ipv6 import valid_str as _is_ipv6
 from teuthology import safepath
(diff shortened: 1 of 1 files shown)


kshtsk:
@susebot run deploy

teuthology/misc.py
```diff
@@ -21,6 +21,8 @@
 import re
 import pprint
 
+from io import BytesIO
```
tchaikov: > For python3 compatibility it is preferable to use BytesIO over StringIO.

why? i think as their names suggest, `StringIO` is for reading/writing strings, while `BytesIO` for bytes. why is this difference related to python3?

also, per https://configobj.readthedocs.io/en/latest/configobj.html#configobj-specifications

> A StringIO instance or file object, or any object with a read method. The filename attribute of your ConfigObj will be None 

so i think `StringIO` is better?

probably we could just pass `path` to it?

kshtsk: @tchaikov in this particular case there was used cStringIO.StringIO, which does not exists in python3, as well as StringIO.StringIO. The only StringIO in py3 is io.StringIO which is for Unicode, the config file we are using is in ascii/utf-8 encoding, which BytesIO more appropriate. So what is the best?

kshtsk: > probably we could just pass `path` to it?

you mean just pass the filename to ConfigObj? this looks to me as a good idea and I thought about it, but there is a `.format` invocation. Though I can not find a `testdir` in 'ceph.conf.template' I thought someone may wish it in the future.

kshtsk: @tchaikov Do you want me just do like that:
```
conf = ConfigObj.ConfigObj(path, file_error=True)
```
Instead of handling all that streaming soup?

kshtsk: @tchaikov like this? #1323

kshtsk: One more else why I used `io.BytesIO` instead of `cStringIO.StringIO`, because
https://docs.python.org/2/library/stringio.html (7.6. cStringIO — Faster version of StringIO):
```
 cStringIO.StringIO([s])
     Return a StringIO-like stream for reading or writing.
     . . .
     Unlike the StringIO module, this module is not able to accept Unicode 
     strings that cannot be encoded as plain ASCII strings.
```
The `io.BytesIO` is present in both py2 and py3.  Any thoughts?

tchaikov: @kshtsk yeah, i'd prefer #1323 than this change.

kshtsk:
Closing in favor of #1323

PR #838, opened 2016-04-05
config: add option for ceph cm ansible git url

roidayan:
This is to be able to supply a custom location without changing
the ceph git base url.

Signed-off-by: Roi Dayan roid@mellanox.com

PR #1321, opened 2019-10-09
config: add option for ceph cm ansible git url

kshtsk:
This is to be able to supply a custom location without changing
the ceph git base url.

Signed-off-by: Roi Dayan <roid@mellanox.com>

Conflicts:
	teuthology/config.py
	teuthology/suite.py
	teuthology/test/test_suite.py


kshtsk:
@zmc @gregsfortytwo I've resurrected the #838. Do you think it will be usable to have merged in master?

PR #1308, opened 2019-09-07
teuthology/ceph.conf.template: do not warn on TOO_FEW_OSDS

smithfarm:
NOTE: merge after the following two PRs are merged:

- [x] mimic https://github.com/ceph/ceph/pull/30180 (merged)
- [x] luminous https://github.com/ceph/ceph/pull/30298 (merged)

---

This is a manual backport of ceph/ceph.git SHA1 0483c1c3e7ffdfa6a6f65c5ef000c45d2f096428
that is needed for mimic.

RATIONALE: The mimic release uses this file in teuthology,
whereas as of nautilus 1ab352dd3105d320bd7ad2c0b37e5de750879dd5 the file
qa/tasks/ceph.conf.template is being used.

See also @neha-ojha 's comment https://github.com/ceph/ceph/pull/30180#issuecomment-528607730

PR #1322, opened 2019-10-09
add note that ceph.conf.template should no longer be modified

batrick:

### DIFF
--- a/teuthology/ceph.conf.template
+++ b/teuthology/ceph.conf.template
@@ -1,3 +1,13 @@
+# XXX
+#
+#   DO NOT MODIFY THIS FILE
+#
+# This file is a legacy ceph.conf template used only when testing older
+# releases of Ceph (pre-Nautilus). The new template exists in ceph.git at
+# qa/tasks/ceph.conf.template
+#
+# XXX
+
 [global]
 	chdir = ""
 	pid file = /var/run/ceph/$cluster-$name.pid


teuthology/ceph.conf.template
```diff
@@ -1,3 +1,13 @@
+# XXX
+#
+#   DO NOT MODIFY THIS FILE
+#
+# This file is a legacy ceph.conf template used only when testing older
+# releases of Ceph (pre-Nautilus). The new template exists in ceph.git at
+# qa/tasks/ceph.conf.template
+#
+# XXX
```
kshtsk: what is that XXX?

batrick: Many text editors show "XXX" in red highlighting to draw attention.

kshtsk: oh... good to know thanks

kshtsk:
wonder, does this mean that #1308 should be closed?

batrick:
> wonder, does this mean that #1308 should be closed?

It should be redone against the ceph.git template.

kshtsk:
@smithfarm I guess for older releases we need to back port:
https://github.com/ceph/ceph/commit/de9b77cd38970106f41376460e0842305594f5ec
and https://github.com/ceph/ceph/commit/1ab352d

Otherwise it will still need to read this file.

smithfarm:
@kshtsk Good idea. See https://tracker.ceph.com/issues/42261

PR #1323, opened 2019-10-09
misc: drop testdir replacement in skeleton_config

kshtsk:
The str.format is not used anymore for ceph.conf.template


kshtsk:
@susebot run deploy

PR #1329, opened 2019-10-11
run: drop StringIO usage for py3 compatibility

kshtsk:

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -1,7 +1,5 @@
 import os
 import yaml
-import StringIO
-import contextlib
 import sys
 import logging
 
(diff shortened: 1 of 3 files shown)


kshtsk:
@susebot run deploy

toabctl:
@kshtsk there is `StringIO` in python3. Maybe it would be simpler to just use https://six.readthedocs.io/#six.StringIO to keep the change minimal so this can get merged quicker?

kshtsk:
> @kshtsk there is `StringIO` in python3. Maybe it would be simpler to just use https://six.readthedocs.io/#six.StringIO to keep the change minimal so this can get merged quicker?

no, because for most cases there is cStringIO used in teuthology which different to StringIO

vasukulkarni:
@rakeshgm @yuriw I think we could start using this branches to run partial rados suite  + smoke and get them merged soon

kshtsk:
@vasukulkarni do we have some small test case to test this PR with 'email-on-error' feature?

vasukulkarni:
> @vasukulkarni do we have some small test case to test this PR with 'email-on-error' feature?

I dont think so , @zmc do you know of any? I think email-on-error might be too much considering how many tests one suite has.

vasukulkarni:
Here we could just run one subset of rados to ensure its good for merge and go ahead with that.

PR #1331, opened 2019-10-11
python3: use compatible urllib packages

kshtsk:
Use py2/py3 compatible functions url* packages.

Python3
    from urllib.parse import parse_qs, urljoin, urlparse, urlencode
    from urllib.request import urlopen, Request
    from urllib.error import HTTPError
Python2
    from urlparse import parse_qs, urljoin, urlparse
    from urllib import urlencode
    from urllib2 import urlopen, Request, HTTPError

### DIFF
--- a/teuthology/lock/query.py
+++ b/teuthology/lock/query.py
@@ -1,6 +1,10 @@
 import logging
 import os
-import urllib
+
+try:
+    from urllib.parse import urlencode
+except ImportError:
+    from urllib import urlencode
 
 import requests
 
@@ -58,7 +62,7 @@ def list_locks(keyed_by_name=False, **kwargs):
     if kwargs:
         if 'machine_type' in kwargs:
             kwargs['machine_type'] = kwargs['machine_type'].replace(',','|')
-        uri += '?' + urllib.urlencode(kwargs)
+        uri += '?' + urlencode(kwargs)
     try:
         response = requests.get(uri)
     except requests.ConnectionError:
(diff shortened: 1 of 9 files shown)


kshtsk:
@susebot run deploy

teuthology/misc.py
```diff
@@ -13,13 +13,19 @@
...
 import re
 import pprint
 
+try:
+    from urllib.parse import urljoin
+    from urllib.request import urlopen
+    from urllib.error import HTTPError
+except ImportError:
+    from urlparse import urljoin
+    from urllib2 import urlopen,  HTTPError
```
smithfarm: `s/urlopen,  HTTPError/urlopen, HTTPError/` to remove excess white space

teuthology/lock/query.py
```diff
@@ -1,6 +1,10 @@
 import logging
 import os
-import urllib
+
+try:
```
toabctl: We could use `six` which is already in requirements.txt to have less code. And later it's easier to grep for it to remove py2 support. Or is six not wanted?

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/lock/query.py
smithfarm: I would prefer to avoid `six`, if only because it's yet another thing I would have to learn how to use.

PR #1340, opened 2019-10-23
docs: Mention step to setup nodes on OpenStack

toabctl:
The --setup step was missing which lead to cryptic error messages when
running the following step (calling the dummy test suite).

### DIFF
--- a/docs/openstack_backend.rst
+++ b/docs/openstack_backend.rst
@@ -31,6 +31,10 @@ Setup
     $ cd teuthology ; ./bootstrap install
     $ source virtualenv/bin/activate
 
+* Setup the teuthology node::
+
+    $ teuthology-openstack --key-filename myself.pem --key-name myself --setup
+
 Get OpenStack credentials and test it
 -------------------------------------
 
--- a/docs/openstack_backend.rst
+++ b/docs/openstack_backend.rst
@@ -62,11 +62,6 @@ Get OpenStack credentials and test it
 Usage
 -----
 
-* Create a passwordless ssh public key::
-
-    $ openstack keypair create myself > myself.pem
-    $ chmod 600 myself.pem
-
 * Run the dummy suite. It does nothing useful but shows all works as
   expected. Note that the first time it is run, it can take a long
   time (from a few minutes to half an hour or so) because it downloads


docs/openstack_backend.rst
```diff
@@ -63,6 +63,10 @@ Usage
     $ openstack keypair create myself > myself.pem
     $ chmod 600 myself.pem
```
smithfarm: This bullet point duplicates the last bullet point of the previous section (`Setup`). @toabctl Could you drop it?

docs/openstack_backend.rst
```diff
@@ -63,6 +63,10 @@ Usage
...
 
+* Setup the teuthology node::
+
+    $ teuthology-openstack --key-filename myself.pem --key-name myself --setup
+
```
smithfarm: This new bullet point actually belongs in the previous section (`Setup`).

docs/openstack_backend.rst
toabctl: Done.

docs/openstack_backend.rst
toabctl: Done

docs/openstack_backend.rst
kshtsk: That section is better to replace with instructions of key creation using `ssh-keygen -m PEM` instead of `openstack keypair create` since newer versions of openstack can create keys with newer format, which is not supported by paramiko at the moment.

toabctl: Which format is not supported?

kshtsk: @toabctl the one having "BEGIN OPENSSH PRIVATE KEY"

kshtsk: see details here https://wiki.sepia.ceph.com/doku.php?id=testnodeaccess#sshexceptionerror_reading_ssh_protocol_banner

toabctl: Would it be possible to get this PR merged and fix stuff that is unrelated to this PR later? Imo this PR is an improvement over the current state.

kshtsk:
ok to test

PR #1341, opened 2019-11-01
openstack: Make used private keys explicit

toabctl:
Instead of appending possible private keyfile paths internally, make
that the default for the --key-filename command line argument so users
see it when looking at the help.

### DIFF
--- a/scripts/openstack.py
+++ b/scripts/openstack.py
@@ -15,7 +15,12 @@ def get_key_parser():
     )
     parser.add_argument(
         '--key-filename',
-        help='path to the ssh private key',
+        help='path to the ssh private key. Default: %(default)s',
+        default=[
+            os.environ['HOME'] + '/.ssh/id_rsa',
+            os.environ['HOME'] + '/.ssh/id_dsa',
+            os.environ['HOME'] + '/.ssh/id_ecdsa'
+        ]
     )
     return parser
 
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -692,11 +692,10 @@ class TeuthologyOpenStack(OpenStack):
         self.setup_logs()
         set_config_attr(self.args)
         log.debug('Teuthology config: %s' % self.config.openstack)
-        for keyfile in [self.args.key_filename,
-                        os.environ['HOME'] + '/.ssh/id_rsa',
-                        os.environ['HOME'] + '/.ssh/id_dsa',
-                        os.environ['HOME'] + '/.ssh/id_ecdsa']:
-            if (keyfile and os.path.isfile(keyfile)):
+        key_filenames = (lambda x: x if isinstance(x, list) else [x]) \
+            (self.args.key_filename)
+        for keyfile in key_filenames:
+            if os.path.isfile(keyfile):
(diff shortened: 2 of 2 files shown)


teuthology/openstack/__init__.py
```diff
@@ -559,7 +559,7 @@ def cloud_init_wait(self, instance):
             'retry': False,
         }
         if self.key_filename:
-            log.debug("using key " + self.key_filename)
+            log.debug("using key " + self.key_fielname)
```
ktdreyer: looks like a typo here

toabctl: Done

kshtsk:
ok to test

scripts/openstack.py
```diff
@@ -15,7 +15,13 @@ def get_key_parser():
     )
     parser.add_argument(
         '--key-filename',
-        help='path to the ssh private key',
+        help='path to the ssh private key. Default: %(default)s',
+        action='append',
```
kshtsk: I'm not sure the feature will work as expected? Do you think `action='append'` will put given parameter to the top of default?
How is your `--help` output look like for this command?

toabctl: you are right - `append` does it a) at the end b) still silently tries the default keys. fixed now. The output of help is:
```
  --key-filename KEY_FILENAME
                        path to the ssh private key. Default:
                        ['/home/xxx/.ssh/id_rsa', '/home/xxx/.ssh/id_dsa',
                        '/home/xxx/.ssh/id_ecdsa']
```

toabctl: Fixed now. The behavior is now:
- if no `--key-filename` is given, the default keys will be used
- If `--key-filename` is given, only that key is used

teuthology/openstack/__init__.py
```diff
@@ -692,11 +692,10 @@ def main(self):
         self.setup_logs()
         set_config_attr(self.args)
         log.debug('Teuthology config: %s' % self.config.openstack)
-        for keyfile in [self.args.key_filename,
-                        os.environ['HOME'] + '/.ssh/id_rsa',
-                        os.environ['HOME'] + '/.ssh/id_dsa',
-                        os.environ['HOME'] + '/.ssh/id_ecdsa']:
-            if (keyfile and os.path.isfile(keyfile)):
+        if not isinstance(self.args.key_filename, list):
+            self.args.key_filename = [self.args.key_filename]
```
kshtsk: setting back args here is not a good style, it is better to use separate variable for key filename list, like:
```
keyfile_list = (lambda x: x if isinstance(x, list) else [x])(self.args.key_filename)
```

kshtsk:
@ktdreyer your request is addressed could you please unblock the PR?

teuthology/openstack/__init__.py
toabctl: done

scripts/openstack.py
toabctl: Resolving the conversation and assuming this is done in the current code.

smithfarm:
I have no objections. I always specify `--key-filename` explicitly, though.

PR #1342, opened 2019-11-01
Fix fetch_branch() check for BranchNotFoundError with newer git

toabctl:
With newer git versions, the string when a branch is not found
is lower case[1].
This fixes the test TestRepoUtils.test_fetch_branch_fake_branch which
was failing due to this problem.

[1] https://github.com/git/git/commit/0b9c3afdbfb629363


kshtsk:
Wonder if we can add a test for this

kshtsk:
ok to test

kshtsk:
@susebot run deploy

toabctl:
> Wonder if we can add a test for this

@kshtsk done

PR #1346, opened 2019-11-06
Disable ceph packages check

rakeshgm:


teuthology/run.py
```diff
@@ -249,6 +249,8 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.setup_additional_repo': None},
             {'kernel.install_latest_rh_kernel': None}
         ])
+        # removing check_packages from init_tasks for now
+        init_tasks.remove({'internal.check_packages': None})
```
kshtsk: 1) Why is this required? 
2) Why just not remove the internal.check_packages at the point where it is added at line 177?

rakeshgm: we are skipping this only rhel as we dont have have EL8 packages available.

vasukulkarni: @kshtsk FYI: this is for 'rh' branch not master, It will affect only downstream runs(not upstream in sepia lab)

kshtsk: @vasukulkarni thanks :-) I was just wondering, sorry for inconvenience

PR #1349, opened 2019-11-12
orchestra: add ceph-daemon DaemonState support

liewegas:


teuthology/orchestra/daemon/cephdaemonunit.py
```diff
@@ -0,0 +1,201 @@
...
+    def check_status(self):
+        """
+        Check to see if the process has exited.
+
+        :returns: The exit status, if any
+        :raises:  CommandFailedError, if the process was run with
+                  check_status=True
+        """
+        raise NotImplementedError
+        proc = self.remote.run(
```
kshtsk: Could you please `remote.sh` instead of `remote.run` in order to make easy to make the code py2/py3 compatible:
```output = self.remote.sh(self.show_cmd + ' | grep -i state')```

File: teuthology/orchestra/daemon/cephdaemonunit.py [hunk 2]
```diff
@@ -0,0 +1,201 @@
...
+            return {key.strip(): value.strip()}
+        show_dict = dict()
+        for line in proc.stdout.readlines():
+            show_dict.update(parse_line(line))
+        active_state = show_dict['ActiveState']
+        sub_state = show_dict['SubState']
+        if active_state == 'active':
+            return None
+        self.log.info("State is: %s/%s", active_state, sub_state)
+        proc = self.remote.run(
```
kshtsk: Same here, use `remote.sh`, so we do not refer cStringIO and StringIO

teuthology/orchestra/daemon/cephdaemonunit.py
```diff
@@ -0,0 +1,201 @@
+import logging
+import re
+from cStringIO import StringIO
```
kshtsk: We should avoid using cString since it is not compatible py2/py3.

teuthology/orchestra/daemon/cephdaemonunit.py
```diff
@@ -0,0 +1,201 @@
...
+        args = ['ps', '-ef',
+                run.Raw('|'),
+                'grep',
+                run.Raw(proc_regex),
+                run.Raw('|'),
+                'grep', '-v',
+                'grep', run.Raw('|'),
+                'awk',
+                run.Raw("{'print $2'}")]
+        proc = self.remote.run(args=args, stdout=StringIO())
```
kshtsk: Same here.

teuthology/orchestra/run.py
```diff
@@ -87,12 +87,8 @@ def execute(self):
         """
         Execute remote command
         """
-        prefix = "Running:"
-        if self.label:
-            prefix = "Running ({label}):".format(label=self.label)
-        log.getChild(self.hostname).info(prefix)
         for line in self.command.split('\n'):
-            log.getChild(self.hostname).info('> %s' % line)
+            log.getChild(self.hostname).info('%s> %s' % (self.label or '', line))
```
kshtsk: Are you sure it is concise duplicating a 'label' for a multi-line script?

liewegas:
@kshtsk updated, and trimmed out most/all of the cruft that got copied over from the systemd file

teuthology/orchestra/run.py
liewegas: the label is generally short.  IMO it'll be more readable.  the main thing is getting rid of the Running: line before every command... in reality the label thing is almost never used

teuthology/orchestra/daemon/group.py
```diff
@@ -57,13 +59,16 @@ def register_daemon(self, remote, type_, id_, *args, **kwargs):
         if id_ in self.daemons[role]:
             self.daemons[role][id_].stop()
             self.daemons[role][id_] = None
+
         klass = DaemonState
-        if remote.init_system == 'systemd':
+        if self.use_ceph_daemon:
+            klass = CephDaemonUnit
+        elif self.use_systemd and \
+             not any(map(lambda i: i == 'valgrind', args)) and \
```
kshtsk: nit: for concise and better readability it is enough ```not any(i == 'valgrind' for i in args)```

teuthology/orchestra/run.py
kshtsk: Ah :-) I found "Running:" useful to quickly jump through the commands, it is pretty opaque to jump with '>'.

kshtsk:
Flake failures:
```
teuthology/orchestra/daemon/cephdaemonunit.py:2:1: F401 're' imported but unused
teuthology/orchestra/daemon/cephdaemonunit.py:4:1: F401 'teuthology.exceptions.CommandFailedError' imported but unused
teuthology/orchestra/daemon/cephdaemonunit.py:5:1: F401 'teuthology.orchestra.run' imported but unused
```

liewegas:
@kshtsk ok can you re-review? i think this is ready to merge

kshtsk:
> @kshtsk ok can you re-review? i think this is ready to merge

looks good to me, except this 'orchestra/run' patch... if the rest of the team is okay I will not insist

kshtsk:
@liewegas how can this patch tested?

kshtsk:
@susebot run deploy

teuthology/orchestra/daemon/group.py
liewegas: fixed

teuthology/orchestra/run.py
liewegas: i'll pull this out into a separate PR since it's unrelated and cosmetic

liewegas:
> @liewegas how can this patch be tested?

I'm testing it locally, and as soon as the next build goes through i'll test it by scheduling a suite.

kshtsk:
@susebot run deploy

PR #1352, opened 2019-11-13
using put_file method instead of python shutil.copyfileobj so it works for python 2 and python 3 both

racpatel:
Signed-off-by: Rachana Patel <racpatel@redhat.com>


kshtsk:
@susebot run deploy

PR #1353, opened 2019-11-13
orchestra/run: more concise logging when we run commands

liewegas:


liewegas:
@kshtsk didn't like that ``:>`` was a less obvious thing to grep/search for.  I just want fewer lines and less cluttered output.  Maybe ``Running>`` prefix or something?  (Or retrain to search for ``:>`` :))

kshtsk:
@susebot run deploy

liewegas:
> Okay, I'll try to train myself to look for `:>`, however there is another advantage of previous scheme is that there was clear when the next multi-line command is separated from from the previous if the latter has no output.

I think the same is still true?  The stdout/stderr doesn't have > or Running there (also, either way, the commands are sometimes backgrounded with wait=False)

PR #1354, opened 2019-11-14
added check_status=False

rakeshgm:
Signed-off-by: rakeshgm <rakeshgm014@gmail.com>

### DIFF
--- a/teuthology/task/internal/redhat.py
+++ b/teuthology/task/internal/redhat.py
@@ -91,7 +91,7 @@ def setup_additional_repo(ctx, config):
             if remote.os.package_type == 'rpm':
                 remote.run(args=['sudo', 'wget', '-O', '/etc/yum.repos.d/rh_add.repo',
                                  add_repo])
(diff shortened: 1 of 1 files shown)


teuthology/task/internal/redhat.py
```diff
@@ -91,7 +91,7 @@ def setup_additional_repo(ctx, config):
             if remote.os.package_type == 'rpm':
                 remote.run(args=['sudo', 'wget', '-O', '/etc/yum.repos.d/rh_add.repo',
                                  add_repo])
-                remote.run(args=['sudo', 'yum', 'update', 'metadata'])
+                remote.run(args=['sudo', 'yum', 'update', 'metadata'], check_status=False)
```
vasukulkarni: I think this command itself is wrong from ages :), It should have been "yum update meta", can  you try on a system and check if that works?

rakeshgm: the output of yum update meta is `No match for argument: meta`, the same is true for `metadata`. This happens only in rhel-8, in older versions, the command works properly. I did not get proper documentation on this.

PR #1356, opened 2019-11-15
[wip] Add basic mypy support

toabctl:
First step for mypy. Still type hints are missing for nearly
everything.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1057,7 +1057,7 @@ def get_valgrind_args(testdir, name, preamble, v):
         return preamble
     if not isinstance(v, list):
         v = [v]
-    val_path = '/var/log/ceph/valgrind'.format(tdir=testdir)
+    val_path = '/var/log/ceph/valgrind'
     if '--tool=memcheck' in v or '--tool=helgrind' in v:
         extra_args = [
             'valgrind',
--- a/teuthology/orchestra/console.py
+++ b/teuthology/orchestra/console.py
@@ -266,8 +266,7 @@ class PhysicalConsole():
         child = self._pexpect_spawn_ipmi('power on')
         child.expect('Chassis Power Control: Up/On', timeout=self.timeout)
         self._wait_for_login()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))
 
     def spawn_sol_log(self, dest_path):
         """
@@ -377,5 +376,4 @@ class VirtualConsole():
         self.vm_domain.info().destroy()
         time.sleep(interval)
         self.vm_domain.info().create()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))
--- a/teuthology/provision/cloud/base.py
+++ b/teuthology/provision/cloud/base.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 
 
 class Provider(object):
-    _driver_posargs = list()
+    _driver_posargs = list() # type: ignore
 
     def __init__(self, name, conf):
         self.name = name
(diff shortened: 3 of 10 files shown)


teuthology/misc.py
```diff
@@ -1057,7 +1057,7 @@ def get_valgrind_args(testdir, name, preamble, v):
         return preamble
     if not isinstance(v, list):
         v = [v]
-    val_path = '/var/log/ceph/valgrind'.format(tdir=testdir)
+    val_path = '/var/log/ceph/valgrind'
```
kshtsk: Is this related to the title of the PR? It looks for me it is not. If so, should be in separate commit.

teuthology/orchestra/console.py
```diff
@@ -266,8 +266,7 @@ def power_off_for_interval(self, interval=30):
         child = self._pexpect_spawn_ipmi('power on')
         child.expect('Chassis Power Control: Up/On', timeout=self.timeout)
         self._wait_for_login()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))
```
kshtsk: Same as above, should go to cleanup commit.

teuthology/orchestra/console.py
```diff
@@ -377,5 +376,4 @@ def power_off_for_interval(self, interval=30):
         self.vm_domain.info().destroy()
         time.sleep(interval)
         self.vm_domain.info().create()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))
```
kshtsk: Same as above.

teuthology/orchestra/console.py
kshtsk: I'm not sure maybe we need to leave 's', and add back it to the message?
Btw, since it is log, there can be better to use "%s" notation, like `log.info('some %s and %s', x, y)`

teuthology/provision/cloud/base.py
```diff
@@ -13,7 +13,7 @@
 
 
 class Provider(object):
-    _driver_posargs = list()
+    _driver_posargs = list() # type: ignore
```
kshtsk: How does it work, why is this required, I am actually not familiar with 'mypy'.

teuthology/test/task/__init__.py
```diff
@@ -1,5 +1,9 @@
 from mock import patch, DEFAULT
 from pytest import raises
+try:
+    from typing import Any
```
kshtsk: why is this required?

teuthology/misc.py
toabctl: It is. Without that change, `mypy` will complain that this code is not correct.

teuthology/orchestra/console.py
toabctl: same here. the current code is incorrect and `mypy` complains about it

teuthology/orchestra/console.py
toabctl: same answer as above

teuthology/test/task/__init__.py
toabctl: It's required for the type hints (even if these are only comments for py2).

toabctl:
`mypy` is a static type checker. So if you add type hints to your classes/methods/functions/variables, `mypy` can check the code (without executing it) if the variables/functions/.. are used correctly.
See http://mypy-lang.org/ and https://docs.python.org/3/library/typing.html
Given that we have nearly no type hints yet in the code, it's not that useful (but also does not harm). But when adding new code, we can also add the type hints directly to get better test coverage.
btw. this is already used in ceph/ceph (eg. ceph-common and ceph-daemon) so it's nothing new to the ceph ecosystem.

teuthology/provision/cloud/base.py
toabctl: See answer below.

toabctl:
@kshtsk it's not WIP imo. I think it's ready to get merged and in follow up commits, the type hints can be added.

PR #1357, opened 2019-11-15
Drop web.py from setup.py install_requires and requirements.txt

toabctl:
It's not used anywhere. requirements.txt got updated with
"pip-compile".

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -55,7 +55,6 @@ markupsafe==1.0           # via jinja2
 monotonic==1.3            # via oslo.utils
 more-itertools==4.3.0     # via pytest
 msgpack-python==0.4.8     # via oslo.serialization
-mysqlclient==1.4.2
 ndg-httpsclient==0.4.2
 netaddr==0.7.19           # via oslo.config, oslo.utils, python-neutronclient
 netifaces==0.10.5         # via oslo.utils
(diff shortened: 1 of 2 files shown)


toabctl:
Looks like `teuthology/coverage.py` imports `MySQLdb` which is not mentioned in `setup.py`. I wonder if `teuthology/coverage.py` is still in use?

kshtsk:
@toabctl please close this PR, it will interfere my upcoming ones, hold on on changes setup.py and requirements for awhile.

kshtsk:
There are some tricks with using 'pip-compile' the code will probably have  issues.

requirements.txt
```diff
@@ -55,7 +55,6 @@ markupsafe==1.0           # via jinja2
 monotonic==1.3            # via oslo.utils
 more-itertools==4.3.0     # via pytest
 msgpack-python==0.4.8     # via oslo.serialization
-mysqlclient==1.4.2
```
kshtsk: This is not related change.

toabctl: It's done automatically by `pip-compile`. It's happening because there is no mysql listed in `install_requires` but it's used in `teuthology/coverage.py`.

toabctl:
Closing as suggested in https://github.com/ceph/teuthology/pull/1357#issuecomment-554252412

PR #1366, opened 2019-11-24
suite: add filter-all parameter

kshtsk:
Add --filter-all parameter to suite to make it possible
precise test case selection, since --filter-in and
--filter-out do not allow narrow test case selection
using AND with filter list.

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -379,6 +379,10 @@ class Run(object):
                 if any(f in path for path in base_frag_paths):
                     return True
                 return False
+            filter_all = self.args.filter_all
+            if filter_all:
+                if not all(matches(f) for f in filter_all):
+                    continue
             filter_in = self.args.filter_in
             if filter_in:
                 if not any(matches(f) for f in filter_in):
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -118,6 +118,9 @@ Scheduler arguments:
   --filter-out KEYWORDS       Do not run jobs whose description contains any of
                               the keywords in the comma separated keyword
                               string specified.
+  --filter-all KEYWORDS       Only run jobs whose description contains each one
+                              of the keywords in the comma separated keyword
+                              string specified.
   --archive-upload RSYNC_DEST Rsync destination to upload archives.
   --archive-upload-url URL    Public facing URL where archives are uploaded.
   --throttle SLEEP            When scheduling, wait SLEEP seconds between jobs.
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -65,7 +65,7 @@ def process_args(args):
         elif key == 'subset' and value is not None:
             # take input string '2/3' and turn into (2, 3)
             value = tuple(map(int, value.split('/')))
-        elif key in ('filter_in', 'filter_out', 'rerun_statuses'):
+        elif key in ('filter_all', 'filter_in', 'filter_out', 'rerun_statuses'):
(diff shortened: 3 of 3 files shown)


kshtsk:
@susebot run deploy

teuthology/suite/run.py
```diff
@@ -359,8 +359,10 @@ def collect_jobs(self, arch, configs, newest=False):
                     'Stopped after {limit} jobs due to --limit={limit}'.format(
                         limit=limit))
                 break
-            # Break apart the filter parameter (one string) into comma
-            # separated components to be used in searches.
+            filter_all = self.args.filter_all
+            if filter_all:
+                if not all(_ in description for _ in filter_all):
```
gregsfortytwo: I'm bad at list comprehensions so I'm having trouble telling if this will work...but shouldn't it be bracketed like the others?

kshtsk: In this case it is not required, without brackets python creates iterator, the all can take utterable objects, so, it is more reasonable to use it not wasting the memory. I know we are not struggle for the performance in our scripting, but this form is legitimate.

kshtsk: Example:
```
>>> a=[1, 2, 3]
>>> b=[2, 2, 2]
>>> all(_==2 for _ in a)
False
>>> all(_==2 for _ in b)
True

```

gregsfortytwo: 👍

tchaikov: i'd suggest `s/_/x/`, because 

1. we are not ignoring `x`
2. for the sake of consistency.

kshtsk: @tchaikov nevertheless someone uses underscore as ignore variables, another good usage example is to utilize it for looping and iterators, since the underscore is the same variable as `x`, or `i` or other, does not have a specific naming, for better abstraction. Secondly, it improves readability especially in comprehensions, we have a lot of example how to do not right code like the one brightest `[param for param in params where param]` :-)
Though, I agree we can try to keep naming consistency.

tchaikov: > for better abstraction

i don't think we need abstraction here. as it's a concrete data structure -- a list of keywords. instead of a general purpose data structure.

>  it improves readability especially in comprehensions

i am not convinced regarding to the readability improvements. if `_` had a name, it helps me to understand what it presents.

kshtsk: > > for better abstraction
> 
> i don't think we need abstraction here. as it's a concrete data structure -- a list of keywords. instead of a general purpose data structure.
> 
> > it improves readability especially in comprehensions
> 
> i am not convinced regarding to the readability improvements. if `_` had a name, it helps me to understand what it presents.

I meant here that abstract context variable name is better for me, I don't care about which exactly name is better x, i, or anything less, this is why '_' is my choice, it is just more readabl, python's `_`  is like wildcard operator in other scripting languages with functional programming abilities,  `$_` in perl, `it` in groovy, `_` in scala.

tchaikov:
@kshtsk ping? what is this change pending on?

kshtsk:
> @kshtsk ping? what is this change pending on?

I wished to test it more.

kshtsk:
@susebot run deploy

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
updated patch to take into account the config fragments since `filter_in` and `filter_out` are using checking it

kshtsk:
@tchaikov could you please take another look

PR #1374, opened 2019-12-10
valgrind.supp is used from ceph/qa

vasukulkarni:
valgrind.supp is used from ceph/qa, we should remove this from setup to avoid

any confusion


kshtsk:
@susebot run deploy

vasukulkarni:
I like this susebot :)

PR #1362, opened 2019-11-20
py3 compatibility

kshtsk:
This PR adds py3 compatibility support
It also enable py3 unit teuthology PR testing

PR #1378, opened 2019-12-11
task/kernel,install: misc fixes for el8

liewegas:
misc fixes for kernel and install tasks for centos8/rhel8

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -1105,15 +1105,6 @@ def get_sha1_from_pkg_name(path):
     return sha1
 
 
-def remove_old_kernels(ctx):
-    for remote in ctx.cluster.remotes.keys():
-        package_type = remote.os.package_type
-        if package_type == 'rpm':
-            log.info("Removing old kernels from %s", remote)
-            args = ['sudo', 'package-cleanup', '-y', '--oldkernels']
-            remote.run(args=args)
-
-
 def task(ctx, config):
     """
     Make sure the specified kernel is installed.
@@ -1217,8 +1208,6 @@ def task(ctx, config):
     need_install = {}  # sha1 to dl, or path to rpm or deb
     need_version = {}  # utsrelease or sha1
     kdb = {}
-
-    remove_old_kernels(ctx)
 
     for role, role_config in config.items():
         # gather information about this remote
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -909,7 +909,7 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
     entry_num = 0
     if '\nmenuitem ' not in grub2conf:
         # okay, do the newer (el8) grub2 thing
-        grub2conf = remote.sh('sudo /bin/ls /boot/loader/entries')
+        grub2conf = remote.sh('sudo /bin/ls /boot/loader/entries || true')
         entry = None
         for line in grub2conf.split('\n'):
             if line.endswith('.conf') and newversion in line:
@@ -923,7 +923,10 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
                     break
                 entry_num += 1
         entry = str(entry_num)
-    remote.run(args=['sudo', grubset, entry])
+    if entry is None:
+        log.warning('Unable to update grub2 order')
+    else:
+        remote.run(args=['sudo', grubset, entry])
 
 
 def generate_legacy_grub_entry(remote, newversion):
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -184,7 +184,8 @@ def need_to_install(ctx, role, version):
         stdout=uname_fp,
         )
     cur_version = uname_fp.getvalue().rstrip('\n')
-    log.debug('current kernel version is {ver}'.format(ver=cur_version))
+    log.debug('current kernel version is {ver} vs {want}'.format(ver=cur_version,
+                                                                 want=version))
 
     if '.' in str(version):
         # version is utsrelease, yay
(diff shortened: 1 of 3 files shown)


teuthology/task/kernel.py
```diff
@@ -905,13 +906,15 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
...
     grub2conf = teuthology.get_file(remote, grubconfig, True)
+    log.debug('grub2conf is:\n%s' % grub2conf)
     entry_num = 0
     for line in grub2conf.split('\n'):
         if line.startswith('menuentry'):
             if newversion in line:
                 break
             entry_num += 1
     remote.run(args=['sudo', grubset, str(entry_num), ])
+    remote.run(args=['sudo', 'cat', '/boot/grub2/grub.cfg'])
```
vasukulkarni: do we need ignore status flag here or is it fine?

liewegas: debugging right now, will drop thsi patch later

teuthology/task/kernel.py
```diff
@@ -910,16 +910,14 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
     entry_num = 0
     if '\nmenuitem ' not in grub2conf:
         # okay, do the newer (el8) grub2 thing
-        proc = remote.run(args=['sudo', 'bash', '-c',
-                                'cat /boot/loader/entries/*.conf'],
+        proc = remote.run(args=['/bin/ls', '/boot/loader/entries'],
```
kshtsk: Please use remote.sh()

vasukulkarni: @kshtsk dont you think its fine to use either remote.run or remote.sh?, just trying to understand (.sh seems to be shortcut method for run and it returnds stdout by default compared to remote object returned by run method)

kshtsk: @vasukulkarni no, I've already cleaned up the kernel task from the cStringIO usage with a purpose of switching to py3, if this will be merged, I will have to fix it again in order py3 tests are passing for py3, I can't merge it right now because this branch is not tested yet, while it can run some of py27 based runs.

vasukulkarni: @kshtsk maybe you missed my question, I am trying to figure out what is the difference between using remote.run and remote.sh method, they have different purpose right? one returns remote object other is a shortcut for the same .run method but returns stdout,  I am not understanding where py3 is playing role?

kshtsk: okay, it is the same but two things: sh always return an 'str' obj, so you do not need provide stdout parameter with a stream obj., and first positional argument is the same as 'args' in run. But yes, they are equivalent.

vasukulkarni: I think I understand your concern now, you commented on wrong remote.run, you should have commented on remote.run that uses StringIO(), thats the one needing change, not this remote.run.

kshtsk: Just for the note: both 'run' and 'sh' supports list and str for args, so the next calls are kinda identical:
```
remote.run(args=['/bin/ls', '/boot/loader/entries'])
remote.run(args='/bin/ls /boot/loader/entries')
remote.sh(['/bin/ls', '/boot/loader/entries'])
remote.sh('/bin/ls /boot/loader/entries')
```
The main advantage of the run at the moment is that it can be post-process proc object of runs executed in background, but I'm thing of improving the sh so you can give an object reference or a callback function. The idea is to avoid bootstrapping, it is good for code readability.

kshtsk: > I think I understand your concern now, you commented on wrong remote.run, you should have commented on remote.run that uses StringIO(), thats the one needing change, not this remote.run.

I commented under 913 line, if you take a look at 914 you'll see StringIO :-D

liewegas: @kshtsk can you merge these py3 fixes?  i'm just copying and modifying the adjacent code

vasukulkarni: I just read bit of kernel task code now, I see its mix of remote.run and .sh, .run has kind of long legacy,  the best thing is to leave it to user to decide based on return values, .run still has advantages where you want to use remote object in the test code.

liewegas: @kshtsk I'd suggest opening a pr with teh changes on a per-task basis so we can merge and work on top of that.  Right now it's a lose-lose because I can't see what the target interface/methods are that I should use, and even if I did, it would still be creating merge conflicts for you and require a rebase if your changes go in after mine.

liewegas:
squashed and cleaned up!

teuthology/task/kernel.py
kshtsk: > @kshtsk I'd suggest opening a pr with teh changes on a per-task basis so we can merge and work on top of that. Right now it's a lose-lose because I can't see what the target interface/methods are that I should use, and even if I did, it would still be creating merge conflicts for you and require a rebase if your changes go in after mine.

This is exactly what is going on, most of the code is merged already. I do my best of fixing task. But the main code of run functions is not well tested yet. Some of the jobs can fails. Corresponding PR is #1362.

For this one change, there will no be merge conflict, we are discussing about your code changes, this is why a do a review of the code to avoid merge conflict, and it is possible to fix that before merge.

teuthology/task/kernel.py
```diff
@@ -906,12 +907,25 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
...
     grub2conf = teuthology.get_file(remote, grubconfig, True)
     entry_num = 0
-    for line in grub2conf.split('\n'):
-        if line.startswith('menuentry'):
-            if newversion in line:
+    if '\nmenuitem ' not in grub2conf:
+        # okay, do the newer (el8) grub2 thing
+        proc = remote.run(args=['sudo', '/bin/ls', '/boot/loader/entries'],
+                          stdout=StringIO())
+        grub2conf = proc.stdout.getvalue()
```
kshtsk:
```
grub2conf = remote.run('sudo /bin/ls /boot/loader/entries')
```

liewegas: fixed.  the change below will conflict tho right?

vasukulkarni: I think you meant remote.sh here.

kshtsk: yeah :-) sorry, I meant `remote.sh` :-)

kshtsk:
@susebot run deploy

liewegas:
@susebot run deploy

liewegas:
@susebot run deploy

kshtsk:
@liewegas you did right thing in rescheduling :-) the latter run is in progress http://51.68.91.103:8081/runner-2019-12-12_20:34:18-smoke-master-distro-basic-ovh/

teuthology/task/kernel.py
```diff
@@ -906,12 +907,26 @@ def grub2_kernel_select_generic(remote, newversion, ostype):
     remote.run(args=['sudo', mkconfig, '-o', grubconfig, ])
     grub2conf = teuthology.get_file(remote, grubconfig, True)
     entry_num = 0
-    for line in grub2conf.split('\n'):
-        if line.startswith('menuentry'):
-            if newversion in line:
+    if '\nmenuitem ' not in grub2conf:
+        # okay, do the newer (el8) grub2 thing
+        grub2conf = remote.sh('sudo /bin/ls /boot/loader/entries || true')
```
kshtsk: I'm not asking for a change, just wonder why there is a full path to `ls` required :-) I believe this is a legacy approach when there was legacy python function required a full path to executable module because it was running without shell :-)

liewegas: just paranoia on my part to make sure there isn't any colorization or ``/`` suffixes or whatever that you get with 'ls' aliases somethings.  those should't come up with a non-tty but just in case

kshtsk:
One of the job is passed another is failed with message:
```
"2019-12-12T21:38:06.003473+0000 mon.b (mon.0) 105 : cluster [WRN] Health check failed: 1 slow ops, oldest one blocked for 33 sec, osd.3 has slow ops (SLOW_OPS)" in cluster log
```
which is not related to the PR.

batrick:
> Worst case we'll fail later because we didn't boot the right kernel.

As foretold: https://tracker.ceph.com/issues/43335

:)

PR #1379, opened 2019-12-12
setup.py: add missing dependencies of tests

tchaikov:
Fixes: https://tracker.ceph.com/issues/43192

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -70,3 +70,16 @@ and should work with teuthology. But if you want to bump up the versions of them
 please use the following command to update this file ::
 
   pip-compile -qo- | sed '/^-e / d' > requirements.txt
+
+Please upgrade pip-tool using following command ::
+
+  pip install pip-tools --upgrade
+
+if the command above fails like::
+
+  Traceback (most recent call last):
+  File "/home/kchai/teuthology/virtualenv/bin/pip-compile", line 5, in <module>
+    from piptools.scripts.compile import cli
+  File "/home/kchai/teuthology/virtualenv/local/lib/python2.7/site-packages/piptools/scripts/compile.py", line 11, in <module>
+    from pip.req import InstallRequirement, parse_requirements
+  ImportError: No module named req
--- a/setup.py
+++ b/setup.py
@@ -92,10 +92,12 @@ setup(
             'nose', # for qa/tasks/rgw_multisite_tests.py',
             'pip-tools',
             'pytest',           # for tox.ini
+            'requests',         # for qa/tasks/mgr/dashboard/helper.py
             'tox',
             # For bucket notification testing in multisite
             'xmltodict',
             'boto3',
+            'PyJWT',            # for qa/tasks/mgr/dashboard/test_auth.py
         ]
     },
 
--- a/setup.py
+++ b/setup.py
@@ -51,21 +51,15 @@ setup(
     install_requires=['gevent',
                       'PyYAML',
                       'argparse >= 1.2.1',
-                      'boto >= 2.0b4',
                       'configobj',
                       'six >= 1.9', # python-openstackclient won't work properly with less
                       'pexpect',
-                      'nose', # for qa/tasks/rgw_multisite_tests.py',
                       'docopt',
                       'psutil >= 2.1.0',
                       'configparser',
                       'ansible>=2.0',
                       'prettytable',
                       'manhole',
-                      # For bucket notification testing in multisite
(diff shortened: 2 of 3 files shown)


tchaikov:
@epuertat good question. i think it includes the tests for ceph and teuthology. i can split them if you think it's necessary to do so.

tchaikov:
but i am merging this PR as it's blocking the dashboard tests.

setup.py
```diff
@@ -93,9 +87,17 @@
...
         ],
         'test': [
-            'tox',
+            'boto >= 2.0b4',       # for qa/tasks/radosgw_*.py
+            'cryptography >= 2.7',  # for qa/tasks/mgr/dashboard/test_rgw.py
+            'nose', # for qa/tasks/rgw_multisite_tests.py',
             'pip-tools',
             'pytest',           # for tox.ini
+            'requests',         # for qa/tasks/mgr/dashboard/helper.py
+            'tox',
```
toabctl: Really tox? Isn't that a pre-requirement to run the tests inside tox?

tchaikov: yeah, that's why i think it belongs to "test".

docs/INSTALL.rst
```diff
@@ -70,3 +70,16 @@ and should work with teuthology. But if you want to bump up the versions of them
 please use the following command to update this file ::
 
   pip-compile -qo- | sed '/^-e / d' > requirements.txt
+
+Please upgrade pip-tool using following command ::
+
+  pip install pip-tools --upgrade
+
+if the command above fails like::
```
kshtsk: @tchaikov I think this commit should not has merged, it would be enough to 'pip-compile -P pip-tools', btw, pip-compile from py2.7 and py3.x produce different output.

tchaikov: is it a chicken and egg problem? also, i am not sure if we are python3 ready? so i assume it's all python2.

kshtsk: @tchaikov we will never be python3 if we will not python3 have ready requirements.txt. 
Btw, https://jenkins.ceph.com/job/teuthology-pull-requests/707/console:
```
___________________________________ summary ____________________________________
  docs: commands succeeded
  py3: commands succeeded
  py27: commands succeeded
  py27-integration: commands succeeded
  flake8: commands succeeded
  openstack: commands succeeded
  congratulations :)
```

setup.py
toabctl: I think it doesn't belong anywhere imo. It's just an external tool that you can use with `tox.ini`

tchaikov: i see. i think we added it here just because it's a convenient way to add stuff to `requirements.txt`. the same applies to manhole, which is for debugging a task.

kshtsk: I saw sometimes issues with system tox :-) so maybe it is a good thing to have it here, however I'm not 100% sure about that

toabctl: it's not useful (imo it's wrong). You already need tox to install requirements.txt in a venv via setuptools/distutils.
And `manhole` is totally different because `manhole` gets imported (`teuthology/__init__.py:import manhole`) so it needs to be in requirements.txt.
`tox` is not imported (or called via eg. subprocess) so it must not be in requirements.txt

kshtsk: > it's not useful (imo it's wrong). You already need tox to install requirements.txt in a venv via setuptools/distutils.
it is not true, you do not need `tox`, you need `pip` to setup virtualenv or whatever what I usually do is:
```
virtualenv v ; . v/bin/activate
pip install -r requirements.txt
```

tchaikov: gentlemen, please note, tox is not added in this change. i was merely reordering it. and i agree with @kshtsk, one don't need to tox to install packages specified requirements.txt.

docs/INSTALL.rst
tchaikov: thanks for the information. but i don't think python3 compatibility is in the scope of this PR. its purpose was to address https://tracker.ceph.com/issues/43192

kshtsk: yes... but if under "it's all python2" means "it's all python2 ONLY" is getting wrong, yes we need to keep it backward compatible with py2, but we should prefer py3 compatibility over py2 only code.

setup.py
toabctl: yeah. I also agree with that. But when you do `pip install -r requirements.txt`, `tox` will be installed which is useless because you already have your venv...

toabctl: anyway. Closing the discussion. Let's create a PR for removing tox if wanted and discuss there..

PR #1380, opened 2019-12-12
py3 requirements

kshtsk:
Fix requirements.txt to satisfy py3 dependencies.

Because pip-compile produces different set of packages
for different python versions it needs to have
a dedicated requirements.txt for py2 and py3.


kshtsk:
@susebot run deploy

requirements3.txt
```diff
@@ -0,0 +1,104 @@
+#
```
toabctl: So which python3 is that now? The output from `pip-compile` might be different depending on the python minor version that was used...

kshtsk: I'm not sure there is a difference between 3.5, 3.6, 3.7 however if there a big difference we can add corresponding file. I was trying to generate it on 3.6 and 3.7 it seemed to me it produced the same content. For now we are resolving more global difference is 2 and 3.

toabctl: ok. good. then let's keep it simple. thanks for checking!

kshtsk: btw, the jenkins.ceph.com slave is running python3.5 and it works fine so far

PR #1381, opened 2019-12-12
orchestra: ceph-daemon -> cephadm

liewegas:

### DIFF
--- a/teuthology/orchestra/daemon/cephadmunit.py
+++ b/teuthology/orchestra/daemon/cephadmunit.py
@@ -4,14 +4,14 @@ from teuthology.orchestra.daemon.state import DaemonState
 
 log = logging.getLogger(__name__)
 
-class CephDaemonUnit(DaemonState):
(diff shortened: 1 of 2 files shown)


liewegas:
passes! http://pulpito.ceph.com/sage-2019-12-12_16:44:23-rados-wip-tdehler-testing-31811-distro-basic-smithi/

teuthology/orchestra/daemon/cephadmunit.py
```diff
@@ -4,14 +4,14 @@
 
 log = logging.getLogger(__name__)
 
-class CephDaemonUnit(DaemonState):
+class CephadmUnit(DaemonState):
```
kshtsk: nit: CephAdmUnit?

liewegas: i went with CephadmUnit because 'cephadm' is one word (not 'ceph-adm').

PR #1382, opened 2019-12-13
py3: cast dict keys to list

kshtsk:


kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/nuke/__init__.py
```diff
@@ -296,7 +296,7 @@ def nuke_one(ctx, target, should_unlock, synch_clocks, reboot_all,
         ret = target
     else:
         if should_unlock:
-            unlock_one(ctx, target.keys()[0], ctx.owner)
+            unlock_one(ctx, list(target.keys())[0], ctx.owner)
```
toabctl: not sure how many elements `target.keys()` has, but if there are a lot, it's faster to really just get the first element with `next(iter(target.keys()))` instead of keeping all keys in memory just to get the first one.
Same comment is true for other places where you do that.
But I'm fine with keeping it as it is if the list is not super large.

teuthology/task/kernel.py
```diff
@@ -682,7 +682,7 @@ def wait_for_reboot(ctx, need_install, timeout, distro=False):
     starttime = time.time()
     while need_install:
         teuthology.reconnect(ctx, timeout)
-        for client in need_install.keys():
+        for client in list(need_install.keys()):
```
toabctl: Is that change needed? Iterating over keys should just work fine in py3, too.

tox.ini
```diff
@@ -52,6 +52,13 @@ deps=
   flake8
 commands=flake8 --select=F,E9 {posargs:teuthology scripts}
 
+[testenv:flake8-py3]
```
toabctl: Do you really want that commit in this PR? I think it's from https://github.com/ceph/ceph/pull/32222

teuthology/kill.py
```diff
@@ -239,7 +239,7 @@ def nuke_targets(targets_dict, owner):
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT)
     for line in iter(proc.stdout.readline, ''):
-        line = line.replace('\r', '').replace('\n', '')
+        line = line.encode().replace('\r', '').replace('\n', '')
```
tchaikov: might want to put
```py
line = line.encode().replace(b'\r', b'').replace(b'\n', b'')
```

teuthology/nuke/__init__.py
kshtsk: You're right, and sometime I use next-iter combination, I think it is not critical here since keys are just an objects.

teuthology/task/kernel.py
kshtsk: Yes, it is needed very much because of:
```
RuntimeError: dictionary changed size during iteration
```

smithfarm:
Could you squash the last two commits into one?

tox.ini
kshtsk: Yes, I want it, as early as possible.

kshtsk:
> Could you squash the last two commits into one?

oh, right, I forgot to do this

teuthology/kill.py
kshtsk: nice catch... need to double check

teuthology/task/kernel.py
kshtsk: btw, there is a shorter way of getting keys as a list from the dict, I would write like `list(need_install)` but this is probably not so intuitive as with exclusively using keys.

teuthology/kill.py
kshtsk: I've dropped this commit, and give it a try in another PR.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

PR #1383, opened 2019-12-13
worker: Drop dead code handling old releases

toabctl:
We no longer need to check for branches like 'argonaut', 'bobtail',
'cuttlefish' or 'dumpling'. These are long gone.
Removing the code simplifies it a bit.

### DIFF
--- a/teuthology/test/test_worker.py
+++ b/teuthology/test/test_worker.py
@@ -1,6 +1,5 @@
 import beanstalkc
 import os
-import subprocess
 
 from mock import patch, Mock, MagicMock
 from datetime import datetime, timedelta
@@ -168,7 +167,7 @@ class TestWorker(object):
             "job_id": "1",
             "worker_log": "worker_log",
             "archive_path": "archive/path",
-            "teuthology_branch": "argonaut"
+            "teuthology_branch": "jewel"
         }
         process = Mock()
         process.poll.return_value = "not None"
@@ -177,13 +176,6 @@ class TestWorker(object):
         m_popen.return_value = m_proc
         worker.run_with_watchdog(process, config)
         m_symlink_log.assert_called_with(config["worker_log"], config["archive_path"])
-        expected_cmd = "teuthology-report -v -D -r the_name -j 1"
-        m_popen.assert_called_with(
-            expected_cmd,
-            shell=True,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT
-        )
 
     @patch("os.path.isdir")
     @patch("teuthology.worker.fetch_teuthology")
--- a/teuthology/worker.py
+++ b/teuthology/worker.py
@@ -316,33 +316,15 @@ def run_with_watchdog(process, job_config):
         report.try_push_job_info(job_info)
         time.sleep(teuth_config.watchdog_interval)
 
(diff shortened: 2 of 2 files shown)


kshtsk:
I guess you've broken integration tests, they are already trying to connect somewhere.

toabctl:
> I guess you've broken integration tests, they are already trying to connect somewhere.

resolved

kshtsk:
> > I guess you've broken integration tests, they are already trying to connect somewhere.
> 
> resolved

no, it is still running too long

teuthology/test/test_worker.py
```diff
@@ -177,13 +177,6 @@ def test_run_with_watchdog_with_reporting(self, m_sleep, m_symlink_log, m_popen)
         m_popen.return_value = m_proc
         worker.run_with_watchdog(process, config)
         m_symlink_log.assert_called_with(config["worker_log"], config["archive_path"])
-        expected_cmd = "teuthology-report -v -D -r the_name -j 1"
-        m_popen.assert_called_with(
-            expected_cmd,
-            shell=True,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT
-        )
```
kshtsk: Not sure I understand why this is removed?

toabctl: Because the command is no longer called because I removed the the special handling which does that call.

toabctl:
> no, it is still running too long

it's resolved now :)

teuthology/worker.py
```diff
@@ -316,33 +316,11 @@ def run_with_watchdog(process, job_config):
...
-        # Let's make sure that paddles knows the job is finished. We don't know
-        # the status, but if it was a pass or fail it will have already been
-        # reported to paddles. In that case paddles ignores the 'dead' status.
-        # If the job was killed, paddles will use the 'dead' status.
-        report.try_push_job_info(job_info, dict(status='dead'))
+    # Let's make sure that paddles knows the job is finished. We don't know
+    # the status, but if it was a pass or fail it will have already been
+    # reported to paddles. In that case paddles ignores the 'dead' status.
+    # If the job was killed, paddles will use the 'dead' status.
+    report.try_push_job_info(job_info, dict(status='dead'))
```
tchaikov: probably we can add 
```py
assert(job_config.get('teuthology_branch') not in ('argonaut', 'bobtail', 'cuttlefish', 'dumpling'))
```
?

tchaikov: just to make sure we don't run into some unexpected issues. also, this line is served as a document that we don't support testing these branches.

toabctl: Feels a bit weird to me but Done

tchaikov: thanks.

toabctl:
@susebot run deploy

toabctl:
@susebot run deploy

toabctl:
> Commit [10a0985](https://github.com/ceph/teuthology/commit/10a09855346c4da9c61f813723ef8255989a5888) is _NOT OK_.
> Check tests results in the Jenkins job: https://ceph-ci.suse.de/job/pr-teuthology-deploy/94/

Failing with `RuntimeError: Could not find an image for centos 8.0`

PR #1385, opened 2019-12-14
task/install: fix UnboundLocalError of local variable repos

shangdehao1:
If overrides is None, repos will don't be assignmented.
In this case, will occur error "UnboundLocalError: local variable repos referenced before assignment"

Signed-off-by: dehao.shang@intel.com


kshtsk:
ok to test

kshtsk:
@susebot run deploy

PR #1387, opened 2019-12-17
setup.py: move more modules back to "install_requires"

tchaikov:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,13 +48,18 @@ setup(
         'Topic :: System :: Distributed Computing',
         'Topic :: System :: Filesystems',
     ],
-    install_requires=['gevent',
+    install_requires=['apache-libcloud',
+                      'gevent',
                       'PyYAML',
                       'argparse >= 1.2.1',
                       'configobj',
                       'six >= 1.9', # python-openstackclient won't work properly with less
                       'pexpect',
                       'docopt',
+                      'netaddr',  # teuthology/misc.py
+                      # only used by orchestra, but we monkey-patch it in
+                      # teuthology/__init__.py
+                      'paramiko',
                       'psutil >= 2.1.0',
                       'configparser',
                       'ansible>=2.0',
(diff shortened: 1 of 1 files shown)


kshtsk:
@susebot run deploy

tchaikov:
@kshtsk could you please rerun deploy test? i just updated the PR.

teuthology/provision/__init__.py
```diff
@@ -20,6 +16,7 @@ def _logfile(ctx, shortname):
 
 
 def reimage(ctx, machine_name):
+    from teuthology.provision import fog
```
kshtsk: Why is this required, that is not a good practice, looks like you're trying to resolve circular imports here, where are you facing this issue.

tchaikov: no. i am not trying to fix cyclic dependency issue. please see the commit message. i am trying to minimize the set of dependencies required for running `vstart_runner.py`, which does not need to talk to a remote service for provision, etc.

kshtsk: I don't see how it could talk to remote service for provision until you call corresponding methods, what's wrong with those dependencies in the vstart? Still I don't think to place imports in the function is a good idea.

tchaikov: no, it does not use the openstack API until the methods are called, but the crux is that i want to avoid the import. because it's a waste of time and resource to install unused modules, just because they are imported in the beginning of the `__init__.py`.

> Still I don't think to place imports in the function is a good idea.

could you be more specific why it is not a good idea? yes, it's unusual. but why is it bad? if you insist, i will have to partially invert my changes in `setup.py` and move some "orchestra" back into "install_requires".

tchaikov:
jenkins retest please

teuthology/provision/__init__.py
tchaikov: seems some unit tests are using them. so i end up moving them to "install_requires".

tchaikov:
@susebot run deploy

kshtsk:
@tchaikov I’m sorry not keeping the discussion since I’m on vacation till the second week of January, so I cannot work on this PR. But I insist on keep import on the beginning of the module, having imports in functions is very special case and it should be very good reason (improving dependency deployment speed for vstart is not the one).

kshtsk:
Btw, do we have a tracker ticket for this?

tchaikov:
@kshtsk no worries. i dropped that change. hopefully, this change is less controversial now. AFAIK, no, we don't have a tracker ticket for it.

callithea:
> @kshtsk no worries. i dropped that change. hopefully, this change is less controversial now. AFAIK, no, we don't have a tracker ticket for it.

@tchaikov @kshtsk https://tracker.ceph.com/issues/43331

kshtsk:
thanks guys, feel free to merge this... looks like the deploy test has got a limited functionality after the default os version changed to centos8 and we do not have centos8 image in our suse tenant of openstack. I can work on fixing it after I get back from vacation.

tchaikov:
thanks @kshtsh. and enjoy your vacation!

PR #1389, opened 2019-12-18
run_task: fix qa/ directory missing

shangdehao1:
In my testing environments, sys.path have "/home/teuthworker/src/github.com_ceph_ceph_xxx/", but don't have "/qa" path.  In this case, will show error "ImportError: No module named tasks.ceph"

if we add "/qa" path to sys.path using suite_path, can fix this error.

### DIFF
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -15,7 +15,7 @@ from teuthology.timer import Timer
 log = logging.getLogger(__name__)
 
 
-def get_task(name):
+def get_task(name, ctx):
     if '.' in name:
         module_name, task_name = name.split('.')
     else:
@@ -25,6 +25,9 @@ def get_task(name):
     module = _import('teuthology.task', module_name, task_name)
     # If it is not found, try qa/ directory (if it is in sys.path)
     if not module:
+        module = _import('tasks', module_name, task_name)
+    if not module:
+        sys.path.append(ctx.config.get("suite_path") + "/qa")
         module = _import('tasks', module_name, task_name, fail_on_import_error=True)
     try:
         # Attempt to locate the task object inside the module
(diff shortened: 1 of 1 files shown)


liewegas:
jenkins test this please

kshtsk:
ok to test

teuthology/run_tasks.py
```diff
@@ -25,6 +25,9 @@ def get_task(name):
     module = _import('teuthology.task', module_name, task_name)
     # If it is not found, try qa/ directory (if it is in sys.path)
     if not module:
+        module = _import('tasks', module_name, task_name)
+    if not module:
+        sys.path.append(ctx.config.get("suite_path") + "/qa")
```
vasukulkarni: the suite_path setup by teuthology has qa, ex: http://qa-proxy.ceph.com/teuthology/teuthology-2019-12-18_07:00:04-smoke-nautilus-testing-basic-smithi/4613617/teuthology.log , shouldn't you do the same when running locally with custom qa path?

shangdehao1: If use teuthology-suite to submit some jobs,  will have change to use --suite-relpath parameter.  But, when i use teuthology-schedule to submit job,  seem unable to use relpath. 

i try to print suite_path in  https://github.com/ceph/teuthology/blob/master/teuthology/worker.py#L165 , and the result of suite_path is /home/teuthworker/src/github.com_ceph_ceph_xxx/.  I don't sure if there are other method to custom qa path. thanks

kshtsk:
Not sure I understand when do you have this issue, could please elaborate your setup?

shangdehao1:
> Not sure I understand when do you have this issue, could please elaborate your s

shangdehao1:
> Not sure I understand when do you have this issue, could please elaborate your setup?

hi, if i directly teuthology-schedule to run a test case, will see this issue. Currently, if use teuthology-suite to run a test case,  will not face this issue. Closed this PR. Thanks.

PR #1392, opened 2020-01-07
use centos 8.0 by default

batrick:
Not sure if this is a good alternative fix or not to https://tracker.ceph.com/issues/43516 . I'm writing a separate fix to have fs:verify set the os_version. (Specifically, I wonder if this will break testing for nautilus/mimic.)


teuthology/orchestra/opsys.py
```diff
@@ -60,7 +60,7 @@
 DEFAULT_OS_VERSION = dict(
     ubuntu="18.04",
     fedora="25",
-    centos="7.6",
+    centos="8.0",
```
tchaikov: this means all jobs whose "os_type" is "centos" without specifying its "os_version" will start using "centos 8.0" after this change. i don't know the exact impaction of this change. but i think the bottom line is that we need to ensure all branches being tested should have EL8 build if we want to make change. but apparently nautilus does not have EL8 build at this moment. see https://shaman.ceph.com/builds/ceph/nautilus/f774f190ec882aa0a38c103cceac793eb28f0e75/

if you want to make this change, at least we should make sure all nautilus teuthology jobs do point their "os_version" to 7.6 if their "os_type" is "centos". i think the same applies to "mimic".

batrick: I personally don't feel it's worth it and perhaps keeping this as an old version will help with converting new and existing QA suites over to explicit versions. So I'm good with closing this.

PR #1394, opened 2020-01-15
bootstrap: install setuptools 44

jan--f:
With setuptools 45 python2 support was dropped.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -141,7 +141,11 @@ SVER=${LVER%%.*}
 ./$VENV/bin/pip install --upgrade pip
 
 # Ensure setuptools is installed
-./$VENV/bin/pip install setuptools --upgrade
+if [ "$SVER" = "2" ]; then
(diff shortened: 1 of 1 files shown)


smithfarm:
@jan--f Would it be appropriate to add `Fixes: https://tracker.ceph.com/issues/43606` ?

bootstrap
```diff
@@ -141,7 +141,11 @@ SVER=${LVER%%.*}
 ./$VENV/bin/pip install --upgrade pip
 
 # Ensure setuptools is installed
-./$VENV/bin/pip install setuptools --upgrade
+if [ "$PYTHON" = "python2" ]; then
```
kshtsk: actually, `"$SVER" = "2"` may work better, since one can use PYTHON=python2.7 for example.

theanalyst:
This works for me locally, LGTM, apart from kyr's comments

PR #1395, opened 2020-01-15
tasks/internal: put kernel.core_pattern in sysctl.conf too

liewegas:
This way if systemd restarts  or is reinstalled (as it is on el8 at the
moment) we won't lose this setting.


djgalloway:
jenkins test

dmick:
Would be even better in /etc/sysctl/conf.d/ceph.conf

liewegas:
or i guess teuthology.conf.  it doesn't matter too much since this is teuthology doing this to a throwaway install that will get reimages in 20 minutes anyway

PR #1400, opened 2020-01-23
selinux: use text processing

batrick:
We ran into some odd grep bug treating the audit.log as binary and
wrongly returning a match result.

Fixes: https://tracker.ceph.com/issues/43797


batrick:
RErunning here to see if the warnings go away: http://pulpito.ceph.com/pdonnell-2020-01-23_23:25:45-kcephfs-wip-pdonnell-testing-20200122.210533-distro-basic-smithi/

PR #1405, opened 2020-02-05
hammer.sh: don't do teuthology-nuke anymore.

rzarzynski:
Apparently `hammer.sh` is pretty old. It calls `teuthology-nuke` expecting the further call to `teuhology` will be able to still reach the nodes. However, it doesn't seem that anybody is turning them back these days.

The commit just drops the `teuthology-nuke` call.

### DIFF
--- a/hammer.sh
+++ b/hammer.sh
@@ -14,8 +14,6 @@ fi
 
 test -e $1
 
-teuthology-nuke -t $job
-
 title() {
 	echo '\[\033]0;hammer '$job' '$N' passes\007\]'
 }


djgalloway:
Does this relock (and thereby reimage) the testnodes as well?  I ask because if it doesn't, then if the systems are getting reused over and over until the job passes, they won't have a clean environment after the first run completes.

rzarzynski:
@djgalloway: no, it doesn't. Thanks for pointing this out! Do we have an already available command for doing the re-lock/? If not, it could be relatively easily developed by reusing the parts of `teuthology-lock`. I'm asking to avoid re-doing an already existing thing. :-)

djgalloway:
I didn't even know `hammer.sh` was a thing so I don't actually know :smiley:

rzarzynski:
@djgalloway: I wasn't aware about the script till yesterday. @jdurgin told me about it. :-)

jdurgin:
it's pre-fog so it assumes that a passing test won't muck up a node enough to affect the next run. in theory that's still true :)

PR #1407, opened 2020-02-11
worker_start: allow ARCHIVE variable override

kshtsk:
The worker_start.sh script is using archive directory inside
the worker user home which cannot be changed, this patches
allows override ARCHIVE environment variable to address
this issue.

### DIFF
--- a/docs/_static/worker_start.sh
+++ b/docs/_static/worker_start.sh
@@ -2,7 +2,7 @@
  
 # A simple script used by Red Hat to start teuthology-worker processes.
  
-ARCHIVE=$HOME/archive
+ARCHIVE=${ARCHIVE:-"$HOME/archive"}
 WORKER_LOGS=$ARCHIVE/worker_logs
  
 function start_workers_for_tube {


kshtsk:
@susebot run deploy

docs/_static/worker_start.sh
```diff
@@ -2,7 +2,7 @@
  
 # A simple script used by Red Hat to start teuthology-worker processes.
  
-ARCHIVE=$HOME/archive
+ARCHIVE=${ARCHIVE:-"$HOME/archive"}
```
tchaikov: could also put

```sh
: ${ARCHIVE:="$HOME/archive"}
```

see https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

kshtsk: Just curious, why do you think it is better than ":-"?

tchaikov: less repeating.

kshtsk:
@susebot run deploy

PR #1408, opened 2020-02-14
task/kernel.py: fix -k distro for opensuse/sle (part 1)

smithfarm:
Part one - this does not get us all the way, but it's a start!

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -713,6 +713,7 @@ def need_to_install_distro(remote):
     :returns: False if running the newest distro kernel. Returns the version of
               the newest if it is not running.
     """
+    dist_release = remote.os.name
     package_type = remote.os.package_type
     output, err_mess = StringIO(), StringIO()
     remote.run(args=['uname', '-r'], stdout=output)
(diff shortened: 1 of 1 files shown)


teuthology/task/kernel.py
```diff
@@ -721,37 +722,44 @@ def need_to_install_distro(remote):
...
-                        'Newest distro kernel already installed and running')
-                    return False
-            else:
-                remote.run(args=['sudo', 'yum', 'reinstall', '-y', 'kernel',
-                                 run.Raw('||'), 'true'])
+        if dist_release in ['opensuse', 'sle']:
+            install_stdout = remote.sh(
+                'sudo zypper --non-interactive install kernel-default'
+                )
+            match = re.search(
```
toabctl: not needed because not used in the sle/opensuse case

teuthology/task/kernel.py
```diff
@@ -1040,7 +1058,7 @@ def get_latest_image_version_rpm(remote):
     for kernel in proc.stdout.getvalue().split():
         if kernel.startswith('kernel'):
             if 'ceph' not in kernel:
-                version = kernel.split('kernel-')[1]
+                version = kernel.split('{}-'.format(kernel_pkg_name))[1]
```
kshtsk: Using 'format` here is overkill.

teuthology/task/kernel.py
kshtsk: agree, looks like it

kshtsk:
I guess this PR is not ready:
```
teuthology/task/kernel.py:844:24: F821 undefined name 'distribution'
teuthology/task/kernel.py:873:24: F821 undefined name 'distribution'
```

smithfarm:
@toabctl @kshtsk Thanks for reviewing. Please take another look.

kshtsk:
@susebot run deploy

smithfarm:
@toabctl Has your change request been addressed?

PR #1409, opened 2020-02-14
requirements: pip-compile -P six

kshtsk:
We need to pump six to the latest 1.14.0 so we have
six.ensure_str() is available


kshtsk:
@susebot run deploy

PR #1410, opened 2020-02-18
task/ceph_ansible: Work with LVM scratch devices

badone:
Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -7,7 +7,7 @@ import yaml
 from teuthology.task import Task
 from tempfile import NamedTemporaryFile
 from teuthology.config import config as teuth_config
-from teuthology.misc import get_scratch_devices
+from teuthology.misc import get_scratch_devices, get_file
 from teuthology import contextutil
 from teuthology.orchestra import run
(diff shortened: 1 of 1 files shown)


teuthology/task/ceph_ansible.py
```diff
@@ -308,7 +308,22 @@ def get_host_vars(self, remote):
...
                               if role.startswith('osd')])
-            host_vars['devices'] = get_scratch_devices(remote)[0:dev_needed]
+            if teuth_config.get('ceph_ansible') and \
+                    self.ctx.machine_type in teuth_config['ceph_ansible']['has_lvm_scratch_disks']:
+                devices = get_file(remote, "/scratch_devs").split()
+                vols = []
+
+                for dev in devices:
+                   if 'vg_nvme' in dev:
+                       splitpath = dev.split('/')
```
kshtsk: This is py3 incompatible, since get_file returns a binary data

badone: @kshtsk I can adda follow-up PR with the following change if that looks right to you?

`devices = get_file(remote, "/scratch_devs").decode().split()`

PR #1411, opened 2020-02-18
kernel.py: fix -k distro for opensuse/sle (part 2)

smithfarm:
Signed-off-by: Nathan Cutler <ncutler@suse.com>

PR #1412, opened 2020-02-18
packaging: add Shaman repo correctly on opensuse

smithfarm:
Fixes: https://tracker.ceph.com/issues/43655
Works Around: https://tracker.ceph.com/issues/44183

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -973,12 +973,40 @@ class ShamanProject(GitbuilderProject):
         return resp.text
 
     def _install_rpm_repo(self):
+        dist_release = self.dist_release
         repo = self._get_repo()
-        sudo_write_file(
-            self.remote,
-            '/etc/yum.repos.d/{proj}.repo'.format(proj=self.project),
-            repo,
-        )
+        if dist_release in ['opensuse', 'sle']:
+            #
+            # Shaman does not currently return opensuse repos in a format that zypper
+            # understands -- see https://tracker.ceph.com/issues/44183 for details.
+            #
+            # So, do some text manipulation to convert the yum-style repo to
+            # zypper-style.
+            #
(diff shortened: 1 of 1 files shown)


smithfarm:
Woohoo, with #1411 and this patch, we get through `kernel.py` *and* `install.py` all the way into `cephadm.py`:

http://pulpito.ceph.com/smithfarm-2020-02-18_16:23:02-custom-master-distro-basic-smithi/

smithfarm:
@gregsfortytwo Can you take another look?

teuthology/packaging.py
```diff
@@ -973,12 +973,40 @@ def _get_repo(self):
...
+            # zypper-style.
+            #
+            # This text manipulation/conversion code should continue to work even
+            # after Shaman is fixed.
+            #
+            munged_repo = ''
+            repo_lines = repo.splitlines()
+            for repo_line in repo_lines:
+                if repo_line.startswith('baseurl='):
+                    if repo_line.endswith('$basearch'):
```
kshtsk: I wonder if you can reuse method `rsstrip` from `teuthology.repo_utils`. Ideally to move it to teuthology.utils of course.

smithfarm: This code is only a temporary workaround. It will be removed once https://tracker.ceph.com/issues/44183 is fixed.

kshtsk:
@susebot run deploy

teuthology/packaging.py
kshtsk: > This text manipulation/conversion code should continue to work even
             # after Shaman is fixed

Means no-one will hurry to remove this code, as practice shows people start copy-pasting what ever code in the new patches, and they usually say they don't want to fix because they just copied from other place of codebase, so they may use this as an example. If it is hard to do, then ok

PR #1414, opened 2020-02-18
task/ceph_ansible: Make use of get_file python3 compatible

badone:
Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

### DIFF
--- a/teuthology/task/ceph_ansible.py
+++ b/teuthology/task/ceph_ansible.py
@@ -310,7 +310,7 @@ class CephAnsible(Task):
                               if role.startswith('osd')])
             if teuth_config.get('ceph_ansible') and \
                     self.ctx.machine_type in teuth_config['ceph_ansible']['has_lvm_scratch_disks']:
-                devices = get_file(remote, "/scratch_devs").split()
+                devices = get_file(remote, "/scratch_devs").decode().split()
                 vols = []
 
                 for dev in devices:


badone:
This should merge at the same time as https://github.com/ceph/teuthology/pull/1362 if possible

teuthology/task/ceph_ansible.py
```diff
@@ -310,7 +310,7 @@ def get_host_vars(self, remote):
                               if role.startswith('osd')])
             if teuth_config.get('ceph_ansible') and \
                     self.ctx.machine_type in teuth_config['ceph_ansible']['has_lvm_scratch_disks']:
-                devices = get_file(remote, "/scratch_devs").split()
+                devices = get_file(remote, "/scratch_devs").decode().split()
```
kshtsk: it is actually is not py2 backward compatible, it is better to use `six.ensure_str` or `teuthology.util.compat.stringify` method from the #1362
@badone no actions required, I am about to close this PR in favor of #1362 which I'll rebase and add corresponding fix.

badone: @kshtsk OK, thanks.

kshtsk: @badone unfortunately older release branches of ceph is still py2

badone: But won't teuthology be either 2 or 3? Can't we have a cutoff version for transition to python3 only? That way we don't have to maintain all the p2 code.

badone: @kshtsk my understanding was that was what py3-migration was all about?

tchaikov: @badone to move to python3, we could have a few options:

- switch teuthology to python3, and then switch all tests to python3. but this would leave us a huge window in which quite a few test fails, including those already in queue. as all qa suites in LTS branches are still python2. and those already scheduled are very likely still in python2
- switch qa tests to python3, and then switch teuthology to python3, again this would leave us a huge window in which quite a few test fails. because teuthology is still python2 before we do the switch and before that the tests in python3 just fail. for the same reason, it's difficult for us to do the hard switch.
- switch tests to python3 but keep them python2 compatible, and then switch teuthology to python3. it's a relatively safe option.

@kshtsk off-topic, probably a python2 compatible teuthology does not really make sense.

badone: @tchaikov why do the tests depend on the version of python teuthology is running on?

tchaikov: teuthology just **imports** the tests for running them if they are written as a python file under `qa/task`.

kshtsk: > @kshtsk off-topic, probably a python2 compatible teuthology does not really make sense.

@tchaikov it makes sense only for transition period until all branches are not py3 ready, as soon as we switch sepia to use py3 we can start and cleanup the py2 related code totally, but if you want to take responsibility for teuthology out of service window who could complain ;-)

tchaikov: @kshtsk i don't understand what the transition period means. if that stands for the period when we don't have all tests moved to python3, then we **cannot** switch teuthology to python3 before that. that's why i think a python2+python3 teuthology does not make sense.

kshtsk: > @kshtsk i don't understand what the transition period means. if that stands for the period when we don't have all tests moved to python3, then we **cannot** switch teuthology to python3 before that. that's why i think a python2+python3 teuthology does not make sense.

my understanding is the next:

0) sepia teuthology is working on py2
1) make master py2/py3 compatible, while keeping switch py3 teuthology merge not merged, try and merge as much as possible py2/py3 teuthology code
2) back port everything to older branches which we are going to support
3) try and switch sepia workers to py3
4) cleanup py2 code from everywhere as far as it reasonable in sense of time/effort balance

kshtsk:
@badone thanks for the submitting this PR, it is squashed to the #1362 so I'm closing it now.

PR #1419, opened 2020-03-10
Drop teuthology-coverage command

toabctl:
It seems to be unused and there was no change in the last 6 years. I
guess it's not used and not needed so drop it to remove the dead
code.


toabctl:
@kshtsk any comments here?

kshtsk:
> @kshtsk any comments here?

coverage stuff is out of my knowledge, I thought it is used by someone, at least I'm trying to not break it and keep all components "working"

toabctl:
@tchaikov do you know if coverage is still in use?

tchaikov:
@toabctl i think Josh is author of these scripts, and since he is fine with this. we can drop it. if there is any needs, we can always revive this feature.

PR #1425, opened 2020-03-17
valgrind: disable openssl's use of avx instructions

liewegas:


liewegas:
we're back to different valgrind failures:
http://pulpito.ceph.com/sage-2020-03-17_14:06:37-rados:verify-wip-sage-testing-2020-03-16-2107-distro-basic-smithi/

PR #1427, opened 2020-03-26
Unlock even when VM destroy fails to free up the paddles entry

vasukulkarni:
Unlock even when VM destroy fails to free up the paddles entry, on new reservation the VM is recreated fresh.

### DIFF
--- a/teuthology/lock/ops.py
+++ b/teuthology/lock/ops.py
@@ -203,7 +203,8 @@ def unlock_one(ctx, name, user, description=None):
     name = misc.canonicalize_hostname(name, user=None)
     if not teuthology.provision.destroy_if_vm(ctx, name, user, description):
         log.error('destroy failed for %s', name)
-        return False
+        # continue unlocking even when destroy fails for VM
+        # this will free up the paddles for new reservation
     request = dict(name=name, locked=False, locked_by=user,
                    description=description)
     uri = os.path.join(config.lock_server, 'nodes', name, 'lock', '')


zmc:
It might free up an entry in the database, but resources to start the VM will still be used. This would likely cause big problems sooner or later

vasukulkarni:
> It might free up an entry in the database, but resources to start the VM will still be used. This would likely cause big problems sooner or later

How would you suggest to remove the entry in case we know there has been some other issue, we dont have cli to clear up that entry.

zmc:
From my memory, when the destroy fails manual intervention has always been necessary. Once that was done, using teuthology-lock to unlock it was able to do the right thing.

kshtsk:
@vasukulkarni do you have a use case for reproducing this?

kshtsk:
I don't remember exactly, but I had similar issue when I entered incorrect credentials or lib cloud experiences a connection issue to openstack and teuthology continuously locks resources not even trying to allocate anything in openstack (like it failed to create a server).

kshtsk:
> > It might free up an entry in the database, but resources to start the VM will still be used. This would likely cause big problems sooner or later
> 
> How would you suggest to remove the entry in case we know there has been some other issue, we dont have cli to clear up that entry.

`teuthology-lock --unlock` was working fine to me, even if openstack does not have corresponding server running.

vasukulkarni:
> @vasukulkarni do you have a use case for reproducing this?

If you provide a image name that doesn't exist, it will fail to create and have the resource locked, without my PR i am not sure how it can get unlocked, since you can see it returns before the node is fully unlocked.

kshtsk:
> > @vasukulkarni do you have a use case for reproducing this?
> 
> If you provide a image name that doesn't exist, it will fail to create and have the resource locked, without my PR i am not sure how it can get unlocked, since you can see it returns before the node is fully unlocked.

Not sure I understand how you can provide non-existing image? Do you mean provide os-type, os-version pair with teuthology-suite or teuthology-lock for which there is no any image in openstack?

vasukulkarni:
> > > @vasukulkarni do you have a use case for reproducing this?
> > 
> > 
> > If you provide a image name that doesn't exist, it will fail to create and have the resource locked, without my PR i am not sure how it can get unlocked, since you can see it returns before the node is fully unlocked.
> 
> Not sure I understand how you can provide non-existing image? Do you mean provide os-type, os-version pair with teuthology-suite or teuthology-lock for which there is no any image in openstack?

I am closing this for now, I think one easy way is to have auth issues and the rest api will throw error in which case we will lock the node but not unlock it. I figured with queued jobs it can actually increase the overhead with my PR, since it will endup in constant loop of unlock/reserve, so better to debug manually and fix the issue.

PR #1428, opened 2020-03-26
suite: debug mgr by default

liewegas:
Like we do the mon and osd.

### DIFF
--- a/teuthology/suite/placeholder.py
+++ b/teuthology/suite/placeholder.py
@@ -63,10 +63,13 @@ dict_templ = {
                     'debug mon': 20,
                     'debug ms': 1,
                     'debug paxos': 20},
+                'mgr': {
+                    'debug mgr': 20,
+                    'debug ms': 1},
                 'osd': {
                     'debug filestore': 20,
                     'debug journal': 20,
-                    'debug ms': 1,
+                    'debug ms': 20,
                     'debug osd': 25
                 }
             },


teuthology/suite/placeholder.py
```diff
@@ -63,10 +63,13 @@ def _substitute(input_dict, values_dict):
                     'debug mon': 20,
                     'debug ms': 1,
                     'debug paxos': 20},
+                'mon': {
+                    'debug mgr': 20,
+		    'debug ms': 1},
```
kshtsk: replace tabs with spaces,

kshtsk:
hm, interesting:
```
teuthology/suite/placeholder.py:62:17: F601 dictionary key 'mon' repeated with different values
teuthology/suite/placeholder.py:66:17: F601 dictionary key 'mon' repeated with different values
```

teuthology/suite/placeholder.py
```diff
@@ -63,10 +63,13 @@ def _substitute(input_dict, values_dict):
                     'debug mon': 20,
                     'debug ms': 1,
                     'debug paxos': 20},
+                'mon': {
```
kshtsk: 'mgr'?

kshtsk: join with previous record?

vasukulkarni: should be mgr, i guess.

PR #1431, opened 2020-03-27
task/kernel.py: should pass args with name 2

kshtsk:
There is leftover, there should be remote.sh used instead of remote.run


tchaikov:
thanks for the fix!

PR #1433, opened 2020-03-30
requirements,setup: add ipy module for cephfs mount.py 

lxbsz:
The IPy module is a tool for handling of IPv4 and IPv6 addresses and
networks, the cephfs mount.py will use it to manage the IP addresses
to isolate the netns for each mount, which could let us get ride of
using ipmi command to hard shutdown the client node for some test cases.

Thanks @tchaikov very much for the help.

### DIFF
--- a/requirements3.txt
+++ b/requirements3.txt
@@ -99,6 +99,7 @@ virtualenv==15.1.0        # via tox
 warlock==1.2.0            # via python-glanceclient
 wrapt==1.10.10            # via debtcollector, positional, python-glanceclient
 xmltodict==0.12.0
+ipy==1.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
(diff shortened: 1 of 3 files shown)


requirements3.txt
```diff
@@ -99,6 +99,7 @@ virtualenv==15.1.0        # via tox
 warlock==1.2.0            # via python-glanceclient
 wrapt==1.10.10            # via debtcollector, positional, python-glanceclient
 xmltodict==0.12.0
+ipy=1.0
```
tchaikov:
```suggestion
ipy==1.0
```

lxbsz: Done, Thanks

lxbsz:
> I think this can be done in a single commit.

Okay, will fold them.
Thanks.

kshtsk:
@susebot run deploy

PR #1437, opened 2020-04-03
teuthology, tasks: fix dict related py3 compatibility issues

tchaikov:


teuthology/lock/ops.py
```diff
@@ -143,7 +143,7 @@ def lock_many(ctx, num, machine_type, user=None, description=None,
                             p.spawn(teuthology.provision.reimage, ctx,
                                     machine, machine_type)
                             reimaged[machine] = machines[machine]
-                reimaged = do_update_keys(reimaged.keys())[1]
+                reimaged = do_update_keys(list(reimaged.keys()))[1]
```
tchaikov, line 146: this should address the test failure of http://pulpito.ceph.com/kchai-2020-04-03_10:09:48-rados:perf-master-distro-basic-smithi/

kshtsk: hm....
```
2020-04-03T10:53:43.638 ERROR:teuthology.run_tasks:Saw exception from tasks.
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/run_tasks.py", line 89, in run_tasks
    manager.__enter__()
  File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/task/internal/lock_machines.py", line 78, in lock_machines
    os_version, arch)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/lock/ops.py", line 146, in lock_many
    reimaged = do_update_keys(reimaged.keys())[1]
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/lock/ops.py", line 286, in do_update_keys
    keys_dict = misc.ssh_keyscan(machines, _raise=_raise)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/misc.py", line 1106, in ssh_keyscan
    raise TypeError("'hostnames' must be a list")
TypeError: 'hostnames' must be a list
```
Maybe we can fix this in teuthology.misc... so it supports iterable, and probably dicts, don't know why it does not want to accept dicts.

tchaikov: @kshtsk i think i am less ambitious than you are. and i really want to limit the scope of the fix so we can migrate to python3 sooner. and then we can revisit the issues after that. but if you believe we do need to update the sanity check in `ssh_keyscan()`, please file a PR to address it. i will be more than happy to review it.

PR #1439, opened 2020-04-05
teuthology: s/cStringIO.StringIO/six.StringIO/

tchaikov:
to be py3 compatible. this addresses the failure of
```
2020-04-05T12:35:02.036 INFO:teuthology.run_tasks:Running task ssh_keys...
2020-04-05T12:35:02.050 ERROR:teuthology.run_tasks:Saw exception from tasks.
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/run_tasks.py", line 86, in run_tasks
    manager = run_one_task(taskname, ctx=ctx, config=config)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/run_tasks.py", line 64, in run_one_task
    task = get_task(taskname)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/run_tasks.py", line 28, in get_task
    module = _import('tasks', module_name, task_name, fail_on_import_error=True)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_py3/teuthology/run_tasks.py", line 52, in _import
    0,
ImportError: No module named 'tasks.ssh_keys'
```

### DIFF
--- a/teuthology/task/mpi.py
+++ b/teuthology/task/mpi.py
@@ -1,7 +1,6 @@
 """
 Start mpi processes (and allow commands to be run inside process)
 """
-from StringIO import StringIO
 import logging
 import re
 
@@ -18,7 +17,7 @@ def _check_mpi_version(remotes):
     """
     versions = set()
     for remote in remotes:
-        version_str = remote.run(args=["mpiexec", "--version"], stdout=StringIO()).stdout.getvalue()
+        version_str = remote.sh("mpiexec --version")
         try:
             version = re.search("^\s+Version:\s+(.+)$", version_str, re.MULTILINE).group(1)
         except AttributeError:
--- a/teuthology/orchestra/test/integration/test_integration.py
+++ b/teuthology/orchestra/test/integration/test_integration.py
@@ -1,7 +1,7 @@
 from teuthology.orchestra import monkey
 monkey.patch_all()
 
-from cStringIO import StringIO
+from six import StringIO
 
 import os
 from teuthology.orchestra import connection, remote, run
--- a/teuthology/task/hadoop.py
+++ b/teuthology/task/hadoop.py
@@ -1,4 +1,4 @@
-from cStringIO import StringIO
+from six import StringIO
 import contextlib
 import logging
 from teuthology import misc as teuthology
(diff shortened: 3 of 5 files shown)


tchaikov:
see http://pulpito.ceph.com/kchai-2020-04-05_11:54:01-fs-master-distro-basic-smithi/4926427/

teuthology/task/mpi.py
```diff
@@ -1,12 +1,12 @@
 """
 Start mpi processes (and allow commands to be run inside process)
 """
-from StringIO import StringIO
 import logging
 import re
 
 from teuthology import misc as teuthology
 from six import string_types as basestring
+from six import StringIO
```
kshtsk, line 9: The String is mentioned only once in this file, could you drop it at all?
Just replace this:
```
version_str = remote.run(args=["mpiexec", "--version"], stdout=StringIO()).stdout.getvalue()
```
with
```
version_str = remote.sh("mpiexec --version")
```

tchaikov: @kshtsk i fixed a couple of regressions introduced by s/run/sh/ . and i'd like to avoid doing this again if possible. we can always revisit these `run()` calls and refactor them after the py3 migration. but at this moment, i'd prefer having a minimum change which only addresses the py3 incompatibility.

kshtsk: I guess it was that typo issues, same can appear with switching `six.StringIO` stuff.

kshtsk: @tchaikov still, could you please make an exclusion for me here, there is no a reason to use StringIO here? thanks

tchaikov: okay, will do.

kshtsk:
@susebot run deploy

teuthology/task/mpi.py
tchaikov: > I guess it was that typo issues, same can appear with switching `six.StringIO` stuff.

i disagree.

tchaikov: updated and repushed.

kshtsk:
``` teuthology: s/sh/run/

to avoid using StringIO directly
```
It vise-versa `s/run/sh`

PR #1443, opened 2020-04-07
suite: Implement --sleep-before-teardown option

kshtsk:
This is another try to merge the useful debugging feature which is supposed to be not breaking 'parallel' and 'sequential' tasks.
The `sleep-before-teardown` only works on the front line of tasks not going deep into parallel and sequential tasks and does not give a sleep within them correspondingly, allowing to finish them correctly before unwinding the main stack of tasks.

Usage hints:
- option can be used with either `teuthology-suite` or `teuthology-openstack`
- use `teuthology-kill` to cleanup the job or run to interrupt the sleep and free resources.

Changes:
- [x] add --non-interactive option
- [x] add force using --limit 1 if it was omitted when --sleep-before-teardown is given  
- [x] add interactive confirmation if --sleep-before-teardown is given and --limit is more than 4
- [x] add --limit check skipping for --sleep-before-teardown when --non-interactive mode is enabled

Notes:
   I had to move job limit check out of the collect_jobs method and newest check loop to avoid asking the same confirmation question over and over again.

### DIFF
--- a/scripts/openstack.py
+++ b/scripts/openstack.py
@@ -187,6 +187,11 @@ def get_suite_parser():
         help=("Use tasks and suite definition in this repository"),
         default=os.getenv('TEUTH_SUITE_REPO', 'https://github.com/ceph/ceph'),
     )
+    parser.add_argument(
+        '--sleep-before-teardown',
+        help='Number of seconds to sleep before tearing down the target VMs',
+        default=0
+    )
     return parser
 
 def get_openstack_parser():
--- a/scripts/openstack.py
+++ b/scripts/openstack.py
@@ -189,7 +189,7 @@ def get_suite_parser():
     )
     parser.add_argument(
         '--sleep-before-teardown',
-        help='Number of seconds to sleep before tearing down the target VMs',
+        help='Number of seconds to sleep before the teardown',
         default=0
     )
     return parser
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -284,6 +284,8 @@ class Run(object):
             job_config.email = self.args.email
         if self.args.owner:
             job_config.owner = self.args.owner
+        if self.args.sleep_before_teardown:
+            job_config.sleep_before_teardown = int(self.args.sleep_before_teardown)
         return job_config
 
     def build_base_args(self):
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -6,6 +6,9 @@ import re
 import time
 import yaml
 
+from six.moves import input
+from humanfriendly import format_timespan
+
 from datetime import datetime
 from tempfile import NamedTemporaryFile
 
@@ -356,14 +359,13 @@ class Run(object):
         if num_jobs:
             self.write_result()
 
-    def collect_jobs(self, arch, configs, newest=False):
+    def collect_jobs(self, arch, configs, newest=False, limit=0):
         jobs_to_schedule = []
         jobs_missing_packages = []
         for description, fragment_paths in configs:
             base_frag_paths = [
                 util.strip_fragment_path(x) for x in fragment_paths
             ]
-            limit = self.args.limit or 0
             if limit > 0 and len(jobs_to_schedule) >= limit:
                 log.info(
                     'Stopped after {limit} jobs due to --limit={limit}'.format(
@@ -518,13 +520,46 @@ class Run(object):
         ).name
         self.base_yaml_paths.insert(0, base_yaml_path)
 
+        # compute job limit in respect of --sleep-before-teardown
+        job_limit = self.args.limit or 0
+        sleep_before_teardown = int(self.args.sleep_before_teardown or 0)
+        if sleep_before_teardown:
+            if job_limit == 0:
+                log.warning('The --sleep-before-teardown option was provided: '
+                            'only 1 job will be scheduled. '
+                            'Use --limit to run more jobs')
+                # give user a moment to read this warning
+                time.sleep(5)
+                job_limit = 1
+            elif self.args.non_interactive:
+                log.warning(
+                    'The --sleep-before-teardown option is active. '
+                    'There will be a maximum {} jobs running '
+                    'which will fall asleep for {}'
+                    .format(job_limit, format_timespan(sleep_before_teardown)))
+            elif job_limit > 4:
+                are_you_insane=(
+                    'There are {total} configs and {maximum} job limit is used. '
+                    'Do you really want to lock all machines needed for '
+                    'this run for {that_long}? (y/N):'
+                    .format(
+                        that_long=format_timespan(sleep_before_teardown),
+                        total=len(configs),
+                        maximum=job_limit))
+                while True:
+                    insane=(input(are_you_insane) or 'n').lower()
+                    if insane == 'y':
+                        break
+                    elif insane == 'n':
+                        exit(0)
+
         # if newest, do this until there are no missing packages
         # if not, do it once
         backtrack = 0
         limit = self.args.newest
         while backtrack <= limit:
             jobs_missing_packages, jobs_to_schedule = \
-                self.collect_jobs(arch, configs, self.args.newest)
+                self.collect_jobs(arch, configs, self.args.newest, job_limit)
             if jobs_missing_packages and self.args.newest:
                 new_sha1 = \
                     util.find_git_parent('ceph', self.base_config.sha1)
--- a/requirements2.txt
+++ b/requirements2.txt
@@ -38,7 +38,9 @@ futures==3.2.0            # via s3transfer
 gevent==1.4.0
 greenlet==0.4.15          # via gevent
 httplib2==0.10.3
+humanfriendly==8.1
 idna==2.5                 # via requests
+ipy==1.0
 ipaddress==1.0.18         # via cryptography
 iso8601==0.1.11           # via keystoneauth1, oslo.utils, python-neutronclient, python-novaclient
 jinja2==2.10.1            # via ansible
@@ -107,7 +109,6 @@ virtualenv==15.1.0        # via tox
 warlock==1.2.0            # via python-glanceclient
 wrapt==1.10.10            # via debtcollector, positional, python-glanceclient
 xmltodict==0.12.0
-ipy==1.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
--- a/requirements3.txt
+++ b/requirements3.txt
@@ -34,7 +34,9 @@ docutils==0.14            # via botocore
 gevent==1.4.0
 greenlet==0.4.15          # via gevent
 httplib2==0.10.3
+humanfriendly==8.1
 idna==2.5                 # via requests
+ipy==1.0
 iso8601==0.1.11           # via keystoneauth1, oslo.utils, python-neutronclient, python-novaclient
 jinja2==2.10.1            # via ansible
 jmespath==0.9.4           # via boto3, botocore
@@ -99,7 +101,6 @@ virtualenv==15.1.0        # via tox
 warlock==1.2.0            # via python-glanceclient
 wrapt==1.10.10            # via debtcollector, positional, python-glanceclient
 xmltodict==0.12.0
-ipy==1.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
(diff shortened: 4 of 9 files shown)


kshtsk:
@susebot run deploy

kshtsk:
retest this please

kshtsk:
This is another try to merge the useful debugging feature which is supposed to be not breaking 'parallel' and 'sequential' tasks.
The sleep-before-teardown only works on the front line of tasks not going deep into parallel and sequential tasks and does not give a sleep within them correspondingly allowing to finish correctly, before unwinding the _main_ task stack.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
@susebot run deploy

scripts/openstack.py
```diff
@@ -187,6 +187,11 @@ def get_suite_parser():
         help=("Use tasks and suite definition in this repository"),
         default=os.getenv('TEUTH_SUITE_REPO', 'https://github.com/ceph/ceph'),
     )
+    parser.add_argument(
+        '--sleep-before-teardown',
+        help='Number of seconds to sleep before tearing down the target VMs',
```
vasukulkarni, line 192: s/before tearing down ../before the teardown of first task in the job/, 

I think we should really add the warning to limit on how many jobs to be run when using this option, As it can hold quite a bit of hardware if not used properly. A warning when the CLI is invoked is needed.

kshtsk: Hm.... not sure I understood the difference from English point of view, but this made me to turn attention to 'target VMs' which is definitely must be change to something appropriate.

smithfarm:
:+1: to @vasukulkarni 's idea to have the feature warn the user.

In fact, it would be worth considering to raise an exception when scheduling a run with over, say, 10 jobs when this option is active.

scripts/openstack.py
kshtsk: @smithfarm I wonder, if should rename the feature to something like --sleep-before-unwinding-tasks (confused). I believe original feature name was call after teuthology-openstack kung-fu. It is okay to me still have a sleep-before-teardown since there is no teardown operation defined in the native teuthology, would this be obvious what does the 'teardown' mean to others?

kshtsk:
> +1 to @vasukulkarni 's idea to have the feature warn the user.
> 
> In fact, it would be worth considering to raise an exception when scheduling a run with over, say, 10 jobs when this option is active.

Wouldn't they be killed by the timeout?

scripts/openstack.py
smithfarm: `--sleep-before-unwind` would be shorter, but isn't this splitting hairs?

To put it another way, what is wrong with "sleep before teardown"? Is the sleep not kicking in before teardown here? (There are only three possibilities: before, at the same time as, or after.)

smithfarm:
> Wouldn't they be killed by the timeout?

Which timeout do you have in mind here?

kshtsk:
> > Wouldn't they be killed by the timeout?
> 
> Which timeout do you have in mind here?

Global job run execution time, don't remember the exact name.

gregsfortytwo:
> > > Wouldn't they be killed by the timeout?
> > 
> > 
> > Which timeout do you have in mind here?
> 
> Global job run execution time, don't remember the exact name.

No the timeouts only apply to individual jobs; there's no total suite timeout limit.

scripts/openstack.py
kshtsk: I'm okay with 'teardown', nevertheless there is no definition of _teardown_ in teuthology, just wondering, since it is not clear what teardown does it mean, if it means target unlock, then it is not right before teardown. As for teuthology-openstack teardown also means teuthology server teardown.

smithfarm:
@gregsfortytwo I think we need to be careful with this, because people will do:

```
teuthology-suite --suite rados --sleep-before-teardown 99999999
```

and lock up a large number of machines for a very long time

kshtsk:
> @gregsfortytwo I think we need to be careful with this, because people will do:
> 
> ```
> teuthology-suite --suite rados --sleep-before-teardown 99999999
> ```
> 
> and lock up a large number of machines for a very long time

Not sure that warning will help a much, since people will ignore it, because we have very noisy logs.

scripts/openstack.py
smithfarm: > there is no definition of teardown in teuthology

This is not a true statement. Teuthology tasks written in the "new style" (using a class inheriting from `Task`)  have a "teardown" phase: https://github.com/ceph/teuthology/blob/master/teuthology/task/__init__.py#L111-L117

Let me repeat: I don't think we should split hairs here. The name "sleep before teardown" is sufficiently clear and easy to understand.

smithfarm:
@kshtsk I think what @vasukulkarni (and certainly what I) had in mind was not to write a warning to the log, but rather **interactively** ask the user if he really wants to potentially lock up lots of machines for a long time.

The warning could translate the number of seconds given by the user into years, months, days, hours, minutes, seconds. So, for example, the user gives `--sleep-before-teardown 99999999` and we ask "Do you really want to lock all machines needed for this run for Y years, M months, D days, H hours, M minutes and S seconds? (y/N)"

In addition to this, to be completely safe, once we know how many jobs are in the run, we can check if it is over a certain number (like 10) and raise an exception in that case, if "sleep before teardown" is in effect.

What do you think?

kshtsk:
The problem is that we can't understand how much jobs will be run before they actually scheduled, this is by teuthology design. The corresponding code will be more complicated. As an option, what if to only allow use --sleep-before-teardown in combination with the --limit?

kshtsk:
> The warning could translate the number of seconds given by the user into years, months, days, hours, minutes, seconds. So, for example, the user gives `--sleep-before-teardown 99999999` and we ask "Do you really want to lock all machines needed for this run for Y years, M months, D days, H hours, M minutes and S seconds? (y/N)"
> 
Nevertheless I like questions like that :-)
I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.

> In addition to this, to be completely safe, once we know how many jobs are in the run, we can check if it is over a certain number (like 10) and raise an exception in that case, if "sleep before teardown" is in effect.
> 
> What do you think?

What about forcing `--limit` parameter to the 1, if --sleep-before-teardown is called? Or/and additionally have `--limit-sleeping-jobs 1` which will override the `--limit` if its value greater?

smithfarm:
> I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.

`--non-interactive` ?

smithfarm:
How about:

1. if `--limit` not given explicitly, set it to 1
2. if `--limit` given explicitly and it's greater than 10, error

kshtsk:
> > I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.
> 
> `--non-interactive` ?



> How about:
> 
>     1. if `--limit` not given explicitly, set it to 1
> 
>     2. if `--limit` given explicitly and it's greater than 10, error

I feel like okay, but why 10? What if I want to run 11? :-D

kshtsk:
What about:
- if `--limit` not given, then set it 1
- if `--limit` is given more than 3, ask interactively for confirmation.

vasukulkarni:
> > > I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.
> > 
> > 
> > `--non-interactive` ?
> 
> > How about:
> > ```
> > 1. if `--limit` not given explicitly, set it to 1
> > 
> > 2. if `--limit` given explicitly and it's greater than 10, error
> > ```
> 
> I feel like okay, but why 10? What if I want to run 11? :-D

Personally I feel there is no use case to run and hold many jobs (since the purpose is to debug live) ?

kshtsk:
> > I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.
> 
> `--non-interactive` ?

We do not have such argument yet, do you mean add it withing this PR?

smithfarm:
> > > I think this is not as good as it could, especially bad when you want to automate scheduling, example, from jenkins.
> > 
> > 
> > `--non-interactive` ?
> 
> We do not have such argument yet, do you mean add it withing this PR?

Yes. It would enable us to implement the proposed interactive warning without causing problems for Jenkins.

This is a technique widely used by command-line tools.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@smithfarm @vasukulkarni I think I've addressed your requests, could you please take another look?

teuthology/suite/run.py
```diff
@@ -516,13 +520,47 @@ def schedule_suite(self):
...
+            elif self.args.non_interactive:
+                log.warning(
+                    'The --sleep-before-teardown is used and there will be maximum '
+                    '{} jobs running which will fall asleep for {}'
+                    .format(job_limit, format_timespan(sleep_before_teardown)))
+            elif job_limit > 4:
+                are_you_insane=(
+                    'There are {total} configs and {maximum} job limit is used. '
+                    'Do you really want to lock all machines needed for '
+                    'this run for {that_long}? Are you insane? (y/N):'
```
smithfarm, line 543: I suggest to drop the additional "Are you insane?" question. The first question is sufficient, no?

teuthology/suite/run.py
```diff
@@ -516,13 +520,47 @@ def schedule_suite(self):
...
+                    .format(
+                        that_long=format_timespan(sleep_before_teardown),
+                        total=len(configs),
+                        maximum=job_limit))
+                while True:
+                    insane=(input(are_you_insane) or 'n').lower()
+                    if insane == 'y':
+                        print("Here you go!")
```
smithfarm, line 551: Could we drop this print statement? It doesn't do anything useful.

teuthology/suite/run.py
kshtsk: :-)

teuthology/suite/run.py
```diff
@@ -516,13 +520,47 @@ def schedule_suite(self):
...
+                        break
+                    elif insane == 'n':
+                        print("Good choice, I'm proud of you")
```
smithfarm, line 554: Again, please drop this print.

kshtsk:
@gregsfortytwo @vasukulkarni after fixes applied to the patch your review is expectedly  disappeared, could you please take another look?

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/suite/run.py
```diff
@@ -516,13 +520,45 @@ def schedule_suite(self):
         ).name
         self.base_yaml_paths.insert(0, base_yaml_path)
 
+        # compute job limit in respect of --sleep-before-teardown
+        job_limit = self.args.limit or 0
+        sleep_before_teardown = int(self.args.sleep_before_teardown or 0)
+        if sleep_before_teardown:
+            if job_limit == 0:
+                log.warning('The --sleep-before-teardown is provided, '
```
smithfarm, line 528: Please change to `The --sleep-before-teardown option was provided: only 1 job will be . . .`

- insert the word "option"
- s/is/was/
- s/,/:/

teuthology/suite/run.py
```diff
@@ -516,13 +520,45 @@ def schedule_suite(self):
...
+                            'only 1 job will be scheduled. '
+                            'Use --limit to run more jobs')
+                # give user a moment to read this warning
+                time.sleep(5)
+                job_limit = 1
+            elif self.args.non_interactive:
+                log.warning(
+                    'The --sleep-before-teardown is used and there will be maximum '
```
smithfarm, line 536: Could it please be `The --sleep-before-teardown option is active. There will be a maximum of {} jobs . . .`

i.e.,

* insert "option"
* s/used/active/
* s/and there/. There/
* insert "a" before "maximum"

kshtsk: As you wish

teuthology/suite/run.py
kshtsk: aye aye

kshtsk:
@smithfarm here you are, new push with your suggestions

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
okay, failure reason is unrelated to the change, it is because of slow osd on the openstack targets.

PR #1445, opened 2020-04-13
teuthology: pass integer as "tries" to safe_while()

tchaikov:
in Python3, `a / b` could return a float, and `safe_while()` uses
`itertools.islice()` under the hood, where `stop` should be None or an
integer. so let's use `a // b` instead.


tchaikov:
this change should address the failure of http://pulpito.ceph.com/kchai-2020-04-13_01:52:52-rados-wip-kefu-testing-2020-04-13-0113-distro-basic-smithi/4948174/

kshtsk:
@susebot run deploy

PR #1448, opened 2020-04-15
py2: sleep before teardown

kshtsk:


kshtsk:
@tchaikov is this how we should backport PRs to py2?

tchaikov:
hi @kshtsk i 

- edit the title of this PR, to add "py2:" before it.
- created "py2" milestone and added this PR to it, to differentiate PR targeting master from "py2" branch.

as you mentioned, py2 branch will be "minimal", and i think it is now in its maintenance mode, ideally only bug fixes will goto that branch.

presumably, `--sleep-before-teardown` option is a feature. but i guess you have good reason to add it to py2 branches as well for tracking bugs.

tchaikov:
@kshtsk are you actively testing LTS branches? i am asking because currently the only guy who is actively testing LTS branches is @yuriw , so i want to minimize the risk that he run into any issue because of the backports. another consumer of py2 branch is cron job.

kshtsk:
@tchaikov depends on what you mean "actively". We still running suse downstream tests in py2 development. Also it is possible to run deploy test on py2 based branches. Good thing that back ports are already tested on py3 tests. If you run ceph named branches, that is another case. Also I would like to keep py2 and py3 be as close as possible, in sense of features and documentation.

tchaikov:
> @tchaikov depends on what you mean "actively". We still running suse downstream tests in py2 development. Also it is possible to run deploy test on py2 based branches. Good thing that back 

that explains. i thought "py2" branch should be in maintenance mode. but since you guys are still using it for testing, it's relatively safe for backporting features.

> ports are already tested on py3 tests. If you run ceph named branches, that is another case. Also I would like to keep py2 and py3 be as close as possible, in sense of features and documentation.

okay, as long as you are still using it. that's a relief. my major concern was we were backporting features and unlike the changes in master, the backport changes were not well tested.

PR #1449, opened 2020-04-16
lock/cli: fix broken teuthology-lock --summary for py3

kshtsk:
Convert None type to empty str in sorted key lambda
The usage of .format method is also adjusted for empty
machine type values

Fixes: https://tracker.ceph.com/issues/45105


tchaikov:
thanks for fixing this, Kyr!

PR #1451, opened 2020-04-19
requirements: Bump cffi version to 1.14.0

toabctl:
python 3.8 needs a newer cffi version. So bump it. Otherwise, calling
"./bootstrap" with python3.8 will fail.
This is also done for the python2 requirements to be in sync.

Fixes: https://tracker.ceph.com/issues/45132


tchaikov:
@susebot run deploy

PR #1453, opened 2020-04-20
travis: Use bionic, no sudo and python 3.6

toabctl:
- switch to Ubuntu Bionic to have a recent version for testing
- Do not use sudo. That will switch from VMs to containers which
  improves build speed on travis a lot
- use python3.6 . python2 is EOL


toabctl:
Travis is disabled again. Seems to be super slow...

PR #1459, opened 2020-04-23
orchestra/remote.py: to be compatible with py3

lxbsz:
```
ERROR: test_filelock (tasks.cephfs.test_client_recovery.TestClientRecovery)
----------------------------------------------------------------------
  File "/data/ceph/qa/tasks/cephfs/test_client_recovery.py", line 387, in test_filelock
    flockable = self._is_flockable()
  File "/data/ceph/qa/tasks/cephfs/test_client_recovery.py", line 361, in _is_flockable
    a_version_str = get_package_version(self.mount_a.client_remote, "fuse")
  File "/data/teuthology/teuthology/packaging.py", line 354, in get_package_version
    if remote.os.package_type == "deb":
  File "/data/teuthology/teuthology/orchestra/remote.py", line 452, in os
    self._os = OS.from_os_release(os_release)
  File "/data/teuthology/teuthology/orchestra/opsys.py", line 161, in from_os_release
    name = cls._get_value(str_, 'ID').lower()
  File "/data/teuthology/teuthology/orchestra/opsys.py", line 204, in _get_value
    match = re.search(regex, str_, flags=re.M)
  File "/usr/lib64/python3.6/re.py", line 182, in search
    return _compile(pattern, flags).search(string)
TypeError: cannot use a string pattern on a bytes-like object
```

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -17,6 +17,7 @@ import time
 import re
 import logging
 from io import BytesIO
+from io import StringIO
 import os
 import pwd
 import tempfile
@@ -448,7 +449,8 @@ class Remote(object):
     def os(self):
         if not hasattr(self, '_os'):
             try:
-                os_release = self.sh('cat /etc/os-release').strip()
+                os_release = self.sh('cat /etc/os-release',
+                                     stdout=StringIO()).strip()
                 self._os = OS.from_os_release(os_release)
                 return self._os
             except CommandFailedError:


teuthology/orchestra/remote.py
```diff
@@ -448,7 +449,8 @@ def get_tar_stream(self, path, sudo=False):
     def os(self):
         if not hasattr(self, '_os'):
             try:
-                os_release = self.sh('cat /etc/os-release').strip()
+                os_release = self.sh('cat /etc/os-release',
+                                     stdout=StringIO()).strip()
```
kshtsk, line 453: what???
remote.sh always returns str.

lxbsz: Yeah, I saw the remote.sh code copied from teuthology, I thought the should be the same, but it is not.
In ceph/qa/vstart_runner it will return what it is.
Let me try to fix it in vstart_runner and to see whether is has any regression for the change.
Thanks.

kshtsk:
@lxbsz could you please describe your environment, we never saw this in many runs based on py3.

lxbsz:
> @lxbsz could you please describe your environment, we never saw this in many runs based on py3.

`python3 ../qa/tasks/vstart_runner.py tasks.cephfs.test_client_recovery.TestClientRecovery.test_filelock`

And together with https://github.com/ceph/ceph/pull/33769.

teuthology/orchestra/remote.py
lxbsz: Tried to make the remote.sh to return str always in qa/tasks/vstart_runner.py and it works and didn't see any regression for other test cases.
I will fix it in ceph/qa code.

lxbsz:
@kshtsk Closing it, and will fix in ceph code.

kshtsk:
@lxbsz really this is very interesting... how is ever possible, I guess something wrong with your local environment...

lxbsz:
> @lxbsz really this is very interesting... how is ever possible, I guess something wrong with your local environment...
@kshtsk 
I will fix it in ceph instead, the teuthology code should be fine here.
Thanks :-)

teuthology/orchestra/remote.py
kshtsk: @lxbsz which is that branch? why do we have a duplication of it vstart? :-)

lxbsz: The vstart_runner.py will have its own remote class, which is a little different from the teuthology's. This is for the local test only for cephfs, more detail please see: https://docs.ceph.com/docs/master/dev/developer_guide/running-tests-locally/.

lxbsz: The local test is a quick test case, I am new in this and I think many code in vstart_runner.py are duplicated to teuthology's we may try to reuse it.

kshtsk: ah... this vstart black magic... I wonder why they duplicating code from teuthology? maybe it is time to get rid of it?

lxbsz: I have no much idea about the history of the vstart_runner, and checked the code it seems we could reuse many code from teuthology, but not all.
For us we usually will do the local test first and then the integration test online, which is more efficient.

PR #1467, opened 2020-05-08
orchestra/remote: revert reconnect code to run

kshtsk:

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -160,22 +160,23 @@ class Remote(object):
 
     @property
     def is_online(self):
-        if self.ssh is None:
+        """
+        This function does not give a comprehensive answer
+        if the connection is still alive, for this purpose
+        the one must use `ensure_online` method.
+        """
+        return self.is_active()
+
+    def is_active(self):
+        if self.ssh and \
+           self.ssh.get_transport() and \
+           self.ssh.get_transport().is_active():
+               return True
+        else:
             return False
-        if self.ssh.get_transport() is None:
-            return False
-        try:
-            self._runner(args="true", client=self.ssh, name=self.shortname)
-        except Exception:
-            return False
-        return self.ssh.get_transport().is_active()
 
     def ensure_online(self):
-        if self.is_online:
-            return
-        self.connect()
-        if not self.is_online:
-            raise Exception('unable to connect')
+        self.run(args="true")
 
     @property
     def system_type(self):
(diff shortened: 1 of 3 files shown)


teuthology/orchestra/remote.py
```diff
@@ -160,22 +160,23 @@ def shortname(self):
...
-            self._runner(args="true", client=self.ssh, name=self.shortname)
-        except Exception:
+        """
+        This function does not give a comprehensive answer
+        if the connection is still alive, for this purpose
+        the one must use `ensure_online` method.
+        """
+        return self.is_active()
+
+    def is_active(self):
```
tchaikov, line 170: why do we need two `is_online` ?

tchaikov: BTW, i think is_online is a better name than is_active. personally, i like `is_reachable` even better.

kshtsk: yeah... it is embarrassing me two... just need to be sure which signature to leave...

kshtsk: I would get rid of `is_online` because looks like it is not possible to make sure it is online, so more correctly to leave `is_active`?

tchaikov: an online remote machine sounds more correct than an active remote box . and again, online is objective. but reachable is more subjective. i'd say we cannot tell if a remote is online/active unless we can be sure about the networking and stuff work. and as you put, 

>  looks like it is not possible to make sure it is online

teuthology/orchestra/remote.py
```diff
@@ -199,8 +200,21 @@ def run(self, **kwargs):
 
         TODO refactor to move run.run here?
         """
-        self.ensure_online()
-        r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
+        if not self.is_active():
+            if not self.reconnect():
+                raise Exception('Unable to reconnect')
+        tries = 2
```
tchaikov, line 206: this is funny. why do you think we should try twice?

kshtsk: @tchaikov on my "testing stand" if I have an open connection and reboot a remote host in 100% cases next try to call self.ssh.get_transport().is_active() returns True, and correspondingly a try to call paramiko execute method on existing session throws an EOFError exception. So this two tries to correctly handle this situation:
1) try run
2) if it fails, and node is online, this means it was not able to actually an issue the command on the remote host, then we need just reestablish a connection and try again
3) if the second try is failed, then their are some another issue, and let the user to handle it as required, like in reboot method, wait and try again.

I am actually not sure that this would work well, maybe I need to think about finding a use case when the first call of _run could possible produce EOFError when the command actually started to execute and then failed

PR #1468, opened 2020-05-08
teuthology/orchestra: use reconnect() in ensure_online()

tchaikov:
* to avoid testing the connection by sending "true" for every command
  sent to remote.
* to reset the connection in ensure_online() to make sure a new
  connection is used

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -917,7 +917,7 @@ def reboot(node, timeout=300, interval=30):
     reboot_start_time = time.time()
     while time.time() - reboot_start_time < timeout:
         time.sleep(interval)
-        if node.is_online or node.reconnect():
+        if node.is_online(do_test=True) or node.reconnect():
             return
     raise RuntimeError(
         "{host} did not come up after reboot within {time}s".format(
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -98,7 +98,7 @@ class Remote(object):
         log.info("Trying to reconnect to host")
         try:
             self.connect(timeout=timeout, context='reconnect')
-            return self.is_online
+            return self.is_online(do_test=True)
         except Exception as e:
             log.debug(e)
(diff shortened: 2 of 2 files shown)


teuthology/misc.py
```diff
@@ -917,7 +917,7 @@ def reboot(node, timeout=300, interval=30):
     reboot_start_time = time.time()
     while time.time() - reboot_start_time < timeout:
         time.sleep(interval)
-        if node.is_online or node.reconnect():
+        if node.is_online(do_test=True) or node.reconnect():
```
kshtsk, line 920: Not sure that is method should do all this checks. Maybe have `remote.ensure_online` do the next:
```
    def ensure_online(self):
        self.run(args="true")
```
since the _run_ is supposed to call reconnect if socket connection failed,
and this loop will look like:
```
 while time.time() - reboot_start_time < timeout:
        time.sleep(interval)
        try:
            node.ensure_online()
            break
        except:
            log.......
```

kshtsk: see #1467

batrick:
@tchaikov I can't seem to schedule a run with this PR. Here's what I get:

http://pulpito.ceph.com/pdonnell-2020-05-08_18:41:06-multimds-master-distro-basic-smithi/

I scheduled with:

`$ teuthology-suite --machine-type smithi --email pdonnell@redhat.com -p 49 --ceph master --ceph-repo https://github.com/ceph/ceph.git -S fe8a93e4f4214467d6288e462ee203baa62fe4d6 --suite-repo https://github.com/ceph/ceph.git -s multimds --filter cephfs_test_snapshots --limit 10 -t pr1468`

kshtsk:
so... this might be closed?

tchaikov:
> so... this might be closed?

okay, i am closing. but i still don't like the repeated logic for checking if `self.ssh` is working. see `is_online()` and `run()`.

PR #1470, opened 2020-05-11
fix some old flake8 issues

vasukulkarni:
For now fixes only flake8 issues seen in new versions, we could as well get rid of flake8 check with py2 but thats up for another discussion.

As seen during flake8 debug

```
flake8 installed: DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021
```

### DIFF
--- a/scripts/queue.py
+++ b/scripts/queue.py
@@ -28,7 +28,7 @@ optional arguments:
   -p, --pause SECONDS   Pause queues for a number of seconds. A value of 0
                         will unpause. If -m is passed, pause that queue,
                         otherwise pause all queues.
-""".format(archive_base=teuthology.config.config.archive_base)
+"""
 
 
 def main():
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1063,7 +1063,7 @@ def get_valgrind_args(testdir, name, preamble, v):
         'env', 'OPENSSL_ia32cap=~0x1000000000000000',
     ])
 
-    val_path = '/var/log/ceph/valgrind'.format(tdir=testdir)
+    val_path = '/var/log/ceph/valgrind'
     if '--tool=memcheck' in v or '--tool=helgrind' in v:
         extra_args = [
             'valgrind',
--- a/teuthology/orchestra/console.py
+++ b/teuthology/orchestra/console.py
@@ -274,8 +274,7 @@ class PhysicalConsole(RemoteConsole):
         child = self._pexpect_spawn_ipmi('power on')
         child.expect('Chassis Power Control: Up/On', timeout=self.timeout)
         self._wait_for_login()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))
 
     def spawn_sol_log(self, dest_path):
         """
@@ -385,5 +384,4 @@ class VirtualConsole(RemoteConsole):
         self.vm_domain.info().destroy()
         time.sleep(interval)
         self.vm_domain.info().create()
-        log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+        log.info('Power off for {i} seconds completed'.format(i=interval))


vasukulkarni:
Unrelated flake8 issue which I am seeing in another PR as well

"""
flake8 run-test: commands[0] | flake8 --select=F,E9 teuthology scripts
[2537] /home/jenkins-build/build/workspace/teuthology-pull-requests$ /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/flake8/bin/flake8 --select=F,E9 teuthology scripts
teuthology/orchestra/connection.py:121:68: E999 SyntaxError: invalid syntax
ERROR: InvocationError for command /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/flake8/bin/flake8 --select=F,E9 teuthology scripts (exited with code 1)
"""

vasukulkarni:
A new version was released today, so that seems to the cause  https://pypi.org/project/flake8/#history

kshtsk:
Please don't merge it, I still want to check py2 compatibility on teuthology master branch.

vasukulkarni:
> Please don't merge it, I still want to check py2 compatibility on teuthology master branch.

This is flake8 check and also due to new flake8 pypi release we are now seeing new issue.

teuthology/orchestra/console.py
```diff
@@ -275,7 +275,7 @@ def power_off_for_interval(self, interval=30):
         child.expect('Chassis Power Control: Up/On', timeout=self.timeout)
         self._wait_for_login()
         log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+            s=self.shortname, i=interval)) # noqa
```
kshtsk, line 278: The should be s=self.shortname remove because it is not used.

teuthology/orchestra/console.py
```diff
@@ -386,4 +386,4 @@ def power_off_for_interval(self, interval=30):
         time.sleep(interval)
         self.vm_domain.info().create()
         log.info('Power off for {i} seconds completed'.format(
-            s=self.shortname, i=interval))
+            s=self.shortname, i=interval)) # noqa
```
kshtsk, line 389: Same here.

scripts/queue.py
```diff
@@ -28,7 +28,7 @@
   -p, --pause SECONDS   Pause queues for a number of seconds. A value of 0
                         will unpause. If -m is passed, pause that queue,
                         otherwise pause all queues.
-""".format(archive_base=teuthology.config.config.archive_base)
+""".format(archive_base=teuthology.config.config.archive_base) # noqa
```
kshtsk, line 31: the '{archive_base}' has been removed at some point, so there is no sense to use it, I guess there new flake8 version released and they raise severity of this warning, instead of `noqa` it is better to just remove `.format`

teuthology/misc.py
```diff
@@ -1063,7 +1063,7 @@ def get_valgrind_args(testdir, name, preamble, v):
         'env', 'OPENSSL_ia32cap=~0x1000000000000000',
     ])
 
-    val_path = '/var/log/ceph/valgrind'.format(tdir=testdir)
+    val_path = '/var/log/ceph/valgrind'.format(tdir=testdir) #noqa
```
kshtsk, line 1066: same here... no need o use format on bare string

scripts/queue.py
vasukulkarni: yeah agree, i think the old flake8 missed those and its safe to remove the format where its not used.

vasukulkarni:
> > Please don't merge it, I still want to check py2 compatibility on teuthology master branch.
> 
> This is flake8 check and also due to new flake8 pypi release we are now seeing new issue.

@kshtsk what concerns do you have with flake8 checks with py2, I think either should do and its not really compatibility check

kshtsk:
@vasukulkarni I have no concerns, I think this is kind of misunderstanding, there both flake8 (py2) and flake8-py3 running against teuthology master, take another look at fox.ini, I found it pretty compatible. Many fixes I from master I would prefer to back port to py2 branch, I hope to keep it compatible py2 as long as we switch ceph branches to use teuthology master, instead.

tox.ini
```diff
@@ -46,12 +46,6 @@ deps=
 commands=py.test --cov=teuthology --cov-report=term -v {posargs:teuthology/test/integration teuthology/orchestra/test/integration}
 basepython=python2.7
 
-[testenv:flake8]
-install_command = pip install --upgrade {opts} {packages}
-deps=
-  flake8
-commands=flake8 --select=F,E9 {posargs:teuthology scripts}
-
```
kshtsk, line 54: @vasukulkarni 
I'll try to explain why this is should not be merged yet:
When new feature is suggested to merge to master and we want it to be back ported to py2 it is much easier to do if the code is py2 compatible in first place. In order to make contributor to have code py2/py3 compatible we still have to run both py2 and py3 based.
So I would hold on with getting rid of py2 from master for awhile?
Does this make sense to you?

kshtsk: The rest of flake8 related fixes I'm happy to merge.

vasukulkarni: I am going to discuss this specific commit in another PR so that I am not blocked on unit test failure. 
It is just flake8 and not sure if it can bring up the compatibility issue, the bootstrap runs in py3 only and py.test I have seen in tox is still using py2

kshtsk: @vasukulkarni 
```
flake8-py3 create: /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox/flake8-py3
[9356] /home/jenkins-build/build/workspace/teuthology-pull-requests/.tox$ /tmp/venv.eacBcHiFr6/bin/python2.7 -m virtualenv --no-download --python /usr/bin/python3 flake8-py3 >flake8-py3/log/flake8-py3-0.log
flake8-py3 installdeps: flake8
```
From this I can assume that tox virtualenv tool is running with py2.7 however the flake8-py3 environment  is already using py3, and correspondingly the flake8 itself is running py3. Thus it is tox only the tox what is running on py2

PR #1473, opened 2020-05-12
suite: schedule first and last in suite jobs once

kshtsk:
Only schedule --first-in-suite and --last-in-suite jobs
once per run when --num is provided for a teuthology-suite.

Fixes: https://tracker.ceph.com/issues/45520

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -294,7 +294,6 @@ class Run(object):
     def build_base_args(self):
         base_args = [
             '--name', self.name,
-            '--num', str(self.args.num),
             '--worker', util.get_worker(self.args.machine_type),
         ]
         if self.args.dry_run:
@@ -411,6 +410,7 @@ class Run(object):
 
             arg = copy.deepcopy(self.base_args)
             arg.extend([
+                '--num', str(self.args.num),
                 '--description', description,
                 '--',
             ])
--- a/teuthology/suite/test/test_run_.py
+++ b/teuthology/suite/test/test_run_.py
@@ -250,6 +250,7 @@ class TestScheduleSuite(object):
         # schedule_jobs() is just neutered; check calls below
 
         self.args.newest = 0
+        self.args.num = 42
         runobj = self.klass(self.args)
         runobj.base_args = list()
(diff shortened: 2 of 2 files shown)


kshtsk:
@susebot run deploy

teuthology/suite/run.py
```diff
@@ -411,6 +410,7 @@ def matches(f):
 
             arg = copy.deepcopy(self.base_args)
             arg.extend([
+                '--num', str(self.args.num),
```
tchaikov, line 413: how is `--num` connected with `write_result()` and `write_rerun_memo()` tasks?

kshtsk: Directly via `build_base_args` as soon as it is has `--num X` in command line to `teuthology-schedule`, the latter actually put jobs in the beanstalk queue, the teuthology-schedule defaults --num to 1. So just removing this parameter will do the deal.

kshtsk:
@tchaikov ping

teuthology/suite/run.py
tchaikov: my question still holds. i still don't understand how they are connected.

kshtsk: Okay, to be precise it is connected not directly, but transitively.
Here are the beginning of `write_result()` a `write_rerun_memo()`:
```
    def write_rerun_memo(self):
        args = copy.deepcopy(self.base_args)
```
and
```
    def write_result(self):
        arg = copy.deepcopy(self.base_args)
```
both of them are deep coppying *self.base_args*, which is set in the beginning of `prepare_schedule()` using `sefl.build_base_args()`:
```
    def prepare_and_schedule(self):
        """
        Puts together some "base arguments" with which to execute
        teuthology-schedule for each job, then passes them and other parameters
        to schedule_suite(). Finally, schedules a "last-in-suite" job that
        sends an email to the specified address (if one is configured).
        """
        self.base_args = self.build_base_args()
```
The `self.build_base_args()` composes argument list, which includes `--num`. This patch, removes the `--num` from it on the line 297, so the base_args do not include it. 
I hope this helps to understand.

PR #1474, opened 2020-05-13
teuthology/run_tasks: do not call sys.exc_clear() for clearing exception

tchaikov:
sys.exc_clear() was removed in Python3, see
https://docs.python.org/3/whatsnew/3.0.html#index-22. so we should not
call it.

### DIFF
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -187,7 +187,6 @@ def run_tasks(tasks, ctx):
                         interactive.task(ctx=ctx, config=None)
                 else:
                     if suppress:
-                        sys.exc_clear()
                         exc_info = (None, None, None)
 
             if exc_info != (None, None, None):


kshtsk:
@tchaikov looks good to me but I wonder if we can label this commit somehow so we do not backport it by mistake to py2.

kshtsk:
@susebot run deploy

kshtsk:
@tchaikov do you have an idea why this might fail on OVH?
https://ceph-ci.suse.de/job/teuthology-deploy-ovh/215/testReport/junit/(root)/teuthology/smoke_basic__clusters__fixed_3_cephfs_yaml_openstack_yaml__objectstore_bluestore_bitmap_yaml_supported_random_distro$__centos_8_yaml__tasks_rados_cls_all_yaml_/attachments/teuthology-smoke-2.log

kshtsk:
@susebot run deploy

tchaikov:
@joshd could you help review this change?

tchaikov:
@neha-ojha could you help take a look?

PR #1477, opened 2020-05-14
py2 backports 20200514

kshtsk:


kshtsk:
@susebot run deploy

tchaikov:
why are you backport py3 related fix to py2?

kshtsk:
> why are you backporting py3 related fix to py2?

as far as I can see those fixes are py2 compatible, and it would be good to have them if it will ease future backport of the conflicting code.

kshtsk:
which ones do you will to drop?

PR #1478, opened 2020-05-14
Revert "teuthology/orchestra/connection: add retry exceptions"

kshtsk:
This reverts commit 1e30d6f624921ae98c27981d55e06c29d1b0fdd0.

### DIFF
--- a/teuthology/orchestra/connection.py
+++ b/teuthology/orchestra/connection.py
@@ -4,22 +4,12 @@ Connection utilities
 import base64
 import paramiko
 import os
-import socket
 import logging
-
-from paramiko import AuthenticationException
-from paramiko.ssh_exception import NoValidConnectionsError
 
 from teuthology.config import config
 from teuthology.contextutil import safe_while
 
 log = logging.getLogger(__name__)
-
-RECONNECT_EXCEPTIONS = (
-  socket.error,
-  AuthenticationException,
-  NoValidConnectionsError,
-)
 
 
 def split_user(user_at_host):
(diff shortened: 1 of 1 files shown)


tchaikov:
see http://pulpito.ceph.com/kchai-2020-05-14_11:55:41-rados-wip-kefu-testing-2020-05-05-1333-distro-basic-smithi/

i was using the latest master of teuthology, but still seeing failures like
```
SSH connection to smithi194 was lost: "/bin/sh -c 'adjust-ulimits ceph-coverage /home/ubuntu/cephtest/archive/coverage rados --no-log-to-stderr --name client.0 -b 65536 -p unique_pool_0 bench 150 write'"
```

batrick:
The issue is this `safe_while`:

https://github.com/ceph/teuthology/blob/5b221634b3746e14b73f0b75f164e1e79bb3e269/teuthology/orchestra/connection.py#L115

Now this code is catching all the network exceptions that were previously **sometimes** caught by the caller (e.g. https://github.com/ceph/teuthology/blob/5b221634b3746e14b73f0b75f164e1e79bb3e269/teuthology/provision/cloud/openstack.py#L307-L315). When the `safe_while` gives up and raises `MaxWhileTries`, the caller doesn't deal with it causing the failure. Normally the caller would wait longer for the machine to come up. So, the answer here is to increase the `safe_while` `tries`/`sleep` sufficiently.


Now, if the preferred way to go about fixing this is to first revert, I don't mind. I don't really have more time to devote to this though. Please go ahead and write up a new fix.

kshtsk:
To summarize the patch is changing behavior of the method, in particular when it was expected by callers to produce connection exceptions it generates maxtry exception, we need to create another PR a test it better adding figuring all the callers of it.

PR #1479, opened 2020-05-14
fixes for redhat internal tasks

rakeshgm:
added couple of the internal tasks for redhat that were missing and fixes for the same. 

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -19,6 +19,8 @@ from teuthology.config import config as teuth_config
 from teuthology.exceptions import VersionNotFoundError
 from teuthology.job_status import get_status, set_status
 from teuthology.orchestra import cluster, remote, run
+# the below import with noqa is to workaround run.py which does not support multilevel submodule import
+from teuthology.task.internal.redhat import setup_cdn_repo, setup_base_repo, setup_additional_repo, setup_stage_cdn  # noqa
 
 log = logging.getLogger(__name__)
 
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -220,11 +220,21 @@ def get_initial_tasks(lock, config, machine_type):
         ])
     init_tasks.append({'internal.timer': None})
 
+    # install_latest_rh_kernel is used for redhat config
+    if 'redhat' not in config:
+        if 'kernel' in config:
+            init_tasks.append({'kernel': config['kernel']})
+
     if 'roles' in config:
         init_tasks.extend([
             {'pcp': None},
             {'selinux': None},
         ])
+
+    if 'redhat' in config:
+        init_tasks.extend([
+            {'internal.setup_stage_cdn': None}])
+
     if config.get('ceph_cm_ansible', True):
         init_tasks.append({'ansible.cephlab': None})
 
@@ -234,6 +244,7 @@ def get_initial_tasks(lock, config, machine_type):
 
     if 'redhat' in config:
         init_tasks.extend([
+            {'internal.git_ignore_ssl': None},
             {'internal.setup_cdn_repo': None},
             {'internal.setup_base_repo': None},
             {'internal.setup_additional_repo': None},
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -17,6 +17,7 @@ log = logging.getLogger(__name__)
 
 
 def get_task(name):
+    # todo: support of submodules
     if '.' in name:
         module_name, task_name = name.split('.')
     else:
--- a/teuthology/task/internal/git_ignore_ssl.py
+++ b/teuthology/task/internal/git_ignore_ssl.py
@@ -1,0 +1,22 @@
+import contextlib
+import logging
+
+from teuthology.orchestra import run
+
+log = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def git_ignore_ssl(ctx, config):
+    """
+    Ignore ssl error's while cloning from untrusted http
+    """
+
+    log.info("ignoring ssl errors while cloning http repo")
+    ctx.cluster.run(
+            args=[
+                  'sudo', 'git', 'config', run.Raw('--system'),
+                  'http.sslverify', 'false'
+                ],
+        )
+    yield
--- a/teuthology/task/internal/redhat.py
+++ b/teuthology/task/internal/redhat.py
@@ -24,7 +24,7 @@ def setup_stage_cdn(ctx, config):
         raise ConfigError("Provide rhbuild attribute")
     teuthconfig.rhbuild = str(rhbuild)
     with parallel() as p:
-        for remote in ctx.cluster.remotes.iterkeys():
+        for remote in ctx.cluster.remotes.keys():
             if remote.os.name == 'rhel':
                 log.info("subscribing stage cdn on : %s", remote.shortname)
                 p.spawn(_subscribe_stage_cdn, remote)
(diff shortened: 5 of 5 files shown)


teuthology/task/internal/__init__.py
```diff
@@ -19,7 +19,7 @@
 from teuthology.exceptions import VersionNotFoundError
 from teuthology.job_status import get_status, set_status
 from teuthology.orchestra import cluster, remote, run
-
+from .redhat import setup_cdn_repo, setup_base_repo, setup_additional_repo, setup_stage_cdn # noqa
```
kshtsk, line 22: Could you please use full path to the redhat module?

kshtsk: And remove # noqa

kshtsk:
@susebot run deploy

kshtsk:
is this needed for anything else but ceph 'master'?

vasukulkarni:
> is this needed for anything else but ceph 'master'?

I am not sure i understand your question,  the task allows to run ceph in octo lab

teuthology/task/internal/__init__.py
vasukulkarni: @rakeshgm can you see if the test passes with noqa and full path?

kshtsk: @vasukulkarni I would prefer to follow one style in imports.

kshtsk: also in some circumstances it can fail

kshtsk:
> > is this needed for anything else but ceph 'master'?
> 
> I am not sure i understand your question, the task allows to run ceph in octo lab

What is 'octo' lab? The question is the committed code supposed to be run against ceph branches which are not switched to py3, if so the is probably need to backport them to py2 branch.

teuthology/task/internal/__init__.py
rakeshgm: Done. relative import added

rakeshgm: test passes with full path.

rakeshgm: tests failed because flake8 reported unused imports

kshtsk: so why this imports needed if they are not used?

rakeshgm: they are actually being called from run.py https://github.com/ceph/teuthology/blob/master/teuthology/run.py#L236.

kshtsk: > they are actually being called from run.py https://github.com/ceph/teuthology/blob/master/teuthology/run.py#L236.

@rakeshgm the code you're referring is not used for calling them, it is just compose the task list, the modules for them are loaded automatically in other place, see as an example other jobs in the task list are not imported directly here in this module.

rakeshgm: @kshtsk , I intended to mention the same. sorry I did not phrase it correctly.

kshtsk:
@susebot run deploy

teuthology/task/internal/__init__.py
rakeshgm: @kshtsk  so to fix the flake8 errors, I can move the tasks from `redhat` and add them in `__init__` which might fix the problem else we could check for flake8 options to ignore such errors i.e if you think its a good idea. 
Or any other suggestions.

kshtsk: @rakeshgm no, just remove this import line, and that's it.

rakeshgm: `setup_cdn_repo, setup_base_repo, setup_additional_repo, setup_stage_cdn` tasks are present in `redhat.py` unlike others in `__init__` for example in run.py we have somthing like this `{'internal.setup_cdn_repo': None}` .
importing this line gets the task.

kshtsk: Ah, I see, why not using `{'internal.redhat.setup_cdn_repo': None}` then?

rakeshgm: I tried that a while ago. the split here fails. 
https://github.com/ceph/teuthology/blob/master/teuthology/run_tasks.py#L21

rakeshgm: I pushed a small [change here](6261eae46675eacc515f00bc4f3f1b070dbeb85b)

rakeshgm: I pushed a small change [here](https://github.com/ceph/teuthology/pull/1479/commits/6261eae46675eacc515f00bc4f3f1b070dbeb85b) to this PR.

rakeshgm: looks like the tests passed.

kshtsk: > I tried that a while ago. the split here fails.
> https://github.com/ceph/teuthology/blob/master/teuthology/run_tasks.py#L21
> returns more than 2.

Oh... shoot, it does not support submodules then... Seems like redhat is the first ever module is trying to do so.
I guess we could add a support for submodules.

kshtsk: > I pushed a small change [here](https://github.com/ceph/teuthology/pull/1479/commits/6261eae46675eacc515f00bc4f3f1b070dbeb85b) to this PR.

That's too hacky, I would go with `# noqa` tag accompanying a comment why it is needed?

kshtsk: however... it's fine....

kshtsk: we can add support for submodules later...

kshtsk:
@susebot run deploy

teuthology/task/internal/__init__.py
vasukulkarni: yeah the split and support for single module was the reason I think we did that, I suggest we keep the original code and add comment and move on until the support exists in teuthology.

vasukulkarni:
> > > is this needed for anything else but ceph 'master'?
> > 
> > 
> > I am not sure i understand your question, the task allows to run ceph in octo lab
> 
> What is 'octo' lab? The question is the committed code supposed to be run against ceph branches which are not switched to py3, if so the is probably need to backport them to py2 branch.

We dont need py2 support, I didnt read your question correctly but octo is internal lab and you can ignore my previous answer :)

teuthology/task/internal/__init__.py
rakeshgm: reverted, add full path import along with `noqa`

vasukulkarni:
@kshtsk do you have any more concerns here?

teuthology/task/internal/__init__.py
```diff
@@ -19,6 +19,7 @@
 from teuthology.exceptions import VersionNotFoundError
 from teuthology.job_status import get_status, set_status
 from teuthology.orchestra import cluster, remote, run
+from teuthology.task.internal.redhat import setup_cdn_repo, setup_base_repo, setup_additional_repo, setup_stage_cdn  # noqa
```
kshtsk, line 22: so could you please a comment here why this import need here? also maybe add a `todo` tag for adding support of submodules

vasukulkarni: yeah good point, @rakeshgm can you fix that

kshtsk:
> @kshtsk do you have any more concerns here?

I'm fine with it, but a comment about the need of the import

kshtsk:
@susebot run deploy

teuthology/task/internal/__init__.py
rakeshgm: done. added comment and todo

PR #1480, opened 2020-05-14
provision.fog: Add optional sentinel file

zmc:
Similar to how provision.cloud.openstack works, wait for a particular
file to be created before considering provisioning to be finished.

Unlike the cloud.openstack module, teuthology doesn't control the entire
process, so to avoid breaking any labs that aren't using a sentinel
file, make the feature optional and default to being inoperative.

### DIFF
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -6,6 +6,7 @@ import re
 
 from datetime import datetime
 from paramiko import SSHException
+from paramiko.ssh_exception import NoValidConnectionsError
 
 import teuthology.orchestra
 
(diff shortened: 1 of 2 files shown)


zmc:
https://tracker.ceph.com/issues/45341

kshtsk:
@susebot run deploy

djgalloway:
Well it didn't NOT work... http://pulpito.ceph.com/dgalloway-2020-05-15_13:51:12-rados-master-distro-basic-smithi/

Kinda hard to test if something's fixed that was only happening in ~10/1000 jobs.

zmc:
This'll work almost exactly like the ready check in `provision.cloud.openstack`, so I'd be surprised if it didn't do what we wanted. I did test in the lab via a simple reimage.

I did not test on python2, but there isn't anything here that's py3-specific so a backport should be trivial.

kshtsk:
The py2 unittests passed though, going to create a backport.

PR #1481, opened 2020-05-15
backport: fog sentinel

kshtsk:
Similar to how provision.cloud.openstack works, wait for a particular
file to be created before considering provisioning to be finished.

Unlike the cloud.openstack module, teuthology doesn't control the entire
process, so to avoid breaking any labs that aren't using a sentinel
file, make the feature optional and default to being inoperative.

(cherry picked from commit 5c66b774517c0f330f23f087b184caa3f9dbd6db)

### DIFF
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -6,6 +6,7 @@ import re
 
 from datetime import datetime
 from paramiko import SSHException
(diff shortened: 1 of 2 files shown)


kshtsk:
@susebot run deploy

PR #1482, opened 2020-05-15
suite/build_matrix: get rid of '.yaml' in descriptions

kshtsk:
Test descriptions have structured representation of yaml
fragments the test config is composed from. Although
having the .yaml extentions in the description looks
redundant and has no use in the code.
It worth to say that no other files except yaml
participate in description composition.

This change removes many '.yaml' sufixes from
descriptions which make it more compact and easy
to read.

### DIFF
--- a/teuthology/suite/build_matrix.py
+++ b/teuthology/suite/build_matrix.py
@@ -185,7 +185,7 @@ def generate_combinations(path, mat, generate_from, generate_to):
     for i in range(generate_from, generate_to):
         output = mat.index(i)
         ret.append((
-            matrix.generate_desc(combine_path, output),
+            matrix.generate_desc(combine_path, output).replace('.yaml', ''),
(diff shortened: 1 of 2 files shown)


kshtsk:
@susebot run deploy

yuriw:
@kshtsk the description is highly valuable in using it as a `--filter` when a job needs to rerun.
I'd say that it still should work if `yaml` is removed, but it's worth checking and making sure that it still works.

kshtsk:
> @kshtsk the description is highly valuable in using it as a `--filter` when a job needs to rerun.
> I'd say that it still should work if `yaml` is removed, but it's worth checking and making sure that it still works.

Yes you're right, the rerun should work as far as you use the same `teuthology-suite` version you've scheduled one.

kshtsk:
image

PR #1484, opened 2020-05-19
teuthology/misc: do not fail when pull_directory(a_dir_does_not_exist)

tchaikov:
there is chance that /var/log/ceph (and other directories) is not
created when cephadm fails, but we always try to collect /var/log/ceph
by tar the directory on the test node and transfer it back for archiving
it. when we fail to do so, an exception is thrown, and this exception is
in turn logged as the cause of the failure instead of the one which
fails the test.

in this change, all `tarfile.TarError`s are caught in
`misc.pull_directory()`, as normally, we don't consider it a fatal
failure when `misc.pull_directory()` fails because of `tar`.

two types of failures are caught here. we could use `tarfile.TarError`
here, as both of them inherit from this class, but it'd be easier to
understand to be explicit. as only these two types are expected.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -764,7 +764,15 @@ def pull_directory(remote, remotedir, localdir):
     if not os.path.exists(localdir):
         os.mkdir(localdir)
     r = remote.get_tar_stream(remotedir, sudo=True)
-    tar = tarfile.open(mode='r|gz', fileobj=r.stdout)
+    try:
+        tar = tarfile.open(mode='r|gz', fileobj=r.stdout)
+    except (tarfile.EmptyHeaderError, tarfile.ReadError) as e:
+        # bail out if remotedir does not exist, otherwise teuthology
+        # will take this exception as the cause of failure instead.
+        log.warn("Failed to pull remote directory '%s': %r",
+                 remotedir, e)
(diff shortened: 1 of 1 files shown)


tchaikov:
see also http://pulpito.ceph.com/nojha-2020-05-19_00:53:41-rados-wip-revert-34894-distro-basic-smithi/5068077/

teuthology/misc.py
```diff
@@ -764,7 +764,15 @@ def pull_directory(remote, remotedir, localdir):
     if not os.path.exists(localdir):
         os.mkdir(localdir)
     r = remote.get_tar_stream(remotedir, sudo=True)
-    tar = tarfile.open(mode='r|gz', fileobj=r.stdout)
+    try:
+        tar = tarfile.open(mode='r|gz', fileobj=r.stdout)
```
kshtsk, line 768: I just wonder why this check should be here? Why not to handle this by caller of `pull_directory`?
The caller should be aware of existence of a remote directory, if it unsure then it should handle the call correctly.
From other side, can the description of the function be updated to reflect the behavior what caller should expect from the function.

tchaikov:
makes sense. will ignore these Error on the caller side.

PR #1488, opened 2020-05-22
skip adding check_packages and buildpackages_prep tasks to init_tasks if redhat is added to config

rakeshgm:
for redhat testing, we test ceph composes which are generated internally. we don't need to do any package checking so removing it from init_tasks only if redhat is in config 

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -170,7 +170,7 @@ def get_initial_tasks(lock, config, machine_type):
     overrides = config.get('overrides')
     having_repos = ('install' in config and 'repos' in config.get('install')) \
         or (overrides and 'install' in overrides and 'repos' in overrides.get('install'))
-    if not having_repos:
+    if not ('redhat' in config or having_repos):
         init_tasks += [
             {'internal.check_packages': None},
             {'internal.buildpackages_prep': None},
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -188,7 +188,6 @@ def run_tasks(tasks, ctx):
                         interactive.task(ctx=ctx, config=None)
                 else:
                     if suppress:
-                        sys.exc_clear()
                         exc_info = (None, None, None)
 
             if exc_info != (None, None, None):


teuthology/run.py
```diff
@@ -250,7 +250,9 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.setup_additional_repo': None},
             {'kernel.install_latest_rh_kernel': None}
         ])
- 
+        # remove check_packages task if redhat is in config
+        init_tasks.remove({'internal.check_packages': None})
```
kshtsk, line 254: Instead of removing, just do not add it in first place.

rakeshgm:
> Instead of removing, just do not add it in first place

you are right, adding 
```having_repos = config.get('redhat') or ('install' in config and 'repos' in config.get('install')) or (overrides and 'install' in overrides and 'repos' in overrides.get('install'))``` might solve it. I'll look into it.

**Comment by kshtsk** [comment]
> > Instead of removing, just do not add it in first place
> 
> you are right, adding
> `having_repos = config.get('redhat') or ('install' in config and 'repos' in config.get('install')) or (overrides and 'install' in overrides and 'repos' in overrides.get('install'))` might solve it. I'll look into it.

I would say it should be:
```if 'redhat' in config or not having_repos:```

**Comment by rakeshgm** [comment]
> if 'redhat' in config or not having_repos:
`if 'redhat' in config or not having_repos:` makes the if statement return True and check_packges task gets added.  I think it can be one of 
1. `if not  'redhat in config or not having_repos`  
 or 
2. `having_repos = ('redhat' in config') or ...`

**Comment by kshtsk** [comment]
> > if 'redhat' in config or not having_repos:
> > `if 'redhat' in config or not having_repos:` makes the if statement return True and check_packges task gets added.  I think it can be one of
> 
> 1. `if not  'redhat in config or not having_repos`
>    or
> 2. `having_repos = ('redhat' in config') or ...`

right, it was my mistake, _not_ is required.

**Comment by kshtsk** [comment]
> > if 'redhat' in config or not having_repos:
> > `if 'redhat' in config or not having_repos:` makes the if statement return True and check_packges task gets added.  I think it can be one of
> 
> 1. `if not  'redhat in config or not having_repos`
>    or
> 2. `having_repos = ('redhat' in config') or ...`

yes, you're correct, either works fine

**Comment by rakeshgm** [comment]
> > > if 'redhat' in config or not having_repos:
> > > `if 'redhat' in config or not having_repos:` makes the if statement return True and check_packges task gets added.  I think it can be one of
> > 
> > 
> > 
> > 1. `if not  'redhat in config or not having_repos`
> >    or
> > 2. `having_repos = ('redhat' in config') or ...`
> 
> yes, you're correct, either works fine

changes added

**Comment by kshtsk** [comment]
@rakeshgm thanks for adding this change, but could you please rework it as you described in our previous commit, `having_repos` is related to specifically validating the structure:
```
install:
  repos:
      . . .
```
and it has nothing to do with 'redhat'.

**Comment by rakeshgm** [comment]
@kshtsk can you elaborate more what you mean. I am sorry I did understand what meant

**Comment by kshtsk** [comment]
> @kshtsk can you elaborate more what you mean. I am sorry I did understand what meant

I mean this, yeah and fixed the logic.
```
if not 'redhat' in config and not having_repos:
```
or
```
having_redhat =  'redhat' in config
if not (having_redhat or having_repos):
```

**Comment by rakeshgm** [comment]
makes sense. adding redhat check for  `having_repos` might not look correct as you mentioned it is for package structure validation. I will add this change 
`if not ('redhat' in config or having_repos)`.  i would like to use this change rather ther creating another variable like `having_redhat`.

**Comment by kshtsk** [comment]
> makes sense. adding redhat check for `having_repos` might not look correct as you mentioned it is for package structure validation. I will add this change
> `if not ('redhat' in config or having_repos)`. i would like to use this change rather ther creating another variable like `having_redhat`.

makes sense

**Comment by rakeshgm** [comment]
> > makes sense. adding redhat check for `having_repos` might not look correct as you mentioned it is for package structure validation. I will add this change
> > `if not ('redhat' in config or having_repos)`. i would like to use this change rather ther creating another variable like `having_redhat`.
> 
> makes sense

changes added.

File: teuthology/run.py [hunk 2]
```diff
@@ -250,7 +250,7 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.setup_additional_repo': None},
             {'kernel.install_latest_rh_kernel': None}
         ])
-
```
kshtsk, line 253: revert this line, please...

kshtsk:
Update commit message

teuthology/run.py
rakeshgm: done.

rakeshgm:
something got messed up in my branch. will close and open another PR.

PR #1489, opened 2020-05-22
run: exclude internal check_packages and buildpackages_prep for redhat

rakeshgm:
for redhat testing, we test ceph composes which are generated internally. we don't need to do any package checking so not adding it from init_tasks only if redhat is in config


kshtsk:
Just want to wait until deploy test finished.

kshtsk:
@susebot run deploy

vasukulkarni:
Lets wait for the results

vasukulkarni:
I have asked @rakeshgm to ammend commit message but will wait for results first.

rakeshgm:
commit message changed

PR #1490, opened 2020-05-24
teuthology-suite: automate -t argument default value

kshtsk:
This change is required to determine which teuthology
branch should be used for scheduling a run for the
given ceph branch.

In order to enable the functionality of this patch we need
submit and merge `qa/.teuthology` to ceph `master` 
with following content:
```
# cherry-pick this for py3 ready branches
teuthology_branch: master
```
And add to `/etc/teuthology.yaml` on sepia server the line
```
# remove this line when all ceph branches get py3 compatible qa tasks
teuthology_branch: py2
```
Obviously, all users have to update their teuthology sandboxes.
As far as other branches

### DIFF
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -25,7 +25,6 @@ def override_arg_defaults(name, default, env=os.environ):
         '--suite-repo'        : 'TEUTH_SUITE_REPO',
         '--ceph-branch'       : 'TEUTH_CEPH_BRANCH',
         '--suite-branch'      : 'TEUTH_SUITE_BRANCH',
-        '--teuthology-branch' : 'TEUTH_BRANCH',
     }
     if name in env_arg and env_arg[name] in env.keys():
         variable = env_arg[name]
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -51,12 +51,6 @@ class Run(object):
         # caches package versions to minimize requests to gbs
         self.package_versions = dict()
 
-        if self.args.suite_dir:
-            self.suite_repo_path = self.args.suite_dir
-        else:
-            self.suite_repo_path = util.fetch_repos(
-                self.base_config.suite_branch, test_name=self.name)
-
         # Interpret any relative paths as being relative to ceph-qa-suite
         # (absolute paths are unchanged by this)
         self.base_yaml_paths = [os.path.join(self.suite_repo_path, b) for b in
@@ -94,9 +88,15 @@ class Run(object):
         # We don't store ceph_version because we don't use it yet outside of
         # logging.
         self.choose_ceph_version(ceph_hash)
-        teuthology_branch = self.choose_teuthology_branch()
         suite_branch = self.choose_suite_branch()
         suite_hash = self.choose_suite_hash(suite_branch)
+        if self.args.suite_dir:
+            self.suite_repo_path = self.args.suite_dir
+        else:
+            self.suite_repo_path = util.fetch_repos(
+                suite_branch, test_name=self.name)
+        teuthology_branch = self.choose_teuthology_branch()
+
 
         if self.args.distro_version:
             self.args.distro_version, _ = \
@@ -196,20 +196,48 @@ class Run(object):
             log.info('skipping ceph package verification')
 
     def choose_teuthology_branch(self):
+        """Select teuthology branch, check if it is present in repo and
+        return the branch name value.
+
+        The branch name value is determined in the following order:
+
+        Use ``--teuthology-branch`` argument value if supplied.
+
+        Use ``TEUTH_BRANCH`` environment variable value if declared.
+
+        If file ``qa/.teuthology_branch`` can be found in the suite repo
+        supplied with ``--suite-repo`` or ``--suite-dir`` and contains
+        non-empty string then use it as the branch name.
+
+        Use ``teuthology_branch`` value if it is set in the one
+        of the teuthology config files ``$HOME/teuthology.yaml``
+        or ``/etc/teuthology.yaml`` correspondingly.
+
+        Use ``master``.
+
+        Generate exception if the branch is not present in the repo.
+        """
         teuthology_branch = self.args.teuthology_branch
-        if teuthology_branch and teuthology_branch != 'master':
-            if not util.git_branch_exists('teuthology', teuthology_branch):
-                exc = BranchNotFoundError(teuthology_branch, 'teuthology.git')
-                util.schedule_fail(message=str(exc), name=self.name)
-        elif not teuthology_branch:
-            # Decide what branch of teuthology to use
-            if util.git_branch_exists('teuthology', self.args.ceph_branch):
-                teuthology_branch = self.args.ceph_branch
-            else:
-                log.info(
-                    "branch {0} not in teuthology.git; will use master for"
-                    " teuthology".format(self.args.ceph_branch))
-                teuthology_branch = 'master'
+        if not teuthology_branch:
+            teuthology_branch = os.environ.get('TEUTH_BRANCH', None)
+        if not teuthology_branch:
+            branch_file_path = self.suite_repo_path + '/qa/.teuthology_branch'
+            log.debug('Check file %s exists', branch_file_path)
+            if os.path.exists(branch_file_path):
+                log.debug('Found teuthology branch config file %s',
+                                                        branch_file_path)
+                with open(branch_file_path) as f:
+                    teuthology_branch = f.read().strip()
+                    if teuthology_branch:
+                        log.debug(
+                            'The teuthology branch is overridden with %s',
+                                                                teuthology_branch)
+                    else:
+                        log.warning(
+                            'The teuthology branch config is empty, skipping')
+        if not teuthology_branch:
+            teuthology_branch = config.get('teuthology_branch', 'master')
+
         teuthology_hash = util.git_ls_remote(
             'teuthology',
             teuthology_branch
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -54,7 +54,13 @@ Standard arguments:
                               [default: basic]
   -t <branch>, --teuthology-branch <branch>
                               The teuthology branch to run against.
-                              [default: {default_teuthology_branch}]
+                              Default value is determined in the next order.
+                              There is TEUTH_BRANCH environment variable set.
+                              There is `qa/.teuthology_branch` present in
+                              the suite repo and contains non-empty string.
+                              There is `teuthology_branch` present in one of
+                              the user or system `teuthology.yaml` configuration
+                              files respectively, otherwise use `master`.
   -m <type>, --machine-type <type>
                               Machine type [default: {default_machine_type}]
   -d <distro>, --distro <distro>
@@ -165,7 +171,6 @@ Scheduler arguments:
     default_suite_repo=defaults('--suite-repo',
                             config.get_ceph_qa_suite_git_url()),
     default_ceph_branch=defaults('--ceph-branch', 'master'),
-    default_teuthology_branch=defaults('--teuthology-branch', 'master'),
 )
 
 
--- a/teuthology/suite/test/test_init.py
+++ b/teuthology/suite/test/test_init.py
@@ -156,6 +156,9 @@ class TestSuiteMain(object):
         def fake_bool(*args, **kwargs):
             return True
 
+        def fake_false(*args, **kwargs):
+            return False
+
         with patch.multiple(
                 'teuthology.suite.run.util',
                 fetch_repos=DEFAULT,
@@ -166,6 +169,9 @@ class TestSuiteMain(object):
             with patch.multiple(
                 'teuthology.suite.run.Run',
                 prepare_and_schedule=prepare_and_schedule,
+            ), patch.multiple(
+                'teuthology.suite.run.os.path',
+                exists=fake_false,
             ):
                 main([
                     '--ceph', 'master',
--- a/teuthology/suite/test/test_run_.py
+++ b/teuthology/suite/test/test_run_.py
@@ -145,8 +145,10 @@ class TestRun(object):
     @patch('teuthology.suite.run.util.git_branch_exists')
     @patch('teuthology.suite.run.util.package_version_for_hash')
     @patch('teuthology.suite.run.util.git_ls_remote')
+    @patch('teuthology.suite.run.os.path.exists')
     def test_regression(
         self,
+        m_qa_teuthology_branch_exists,
         m_git_ls_remote,
         m_package_version_for_hash,
         m_git_branch_exists,
@@ -157,6 +159,7 @@ class TestRun(object):
         m_package_version_for_hash.return_value = 'ceph_hash'
         m_git_branch_exists.return_value = True
         m_git_ls_remote.return_value = "suite_branch"
+        m_qa_teuthology_branch_exists.return_value = False
         self.args_dict = {
             'base_yaml_paths': [],
             'ceph_branch': 'master',


sebastian-philipp:
Hm. This looks suspiciously similar to a requirements.txt file in ceph/qa to me.

Researching a bit more and it turns out that we already have an explicitly dependency to teuthology in the tox.ini. 

I suspect a proper requirements.txt won't work, right?

teuthology/suite/run.py
```diff
@@ -196,15 +196,59 @@ def choose_ceph_version(self, ceph_hash):
...
-                teuthology_branch = self.args.ceph_branch
+            check_branch = None
+            branch_config = self.suite_repo_path + '/qa/.teuthology'
+            if os.path.exists(branch_config):
+                log.debug('Found teuthology config in suite branch %s', branch_config)
+                with open(branch_config) as f:
+                    y = yaml.safe_load(f)
+                    check_branch = y.get('teuthology_branch', None)
+                    if check_branch:
+                        log.debug('Teuthology branch is overridedn with %s',
```
badone, line 244: a/overridedn/overridden/

teuthology/suite/run.py
```diff
@@ -196,15 +196,59 @@ def choose_ceph_version(self, ceph_hash):
...
+        or ``/etc/teuthology.yaml`` correspondingly.
+
+        Use ``--suite-branch`` argument value if supplied.
+
+        Use ``--ceph-branch`` argument value if supplied.
+
+        Use ``master`` if branch determined with ``--suite-branch``
+        or ``--ceph-branch`` are missing in the teuthology repo.
+
+        Generate exception if the determine branch is not present
```
badone, line 223: "determined" ?

kshtsk: @badone is _supplied with_ better?

kshtsk:
> Hm. This looks suspiciously similar to a requirements.txt file in ceph/qa to me.
I'm not following what you're referring to?
> 
> Researching a bit more and it turns out that we already have an explicitly dependency to teuthology in the tox.ini.
> 
> I suspect a proper requirements.txt won't work, right?
I have no idea what you're asking for, teuthology does not use requirements.txt at the moment, for py3 it uses requirements3.txt and for py2 it uses requirements2.txt

teuthology/suite/run.py
kshtsk: I've used this definition:
https://www.macmillandictionary.com/dictionary/british/determine
_3. INTRANSITIVE/TRANSITIVE to calculate something, or to discover it by examining evidence_

teuthology/suite/run.py
kshtsk: thanks, fixed

kshtsk:
> FWIW I don't think this is worth it, both the override file and the tying of `teuthology_branch` to `suite_branch` or `ceph_branch`. Particularly the latter -- teuthology repo is full of stale branches with fairly generic names (wip-fixes and alike).

Regarding suite_branch and ceph_branch functionality it was already in the code, I just added a description of this hidden logic. Do you request to delete it? @djgalloway @gregsfortytwo do we need to keep this?

> 
> I think having to specify `--teuthology-branch py2` is just fine, but if the consensus is that py2/py3 automation is necessary, then only py2 branch should be treated specially IMO. 

The point is that to avoid pointing to the right branch by a user, especially for users who do not have a clue which branch to use. 

> Instead of adding an override file, perhaps there is a py2-specific marker that can be grepped for in qa/tasks directory?

This is overkill, it much easier to add single file.

idryomov:
> > FWIW I don't think this is worth it, both the override file and the tying of `teuthology_branch` to `suite_branch` or `ceph_branch`. Particularly the latter -- teuthology repo is full of stale branches with fairly generic names (wip-fixes and alike).
> 
> Regarding suite_branch and ceph_branch functionality it was already in the code, I just added a description of this hidden logic. Do you request to delete it? @djgalloway @gregsfortytwo do we need to keep this?

Was `suite_branch` already there?  I thought it was just `ceph_branch`, which goes back to the early days of ceph, when all tasks lived in the teuthology repo and runs were scheduled with a shell script.

I'm not requesting that you delete it, but merely trying to discourage adding more tangling.  It would be nice to look at whether `ceph_branch` logic is still needed, though -- there shouldn't be any *good* reason for it now.  It's certainly not needed for any of the recent releases, as we are using the same py2 branch for them.

> 
> > I think having to specify `--teuthology-branch py2` is just fine, but if the consensus is that py2/py3 automation is necessary, then only py2 branch should be treated specially IMO.
> 
> The point is that to avoid pointing to the right branch by a user, especially for users who do not have a clue which branch to use.
> 
> > Instead of adding an override file, perhaps there is a py2-specific marker that can be grepped for in qa/tasks directory?
> 
> This is overkill, it much easier to add single file.

Well, conceptually this file is an override for /etc/teuthology.yaml.  I would argue that if you introduce an override mechanism, it should handle all options, not just one of them.  And *that* is definitely an overkill for the problem at hand because all we need is for teuthology to detect that py2 tasks would be used and "switch" from py3 to py2.  No additional surface area and no configuration files committed anywhere.

kshtsk:
> Was `suite_branch` already there? I thought it was just `ceph_branch`, which goes back to the early days of ceph, when all tasks lived in the teuthology repo and runs were scheduled with a shell script.
> 
The ceph_branch what I saw has nothing todo with teuthology repo and uses the ceph repo.
If it is not the case, then I misinterpreted the usage of ceph_branch, because in other place the suite_branch is used for for checking out the code of tests from ceph repo, not teuthology. And following that logic I just complemented with logic reflecting ceph_branch and suite_branch substition of options `--ceph` and `--suite-branch` arguments.

> I'm not requesting that you delete it, but merely trying to discourage adding more tangling. It would be nice to look at whether `ceph_branch` logic is still needed, though -- there shouldn't be any _good_ reason for it now. It's certainly not needed for any of the recent releases, as we are using the same py2 branch for them.
> 
> > > I think having to specify `--teuthology-branch py2` is just fine, but if the consensus is that py2/py3 automation is necessary, then only py2 branch should be treated specially IMO.
> > 
> > 
> > The point is that to avoid pointing to the right branch by a user, especially for users who do not have a clue which branch to use.
> > > Instead of adding an override file, perhaps there is a py2-specific marker that can be grepped for in qa/tasks directory?
> > 
> > 
> > This is overkill, it much easier to add single file.
> 
> Well, conceptually this file is an override for /etc/teuthology.yaml. I would argue that if you introduce an override mechanism, it should handle all options, not just one of them. 

It could look like an override for teuthology.yaml but it is not, it should not be, nevertheless, it has same yaml format, it only means to take only teuthology branch parameter.
> And _that_ is definitely an overkill for the problem at hand because all we need is for teuthology to detect that py2 tasks would be used and "switch" from py3 to py2. No additional surface area and no configuration files committed anywhere.

Could you elaborate, I'm not following, would it be better to you instead of `yaml.safe_load('qa/.teuthology')`, use misc.sh('grep py2 qa/.teuthology'). Or you want to try run `python3 somepy3only.script`?

idryomov:
> > Was `suite_branch` already there? I thought it was just `ceph_branch`, which goes back to the early days of ceph, when all tasks lived in the teuthology repo and runs were scheduled with a shell script.
> 
> The ceph_branch what I saw has nothing todo with teuthology repo and uses the ceph repo.
> If it is not the case, then I misinterpreted the usage of ceph_branch, because in other place the suite_branch is used for for checking out the code of tests from ceph repo, not teuthology. And following that logic I just complemented with logic reflecting ceph_branch and suite_branch substition of options `--ceph` and `--suite-branch` arguments.

I'm referring to this piece of code, where in the absence of explicit `--teuthology-branch foo` on the command line (or in the environment) teuthology switches to a teuthology branch with same name as `ceph_branch`:

```
...
elif not teuthology_branch:
    # Decide what branch of teuthology to use
    if util.git_branch_exists('teuthology', self.args.ceph_branch):
        teuthology_branch = self.args.ceph_branch
```

This behaviour goes way back, when all tasks lived in the teuthology repo and one had to use the same branch of ceph and teuthology.  Those days are gone and I think we should look into dropping this logic.

This PR adds similar behaviour for `suite_branch`.  I don't see what use case is being addressed by this change and I think it should be dropped regardless of how you choose to proceed with py2/py3.

kshtsk:
> This PR adds similar behaviour for `suite_branch`. I don't see what use case is being addressed by this change and I think it should be dropped regardless of how you choose to proceed with py2/py3.

thanks for the extra explanation, I'm actual fine with dropping it, just don't want to do it by my own will...

idryomov:
> > > > Instead of adding an override file, perhaps there is a py2-specific marker that can be grepped for in qa/tasks directory?
> > > 
> > > 
> > > This is overkill, it much easier to add single file.
> > 
> > 
> > Well, conceptually this file is an override for /etc/teuthology.yaml. I would argue that if you introduce an override mechanism, it should handle all options, not just one of them.
> 
> It could look like an override for teuthology.yaml but it is not, it should not be, nevertheless, it has same yaml format, it only means to take only teuthology branch parameter.
> 
> > And _that_ is definitely an overkill for the problem at hand because all we need is for teuthology to detect that py2 tasks would be used and "switch" from py3 to py2. No additional surface area and no configuration files committed anywhere.
> 
> Could you elaborate, I'm not following, would it be better to you instead of `yaml.safe_load('qa/.teuthology')`, use misc.sh('grep py2 qa/.teuthology').

No.

> Or you want to try run `python3 somepy3only.script`? What if those files are missing in code base? The config file is easy to cherrypick. It would be greate to see a PR with your idea how it is the best to detect py2 tasks? ;-)

Yes, I think it should be relatively easy to detect that py2 branch should be used without any new configuration files, whether by trying to import/parse as python3 or just a grep for some pattern in one of .py files under qa/tasks.  If that file is not there, do nothing (i.e. use master).

kshtsk:
> Yes, I think it should be relatively easy to detect that py2 branch is be used without any new configuration files, whether by trying to import/parse as python3 or just a grep for some pattern in one of .py files under qa/tasks. If that file is not there, do nothing (i.e. use master).

I disagree and think this is incorrect approach an awkward... teuthology code should be related on any client python code, you anyway need to track exact file, in many branches, you anyway have to made a change, you have much more file operation to make that grep, than just read this config file. Given that most of tasks py for example. In fact, grepping is logically the same to loading of exact yaml file, but computationally much lighter.

teuthology/suite/run.py
badone: That's fine, but it should be "determined branch", not "determine branch".

djgalloway:
I don't have a dog in this fight as I don't use teuthology as a developer.  Basically strictly a machine locking tool as far as I'm concerned.

kshtsk:
I don't see input from @gregsfortytwo who actually requested the feature.
I'm actually discouraged so I'm ok to close it.

badone:
@gregsfortytwo ping?

badone:
@kshtsk I, for one, think we should persist with this because I am seeing confusion about this in day-to-day operations. Let's just see if we can get more eyes on this.

teuthology/suite/run.py
```diff
@@ -196,15 +196,58 @@ def choose_ceph_version(self, ceph_hash):
...
+                with open(branch_config) as f:
+                    y = yaml.safe_load(f)
+                    check_branch = y.get('teuthology_branch', None)
+                    if check_branch:
+                        log.debug('Teuthology branch is overridden with %s',
+                                  check_branch)
+            if not check_branch:
+                check_branch = config.get('teuthology_branch', None)
+            if not check_branch:
+                check_branch = self.args.suite_branch or self.args.ceph_branch
```
jdurgin, line 248: I agree with Ilya about dropping suite_branch and ceph_branch fallbacks - they're no longer necessary

teuthology/suite/run.py
```diff
@@ -196,15 +196,58 @@ def choose_ceph_version(self, ceph_hash):
...
+        Use ``--teuthology-branch`` argument value if supplied.
+
+        Use ``TEUTH_BRANCH`` environment variable value if declared.
+
+        If yaml file ``qa/.teuthology`` can be found in suite repo
+        supplied with ``--suite-repo`` or ``--suite-dir`` and
+        has ``teuthology_branch`` record then use its value.
+
+        Use ``teuthology_branch`` value if it is set in the one
+        of the teuthology config files ``$HOME/teuthology.yaml``
```
jdurgin, line 213: looks like ```$HOME/teuthology.yaml``` and ```/etc/teuthology.yaml``` aren't being read in this version of the patch

teuthology/suite/run.py
```diff
@@ -196,15 +196,58 @@ def choose_ceph_version(self, ceph_hash):
...
+                    exc = BranchNotFoundError(teuthology_branch, 'teuthology.git')
+                    util.schedule_fail(message=str(exc), name=self.name)
+        elif 'TEUTH_BRANCH' in os.environ:
+            teuthology_branch = os.environ.get('TEUTH_BRANCH')
+        else:
             # Decide what branch of teuthology to use
-            if util.git_branch_exists('teuthology', self.args.ceph_branch):
-                teuthology_branch = self.args.ceph_branch
+            check_branch = None
+            branch_config = self.suite_repo_path + '/qa/.teuthology'
```
jdurgin, line 236: to avoid the impression that this file could contain other options, perhaps it could be named something more specific, like /qa/.teuthology_branch

teuthology/suite/run.py
kshtsk: It is done at 246 line with this code:
```
check_branch = config.get('teuthology_branch', None)
```

teuthology/suite/run.py
idryomov: ... and changed to contain just the branch name instead of "teuthology_branch: foo" key/value pair.

kshtsk: that makes sense

teuthology/suite/run.py
jdurgin: ah, got it thanks

teuthology/suite/run.py
kshtsk: pr updated

teuthology/suite/run.py
kshtsk: dropped

teuthology/suite/run.py
kshtsk: but @jdurgin @idryomov nevertheless I changed it, I like more to have `.teuthology` file, this is a standard practice for modern tools to put repo configuration, like git, jenkins, travis, etc. And I need more configuration options rather than teuthology_branch in it.

idryomov: Just to be clear, I didn't object to having a *real* override file.  What I picked on was the fact that `.teuthology` looked like an override file but only handled one option.

At a high level, there are two ways to do this: either something really specific to `teuthology_branch` option or a general solution that would handle all options.  Your original approach was kind of in the middle: it had the appearance of a general solution but couldn't handle other options.

teuthology/suite/__init__.py
```diff
@@ -25,7 +25,6 @@ def override_arg_defaults(name, default, env=os.environ):
         '--suite-repo'        : 'TEUTH_SUITE_REPO',
         '--ceph-branch'       : 'TEUTH_CEPH_BRANCH',
         '--suite-branch'      : 'TEUTH_SUITE_BRANCH',
-        '--teuthology-branch' : 'TEUTH_BRANCH',
```
idryomov, line 28: I'm probably missing something obvious, but what is the reason for getting rid of this and doing os.environ.get() later?

(Same question for `default_teuthology_branch` above.)

teuthology/suite/run.py
```diff
@@ -196,20 +196,55 @@ def choose_ceph_version(self, ceph_hash):
...
+                    if teuthology_branch:
+                        log.debug(
+                            'The teuthology branch is overridden with %s',
+                                                                teuthology_branch)
+                    else:
+                        log.warning(
+                            'The teuthology branch config is empty, skipping')
+
+            if not teuthology_branch:
+                teuthology_branch = config.get('teuthology_branch', 'master')
```
idryomov, line 247: Should the branch be checked for existence here?  It looks like only `--teuthology-branch` command line argument is checked and not `TEUTH_BRANCH`, `.teuthology_branch` or the one from the configuration file.

teuthology/suite/__init__.py
kshtsk: this overall code and `default_teuthology_branch` was introduced to show the default value for `-t` option, but since the algorithm of determination (which is described in the `-t`'s help now)  is changed and in some cases requires to literally checkout suite-repo, there is no sense to do that, because it will confuse the user.

teuthology/suite/run.py
kshtsk: The teuthology-branch is checked later at line 213 (new 248):
```
teuthology_hash = util.git_ls_remote(
             'teuthology',
             teuthology_branch
        )
        if not teuthology_hash:
            exc = BranchNotFoundError(teuthology_branch, build_git_url('teuthology'))
            util.schedule_fail(message=str(exc), name=self.name)
```

idryomov: Then is there a point in checking it separately for `--teuthology-branch` above (lines 223-225)?

kshtsk: This is previous behavior and was thinking to delete that code, but I was shy to do it.

kshtsk: Dropped the code.

idryomov:
Hrm, why is master protected to the point that not everyone can merge?

kshtsk:
> Hrm, why is master protected to the point that not everyone can merge? This didn't use to be the case and not a practice we follow in other repos...

as far as I understand the history, only admins allowed, because not everyone wants take responsibility of breaking test infrastructure in sepia lab?

jdurgin:
@idryomov added you to the list of folks who can merge. perhaps we should revisit that and allow more broad access again

kshtsk:
Now we need:
1. Follow up commit to ceph/master the file qa/.teuthology_branch with the contents:
```
master
```
2. Ask @djgalloway to add `teuthology_branch: py2` to `/etc/teuthology.yaml`

kshtsk:
> Step 2 is done.

The order is important, sorry, forgot to notice about that.

djgalloway:
Uncommented now that https://github.com/ceph/ceph/pull/35349 is merged.

kshtsk:
nice, will see how it goes

PR #1493, opened 2020-05-26
setup.py: add toml for qa/tasks/cephadm.py

sebastian-philipp:

### DIFF
--- a/requirements2.txt
+++ b/requirements2.txt
@@ -102,6 +102,7 @@ scandir==1.8              # via pathlib2
 simplejson==3.10.0        # via osc-lib, python-cinderclient, python-neutronclient, python-novaclient
 six==1.14.0               # via bcrypt, cliff, cmd2, configobj, cryptography, debtcollector, keystoneauth1, more-itertools, openstacksdk, osc-lib, oslo.config, oslo.i18n, oslo.serialization, oslo.utils, pathlib2, pip-tools, pynacl, pyopenssl, pytest, python-cinderclient, python-dateutil, python-glanceclient, python-keystoneclient, python-neutronclient, python-novaclient, python-openstackclient, stevedore, teuthology, tox, warlock
 stevedore==1.21.0         # via cliff, keystoneauth1, openstacksdk, osc-lib, oslo.config, python-keystoneclient
+toml==0.10.1              # via teuthology
 tox==3.0.0                # via teuthology
 typing==3.7.4.1           # via apache-libcloud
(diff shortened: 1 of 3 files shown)


tchaikov:
@susebot run deploy

tchaikov:
http://pulpito.front.sepia.ceph.com/swagner-2020-05-26_11:34:30-rados:cephadm-wip-swagner-testing-2020-05-25-0943-distro-basic-smithi/

sebastian-philipp:
http://pulpito.ceph.com/swagner-2020-05-26_11:58:14-rados:cephadm-wip-swagner-testing-2020-05-25-0943-distro-basic-smithi/

toabctl:
@sebastian-philipp  Out of interest - why does cephadm wants to modify the registries.conf file? This file might be under control of other systems (like ceph-salt).

sebastian-philipp:
> @sebastian-philipp Out of interest - why does cephadm wants to modify the registries.conf file? This file might be under control of other systems (like ceph-salt).

we need to add a caching registry proxy (aka "mirror"), cause dockerhub.io is blocking our requests due to rate limiting.

kshtsk:
@sebastian-philipp what about adding "podman" module to teuthology code base instead of adding some strange stuff to cephadm?

kshtsk:
or is this more related to docker... I can see that on opensuse-15.1 the conf file you're modifying is provided by `libcontainers-common`

sebastian-philipp:
> or is this more related to docker... I can see that on opensuse-15.1 the conf file you're modifying is provided by `libcontainers-common`

yes, I need to patch the existing docker.io registry entry. :+1: for moving this to a different location. But my top priority is to make the qa runs in  sepia working again. there is a lot that we need to refactor here.

tchaikov:
i don't think the difference docker versus podman makes sense here. what we are after is a way to talk to a mirror registry. the client side can be podman or docker.

kshtsk:
> i don't think the difference between docker and podman makes sense here. what we are after is a way to talk to a mirror registry. the client side can be podman or docker.

right, this is why it is better to use `containers` as shorthand for `containers_common` instead of `docker.io`

toabctl:
> @sebastian-philipp Out of interest - why does cephadm wants to modify the registries.conf file? This file might be under control of other systems (like ceph-salt).

sorry for the noise. I thought this is ceph/ceph , not ceph/teuthology. so makes sense now.

PR #1495, opened 2020-05-27
provision/pelagos: improve logging

kshtsk:

### DIFF
--- a/teuthology/provision/pelagos.py
+++ b/teuthology/provision/pelagos.py
@@ -79,27 +79,29 @@ class Pelagos(object):
             raise RuntimeError("Pelagos is not configured!")
         location = None
         try:
+            params=dict(os=self.os_name, node=self.name)
             response = self.do_request('node/provision',
-                                      data={'os': self.os_name,
-                                            'node': self.name},
-                                      method='POST')
+                                data=params, method='POST')
             location = response.headers.get('Location')
-            self.log.info("Waiting for deploy to finish")
-            self.log.info("Observe location: '%s'", location)
+            self.log.debug("provision task: '%s'", location)
+            # gracefully wait till provision task gets created on pelagos
             time.sleep(2)
-            with safe_while(sleep=15, tries=60) as proceed:
+            self.log.info("Waiting for deploy to finish")
+            sleep_time=15
+            with safe_while(sleep=sleep_time, tries=60) as proceed:
                 while proceed():
                     if not self.is_task_active(location):
                         break
+                    self.log.info('Sleeping %s seconds' % sleep_time)
         except Exception as e:
             if location:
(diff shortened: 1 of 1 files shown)


ryg-:
Looks good to me, please commit

kshtsk:
@susebot run deploy

teuthology/provision/pelagos.py
```diff
@@ -79,36 +79,39 @@ def create(self):
...
+                    if self.deploy_task_active(location):
+                        self.log.info('Sleeping %s seconds' % sleep_time)
+                        continue
+                    break
         except Exception as e:
             if location:
                 self.cancel_deploy_task(location)
             else:
-                self.log.error("No task started")
+                self.log.error("Failed to start deploy tasks!")
```
toabctl, line 101: would it make sense to log the exeception here, too? like `self.log.exception("Failed to start deploy tasks!")`

teuthology/provision/pelagos.py
```diff
@@ -79,36 +79,39 @@ def create(self):
...
-            raise Exception("Provisioning failed")
+            raise Exception("Deploy failed")
         return self.task_status_response
 
     def cancel_deploy_task(self,  task_id):
         # TODO implement it
         return
 
-    def is_task_active(self, task_url):
+    def deploy_task_active(self, task_url):
```
toabctl, line 112: rename is unrelated to the commit message (which is about improving logging)

teuthology/provision/pelagos.py
```diff
@@ -117,41 +120,43 @@ def is_task_active(self, task_url):
...
                 return True
         return False
 
-    def do_request(self, url_suffix, url="" , data=None, method='GET', verify=True):
-        """
-        A convenience method to submit a request to the Pelagos server
-        :param url_suffix: The portion of the URL to append to the endpoint,
-                           e.g.  '/system/info'
-        :param data: Optional JSON data to submit with the request
+    def do_request(self, api='', url='', data=None, method='GET', verify=True):
```
toabctl, line 133: same here - unrelated to "improve logging"

teuthology/provision/pelagos.py
kshtsk: The exception is raised again next line, so... I think it should be decided by caller.

toabctl: yeah. but you are logging just half of the information. when you need to debug it, you'll be happy to know **what** exception was raised. especially because it's just catching the generic `Exception`

kshtsk: It is not me, it is author's choice. I was not going to polish the code, just to improve it no my taste. Just giving you a chance to improve it even more with your own PR.

kshtsk:
@toabctl thanks

PR #1498, opened 2020-05-27
console.py: Wait longer for IPMI timeouts

djgalloway:
The smithi have really crappy/flaky BMCs.  I occasionally see `'wait for power on' reached maximum tries (5) after waiting for 20.0 seconds` but when I go to check the BMC, it just needed some more time.  So let's give it more time.


djgalloway:
@susebot deploy?

PR #1499, opened 2020-05-28
orchestra/run: py2 fix for copy_and_close

sebastian-philipp:
needed by  https://github.com/ceph/ceph/pull/35260

### DIFF
--- a/teuthology/orchestra/test/test_run.py
+++ b/teuthology/orchestra/test/test_run.py
@@ -6,7 +6,7 @@ import socket
 from mock import MagicMock, patch
 from pytest import raises
 
-from six import ensure_str, ensure_binary
+from six import ensure_str, ensure_binary, PY2
 
 from teuthology.orchestra import run
 from teuthology.exceptions import (CommandCrashedError, CommandFailedError,
@@ -259,6 +259,14 @@ class TestRun(object):
         code = proc.wait()
         assert code == 0
         assert proc.exitstatus == 0
+
+    def test_copy_and_close(self):
+        run.copy_and_close(None, MagicMock())
+        run.copy_and_close('', MagicMock())
+        run.copy_and_close(b'', MagicMock())
+        if PY2:
+            run.copy_and_close(u'', MagicMock())
+
 
 
 class TestQuote(object):
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -328,6 +328,8 @@ def copy_and_close(src, fdst):
                 src = io.StringIO(src)
             else:
                 src = io.BytesIO(src)
+        elif not PY3 and isinstance(src, unicode):  # noqa: F821
+            src = io.StringIO(src)
         shutil.copyfileobj(src, fdst)
     fdst.close()
 


kshtsk:
Please no! Don't use io.StringIO for py2

sebastian-philipp:
> Please no! Don't use io.StringIO for py2, use io.BytesIO on caller side.

fair enough. Will do. If I may ask: why?

kshtsk:
> > Please no! Don't use io.StringIO for py2, use io.BytesIO on caller side.
> 
> fair enough. Will do. If I may ask: why?

Same question, why do you need to introduce unicode usage, where it is not required?
Also, it is very risky to make changes like this to core functionality which is tested only well for bytes on py2, I can understand why you guys might want to use StringIO on py3 since it is standard type `str`, but for py2 it is unicode, which makes only troubles.

sebastian-philipp:
Cause the table of supported inputs of `copy_and_close` looks like so:

| *        | str           | unicode      | bytes       |
| -------- | ------------- | ------------ | ----------- |
| Py3      | supported     |  same as str | supported   |
| Py2      | supported     |  **missing** | supported   |

This PR just fills the missing gap

kshtsk:
> Cause the table of supported inputs of `copy_and_close` looks like so:
> * 	str 	unicode 	bytes
> Py3 	supported 	same as str 	supported
> Py2 	supported 	**missing** 	supported
> 
> This PR just fills the missing gap

We do not need to support py2 code, and we don't want to spend much time for this, it is now only in maintenance mode, it highly desired to switch all branches to py3.
The unicode py2 is something ugly thing, you can't even check it's type correctly:
```
>>> isinstance('a', str)
True
>>> isinstance('a', bytes)
False
>>> isinstance(u'a', bytes)
False
>>> isinstance(u'a', str)
True
>>> isinstance(u'a', unicode)
NameError: name 'unicode' is not defined

```

kshtsk:
> This PR just fills the missing gap

it is only introduce more troubles than you think, for example, you need to fix the tests, this will increase divergence with py3 branch, which will complicate backports, I'm not sure you will want to spend time to support this branch?

kshtsk:
@sebastian-philipp if you want to go this way and take responsibility for supporting py2 branch. Here you go. Removing DNM label ;-)

teuthology/orchestra/test/test_run.py
```diff
@@ -6,7 +6,7 @@
 from mock import MagicMock, patch
 from pytest import raises
 
-from six import ensure_str, ensure_binary
+from six import ensure_str, ensure_binary, PY2
```
kshtsk, line 9: The PY2 should go from teuthology.util.compat as it is in the run.

sebastian-philipp:
> @sebastian-philipp if you want to go this way and take responsibility for supporting py2 branch. Here you go. Removing DNM label ;-)

Ok, here is the alternative to this PR: 

https://github.com/ceph/ceph/pull/35260/commits/9b4d9554121214623be4d5a5e6b7d9180acaf618

Both are ugly, but I have the impression that making the change here is a bit less intrusive.

kshtsk:
> Both are ugly, but I have the impression that making the change here is a bit less intrusive.

It is mistaken feeling ;-) especially in the run methods :-D

sebastian-philipp:
@kshtsk this PR is no longer a requirement. feel free to close this one now.

kshtsk:
@susebot run deploy

PR #1501, opened 2020-05-29
teuthology/config: Add container_images to site configs.

sebastian-philipp:
Every teuthology installation needs to have a different mirror for docker.io and
for the CI images. This means we shoudn't put the configs into any ceph git repos,
but instead in the site configs.

Later, if this is also backported to py2, we can then change the container sources for all branches at once.

See https://github.com/ceph/ceph/pull/35308 for the ceph side

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -242,3 +242,15 @@ Here is a sample configuration with many of the options set and documented::
     pelagos:
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Defines the bases for container images
+    container_images:
+      # This is a read-only mirror to limit the number of requests for downloading
+      # prometheus, grafana, node-exporter and alertmanager. Typically, this is a mirror
+      # for docker.io, but could be a mirror for any search registry, as the images are
+      # referenced without an explicit registry name.
+      base_registry_mirror: 'vossi04.front.sepia.ceph.com:5000'
+
+      # Container image name for CI images.
+      ci_image: 'quay.io/ceph-ci/ceph'
+
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -186,6 +186,10 @@ class TeuthologyConfig(YamlConfig):
             },
         },
         'sleep_before_teardown': 0,
+        'container_images': {
+            'base_registry_mirror': 'vossi04.front.sepia.ceph.com:5000',
+            'ci_image': 'quay.io/ceph-ci/ceph'
+        }
     }
 
     def __init__(self, yaml_path=None):


docs/siteconfig.rst
```diff
@@ -242,3 +242,15 @@ Here is a sample configuration with many of the options set and documented::
     pelagos:
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Defines the bases for container images
+    container_images:
```
kshtsk, line 247: Change this to `containers`

docs/siteconfig.rst
```diff
@@ -242,3 +242,15 @@ Here is a sample configuration with many of the options set and documented::
...
+      # This is a read-only mirror to limit the number of requests for downloading
+      # prometheus, grafana, node-exporter and alertmanager. Typically, this is a mirror
+      # for docker.io, but could be a mirror for any search registry, as the images are
+      # referenced without an explicit registry name.
+      base_registry_mirror: 'vossi04.front.sepia.ceph.com:5000'
+
+      # Container image name for CI images.
+      ci_image: 'quay.io/ceph-ci/ceph'
```
kshtsk, line 255: `test_image`?

docs/siteconfig.rst
```diff
@@ -242,3 +242,15 @@ Here is a sample configuration with many of the options set and documented::
...
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Defines the bases for container images
+    container_images:
+      # This is a read-only mirror to limit the number of requests for downloading
+      # prometheus, grafana, node-exporter and alertmanager. Typically, this is a mirror
+      # for docker.io, but could be a mirror for any search registry, as the images are
+      # referenced without an explicit registry name.
+      base_registry_mirror: 'vossi04.front.sepia.ceph.com:5000'
```
kshtsk, line 252: Just `registry`?

kshtsk:
Since you're moving the containers config to teuthology, I think it is strange to have the code in ceph, maybe it is time to introduce `containers` module to teuthology?

docs/siteconfig.rst
sebastian-philipp: well, this section doesn't deal with containers, but container images

docs/siteconfig.rst
sebastian-philipp: * it's not a registry, but a mirror of a registry
* it can also contain a port

docs/siteconfig.rst
kshtsk: container images are related directly to containers, because the image will be used with containers.

kshtsk:
I think having config data in teuthology for some specific test which is not reachable from teuthology is absolutely a mess, especially in the root namespace. Teuthology should contain code which can be reusable by any caller.

For now it looks like teuthology has nothing todo with containers, because it is not used by anything in teuthology code base. The reason only can appear if you move containers-common config file as a module to teuthology, and use it in your code like:
```
   from teuthology.orchestra.remote.cluster import set_mirror
   set_mirror(cluster, mirror_url)
```
That would make sense.

If you want some specific config base on code of specific code, like treat cephadm test as a separate entity from teuthology, maybe think about its own config file, like /etc/cephadm.conf?

docs/siteconfig.rst
kshtsk: 1. mirror is an equivalent of registry, if there is a real distinction, name it `registry_mirror_host`.
2. you can see another example of `_host` in the config which also have ports and protocol, it is used for concise notation, we may follow the style of teuthology config I guess.

kshtsk: could you explain why `base_` prefix required?

docs/siteconfig.rst
```diff
@@ -242,3 +242,15 @@ Here is a sample configuration with many of the options set and documented::
     pelagos:
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Defines the bases for container images
+    container_images:
+      # This is a read-only mirror to limit the number of requests for downloading
+      # prometheus, grafana, node-exporter and alertmanager. Typically, this is a mirror
+      # for docker.io, but could be a mirror for any search registry, as the images are
+      # referenced without an explicit registry name.
```
kshtsk, line 251: This overall comment is useless here.

docs/siteconfig.rst
sebastian-philipp: > could you explain why `base_` prefix required?

Right. it should be changed into a dict

```yaml
container_images:
  registry_mirrors:
    docker.io: 'vossi04.front.sepia.ceph.com:5000'
```

sebastian-philipp:
> If you want some specific config base on code of specific code, like treat cephadm test as a separate entity from teuthology, maybe think about its own config file, like /etc/cephadm.conf?

ok, but then how do I access this from the cephadm task?

docs/siteconfig.rst
tchaikov: i am not sure what you want to explain here. but i am sure`container_images` is more correct in this context. `base_registry_mirror` defines where we pull the images.

docs/siteconfig.rst
tchaikov: i think `ci_image` is better than `image` or `test_image`. in the context of testing ceph-ci branch, `ci_image` can be easily connected to `ceph-ci`. but i think `ceph_ci_image` might be better.

kshtsk:
> > I think having config data in teuthology for some specific test which is not reachable from teuthology is absolutely a mess, especially in the root namespace. Teuthology should contain code which can be reusable by any caller.
> 
> at the first glance, it might seems to be. but we should not encode lab specific settings in the test.
> 

We should not encode test specific stuff in teuthology.config. I say more, I don't like the idea of having toml in setup.py, which is not a requiremento of teuthology as a framework.

> > For now it looks like teuthology has nothing todo with containers, because it is not used by anything in teuthology code base. The reason only can appear if you move containers-common config file as a module to teuthology, and use it in your code like:
> > ```
> >    from teuthology.orchestra.cluster import set_mirror
> >    set_container_mirror(cluster, mirror_url)
> > ```
> > 
> > 
> > That would make sense.
> > If you want some specific config base on code of specific code, like treat cephadm test as a separate entity from teuthology, maybe think about its own config file, like /etc/cephadm.conf?
> 
> this is overkill. and in future, we will deploy and test crimson using cephadm as well. so i think `teuthology.yaml` is a compromise which sucks less.
I don't agree this is a compromise, the compromise is using cephadm.conf 

I don't like the idea of making recyclebin from teuthology.yaml, if you really want to go this way, then it should go like this in teuthology:
```
tasks: # this is not coincedent, it corresponds to module name `tasks` in ceph `qa`
    cephadm:
         #add whatever tasks specific config you want    
```
or follow overrides style:
```
overrides:
    cephadm:
         #add whatever tasks specific config you want    
```
But this is a new mechanism which should be properly documented in teuthology.

sebastian-philipp:
> I don't like the idea of having toml in setup.py, which is not a requiremento of teuthology as a framework.

A good way to avoid this would be to make ceph/qa a dedicated python package with it's own dependencies.

kshtsk:
> > I don't like the idea of having toml in setup.py, which is not a requiremento of teuthology as a framework.
> 
> A good way to avoid this would be to make ceph/qa a dedicated python package with it's own dependencies.

I like this idea, I was actually, thinking about it, like adding a `qa/requirements.txt` which will try and use it after bootstrap by the worker, and this is another good feature for teuthology, not trivial though, since to make it correctly handled we need to refactor worker a bit :-) however workarounds also would do the deal.
The current problem is that the only way to figure out what is wrong with requirements.txt is to look at worker.log which is not obvious from pulpito.

kshtsk:
and off course the layout of the `cephadm` config file should be well documented in the cephadm.py task class, and it is not required to change teuthology to know anything about cephadm params, because it should not care about it at all.

docs/siteconfig.rst
kshtsk: > i am not sure what you want to explain here. but i am sure`container_images` is more correct in this context. `base_registry_mirror` defines where we pull the images.

I am trying to say that:
```
containers:
   registry: vossi04.front.sepia.ceph.com:5000
   image: quay.io/ceph-ci/ceph
```
is concise and self-explanation notation, in the context of test framework.
It is absolutely clear, that for running containers there will be used images, which supposed to be taken from the given registry. Will this be using a mirror or something other under the hood I absolutely don't care.

Where:
```
container_images:
   base_registry_mirror: vossi04.front.sepia.ceph.com:5000
   ci_image: quay.io/ceph-ci/ceph
```
Looks clumsy and awkward to me, and hurts my feeling of beauty. Also, there can be a confusion, when this ceph-ci images be used with ceph/ceph repo.
Regarding the given names for CI, CI is a process, it is not a job, or repo, or peace of a software, this is why it is not correct thing to call it like this. In classic theory the ci determined as a specific process of continuous runs of integration tests of mainline (shared branches among the developers) in opposite for private branches, to make sure the code is still deployable, it is usually more advanced testing that running tests locally or in private branches. This why I would avoid to over-exploit CI abbreviation.

docs/siteconfig.rst
kshtsk: Looking at `ceph_ci_image` I understand that it should be used only with ceph-ci repo, if I want another image mirror and another image names, it will loose connection very easily.
And there should be also present `image` without _ci_ which will use images in other cases rather then ci or what?

docs/siteconfig.rst
sebastian-philipp: > is concise and self-explanation notation

only, till you know the details. 

We have minimum three container registries for teuthology:

* `docker.io` hosting prometheus, grafana, node-exporter, alertmanager and the ceph images
* `vossi04.front.sepia.ceph.com:5000` is a read-only mirror for `docker.io`
* `quay.io` hosting the ceph-ci images.
* Another site will host prometheus, grafana, node-exporter, alertmanager and the ceph image on another registry, like e.g. `registry.suse.com`. Thus I named it the "base" registry. 

Calling the mirror registry simply `registry` is just plain wrong. 

Calling the ceph-ci image simply `image` is also plain wrong as we're not referencing the ceph image here.

kshtsk: > > is concise and self-explanation notation
> 
> only, till you know the details.
> 
> We have minimum three container registries for teuthology:
> 
> * `docker.io` hosting prometheus, grafana, node-exporter, alertmanager and the ceph images
> * `vossi04.front.sepia.ceph.com:5000` is a read-only mirror for `docker.io`
> * `quay.io` hosting the ceph-ci images.
> * Another site will host prometheus, grafana, node-exporter, alertmanager and the ceph image on another registry, like e.g. `registry.suse.com`. Thus I named it the "base" registry.
> 
> Calling the mirror registry simply `registry` is just plain wrong.

Can those three registries be used simultaneously at the given moment of time? If no, for the code it is useless to call differently. So it is not wrong. Each of them are just "variable value".

> Calling the ceph-ci image simply `image` is also plain wrong as we're not referencing the ceph image here.
> 

Even if the image ci, it is still the image.

> > Regarding the given names for CI, CI is a process
> 
> it's also a given name to the `ceph-ci` image. I'm not referencing the process, but the name of the image, which is (for better or worse) called `ceph-ci`

In general you can refer to any image in cephadm test, the code itself does not care is it built from ceph-ci or on ceph.

Anyways, this discussion is useless, because in any case the test specific config should not be introduced to teuthology.conf as its own placeholder. It is important also because different tests can have a conflict.

I already mentioned the alternative way you either use a separate file `cephadm.conf` (or whatever you call), or use specific test subsection like:
```
tests:
    cephadm:
```
or
```
overrides:
    cephadm:
```
In both cases you have to put valuable description in ceph module __doc__ explaining what are those parameters and how to use it.
No `teuthology/config.py` changes, hardcoded mirror locations or "unknown" test config should be provided.

teuthology/config.py
```diff
@@ -186,6 +186,10 @@ class TeuthologyConfig(YamlConfig):
             },
         },
         'sleep_before_teardown': 0,
+        'container_images': {
+            'base_registry_mirror': 'vossi04.front.sepia.ceph.com:5000',
+            'ci_image': 'quay.io/ceph-ci/ceph'
+        }
```
kshtsk, line 192: As we figured out earlier this is test specific config, it should not be a part of teuthology code base.

sebastian-philipp:
@tchaikov turns out, I also need this for https://github.com/ceph/ceph/blob/master/qa/suites/rados/cephadm/workunits/task/test_adoption.yaml which is not using the qa/tasks/cephadm.py

kshtsk:
> @tchaikov turns out, I also need this for https://github.com/ceph/ceph/blob/master/qa/suites/rados/cephadm/workunits/task/test_adoption.yaml which is not using the qa/tasks/cephadm.py

@sebastian-philipp this is test specific configuration and should be managed on the tasks side.

tchaikov:
i still think it's a compromise. this is probably not the best solution, but it's the most  practical one we have now. i'd be happy to review your PR, if you don't like it.

teuthology/config.py
tchaikov: another way to look these settings is see it as a freestyle setting which can be consumed by any test(s). we could put these settings in their own namespace though, if that's better.

kshtsk:
> i still think it's a compromise. this is probably not the best solution, but it's the most practical one we have now. i'd be happy to review your PR, if you don't like it.

As I said, there is no need to merge these changes to teuthology in order to get cephadm task working.
Here is an example of patch you may work in cephadm: https://github.com/ceph/ceph/pull/35395

kshtsk:
Also you can have any naming you want withing `overrides/cephadm` section, that would be flexible enough for your needs, just don't add anything not related to the teuthology core code base to `teuthology/config.py`.

PR #1502, opened 2020-06-03
tox: Avoid system site-packages

zmc:
This fixes a failure to locate `py.test` using python3 on MacOS:
```
py3 runtests: commands[0] | py.test --cov=teuthology --cov-report=term -v teuthology scripts
  File "/Users/zack/dev/teuthology/.tox/py3/bin/py.test", line 5, in <module>
    from pytest import main
ModuleNotFoundError: No module named 'pytest'
```


tchaikov:
@susebot run deploy

PR #1503, opened 2020-06-04
teuthology/misc: Valgrind exits on first detected error.

rzarzynski:
Valgrind v3.14 brought the `--exit-on-first-error` option which causes termination of the process on first encountered error. Although it does nothing for memory leaks, it shouldhelp with pin-pointing e.g. `UninitValue` without imposing the costly `--track-origins=yes`.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1075,6 +1075,9 @@ def get_valgrind_args(testdir, name, preamble, v):
             '--xml-file={vdir}/{n}.log'.format(vdir=val_path, n=name),
             '--time-stamp=yes',
             '--vgdb=yes',
+            # at least Valgrind 3.14 is required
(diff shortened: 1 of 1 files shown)


jdurgin:
all valgrind suites use centos 8, where this is available

cbodley:
valgrind was disabled in the rgw suite for most of this pacific cycle as fallout from the python3 work. we're trying to turn it back on, but this change has made it really hard to get a full list of issues. radosgw doesn't even start, so we don't run any tests and get a single valgrind issue as a result

i really don't think this flag should be on by default. when necessary, you can add it to a specific suite in yaml:
```
valgrind:
  osd: [--tool=memcheck,--exit-on-first-error=yes,--error-exitcode=42]
```

PR #1505, opened 2020-06-08
bootstrap: python3 packages for fedora and redhat based systems

rakeshgm:
adding python3 packages for fedora and RedHat based systems. 

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -44,13 +44,18 @@ Linux)
             fi
         fi
         ;;
-    RedHatEnterpriseWorkstation|RedHatEnterpriseServer|CentOS)
-        for package in python2-pip python-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel; do
-	    if [ "$(rpm -q $package)" == "package $package is not installed" ]; then
-		missing="${missing:+$missing }$package"
-	    fi
+    RedHatEnterpriseWorkstation|RedHatEnterpriseServer|RedHatEnterprise|CentOS)
+        if [[ "$PYTHON" =~ "python2" ]]; then
+            deps=(python-pip python-devel python-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel)
+        else
+            deps=(python3-pip python3-devel python3-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel)
+        fi
+        for package in ${deps[@]}; do
+          if [ "$(rpm -q $package)" == "package $package is not installed" ]; then
+              missing="${missing:+$missing }$package"
+          fi
         done
-	if [ -n "$missing" ]; then
+        if [ -n "$missing" ]; then
             echo "$0: missing required packages:" 1>&2
             echo "$missing"
             if [ "$install" = true ]; then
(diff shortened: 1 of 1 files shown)


bootstrap
```diff
@@ -45,12 +45,17 @@ Linux)
         fi
         ;;
     RedHatEnterpriseWorkstation|RedHatEnterpriseServer|CentOS)
-        for package in python2-pip python-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel; do
-	    if [ "$(rpm -q $package)" == "package $package is not installed" ]; then
-		missing="${missing:+$missing }$package"
-	    fi
+        if [[ "$PYTHON" =~ "python2" ]]; then
+            packages=(python-pip python-devel python-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel)
```
tchaikov, line 49: what's the advantage of using array here?

bootstrap
```diff
@@ -45,12 +45,17 @@ Linux)
...
+        else
+            pacakges=(python3-pip python3-devel python3-virtualenv mariadb-devel libev-devel libvirt-devel libffi-devel)
+        fi
+        for package in  ${packages[@]}; do
```
tchaikov, line 53: please remove the extra space.

rakeshgm: done

bootstrap
rakeshgm: I don't this is an array here. I think this is a conditional construct. [ref here](http://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Conditional-Constructs).

tchaikov: i don't think i am talking about `if`. what i am referencing is the above line.
```sh
packages=(....)
```

> Arrays are assigned to using compound assignments of the form
> name=(value1 value2 … )

see https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Arrays

rakeshgm: ok. correct me If I am wrong. the pattern I followed is similar to that one where the comparison for previous distros already exists. [some thing like this](https://github.com/ceph/teuthology/blob/master/bootstrap#L24)

kshtsk: > what's the advantage of using array here?

Why not using array for lists? It is natural and allows to generalize iteration algorithms:
```
a=("abc" "a b c")

for i in "${a[@]}" ; do
  echo - $i
done

b='xyz "x y z"'
for i in $b ; do
  echo + $i
done
```
will give:
```
- abc
- a b c
+ xyz
+ "x
+ y
+ z"
```

Also, each element of array can be taken by index.

kshtsk:
@rakeshgm Overall change looks fine to me, I believe you tested this on fedora an rhel?

bootstrap
kshtsk: > ok. correct me If I am wrong. the pattern I followed is similar to that one where the comparison for previous distros already exists. [some thing like this](https://github.com/ceph/teuthology/blob/master/bootstrap#L24)

Maybe use the same pattern as you refer to, like use `$deps` instead `$packages`?

rakeshgm: done. changed `deps`

rakeshgm:
> @rakeshgm Overall change looks fine to me, I believe you tested this on fedora an rhel?

yes.

kshtsk:
@susebot run deploy

kshtsk:
The failure is not related to the change, though, scheduling another run...

kshtsk:
@susebot run deploy

bootstrap
vasukulkarni: Resolving, I think rakesh is following same patter what we have in the rest of the places in bootstrap.

PR #1515, opened 2020-06-19
drop requirements.in

tchaikov:
dependabot adds an entry in generated requirements.txt like
```
-e file:///home/dependabot/dependabot-updater/tmp/dependabot_20200617-72-1n8af4b  # via -r requirements.in
```

this is expected. as dependabot does not read docs/INSTALL.rst, hence
failed to apply `sed '/^-e / d'` to the generated file.

instead of teaching dependabot to read the manual, it's simpler
just ditch requirements.in .

see also #1367

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -68,9 +68,9 @@ Update Dependencies
 We track the dependencies using ``requirements.txt`` 
 for python2 and python3 respectively. These packages are tested, and should work 
 with teuthology. But if you want to bump up the versions of them, please use the 
-following command to update these files. For python3 ::
+following command to update these files::
 
-  pip-compile -qo- | sed '/^-e / d' > requirements.txt
+  ./update-requirements.sh
 
 Please upgrade pip-tool using following command ::
 
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -65,10 +65,9 @@ interacting with the services to schedule tests and to report the test results.
 Update Dependencies
 -------------------
 
-We track the dependencies using ``requirements.txt`` 
-for python2 and python3 respectively. These packages are tested, and should work 
-with teuthology. But if you want to bump up the versions of them, please use the 
-following command to update these files::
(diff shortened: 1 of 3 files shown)


docs/INSTALL.rst
```diff
@@ -70,7 +70,7 @@ for python2 and python3 respectively. These packages are tested, and should work
 with teuthology. But if you want to bump up the versions of them, please use the 
 following command to update these files. For python3 ::
 
-  pip-compile -qo- | sed '/^-e / d' > requirements.txt
+  echo "-e .[orchestra,test]" | pip-compile - -qo- | sed '/^-e / d' > requirements.txt
```
kshtsk, line 73: Instead of teaching users to read the docs, we could create a shorthand script for updating requirements, like `/update_requirements` or `/pip-compile`:
```
echo "-e .[orchestra,test]" | pip-compile - -qo- | sed '/^-e / d' > requirements.txt
```

kshtsk: More advanced script which can bypass arguments to pip-compile:
```
export DIR=$(cd $(dirname $BASH_SOURCE); echo $PWD)
cd $DIR
echo "-e .[orchestra,test]" | pip-compile $@ - -qo- | sed '/^-e / d' > requirements.txt

```

tchaikov: again, then he/she needs to read read document to understand that `update_requirements` or `pip-compile` should be used for generating `requirements.txt`. due to the activity level and popularity of this project, i don't think it's worthwhile.

kshtsk: I agree, but this makes much easier to copy-paste pip-compile, I found myself boring to got to docs and search for the right command to run, instead it is much easier to recall there is a script in the root of the repo.

tchaikov: @kshtsk updated and repushed.

kshtsk:
I guess it should be rebased to master, since run.py fix is merged there.

tchaikov:
@kshtsk rebased and repushed.

kshtsk:
@susebot run deploy

PR #1517, opened 2020-06-19
py2 backport: teuthology/task/clock: increase timeout

kshtsk:
Increase timeout because 1 minute is not always enough
to configure ntp.

(cherry picked from commit a84c87eae280c7c692f8ecc83f5647353834938d)


kshtsk:
@susebot run deploy

PR #1519, opened 2020-06-23
teuthology/suite: use rhel8 for querying package version

tchaikov:
as we don't build packages for el7 on master anymore. but we do build
packages for nautilus and octopus.


tchaikov:
tested at http://pulpito.ceph.com/kchai-2020-06-23_01:29:10-rados-master-distro-basic-smithi/

PR #1520, opened 2020-06-23
suite/run.py: check config['verify_ceph_hash'] before verifying ceph …

tchaikov:
…packages

there is chance that we don't have ceph packages built for the
combination of specified os_type and os_version, for instance, in the
case of rados/thash_old_clients, older ceph clients are installed on
el7, but the ceph cluster is deployed using cephadm, which in turn pull
ceph container images built using the ceph being tested and el8.

since we've dropped the build of master on el7, there is no need to
verify if ceph package is available if cephadm is used for deploying the
cluster.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -170,7 +170,13 @@ def get_initial_tasks(lock, config, machine_type):
     overrides = config.get('overrides', {})
     having_repos = ('repo' in config.get('install', {}) or
                     'repo' in overrides.get('install', {}))
-    if not ('redhat' in config or having_repos):
+    if 'redhat' in config:
+        pass
+    elif having_repos:
+        pass
+    elif not config.get('verify_ceph_hash', True):
+        pass
+    else:
         init_tasks += [
             {'internal.check_packages': None},
             {'internal.buildpackages_prep': None},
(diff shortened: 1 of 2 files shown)


tchaikov:
being tested at http://pulpito.ceph.com/kchai-2020-06-23_15:15:59-rados-wip-kefu-testing-2020-06-23-1353-distro-basic-smithi/

tchaikov:
> We should add a check for cephadm to see whether the containers it uses exist as well, in a future PR

i checked `podman manifest` and `docker manifest inspect`. `podman manifest` was added couple months ago. i think we need wait a bit, before that feature is packaged in downstream. while `docker manifest inspect` requires user to enable experimental features. guess the same applies to it.

so, in short, let's revisit it after, probably 6 months.

kshtsk:
@tchaikov this PR is breaking install.repos mechanism

teuthology/run.py
```diff
@@ -167,10 +167,16 @@ def validate_tasks(config):
 
 def get_initial_tasks(lock, config, machine_type):
     init_tasks = []
-    overrides = config.get('overrides')
-    having_repos = ('install' in config and 'repos' in config.get('install')) \
-        or (overrides and 'install' in overrides and 'repos' in overrides.get('install'))
-    if not ('redhat' in config or having_repos):
+    overrides = config.get('overrides', {})
+    having_repos = ('repo' in config.get('install', {}) or
```
kshtsk, line 171: Here is unfortunate typo, this should be 'repos' instead of 'repo'.

kshtsk:
Related fix #1521

PR #1523, opened 2020-06-24
task/install: add check for repos

kshtsk:


teuthology/task/install/__init__.py
```diff
@@ -441,6 +441,18 @@ def task(ctx, config):
              - url: "http://location.repo"
                name: "ceph_repo"
 
+    Add repo before trying to install any package, the shaman
+    related tasks will be ignored.
```
smithfarm, line 445: Add repo before trying to install any package (all Shaman-related tasks will be ignored):

teuthology/task/install/__init__.py
```diff
@@ -441,6 +441,18 @@ def task(ctx, config):
...
+
+    Note: This only valid for SUSE based distro, other systems
+    are welcome to support.
```
smithfarm, line 448: Note: This is valid for SUSE-based distros only: other distros are welcome to support.

kshtsk:
@smithfarm suggestions are applied and the commit is re-pushed.

PR #1525, opened 2020-06-25
orchestra/daemon/cephadmunit: add journalctl suffix

kshtsk:
The logger is running in background and the output is mixed
with other loggers. Adding a recognizable suffix will
help to distinguish logging threads.

### DIFF
--- a/teuthology/orchestra/daemon/cephadmunit.py
+++ b/teuthology/orchestra/daemon/cephadmunit.py
@@ -50,7 +50,7 @@ class CephadmUnit(DaemonState):
                   '-u',
                   'ceph-%s@%s.service' % (self.fsid, name)
             ],
-            logger=logging.getLogger(self.cluster + '.' + name),
+            logger=logging.getLogger('journalctl@' + self.cluster + '.' + name),
             label=name,
             wait=False,
             check_status=False,


kshtsk:
@susebot run deploy

smithfarm:
@sebastian-philipp Does this look good to you?

teuthology/orchestra/daemon/cephadmunit.py
```diff
@@ -50,7 +50,7 @@ def _start_logger(self):
                   '-u',
                   'ceph-%s@%s.service' % (self.fsid, name)
             ],
-            logger=logging.getLogger(self.cluster + '.' + name),
+            logger=logging.getLogger('journalctl@' + self.cluster + '.' + name),
```
sebastian-philipp, line 53: The log prefix is already pretty big:

`2020-06-29T10:22:09.539 INFO:ceph.mon.c.smithi105.stdout:Jun 29 10:22:09 smithi105 bash[26241]:`

What's the problem you're facing? As we don't have any ceph process running outside of containers, I'd thought that something like `ceph.mon.c` would be enough to distinguish teuthology log output from daemon output.

kshtsk: @sebastian-philipp `ceph.mon.c` is service name which is useless for separating all journalctl threads, and it is differs from suite to suite, so we want to have more standard way to remove the logger noise.

sebastian-philipp: interesting use case. For me, those are the interesting logs and everything else is noise to some degree :smile: . 

```suggestion
            logger=logging.getLogger('cluster.' + self.cluster + '.' + name),
```

works for you?

kshtsk: @sebastian-philipp no, I prefer `journalctl@` since it makes it clear what is the source of the log. The `cluster.` does not makes sense to me at all, we already have `ceph.` prefix which makes the logs indistinct. Especially when they are generated even after all ceph related tasks finished.

sebastian-philipp: > @sebastian-philipp no, I prefer `journalctl@` since it makes it clear what is the source of the log. 

Hm. isn't journald just a delivery mechanism and not a source?

> The `cluster.` does not makes sense to me at all, we already have `ceph.` prefix which makes the logs indistinct. 

indistinct from what? 

> Especially when they are generated even after all ceph related tasks finished. Particular case is when the job is hung or in sleep before teardown and there are tons of logs generated by journalctl background processes.

But then we're talking about logs of the ceph cluster, right? Hence `cluster.ceph`.

kshtsk: > > @sebastian-philipp no, I prefer `journalctl@` since it makes it clear what is the source of the log.
> 
> Hm. isn't journald just a delivery mechanism and not a source?
> 
Journal is not a delivery mechanism, the ssh is delivery mechanism, the `journalctl` remote process is literally generating the logs.

> > The `cluster.` does not makes sense to me at all, we already have `ceph.` prefix which makes the logs indistinct.
> 
> indistinct from what?
From logs from other commands of the main process.
> 
> > Especially when they are generated even after all ceph related tasks finished. Particular case is when the job is hung or in sleep before teardown and there are tons of logs generated by journalctl background processes.
> 
> But then we're talking about logs of the ceph cluster, right? Hence `cluster.ceph`.

Not sure I understand what you don't like from in the change and why do you think `cluster.ceph.mon.X` makes it more distinguishable than `ceph.mon.Y` from other commands.

I just want to `grep -v 'journalctl@'` or `grep 'journalctl@'` to hide or filter out those logs.

sebastian-philipp:
```suggestion
            logger=logging.getLogger('journald@' + self.cluster + '.' + name),
```

reading the logs is already painful enough. I don't wan the prefix to be any longer.

sebastian-philipp: image

kshtsk: > reading the logs is already painful enough. I don't wan the prefix to be any longer.

we are fighting for two letters now? :-D I would prefer to use journalctl since it is what is called on the remote hosts. Originally I was thinking about `logger@` or even `log@` to add as short _tag_ as possible, but having _journalctl_ is making a hint what to refer in the source code of teuthology.

kshtsk: Also I would say it is painful much more because of oneline json soup but not the single tag.

kshtsk: and you can just strip the `journalctl@` with a simple sed now.

PR #1531, opened 2020-07-06
teuthology-suite: allow dump job configs to a file instead of beanstalkd

kshtsk:
This will allow to see what job config contents can be submitted to beanstalkd by storing them into the separate file in yaml multi-document format.
Having `--queue-backend` option can be later used for queuing the jobs into alternative to beanstalkd service, for example, we can queue jobs to the paddles database directly, or any other.
The yaml file can be easily read and any job config can be taken from it and used with teuthology-run directly to run tests without involving beanstalkd and paddles.

### DIFF
--- a/scripts/schedule.py
+++ b/scripts/schedule.py
@@ -16,6 +16,11 @@ positional arguments:
 optional arguments:
   -h, --help                           Show this help message and exit
   -v, --verbose                        Be more verbose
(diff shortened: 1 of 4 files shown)


kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

PR #1532, opened 2020-07-07
teuthology: don't use git to define version for released teuthology

kshtsk:
After installing teuthology using pip there can be found error messages
in the log when running any command, for example 'teuthology --version':

  fatal: not a git repository (or any parent up to mount point /)
  Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
  Can't get version from git rev-parse Command '['git', 'rev-parse', '--short', 'HEAD']' returned non-zero exit status 128.
  1.0.0

### DIFF
--- a/teuthology/__init__.py
+++ b/teuthology/__init__.py
@@ -37,10 +37,15 @@ __version__ = '1.0.0'
 # do our best, but if it fails, continue with above
 
 try:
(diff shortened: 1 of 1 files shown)


kshtsk:
@susebot run deploy

PR #1533, opened 2020-07-07
repo_utils: exclude refs/for branches when ls-remote

kshtsk:
If we do not request refs/for branch specifically,
we should exclude them since sometimes when we try
to ls the 'master' branch the ls-remote returns
two branches 'refs/for/master' and 'refs/heads/master',
though the latter is expected only.

### DIFF
--- a/teuthology/repo_utils.py
+++ b/teuthology/repo_utils.py
@@ -62,6 +62,15 @@ def ls_remote(url, ref):
     """
     sha1 = None
     cmd = "git ls-remote {} {}".format(url, ref)
+    # workaround for the refs/for/master branch, sometimes happens when
+    # we run ls-remote against master and we get refs/for/master in
+    # first place instead of only refs/heads/master ref, for example:
+    #
+    # > git ls-remote https://github.com/ceph/ceph.git master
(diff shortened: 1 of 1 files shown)


vasukulkarni:
There is -h option to list only heads, I think that is better option.

kshtsk:
> There is -h option to list only heads, I think that is better option.
> 
> ```
> git ls-remote -h https://github.com/ceph/ceph.git master 
> 03dbeebab360b8b4e1bdf7d73f42552e1be88557	refs/heads/master
> ```

Maybe to use -h with non-refs branches makes sense.

neha-ojha:
@kshtsk The original problem has been fixed after deleting the invalid reference to `refs/for/master`
```
git ls-remote https://github.com/ceph/ceph.git master 
02fc51deb7f3f635fef41321da03283b165eb7ef	refs/heads/master
```

kshtsk:
@neha-ojha thanks, I let myself close this PR then...

kshtsk:
Since there are some failures in tests it requires more time, but since the issue is not happening after fixing github repo I close it, we can create another ticket with corresponding fix.

PR #1537, opened 2020-07-16
misc: upgrade sudo_write_file()

rishabh-d-dave:
Set `omit_sudo` to `False` and return the return value of `remote.run()`.

Related to `tasks.cephfs.mount.CephFSMount.write_file()` that is added on PR https://github.com/ceph/ceph/pull/32581. 

I intended to upgrade rest of the methods here later.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -549,7 +549,7 @@ def sudo_write_file(remote, path, data, perms=None, owner=None):
     owner_args = []
     if owner:
         owner_args = [run.Raw('&&'), 'sudo', 'chown', owner, path]
-    remote.run(
+    return remote.run(
         args=[
             'sudo',
             'sh',
@@ -557,6 +557,7 @@ def sudo_write_file(remote, path, data, perms=None, owner=None):
             'cat > ' + path,
         ] + owner_args + permargs,
         stdin=data,
+        omit_sudo=False
     )
 
 


kshtsk:
I guess we need to deprecate methods *misc.sudo_write_file* and *misc.write_file* in favor of *remote.write_file*

kshtsk:
Yeah, and we are back to our discussion of #1297

kshtsk:
Also, I would prefer to use 'dd' or 'tee' as a backend, since they have better syntax and easier to sudo.

teuthology/misc.py
```diff
@@ -549,14 +549,15 @@ def sudo_write_file(remote, path, data, perms=None, owner=None):
     owner_args = []
     if owner:
         owner_args = [run.Raw('&&'), 'sudo', 'chown', owner, path]
-    remote.run(
+    return remote.run(
```
kshtsk, line 552: Instead of this you need to use try-catch and check CommandFailedError.exitstatus value

kshtsk:
We should not use omit_sudo in teuthology code base, since it is not used. This commit bdb9d1d was merged without discussion and in my view is wrong solution. The core teuthology code does not know anything about external hack like vstart_runner.

rishabh-d-dave:
@kshtsk Without setting `omit_sudo` to `False` certain testcases would fail unconditionally.

kshtsk:
> @kshtsk Without setting `omit_sudo` to `False` certain testcases would fail unconditionally.

which test cases? I'm trying to grep it in teuthology code base:
```> grep -R omit_sudo teuthology
teuthology/orchestra/run.py:    # omit_sudo is used by vstart_runner.py
teuthology/orchestra/run.py:    omit_sudo=False
```
This looks like dead code for teuthology remote.run, I could imaging it was added by some purpose of external tools which are using teuthology incorrectly.

rishabh-d-dave:
> > @kshtsk Without setting `omit_sudo` to `False` certain testcases would fail unconditionally.
> 
> @rishabh-d-dave which test cases? I'm trying to grep it in teuthology code base:
> 
> ```

Sorry for the lack of context, I meant CephFS testcases. [mount.py](https://github.com/ceph/ceph/blob/master/qa/tasks/cephfs/mount.py) uses omit_sudo pretty often and mount.py is run for all CephFS testcases

> > grep -R omit_sudo teuthology
> teuthology/orchestra/run.py:    # omit_sudo is used by vstart_runner.py
> teuthology/orchestra/run.py:    omit_sudo=False
> ```
> 
> This looks like dead code for teuthology remote.run, I could imaging it was added by some purpose of external tools which are using teuthology incorrectly.

No, it isn't dead code; it was added to make `teuthology.orchestra.remote.run`'s signature same as `vstart_runner.LocalRemote.run`. This is vital to making CephFS QA compatible with both vstart_runner.py and teuthology. 

> And this is a very bad practice.

kshtsk:
Hi thanks for giving more context.

> > > @kshtsk Without setting `omit_sudo` to `False` certain testcases would fail unconditionally.
> > 
> > 
> > @rishabh-d-dave which test cases? I'm trying to grep it in teuthology code base:
> 
> Sorry for the lack of context, I meant CephFS testcases. [mount.py](https://github.com/ceph/ceph/blob/master/qa/tasks/cephfs/mount.py) uses omit_sudo pretty often and mount.py is run for all CephFS testcases

Do you have an example of jobs failing without `omit_sudo`?
> 
> > > grep -R omit_sudo teuthology
> > > teuthology/orchestra/run.py:    # omit_sudo is used by vstart_runner.py
> > > teuthology/orchestra/run.py:    omit_sudo=False
> > 
> > 
> > ```
> > 
> > This looks like dead code for teuthology remote.run, I could imaging it was added by some purpose of external tools which are using teuthology incorrectly.
> > ```
> 
> No, it isn't dead code; it was added to make `teuthology.orchestra.remote.run`'s signature same as `vstart_runner.LocalRemote.run`. This is vital to making CephFS QA compatible with both vstart_runner.py and teuthology.
> 

This is exactly a dead code for teuthology.orchestra.remote.run it is not used anyway in this method.
I have no idea why it is decided to align teuthology core methods with the code of vstart_runner.LocalRemote.run, it is wrong approach of adding arguments to function which is not used. It was a mistake by review to aprove this change to teuthology.
We need to workout a better solution.

> > And this is a very bad practice.

teuthology/misc.py
rishabh-d-dave: Hmm, agreed. But I would like to say that using try-except is not as elegant as -
```
    retval = remote.run(args=['some', 'cmd']).returncode
   self.assertEqual(retval, 1)
```

rishabh-d-dave:
> Hi thanks for giving more context.
> 
> > > > @kshtsk Without setting `omit_sudo` to `False` certain testcases would fail unconditionally.
> > > 
> > > 
> > > @rishabh-d-dave which test cases? I'm trying to grep it in teuthology code base:
> > 
> > 
> > Sorry for the lack of context, I meant CephFS testcases. [mount.py](https://github.com/ceph/ceph/blob/master/qa/tasks/cephfs/mount.py) uses omit_sudo pretty often and mount.py is run for all CephFS testcases
> 
> Do you have an example of jobs failing without `omit_sudo`?
> 

Not right now. I would need to look or maybe even reproduce a failure. It's pretty obvious though, for example, call on this line would run successfully with vstart_runner.py but not with teuthology if omit_sudo is remove teuth's remote.run's signature  - https://github.com/ceph/ceph/blob/master/qa/tasks/cephfs/mount.py#L118.

> > > > grep -R omit_sudo teuthology
> > > > teuthology/orchestra/run.py:    # omit_sudo is used by vstart_runner.py
> > > > teuthology/orchestra/run.py:    omit_sudo=False
> > > 
> > > 
> > > ```
> > > 
> > > This looks like dead code for teuthology remote.run, I could imaging it was added by some purpose of external tools which are using teuthology incorrectly.
> > > ```
> > 
> > 
> > No, it isn't dead code; it was added to make `teuthology.orchestra.remote.run`'s signature same as `vstart_runner.LocalRemote.run`. This is vital to making CephFS QA compatible with both vstart_runner.py and teuthology.
> 
> This is exactly a dead code for teuthology.orchestra.remote.run it is not used anyway in this method.

Agreed that it's not being used but it's vital to compatibility between teuthology and vstart_runner.py.

> I have no idea why it is decided to align teuthology core methods with the code of vstart_runner.LocalRemote.run, 

It was done to maintain compatibility for CephFS testscases between vstart_runner,py and teuthology.

> it is wrong approach of adding arguments to function which is not used. It was a mistake by review to aprove this change to teuthology.
> We need to workout a better solution.
> 

If there's a better solution, I'll be happy to get that done.

rishabh-d-dave:
> I guess we need to deprecate methods misc.sudo_write_file and misc.write_file in favor of remote.write_file

Do we have `remote.write_file`? I couldn't spot it. If you were proposing adding `remote.write_file`, I fully agree.

kshtsk:
> > I guess we need to deprecate methods misc.sudo_write_file and misc.write_file in favor of remote.write_file
> 
> Do we have `remote.write_file`? I couldn't spot it. If you were proposing adding `remote.write_file`, I fully agree.

created PR #1538 with it

kshtsk:
> Not right now. I would need to look or maybe even reproduce a failure. It's pretty obvious though, for example, call on this line would run successfully with vstart_runner.py but not with teuthology if omit_sudo is remove teuth's remote.run's signature - https://github.com/ceph/ceph/blob/master/qa/tasks/cephfs/mount.py#L118.
>

This looks like bugs in vstart_runner, or bad design with those: "We don't need no stinkin' sudo"
In many commands sudo is given in purpose, we can not selectively remove it from API because this generates new bugs.

kshtsk:
The vstart script is always run as root, this is why they wanted to remove 'sudo' from all commands, but this is slightly changes the logic in all commands. So instead of adding `omit_sudo` argument to many run calls we just need to remove erasure of 'sudo' from vstat:
```
         # We don't need no stinkin' sudo
        if omit_sudo:
            args = [a for a in args if a != "sudo"]
```

teuthology/misc.py
kshtsk: I'm sorry, but in my taste this is not super elegant. That's more like c-style coding, I don't mind, but we have OOP language which allows to use try-catch constructions. From other side, why do you need assertEqual which generating corresponding exception anyways, and it mask original exception which makes a step back from usability of issue investigation. Also in case of error remote.run will not return any object by default but will throw an exception which you have to handle. If you don't want exception to be occur, you need to pass special parameter to run:
```
        :param check_status: Whether to raise CommandFailedError on non-zero
                             exit status, and . Defaults to True. All signals
                             and connection loss are made to look like SIGHUP.
```
But this only usable if you want to ignore exceptions.
Third, when we use write_file function, we do not expect there is an error so we need to interrupt program, and either use sudo_write_file or we have some a bug definitely.

rishabh-d-dave:
> The vstart script is always run as root, this is why they wanted to remove 'sudo' from all commands,

Really? It's written so that it can be used without being the root user. So, is the case with vstart_runner.py -- we omit sudo so that permissions on files don't change and therefore it can be used _without_ being root.

> but this is slightly changes the logic in all commands. So instead of adding `omit_sudo` argument to many run calls we just need to remove erasure of 'sudo' from vstat:

I don't understand. Can you please elaborate on how removing erasure of 'sudo' is different from just omitting it?

rishabh-d-dave:
> > > I guess we need to deprecate methods misc.sudo_write_file and misc.write_file in favor of remote.write_file
> > 
> > 
> > Do we have `remote.write_file`? I couldn't spot it. If you were proposing adding `remote.write_file`, I fully agree.
> 
> created PR #1538 with it

Okay, thanks!

teuthology/misc.py
rishabh-d-dave: > I'm sorry, but in my taste this is not super elegant. 

Really? Compared to following -

```
try:
    remote.run(args=['some', 'cmd])
except CommandFailedError as e:
    self.assertEqual(e.existatus, 1)
```

I can use check_status, agreed but I still think passing the object representing the subprocess is better than returning void. For example, access to stdout and stderr can help better error diagnosing/more accurate test sometimes.

kshtsk: You do not need to have access to stdout or stderr if you are not going to analyze the errors programatically. For example in case of write_file there is no need to do this, since you are not going to continue execution, you just need to stop the test as fast as possible. All stdout and stderr can be found in the logs by anyone who read the docs.
See, the exit status is printed in output:
```
>>> from teuthology.orchestra.remote import Remote as r ; l=r('localhost')
>>> l.write_file('/tmp/bla/bla', 'blabla')
2020-07-17 17:51:26,461.461 INFO:teuthology.orchestra.remote:Trying to reconnect to host
2020-07-17 17:51:26,686.686 INFO:teuthology.orchestra.run.localhost:> true
2020-07-17 17:51:26,804.804 INFO:teuthology.orchestra.run.localhost:> set -ex
2020-07-17 17:51:26,804.804 INFO:teuthology.orchestra.run.localhost:> dd of=/tmp/bla/bla
2020-07-17 17:51:26,855.855 INFO:teuthology.orchestra.run.localhost.stderr:+ dd of=/tmp/bla/bla
2020-07-17 17:51:26,856.856 INFO:teuthology.orchestra.run.localhost.stderr:dd: failed to open '/tmp/bla/bla': No such file or directory
  File "/home/kyr/kshtsk/teuthology/teuthology/orchestra/remote.py", line 481, in write_file
    self.run(args=args, stdin=data)
  File "/home/kyr/kshtsk/teuthology/teuthology/orchestra/remote.py", line 204, in run
    r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
  File "/home/kyr/kshtsk/teuthology/teuthology/orchestra/run.py", line 446, in run
    r.wait()
  File "/home/kyr/kshtsk/teuthology/teuthology/orchestra/run.py", line 160, in wait
    self._raise_for_status()
  File "/home/kyr/kshtsk/teuthology/teuthology/orchestra/run.py", line 182, in _raise_for_status
    node=self.hostname, label=self.label
teuthology.exceptions.CommandFailedError: Command failed on localhost with status 1: 'set -ex\ndd of=/tmp/bla/bla'
```
The write_file should not return subprocess, it is a blackbox, I can agree that CommandFailedError could include the stderr and stdout in some cases but this not optimal approach because stdout and stderr in general can be very very big, especially for long running processes. Probably this is why it is not included as a member to CommandFailedError. But when it comes for writing file we do not need so much info, since we usually want to be written always, and if not, then stop.

From your code:
```
except CommandFailedError as e:
    self.assertEqual(e.existatus, 1)
```
I don't understand why do you need to ignore error with status not equal to 1? If exitstatus is Zero the exception is not even raised.
So, writing:
```
remote.run(args=['some', 'cmd'])
```
is more elegant than:
```
 retval = remote.run(args=['some', 'cmd']).returncode
   self.assertEqual(retval, 1)
```
You just don't need to call assertEqual and do you need to call assertEqual at all?

kshtsk: I'm not sure I understand which goal are you trying to reach.

kshtsk:
> Really? It's written so that it can be used without being the root user. So, is the case with vstart_runner.py -- we omit sudo so that permissions on files don't change and therefore it can be used _without_ being root.

If you say this, then the problem in the test that is using sudo in api.

> > but this is slightly changes the logic in all commands. So instead of adding `omit_sudo` argument to many run calls we just need to remove erasure of 'sudo' from vstat:
> 
> I don't understand. Can you please elaborate on how removing erasure of 'sudo' is different from just omitting it?

This is very odd to right commands like "sudo_do_something(do_sudo_command, omit_sudo=True)", like you gave an example, running "sudo -u user". If command can be run successfully without sudo, it should be run without sudo, and we need to fix test or command, but not adding workarounds like: monkey replacing of all "sudo" from the commands, or even scripts, and later adding an options to omit_sudo to overcome bugs produced by the first sudo replacement.

I may suggest that this replacement added as a proof of concept that we can run sometimes in vstart, it is okay. It is okay to merge it for the first time, but when we found a bug, and went breaking everything else is not okay.

kshtsk:
@rishabh-d-dave I guess we can close it now after merging the #1538

PR #1538, opened 2020-07-16
Add remote file operations

kshtsk:


kshtsk:
```>>> from teuthology.orchestra.remote import Remote as r
>>> l=r('localhost')
>>> l.write_file('/tmp/dir/subdir/deep/deep/subdir/file', b'zzz', mode='0600', mkdir=True)
2020-07-16 18:33:34,795.795 INFO:teuthology.orchestra.remote:Trying to reconnect to host
2020-07-16 18:33:34,873.873 INFO:teuthology.orchestra.run.localhost:> true
2020-07-16 18:33:35,073.073 INFO:teuthology.orchestra.run.localhost:> set -ex
2020-07-16 18:33:35,073.073 INFO:teuthology.orchestra.run.localhost:> mkdir -p /tmp/dir/subdir/deep/deep/subdir
2020-07-16 18:33:35,074.074 INFO:teuthology.orchestra.run.localhost:> dd of=/tmp/dir/subdir/deep/deep/subdir/file
2020-07-16 18:33:35,074.074 INFO:teuthology.orchestra.run.localhost:> chmod 0600 /tmp/dir/subdir/deep/deep/subdir/file
2020-07-16 18:33:35,214.214 INFO:teuthology.orchestra.run.localhost.stderr:+ mkdir -p /tmp/dir/subdir/deep/deep/subdir
2020-07-16 18:33:35,217.217 INFO:teuthology.orchestra.run.localhost.stderr:+ dd of=/tmp/dir/subdir/deep/deep/subdir/file
2020-07-16 18:33:35,219.219 INFO:teuthology.orchestra.run.localhost.stderr:0+1 records in
2020-07-16 18:33:35,219.219 INFO:teuthology.orchestra.run.localhost.stderr:0+1 records out
2020-07-16 18:33:35,219.219 INFO:teuthology.orchestra.run.localhost.stderr:3 bytes copied, 8.0701e-05 s, 37.2 kB/s
2020-07-16 18:33:35,219.219 INFO:teuthology.orchestra.run.localhost.stderr:+ chmod 0600 /tmp/dir/subdir/deep/deep/subdir/file
```
and:
```> ls -la /tmp/dir/subdir/deep/deep/subdir/file
-rw------- 1 kyr users 3 Jul 16 18:33 /tmp/dir/subdir/deep/deep/subdir/file
```

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/orchestra/remote.py
```diff
@@ -449,6 +449,43 @@ def get_tar_stream(self, path, sudo=False):
...
+        if mode:
+            chmod = 'sudo chmod' if sudo else 'chmod'
+            args += '\n' + chmod + ' ' + mode + ' ' + path
+        if owner:
+            chown = 'sudo chown' if sudo else 'chown'
+            args += '\n' + chown + ' ' + owner + ' ' + path
+        args = 'set -ex' + '\n' + args
+        self.run(args=args, stdin=data)
+
+    def sudo_write_file(self, path, data, **kwargs):
```
rishabh-d-dave, line 483: IMHO, `sudo_write_file()` is not super-convenient compared to `write_file(path, data,  sudo=True)`. So I won't mind not having `sudo_write_file()`.

But, yes, not having it would break many tests.

teuthology/orchestra/remote.py
```diff
@@ -449,6 +449,43 @@ def get_tar_stream(self, path, sudo=False):
...
+            if dirpath:
+                args = mkdirp + ' ' + dirpath + '\n' + args
+        if mode:
+            chmod = 'sudo chmod' if sudo else 'chmod'
+            args += '\n' + chmod + ' ' + mode + ' ' + path
+        if owner:
+            chown = 'sudo chown' if sudo else 'chown'
+            args += '\n' + chown + ' ' + owner + ' ' + path
+        args = 'set -ex' + '\n' + args
+        self.run(args=args, stdin=data)
```
rishabh-d-dave, line 481: Oh, I must also mention that I would prefer returning the return value of `self.run()` since access to the object representing the process can allow better testing and troubleshooting. Applies to this method as well as to other methods here and in `misc.py`.

teuthology/orchestra/remote.py
kshtsk: I can delete it, I don't it can break anything because it is new function, I added because someone might think it handy to use `sudo_write_file` because of shorter syntax.

kshtsk: > But, yes, not having it would break many tests.

this is why we do not delete `teuthology.misc.sudo_write_file` but make it use `remote.write_file(sudo=True)`

teuthology/orchestra/remote.py
kshtsk: As I said, we don't need trouble shoot write_file, we need to stop as fast as possible and analize logs, even more, it is bad to continue execution because it make more troubles to figure out the first source of the failure.

kshtsk: So, if we want more control on error handling I would prefer to specific exception produced by the file operation, so the user don't need to check output of the script of write_file. We need to declare all specific to the function exception in the __doc__ for the function and we must not let user to do this instead of us, because the code of script can be changed and will be very difficult to support. Basically user should not use anything about internals of functions like write_file and treat it as blackbox.

setup.py
```diff
@@ -49,6 +49,7 @@
         'Topic :: System :: Filesystems',
     ],
     install_requires=['apache-libcloud',
+                      'Deprecated >= 1.2.10',
```
tchaikov, line 52: instead of introducing yet another dependency, i'd suggest just update all the callers of the deprecated functions, and then drop them.

teuthology/orchestra/remote.py
tchaikov: i'd like to see PRs on ceph side which use the newly added functions. otherwise these new methods are just unused and untested.

kshtsk: @tchaikov it is already used there, through misc.sudo_write_file, check there logs:
https://pulpito.ceph.com/kyr-2020-07-16_20:09:04-rados:cephadm-master-distro-basic-smithi/
You can find there a 'dd' command.
Even if you don't see it used in upstream ceph, it does not mean we (mostly me) do not plan to use it downstream.
I agree we can add more test cases, however it is "tested" through misc api.

I don't want to refactor and go through the PRs on ceph side now and go through the all cycles of merging and backporting. 
I'm more comfortable to do this later, this is why a deprication message.

setup.py
kshtsk: what is the problem with introducing new dependency which is useful?

tchaikov: > i'd suggest just update all the callers of the deprecated functions, and then drop them.

teuthology/orchestra/remote.py
tchaikov: > I don't want to refactor and go through the PRs on ceph side now and go through the all cycles of merging and backporting.
> I'm more comfortable to do this later, this is why a deprication message.

please take your time. we can have all of them once they are ready. 

[OT] also, please let me know if you are going to work on https://github.com/ceph/ceph/pull/35674#discussion_r442871795 ? if not, i will cherry-pick the fixes.

kshtsk: @tchaikov what do you have on mind, to backport it to 'octopus', 'nautilus', or also py2 depended branches 'luminous' and correspondingly backport this PR to teuthology py2?

kshtsk: @tchaikov is that kind of PR do you expect? https://github.com/ceph/ceph/pull/36211

kshtsk: > [OT] also, please let me know if you are going to work on [ceph/ceph#35674 (comment)](https://github.com/ceph/ceph/pull/35674#discussion_r442871795) ? if not, i will cherry-pick the fixes.

Do you suggest to finish it before merging this PR?

tchaikov: > > [OT] also, please let me know if you are going to work on [ceph/ceph#35674 (comment)](https://github.com/ceph/ceph/pull/35674#discussion_r442871795) ? if not, i will cherry-pick the fixes.
> 
> Do you suggest to finish it before merging this PR?

no, it's note related to this changeset. i marked "[OT]" (short for off-topic) here. as per your reply, i just realized that you are not likely to work on backports. hence the question.

kshtsk: > no, it's not related to this changeset. i marked "[OT]" (short for off-topic) here. as per your reply, i just realized that you are not likely to work on backports. hence the question.

thanks for the remark, sorry sometimes I'm bad to decode things like "ot". Regarding backport, sooner or later I might work on it. But not right now. So go ahead and cherry-pick it.

kshtsk:
retest this please

kshtsk:
retest this please

setup.py
kshtsk: @tchaikov I've dropped the dependency.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

tchaikov (approved):

teuthology/orchestra/remote.py
```diff
@@ -458,6 +458,131 @@ def get_tar_stream(self, path, sudo=False):
...
+            if dirpath:
+                args = mkdirp + ' ' + dirpath + '\n' + args
+        if mode:
+            chmod = 'sudo chmod' if sudo else 'chmod'
+            args += '\n' + chmod + ' ' + mode + ' ' + dst
+        if owner:
+            chown = 'sudo chown' if sudo else 'chown'
+            args += '\n' + chown + ' ' + owner + ' ' + dst
+        args = 'set -ex' + '\n' + args
+        self.run(args=args)
```
tchaikov, line 490: i'd suggest rewrite this using a helper function for adding optional `sudo` and f-string, but my suggestion was dropped by accident when posting it. i will add a follow-up PR later on.

kshtsk: thanks for the heads up, I also would like to backport this patch to py2, however I'm not sure that I will do that, hopefully we will not need py2 based branches of teuthology downstream in couple of months.

PR #1540, opened 2020-07-28
[WIP] add-teuthology-dispatcher

ShraddhaAg:
Related PRs: [#36352](https://github.com/ceph/ceph/pull/36352)

**Motivation**
[Project Idea Link](https://ceph.io/gsoc-2020/#teuthology-scheduling%20Improvements) The existing worker framework leads to jobs, that require large number of machines to run on, waiting for a long time (often indefinitely). This also leads to running jobs in unreliable priority order. To solve the above, this PR is is the first step. It replaces the existing worker framework with a dispatcher that takes jobs from a queue, allocates nodes for them to run on, and executes them. 

Flow charts for:
*  [teuthology-dispatcher](https://docs.google.com/drawings/d/1JTRcqnzZ1YdKv0IDADyMTWffCoU8SJaD3bVIdWO5lww/edit?usp=sharing)
* [modified teuthology-worker](https://docs.google.com/drawings/d/1bkwGI8ar2FL18ohIyjv8USOA78Mp4bFRnu6DTy6cIhY/edit?usp=sharing)

**Changes in the current working:** 

* `teuthology-dispatcher` takes jobs from the queue, locks machines, populates job config with targets and invokes `teuthology-worker`. Without waiting for `teuthology-worker` child process to finish, it moves on to take another job from the queue and repeat the above process.
* `teuthology-worker` is now job specific. It sets up a worker new log file in suites archive dir named worker.{job_id}. It then reimages the target machines, constructs the `teuthology` cmd and invokes it. In case of job timeout, It also archives log dirs from target machines before nuking them. For the same, the [teuthology task](https://github.com/ceph/teuthology/pull/1540/files#diff-57bfcbe1ad60f3514b043712cf1b62e1) and [ceph task](opened PR [#36352](https://github.com/ceph/ceph/pull/36352)) are modified to add the log dirs to info.yaml. When job times out, the dirs to archive are read from here.    
* `teuthology` cmd: now adds an internal task, [unlock_targets](https://github.com/ceph/teuthology/pull/1540/files#diff-1ddc90aa37564b7357f276dfd792da67) in [run.py](https://github.com/ceph/teuthology/pull/1540/files#diff-438dc42f85af0a957680785784f8417dR256-R259) that unlocks the target machines (locked by dispatcher) at the end of the job run. 

**Development Lab Setup**: 

I have setup a development lab for teuthology, using some mira nodes from the current cluster, to test out the above mentioned changes. I have done so using:
* Master node (teuthworker/teuthology node): mira045
* Paddles/Pulpito node: mira119
* Test nodes: mira062, mira072, mira085

Pulpito can be accessed [here](http://mira119.front.sepia.ceph.com:8081/). 
archive logs are on the master node (mira045) in location: /home/teuthworker/archive

**Test performed**

* Job meant to pass: [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-07-28_02:55:19-:tmp:test1.yaml-wip-jd-testing-distro-basic-mira/)
* Job meant to fail: [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-07-28_03:01:40-:tmp:test2.yaml-wip-jd-testing-distro-basic-mira/)

**Changes Made**
* [X] Add a new dispatcher command, `teuthology-dispatcher`, that runs each job in the queue as its subprocess by invoking `teuthology-worker`. 
* [X] Modify `teuthology-worker` such that it only runs the job whose config is provided to it and ported to `docopt`.
* [X] Add an internal task, task/internal/unlock_targets.py to unlock target machines after the job run is complete.
* [X] Modify `teuthology` command by adding a `--unlock` flag to to unlock target machines if specified in config/s. 
* [X] Enables graceful job timeout by archiving logs before nuking machines.
* [X] Only lock machines in dispatcher and move reimaging machines to worker. 
* [X] Update documentation
* [ ] Update unit tests for worker
* [ ] Port `teuthology-dispatcher` to `docopt`.

### DIFF
--- a/docs/COMPONENTS.rst
+++ b/docs/COMPONENTS.rst
@@ -48,10 +48,11 @@ database. Depending on the requirements (in terms of requested machines), the
 scheduler eventually determines when a job can get executed. At this point the
 master communicates with ``teuthology-dispatcher``, which checks the lock
 status of the requested machines (``teuthology-lock``) by querying ``paddles``,
-acquires locks of the nodes if they are available and invokes ``teuthology-worker``
-which in turn invokes ``teuthology`` (the command) which proceeds to execute
-the job (execute every task in the YAML job description). Results from the job
-are stored in the archive directory of the worker for forensic analysis.
+acquires locks of the nodes if they are available and invokes ``teuthology-worker``.
+``teuthology-worker`` reimages target machines and then invokes ``teuthology``
+(the command) which proceeds to execute the job (execute every task in the
+YAML job description). Results from the job are stored in the archive directory
+of the worker for forensic analysis.
 
 Since `QA suites <https://github.com/ceph/ceph-qa-suite>`__ usually
 specify ``install`` and ``ceph`` tasks, we briefly describe what they do. When
--- a/docs/COMPONENTS.rst
+++ b/docs/COMPONENTS.rst
@@ -21,47 +21,47 @@ a scheduler (a.k.a. master node), worker(s), package builder
 
 In the figure above, every service appears on a separate machine but this is
 not a requirement (see :ref:`lab_setup` for an alternative setup). Jobs are
-submitted to the scheduler, which are then processed by workers. One worker
-processes and keeps track of a job (one at a time). The output of the job (logs
-and files associated to an execution) is stored in the archive, which
-is a path in the file system where the worker is running. The job database
+submitted to the scheduler, which are then processed by dispatcher. Dispatcher
+takes job from the queue, allocates nodes for them to run on and initiates
+workers as its sub process to run the job. The output of the job (logs and
+files associated to an execution) is stored in the archive, which is a path
+in the file system where the dispatcher is running. The job database
 contains information about the status of jobs and test nodes, as well as
 results of executions (# of tests passed, failed, etc.). All this information
 can be visualized in ``pulpito``, the web UI. For an example, see Ceph
 community's Lab `here <http://pulpito.ceph.com>`__.
 
-Test nodes can be grouped in classes (referred to as ``machine-type``), 
-allowing teuthology schedule jobs across multiple hardware setups, 
+Test nodes can be grouped in classes (referred to as ``machine-type``),
+allowing teuthology to schedule jobs across multiple hardware setups,
 provided they're visible to the master node.
 
 Life of a Teuthology Job
 ========================
 
 The teuthology scheduler exposes a work queue (using `beanstalkd
-<https://kr.github.io/beanstalkd/>`__) where jobs are submitted. The life of a 
-job begins when ``teuthology-suite`` is executed, which is when a job is 
-prepared and queued (``teuthology-schedule`` is implicitly invoked). When a job 
-is created (or whenever the status of a job is changed, e.g. from queued to 
-started), information about the job is recorded in ``paddles``'s internal 
-database. Depending on the requirements (in terms of requested machines), the 
-scheduler eventually determines when a job can get executed. At this point the 
-master communicates with ``teuthology-worker``, which in turn invokes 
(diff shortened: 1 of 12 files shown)


kshtsk:
1) Can we move the linter patches to a separate PR? It would much better to have the code compact out of non-necessary commits for dispatcher.
2) Could you please add a document describing the high level architecture of the dispatcher.
3) Could we refactor the code so the older version of worker are not modified so we still continue use them for some time before dispatcher become mature enough?
4) Could you please move --unlock patches to a separated PR, I don't think they are needed for the dispatcher itself, since the teuthology is already unlocking machines if the --lock option is provided and it is useless when --lock is not provided.

kshtsk:
Also, if you still wish teuthology-worker code modified, there is another idea to not create extra teuthology-worker command, but implement this as special mode of worker, let say `--dispatch` option.

ShraddhaAg:
>     1. Can we move the linter patches to a separate PR? It would much better to have the code compact out of non-necessary commits for dispatcher.
> 

Yes, I'll do that.

>     2. Could you please add a document describing the high level architecture of the dispatcher.
> 

Yeah, I was working on updating the docs/COMPONENTS.rst for the same :)

>     3. Could we refactor the code so the older version of worker are not modified so we still continue use them for some time before dispatcher become mature enough?
> 

Yes, that makes sense, I'll do that.

>     4. Could you please move --unlock patches to a separated PR, I don't think they are needed for the dispatcher itself, since the teuthology is already unlocking machines if the --lock option is provided and it is useless when --lock is not provided.

Since we are locking machines in the dispatcher, config will have `targets` dict. If we pass `--lock` to `teuthology` it would would fail [here](https://github.com/ceph/teuthology/blob/master/teuthology/run.py#L187). We need `--unlock` to unlock the target machines.

kshtsk:
> Since we are locking machines in the dispatcher, config will have `targets` dict. If we pass `--lock` to `teuthology` it would would fail [here](https://github.com/ceph/teuthology/blob/master/teuthology/run.py#L187). We need `--unlock` to unlock the target machines.

If we lock machines in the dispatcher, we just do not need to add --lock option for teuthology, and in such a way we do not need unlock them.

kshtsk:
btw, I'm looking into locking mechanism and we just need to refactor it.

ShraddhaAg:
> If we lock machines in the dispatcher, we just do not need to add --lock option for teuthology, and in such a way we do not need unlock them, because lock internal task is not even added.

yes, but then machines will not be unlocked after the job finishes the run. Do you mean to move the logic that enforces unlocking machines (after the job run is complete) when using dispatcher to a new PR?

kshtsk:
> > If we lock machines in the dispatcher, we just do not need to add --lock option for teuthology, and in such a way we do not need unlock them, because lock internal task is not even added.
> 
> yes, but then machines will not be unlocked after the job finishes the run. Do you mean to move the logic that enforces unlocking machines (after the job run is complete) when using dispatcher to a new PR?

the machines should by unlocked by those who is locking them, in our case is it dispatcher/worker who locks it?

ShraddhaAg:
> > > If we lock machines in the dispatcher, we just do not need to add --lock option for teuthology, and in such a way we do not need unlock them, because lock internal task is not even added.
> > 
> > yes, but then machines will not be unlocked after the job finishes the run. Do you mean to move the logic that enforces unlocking machines (after the job run is complete) when using dispatcher to a new PR?
> 
> the machines should by unlocked by those who is locking them, in our case is it dispatcher/worker who locks it? 

Yes, here the dispatcher is locking the machines.

kshtsk:
JFYI, our deployment test which is using ceph-cm-ansible is using multiple workers.

ShraddhaAg:
closed in favour of #1546

PR #1541, opened 2020-07-31
orchestra/cluster: be explicit that a positional argument is needed

tchaikov:
otherwise the caller may run into

sh() missing 1 required positional argument: 'script'

if the developer only check the function signature of Cluster.sh().


kshtsk:
just an idea, maybe we need to add reference to "teuthology.orchestra.run.run" in the doc for all methods from cluster.run, cluster.sh, remote.run, remote.sh methods.

tchaikov:
> just an idea, maybe we need to add reference to "teuthology.orchestra.run.run" in the doc for all methods from cluster.run, cluster.sh, remote.run, remote.sh methods.

i'd prefer updating the APIs to make them self-documented.

PR #1543, opened 2020-08-03
read redhat repos list from downstream.yaml

rakeshgm:
Read redhat repos list from downstream.yaml instead of hard coding here. 
 

rakeshgm:
one successful log here http://pulpito.ceph.redhat.com/rakesh-2020-08-03_10:24:33-rgw:system-nautilus-distro-basic-clara/

PR #1544, opened 2020-08-07
added user option to perform rh kernel update

sunilkumarn417:
Signed-off-by: sunilkumarn417 <sunnagar@redhat.com>

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -249,9 +249,11 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.git_ignore_ssl': None},
             {'internal.setup_cdn_repo': None},
             {'internal.setup_base_repo': None},
-            {'internal.setup_additional_repo': None},
-            {'kernel.install_latest_rh_kernel': None}
+            {'internal.setup_additional_repo': None}
         ])
+        # Install latest kernel task for redhat downstream runs
+        if config.get('redhat').get('install_latest_rh_kernel', False):
+            init_tasks.extend({'kernel.install_latest_rh_kernel': None})
  
     return init_tasks
 


teuthology/run.py
```diff
@@ -249,9 +249,11 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.git_ignore_ssl': None},
             {'internal.setup_cdn_repo': None},
             {'internal.setup_base_repo': None},
-            {'internal.setup_additional_repo': None},
-            {'kernel.install_latest_rh_kernel': None}
+            {'internal.setup_additional_repo': None}
         ])
+        # update kernel
+        if config.get('redhat').get('kernel_update', False):
```
kshtsk, line 255: what if 'redhat' is undefined?

kshtsk:
ok to test

teuthology/run.py
```diff
@@ -249,9 +249,11 @@ def get_initial_tasks(lock, config, machine_type):
             {'internal.git_ignore_ssl': None},
             {'internal.setup_cdn_repo': None},
             {'internal.setup_base_repo': None},
-            {'internal.setup_additional_repo': None},
-            {'kernel.install_latest_rh_kernel': None}
+            {'internal.setup_additional_repo': None}
         ])
+        # Install latest kernel task for redhat downstream runs
+        if config.get('redhat').get('install_latest_rh_kernel', False):
```
vasukulkarni, line 255: In the email i saw a comment from @kshtsk  but not here, I think we need to handle when 'redhat' is none.

vasukulkarni: well the diff is confusing, the outer block has check so its safe.

PR #1545, opened 2020-08-08
roles/overrides, placeholder.py: add log-ignorelist

neha-ojha:
Corresponding ceph change https://github.com/ceph/ceph/pull/36527

Signed-off-by: Neha Ojha <nojha@redhat.com>


neha-ojha:
> @neha-ojha is it ready to merge?

I think so.

PR #1546, opened 2020-08-10
Add teuthology-dispatcher

ShraddhaAg:
Related PRs: 
* [#36718](https://github.com/ceph/ceph/pull/36718)
* [#36738](https://github.com/ceph/ceph/pull/36738)

**Motivation**
[Project Idea Link](https://ceph.io/gsoc-2020/#teuthology-scheduling%20Improvements) The existing worker framework leads to jobs, that require large number of machines to run on, waiting for a long time (often indefinitely). This also leads to running jobs in unreliable priority order. This PR tries to solve the above. It the first step towards replacing the existing worker framework with a dispatcher. This PR adds a dispatcher that takes jobs from a queue, allocates nodes for them to run on, and executes them as its subprocess. 

Flow charts for:
*  [teuthology-dispatcher](https://docs.google.com/drawings/d/1JTRcqnzZ1YdKv0IDADyMTWffCoU8SJaD3bVIdWO5lww/edit?usp=sharing)
* [teuthology-dispacther supervisor mode](https://docs.google.com/drawings/d/1bkwGI8ar2FL18ohIyjv8USOA78Mp4bFRnu6DTy6cIhY/edit?usp=sharing)

**Changes in the current working:** 

* `teuthology-dispatcher` takes jobs from the queue, locks machines without reimaging them, populates job config with target machines and invokes `teuthology-dispacther` in `--supervisor` mode. Without waiting for its child process (`teuthology-dispacther` in `--supervisor` mode) to finish, it moves on to take another job from the queue and repeat the above process.
* `teuthology-dispacther --supervisor mode` is a job supervisor process and thus is job specific. It sets up a supervisor log file in suites' archive dir named supervisor.{job_id}. It then reimages the target machines, constructs the `teuthology` cmd and invokes it. At the end of the job run, the supervisor checks the job status and depending on whether it was successfully completed or not it unlocks target machines or nukes them.
* graceful job timeout: When a job times out, before nuking nodes in cluster, the supervisor first compresses and archives logs from each node and then nukes them.  

**Development Lab Setup**: 

I have setup a development lab for teuthology, using some mira nodes from the current cluster, to test out the above mentioned changes. I have done so using:
* Master node (teuthworker/teuthology node): mira045
* Paddles/Pulpito node: mira119
* Test nodes: mira062, mira072, mira085

Pulpito can be accessed [here](http://mira119.front.sepia.ceph.com:8081/). 
archive logs are on the master node (mira045) in location: /home/teuthworker/archive

**Test performed**

* Job meant to pass: [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-08-20_19:34:34-:tmp:test1.yaml-master-distro-basic-mira/)
* Job meant to fail: [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-08-20_19:34:21-:tmp:test2.yaml-master-distro-basic-mira/)
* Job successfully transfering logs from remote machine to archive dir in case of job timeout:
  1. (teuthology init + ceph task): [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-08-19_21:36:20-rados:singleton:all:admin-socket.yaml-wip-jd-testing-distro-basic-mira/)
  2. (teuthology init + cephadm task): [link](http://mira119.front.sepia.ceph.com:8081/teuthology-2020-08-20_18:36:55-rados:cephadm:upgrade-master-distro-basic-mira/)

**Checklist**
* [X] Add a new dispatcher command, `teuthology-dispatcher`, that runs each job in the queue as its subprocess by invoking `teuthology-dispatcher` in `--supervisor` mode. 
* [X] Add `supervisor` mode to `teuthology-dispatcher` such that it reimages target machines, runs the job whose config is provided to it and unlocks/nukes targets at the end of the job run. 
* [X] move reimaging and unlocking targets in `supervisor` subprocess.
* [X] Add graceful job timeout by archiving logs before nuking machines commit. 
* [x] Use same teuthology workspace for entire test suite.
* [ ] Update documentation


scripts/dispatcher.py
```diff
@@ -0,0 +1,33 @@
...
+
+standard arguments:
+  -h, --help                     show this help message and exit
+  -v, --verbose                  be more verbose
+  -t, --tube TUBE                which beanstalk tube to read jobs from
+  -l, --log-dir LOG_DIR          path in which to store logs
+  --archive-dir ARC_DIR          path to archive results in
+  --supervisor                   run dispactcher in job supervisor mode
+  --bin-path BIN_PATH            teuthology bin path
+  --config-fd FD                 file descriptor of job's config file
```
kshtsk, line 23: Can we use file path insteaf of file descriptor?

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+
+        run_args = [
+            os.path.join(teuth_bin_path, 'teuthology-dispatcher'),
+            '--supervisor',
+            '-v',
+            '--bin-path', teuth_bin_path,
+            '--archive-dir', archive_dir,
+        ]
+
+        with tempfile.NamedTemporaryFile(prefix='teuthology-dispatcher.',
```
kshtsk, line 137: Instead of temporary file can we create this file in the job archive_dir for debugging, reproducibility and simplifying issue analysis.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,202 @@
...
+    arg.extend([
+        '--owner', job_config['owner'],
+        '--archive', job_config['archive_path'],
+        '--name', job_config['name'],
+    ])
+    if job_config['description'] is not None:
+        arg.extend(['--description', job_config['description']])
+    arg.append('--')
+
+    with tempfile.NamedTemporaryFile(prefix='teuthology-worker.',
```
kshtsk, line 136: Instead of temporary file, could you please create teuthology worker log in the job archive directory.

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+                                         suffix='.tmp', mode='w+t') as tmp:
+            yaml.safe_dump(data=job_config, stream=tmp)
+            tmp.flush()
+            run_args.extend(["--config-fd", str(tmp.fileno())])
```
kshtsk, line 141: Correspondingly pass the full path to the config file from archive_dir.

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+            job_proc = subprocess.Popen(run_args, pass_fds=[tmp.fileno()])
+
+        log.info('Job subprocess PID: %s', job_proc.pid)
```
kshtsk, line 144: Can we store this pid in somewhere in `/var` so we can return control over the job if dispatchers occasionally restart.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,202 @@
...
+from teuthology import setup_log_file, install_except_hook
+from teuthology.lock.ops import reimage_many
+from teuthology.misc import get_user
+from teuthology.config import FakeNamespace
+from teuthology.worker import run_with_watchdog, symlink_worker_log
+
+log = logging.getLogger(__name__)
+start_time = datetime.utcnow()
+restart_file_path = '/tmp/teuthology-restart-workers'
+stop_file_path = '/tmp/teuthology-stop-workers'
```
kshtsk, line 26: This should be useless for the dispatcher, no?

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,202 @@
...
+def main(args):
+
+    verbose = args["--verbose"]
+    archive_dir = args["--archive-dir"]
+    teuth_bin_path = args["--bin-path"]
+    config_fd = int(args["--config-fd"])
+
+    with open(config_fd, 'r') as config_file:
+        config_file.seek(0)
+        job_config = yaml.safe_load(config_file)
```
kshtsk, line 38: Please, use file path instead of file descriptor.

teuthology/task/internal/lock_machines.py
```diff
@@ -23,6 +23,14 @@ def lock_machines(ctx, config):
...
     machines and placed those keys in the Targets section of a yaml file.
     """
+    lock_machines_helper(ctx, config)
+    try:
+        yield
+    finally:
+        unlock_machines(ctx)
+
+
+def lock_machines_helper(ctx, config, reimage=True):
```
kshtsk, line 33: I think we decided to not touch internal tasks for now. The dispatcher should not use lock_machines module in any ways.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,202 @@
...
+            except Exception:
+                log.exception("run_with_watchdog had an unhandled exception")
+                raise
+        else:
+            log.info("Running without watchdog")
+            # This sleep() is to give the child time to start up and create the
+            # archive dir.
+            time.sleep(5)
+            symlink_worker_log(job_config['worker_log'],
+                               job_config['archive_path'])
```
kshtsk, line 163: There is no sense in this code since we can redirect output of corresponding dispatcher log directly to archive_path.

kshtsk:
Good job overall, see inline comments.

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+        # beanstalkc throws a SocketError
+        try:
+            job.delete()
+        except Exception:
+            log.exception("Saw exception while trying to delete job")
+
+
+def lock_machines(job_config):
+    fake_ctx = supervisor.create_fake_context(job_config, block=True)
+    lock_machines_helper(fake_ctx, [len(job_config['roles']),
```
kshtsk, line 156: Instead of using internal tasks we should use lock module directly.

kshtsk:
ok to test

kshtsk:
@susebot run deploy

teuthology/dispatcher/__init__.py
jdurgin: not sure what you're getting at here - the supervisor process started here isn't monitored or interacted with again by the dispatcher after it's started

teuthology/dispatcher/supervisor.py
jdurgin: agreed, no need to restart the supervisor

teuthology/dispatcher/__init__.py
kshtsk: ah.... forget about this

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,222 @@
...
+    with open(config_fd, 'r') as config_file:
+        config_file.seek(0)
+        job_config = yaml.safe_load(config_file)
+
+    loglevel = logging.INFO
+    if verbose:
+        loglevel = logging.DEBUG
+    log.setLevel(loglevel)
+
+    suite_dir = os.path.join(archive_dir, job_config['name'])
```
kshtsk, line 45: This looks like not a suite dir, maybe *run_dir* is more suitable name.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,222 @@
...
+    if (not os.path.exists(suite_dir)):
+        os.mkdir(suite_dir)
+    log_file_path = os.path.join(suite_dir, 'supervisor.{job_id}'.format(
```
kshtsk, line 48: Can we put the supervisor log into the job directory, i.e. `f'{run_dir}/{job_id}/supervisor.log'`

ShraddhaAg: The only reason I didn't put it there is that we will have unnecessary folders for first and last jobs in a run. If that is not an issue, this makes more sense.

teuthology/dispatcher/__init__.py
ShraddhaAg: We create the same file (`orig.config.yaml`) in the job archive directory when `teuthology` cmd is invoked by the supervisor while initiating run in run.py [here](https://github.com/ceph/teuthology/blob/master/teuthology/run.py#L42). I'll refactor so that the file is created here.

teuthology/dispatcher/supervisor.py
ShraddhaAg: This is again the job config itself. I'll pass the same file here as in dispatcher.

teuthology/dispatcher/supervisor.py
ShraddhaAg: ah sorry, I missed this

teuthology/dispatcher/supervisor.py
kshtsk: Yes... right, the first and last in suite are specific jobs, and they should be handled differently, since they do not start any teuthology job.

teuthology/dispatcher/__init__.py
kshtsk: Right, just store under different name, no need to refactor teuthology.run right now, it would be better to drop saving the `orig.config.yaml` in follow up PR after teuthology-dispatcher. Another reason to minimize invasion to teuthology.run is that some system can still use older py2 branch for running teuthology test, so we want to have teuthology-worker to be backwards compatible for some time.

teuthology/dispatcher/supervisor.py
ShraddhaAg: Done. Removing the unnecessary dirs created at the end of execution in case the job is first or last in suite [here](https://github.com/ceph/teuthology/pull/1546/files#diff-159cd41bfcdb054c0e04dc2650bfa033R93-R96).

teuthology/dispatcher/supervisor.py
ShraddhaAg: Changed it to `run_archive`.

teuthology/dispatcher/supervisor.py
ShraddhaAg: Done.

teuthology/dispatcher/__init__.py
ShraddhaAg: Created `orig.config.yaml` in dispatcher itself (along with creating archive dirs for run and job) without refactoring run.py to maintain backward compatibility.

scripts/dispatcher.py
ShraddhaAg: Done.

teuthology/dispatcher/__init__.py
ShraddhaAg: Done.

teuthology/dispatcher/supervisor.py
ShraddhaAg: Removed this symlink since the supervisor log is already inside the job archive now.

teuthology/dispatcher/supervisor.py
```diff
@@ -21,39 +18,33 @@
...
-stop_file_path = '/tmp/teuthology-stop-workers'
 
 
 def main(args):
 
     verbose = args["--verbose"]
     archive_dir = args["--archive-dir"]
     teuth_bin_path = args["--bin-path"]
-    config_fd = int(args["--config-fd"])
+    config_file_path = args["--config-file"]
```
kshtsk, line 28: Could you please rename the parameter to `--job-config` so we do not misinterpret it with teuthology config.

ShraddhaAg: Right, done!

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,204 @@
...
+    # Reimage the targets specified in job config
+    # and update their keys in config after reimaging
+    ctx = create_fake_context(job_config)
+    # change the status during the reimaging process
+    report.try_push_job_info(ctx.config, dict(status='waiting'))
+    targets = job_config['targets']
+    reimaged = reimage_many(ctx, targets, job_config['machine_type'])
+    ctx.config['targets'] = reimaged
+    # change the status to running after the reimaging process
+    report.try_push_job_info(ctx.config, dict(status='waiting'))
```
kshtsk, line 165: The comment above this line states it should set status to 'running' but you're trying to set it to 'waiting'

ShraddhaAg: Oops, changed it.

teuthology/task/internal/lock_machines.py
ShraddhaAg: It doesn't change the working of the internal task. Just refactors it such that the same logic can be used in dispatcher to wait to for required number of machines to be available and then lock them (using `lock_many` of lock module). Since the waiting logic is not in the lock module, I thought it's best to reuse the logic from here to avoid repetitive code and still maintaining backward compatibility. Please let me know if this needs to be done differently.

kshtsk: I can see it does not change the behavior of the code, it just wrong place for this can of refactoring right now, overall locking mechanism should be rewritten in future, but for the goals of this PR those changes are not required. I would prefer to leave internal tasks untouched right now. Later we should refactor the lock_machines tasks to use teuthology.lock.ops code. The dispatcher should not user anything from the directory teuthology/task, and use lock api directly. The tasks should only be invoked from the run itself.

ShraddhaAg: Added func [`block_and_lock_machines`](https://github.com/ceph/teuthology/pull/1546/files#diff-0c1bc86b25e221cf13dbc1b8dee22354R318) in lock module which is used by both the `lock_machines` task and dispatcher. Is this approach okay?

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def sentinel(path):
+    if not os.path.exists(path):
+        return False
+    file_mtime = datetime.utcfromtimestamp(os.path.getmtime(path))
+    if file_mtime > start_time:
+        return True
+    else:
+        return False
```
tchaikov, line 32:
```suggestion
    return file_mtime > start_time
```

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+    tube = args["--tube"]
+    log_dir = args["--log-dir"]
+    archive_dir = args["--archive-dir"]
+
+    # setup logging for disoatcher in {log_dir}
+    loglevel = logging.INFO
+    if verbose:
+        loglevel = logging.DEBUG
+    log.setLevel(loglevel)
+    log_file_path = os.path.join(log_dir, 'dispatcher.{tube}.{pid}'.format(
```
tchaikov, line 74: nit, could use f-string, as the teuthology is using python3.6 now.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,324 @@
...
+    # Let's make sure that paddles knows the job is finished. We don't know
+    # the status, but if it was a pass or fail it will have already been
+    # reported to paddles. In that case paddles ignores the 'dead' status.
+    # If the job was killed, paddles will use the 'dead' status.
+    report.try_push_job_info(job_info, dict(status='dead'))
+
+
+def create_fake_context(job_config, block=False):
+    if job_config['owner'] is None:
+        job_config['owner'] = get_user()
```
tchaikov, line 242: why are you mutating the `job_config`? do we expect a changed `job_config` after calling `create_fake_context()`?

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,324 @@
...
+
+    if 'os_version' in job_config:
+        os_version = job_config['os_version']
+    else:
+        os_version = None
```
tchaikov, line 247:
```suggestion
    os_version = job_config.get('os_version'], None)
```

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,324 @@
...
+        'owner': job_config['owner'],
+        'archive': job_config['archive_path'],
+        'machine_type': job_config['machine_type'],
+        'os_type': job_config['os_type'],
+        'os_version': os_version,
+        'name': job_config['name'],
+    }
+
+    fake_ctx = FakeNamespace(ctx_args)
+    return fake_ctx
```
tchaikov, line 261:
```suggestion
    return FakeNamespace(ctx_args)
```

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def archive_logs(ctx, remote_path, log_path):
+    """
+    Archive directories from all nodes in a cliuster. It pulls all files in
+    remote_path dir to job's archive dir under log_path dir.
+    """
+    path = os.path.join(ctx.archive, 'remote')
+    if (not os.path.exists(path)):
+        os.mkdir(path)
```
tchaikov, line 288:
```suggestion
    os.mkdirs(path, exist_ok=True)
```

teuthology/misc.py
```diff
@@ -1345,3 +1345,18 @@ def sh(command, log_limit=1024, cwd=None, env=None):
...
+    Add key/value pair (local_dir: remote_dir) to job's info.yaml.
+    These key/value pairs are read to archive them in case of job timeout.
+    """
+    with open(os.path.join(ctx.archive, 'info.yaml'), 'r+') as info_file:
+        info_yaml = yaml.safe_load(info_file)
+        info_file.seek(0)
+        if 'archive' not in info_yaml:
+            info_yaml['archive'] = {local_dir: remote_dir}
+        else:
+            info_yaml['archive'][local_dir] = remote_dir
```
tchaikov, line 1361: please transpose these two branches, like

```suggestion
        if 'archive' in info_yaml:
            info_yaml['archive'][local_dir] = remote_dir
        else:
            info_yaml['archive'] = {local_dir: remote_dir}
```

kshtsk: I'm discovering new github features, if PR owner press "commit suggestion" it will merge or add another commit to the PR? What is the difference to "Add suggestion to batch"?
@tchaikov how do you add such kind of suggestions?

ShraddhaAg: @tchaikov done!

teuthology/dispatcher/supervisor.py
ShraddhaAg: ah, my mistake. Reverted.

teuthology/dispatcher/supervisor.py
```diff
@@ -0,0 +1,270 @@
...
+            verbose
+        )
+    except SkipJob:
+        return
+
+
+def run_job(job_config, teuth_bin_path, archive_dir, verbose):
+    safe_archive = safepath.munge(job_config['name'])
+    if job_config.get('first_in_suite') or job_config.get('last_in_suite'):
+        suite_archive_dir = os.path.join(archive_dir, safe_archive)
```
kshtsk, line 68: It is run_archive_dir, not a suite_archive_dir.

kshtsk, line 68: I suppose this is job archive directory not a suite one.

ShraddhaAg: Ah sorry, changed it to `job_archive`.

teuthology/misc.py
```diff
@@ -1345,3 +1347,62 @@ def sh(command, log_limit=1024, cwd=None, env=None):
...
+                '-print0',
+                run.Raw('|'),
+                'sudo',
+                'xargs',
+                '-0',
+                '--no-run-if-empty',
+                '--',
+                'gzip',
+                '--',
+            ],
```
kshtsk, line 1405: It would be better when you use more human readable form of run command which is easier to maintain, like:
```
ctx.cluster.run(
    args=(f"sudo find {remote_dir} -name *.log -print0 | "
          f"sudo xargs -0 --no-run-if-empty -- gzip --"),
    wait=False
)
```

scripts/dispatcher.py
```diff
@@ -0,0 +1,33 @@
...
+standard arguments:
+  -h, --help                     show this help message and exit
+  -v, --verbose                  be more verbose
+  -t, --tube TUBE                which beanstalk tube to read jobs from
+  -l, --log-dir LOG_DIR          path in which to store logs
+  --archive-dir ARC_DIR          path to archive results in
+  --supervisor                   run dispactcher in job supervisor mode
+  --bin-path BIN_PATH            teuthology bin path
+  --job-config CONFIG                 file descriptor of job's config file
+"""
```
kshtsk, line 24: When you run teuthoogy-worker:
```
teuthology-worker --help
usage: teuthology-worker [-h] [-v] --archive-dir DIR -l LOG_DIR -t TUBE

Grab jobs from a beanstalk queue and run the teuthology tests they describe. One job is run at a time.

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         be more verbose
  --archive-dir DIR     path under which to archive results
  -l LOG_DIR, --log-dir LOG_DIR
                        path in which to store logs
  -t TUBE, --tube TUBE  which beanstalk tube to read jobs from
```
or teuthology:
```
teuthology --help
usage: teuthology --help
       teuthology --version
       teuthology [options] [--] <config>...

Run ceph integration tests

positional arguments:
  <config> one or more config files to read

optional arguments:
  -h, --help                     show this help message and exit
  -v, --verbose                  be more verbose
  --version                      the current installed version of teuthology
  -a DIR, --archive DIR          path to archive results in
  --description DESCRIPTION      job description
  --owner OWNER                  job owner
  --lock                         lock machines for the duration of the run
  --machine-type MACHINE_TYPE    Type of machine to lock/run tests on.
  --os-type OS_TYPE              Distro/OS of machine to run test on.
  --os-version OS_VERSION        Distro/OS version of machine to run test on.
  --block                        block until locking machines succeeds (use with --lock)
  --name NAME                    name for this teuthology run
  --suite-path SUITE_PATH        Location of ceph-qa-suite on disk. If not specified,
                                 it will be fetched
```

You may see there is archive dir, like:
```
 -a DIR, --archive DIR
```
Could you please align teuthology-dispatcher to use the same style, i.e. use `DIR` instead of `ARC_DIR` and, it ould be great to have short option for it.

scripts/dispatcher.py
```diff
@@ -0,0 +1,33 @@
...
+
+standard arguments:
+  -h, --help                     show this help message and exit
+  -v, --verbose                  be more verbose
+  -t, --tube TUBE                which beanstalk tube to read jobs from
+  -l, --log-dir LOG_DIR          path in which to store logs
+  --archive-dir ARC_DIR          path to archive results in
+  --supervisor                   run dispactcher in job supervisor mode
+  --bin-path BIN_PATH            teuthology bin path
+  --job-config CONFIG                 file descriptor of job's config file
```
kshtsk, line 23: Fix indentation.

scripts/dispatcher.py
```diff
@@ -0,0 +1,33 @@
...
+target machines and invoke teuthology command. Unlock the target machines
+at the end of the run.
+
+standard arguments:
+  -h, --help                     show this help message and exit
+  -v, --verbose                  be more verbose
+  -t, --tube TUBE                which beanstalk tube to read jobs from
+  -l, --log-dir LOG_DIR          path in which to store logs
+  --archive-dir ARC_DIR          path to archive results in
+  --supervisor                   run dispactcher in job supervisor mode
```
kshtsk, line 21: fix typo in dispatcher

teuthology/lock/ops.py
```diff
@@ -297,3 +285,153 @@ def push_new_keys(keys_dict, reference):
...
+            log.info("Start node '%s' reimaging", machine)
+            update_nodes([machine], True)
+            p.spawn(teuthology.provision.reimage, ctx,
+                    machine, machine_type)
+            reimaged[machine] = machines[machine]
+            log.info("Node '%s' reimaging is complete", machine)
+    return reimaged
+
+
+def reimage_many(ctx, machines, machine_type):
```
kshtsk, line 303: This is not `reimage_many`, giving sufix _many_ makes function to expect a number of machines for a given type. I suggest to rename it to something `reimage` or `reimage_machines`.

teuthology/lock/ops.py
```diff
@@ -297,3 +285,153 @@ def push_new_keys(keys_dict, reference):
...
+    reimaged = dict()
+    with teuthology.parallel.parallel() as p:
+        for machine in machines:
+            log.info("Start node '%s' reimaging", machine)
+            update_nodes([machine], True)
+            p.spawn(teuthology.provision.reimage, ctx,
+                    machine, machine_type)
+            reimaged[machine] = machines[machine]
+            log.info("Node '%s' reimaging is complete", machine)
+    return reimaged
```
kshtsk, line 300: Why this methods was split off the main reimage code?

teuthology/lock/ops.py
```diff
@@ -297,3 +285,153 @@ def push_new_keys(keys_dict, reference):
...
+                 for machine in machines],
+    )
+    with console_log.task(ctx, console_log_conf):
+        reimaged = reimage(ctx, machines, machine_type)
+    reimaged = do_update_keys(list(reimaged.keys()))[1]
+    update_nodes(reimaged)
+    return reimaged
+
+
+def block_and_lock_machines(ctx, total_requested, machine_type, reimage=True):
```
kshtsk, line 318: Why this so strange name? block_and_lock_machines? what doe it mean block?

kshtsk: I suppose you meant that lock_machines will be trying to lock in loop even if it was not able to actually lock or provision the node successfully.
This is probably out of the scope for this PR but this method should be combined in future with lock_many, and lock_one, and present the same method which is used by: dispatcher, the task, and lock cli.

kshtsk:
Wonder if worker still works.

kshtsk:
@susebot run deploy

scripts/dispatcher.py
ShraddhaAg: done.

ShraddhaAg: Done.

teuthology/misc.py
ShraddhaAg: right, much more readable. Done.

teuthology/lock/ops.py
ShraddhaAg: Yeah, I used `block_and_lock_machines` since we are initially waiting to block machines and then lock them. 

Like you said, the locking mechanism is definitely split in a lot of places and could use some refactoring.

teuthology/lock/ops.py
ShraddhaAg: I wanted to use the method in cli and the `teuthology-reimage` cmd in a future PR since they are essentially repeating the same.

teuthology/lock/ops.py
ShraddhaAg: Changed it to `reimage_machines`.

kshtsk:
@susebot run deploy

teuthology/misc.py
tchaikov: image

kshtsk:
We need to update manual within the docs directory how to use and deploy the dispatcher. Also, do we plan any works on enabling a dispatcher in the sepia production server?

jdurgin:
Yes, would like to deploy to sepia as soon as we can to work out any remaining bugs. We could start by using it just for one machine time, like mira, that is less critical, fix any bugs found, then deploy it for smithi (where all the automated jobs are run).

ShraddhaAg:
> We need to update manual within the docs directory how to use and deploy the dispatcher. 

I have added the `teuthology-dispatcher` cmd page and updated COMPONENTS.rst with the new architecture and working details. Since dispatcher will not be directly interacted with by the users of teuthology, I'm not sure if more details are required on how to use it and if so, where should I add them? 

Regarding how to deploy the dispatcher, other than changing the cmd `teuthology-worker` to `teuthology-dispacther`, not much has changed syntactically. I do although would like to update:
* [Teuthology lab setup notes](https://docs.ceph.com/teuthology/docs/LAB_SETUP.html)
* the script, [worker_start.sh](https://raw.githubusercontent.com/ceph/teuthology/master/docs/_static/worker_start.sh) (mentioned in the above doc to start workers) to start single dispatchers for various machine types. 

Please let me know if the above sounds right.

kshtsk:
I was going to try this changes on my desktop, but I need to merge the #1548 first.

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+                           job_config['archive_path'],
+                           archive_dir)
+        job_config_path = os.path.join(job_config['archive_path'], 'orig.config.yaml')
+
+        # Write initial job config in job archive dir
+        with open(job_config_path, 'w') as f:
+            yaml.safe_dump(job_config, f, default_flow_style=False)
+
+        run_args.extend(["--job-config", job_config_path])
+        job_proc = subprocess.Popen(run_args)
```
jdurgin, line 146: looks like we need some error handling here in case spawning the process fails for some reason, e.g.:

```Traceback (most recent call last):
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_wip-minimal-dispatcher/virtualenv/bin/teuthology-dispatcher", line 33, in <module>
    sys.exit(load_entry_point('teuthology', 'console_scripts', 'teuthology-dispatcher')())
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_wip-minimal-dispatcher/scripts/dispatcher.py", line 33, in main
    teuthology.dispatcher.main(args)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_wip-minimal-dispatcher/teuthology/dispatcher/__init__.py", line 146, in main
    job_proc = subprocess.Popen(run_args)
  File "/usr/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/home/teuthworker/src/git.ceph.com_git_teuthology_master/virtualenv/bin/teuthology-dispatcher': '/home/teuthworker/src/git.ceph.com_git_teuthology_master/virtualenv/bin/teuthology-dispatcher'```

To keep things simple we could log the error, delete the job, and continue. Ideally we could mark the job failed with the exception we got too.

Comment (ShraddhaAg) [reply teuthology/dispatcher/__init__.py]: Ah right. Now logs error message along with the exception raised, unlocks targets and reports status.

**Comment by jdurgin** [comment]
thanks, that looks good. only other thing I noticed was due to problems with checking out the teuthology branch again, which would be resolved by the separate teuthology workspaces in the future.

File: docs/COMPONENTS.rst [hunk 31]
```diff
@@ -21,47 +21,51 @@ a scheduler (a.k.a. master node), worker(s), package builder
...
 
 The teuthology scheduler exposes a work queue (using `beanstalkd
-<https://kr.github.io/beanstalkd/>`__) where jobs are submitted. The life of a 
-job begins when ``teuthology-suite`` is executed, which is when a job is 
-prepared and queued (``teuthology-schedule`` is implicitly invoked). When a job 
-is created (or whenever the status of a job is changed, e.g. from queued to 
-started), information about the job is recorded in ``paddles``'s internal 
-database. Depending on the requirements (in terms of requested machines), the 
-scheduler eventually determines when a job can get executed. At this point the 
-master communicates with ``teuthology-worker``, which in turn invokes
```
kshtsk, line 48: @jdurgin so with this change you guys are dropping any reference to teuthology-worker?

teuthology/dispatcher/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+
+def main(args):
+    # run dispatcher in job supervisor mode if --supervisor passed
+    if args["--supervisor"]:
+        return supervisor.main(args)
+
+    verbose = args["--verbose"]
+    tube = args["--tube"]
+    log_dir = args["--log-dir"]
+    archive_dir = args["--archive-dir"]
```
kshtsk, line 65: Can we add the default value for archive taken from teuthology config, that will make --archive-dir optional.

kshtsk:
@jdurgin were you able to test it appropriately?

jdurgin:
@kshtsk hey, sorry I forgot to update the PR. Discussed with @ShraddhaAg on IRC a bit ago, I saw errors from using the same teuthology checkout as it was being changed/rechecked out, so it seems we need the independent checkouts for this to work well.

@ShraddhaAg do you have time to look at that? If not someone else can finish that part, and then we can re-test.

docs/COMPONENTS.rst
jdurgin: yes, I think it'll be obsolete once the dispatcher si stabilized

ShraddhaAg:
@jdurgin yes, I'll add teuthology workspace bit.

jdurgin:
I've been testing this with the mira queue in sepia, and it's been working well with https://github.com/ceph/teuthology/pull/1599 . It requires jobs to be scheduled with teuthology_branch: wip-dispatcher to work right now.

I'd like to merge and start trying it with smithi next. @kshtsk what do you think?

jdurgin (approved):

kshtsk:
@susebot run deploy

kshtsk:
I guess we need to remove [WIP] tag before merge.

PR #1548, opened 2020-08-13
Setup development environment for teuthology on your laptop

kshtsk:


jdurgin:
very exciting! was just discussing similar ideas with @rzarzynski

kshtsk:
> Great work! Looks like a nice way to setup a teuthology dev environment.
> 
> It'd be nice to get all these steps into a script of some sort, so it could be created with one command.

Yeah, this is what I am thinking about too. Shell script sounds fastest solution, but probably we can introduce a new tool, like teuthology-sandbox or similar.

> IIRC you mentioned downburst may need some changes to work with centos or fedora. 

Not sure I was saying about the downburst itself. But that make sense to, however I expect it be working.

> Being able to use podman instead of docker would be nice too.

I was thinking about docker compose, but my colleges also want to podman, I will look into that.

kshtsk:
@jdurgin I've made some docs cleanup, could you please give another look.

kshtsk:
@susebot run deploy

docs/laptop/README.md
```diff
@@ -0,0 +1,434 @@
...
+insert into nodes (name, machine_type, is_vm, locked, up, mac_address, vm_host_id) values ('target-02.local', 'vps', true, false, false, '52:54:00:00:00:02', (select id from nodes where name='localhost'));
+insert into nodes (name, machine_type, is_vm, locked, up, mac_address, vm_host_id) values ('target-03.local', 'vps', true, false, false, '52:54:00:00:00:03', (select id from nodes where name='localhost'));
+```
+or just use the following command:
+
+```bash
+psql -h localhost -U paddles -d paddles < docs/laptop/targets.sql
+```
+
+Add libirt config file so downburst able to use 'localhost' node to connect to:
```
jdurgin, line 130:
```suggestion
Add libvirt config file so downburst able to use 'localhost' node to connect to:
```

kshtsk: Updated

kshtsk:
@jdurgin I've addressed your comment, but the review is reset.

PR #1550, opened 2020-08-19
teuthology-suite: Check the priority of jobs to be run

varshar16:
Check if the passed testing priority is according to the range mentioned in
developer guide[1]. This patch also adds '--force-priority' flag which allows to skip
the priority check.

[1] https://docs.ceph.com/docs/master/dev/developer_guide/tests-integration-tests/#testing-priority

Signed-off-by: Varsha Rao <varao@redhat.com>

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -168,6 +168,7 @@ Scheduler arguments:
                               with --rerun argument. This number can be found
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
+ --force-priority             Skip the priority check.
 
 """.format(
     default_machine_type=config.default_machine_type,
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -501,6 +501,22 @@ class Run(object):
                 log.info("pause between jobs : --throttle " + str(throttle))
                 time.sleep(int(throttle))
 
+    def check_priority(self, jobs_to_schedule):
+        priority = self.args.priority
+        msg='''Use the following testing priority
+10 to 49: Tests which are urgent and blocking other important development.
+50 to 74: Testing a particular feature/fix with less than 25 jobs and can also be used for urgent release testing.
+75 to 99: Tech Leads usually schedule integration tests with this priority to verify pull requests against master.
+100 to 149: QE validation of point releases.
+150 to 199: Testing a particular feature/fix with less than 100 jobs and results will be available in a day or so.
+200 to 1000: Large test runs that can be done over the course of a week.
+Note: To force run, use --force-priority'''
+        if priority < 50:
+            util.schedule_fail(msg)
+        elif priority < 75 and jobs_to_schedule > 25:
+            util.schedule_fail(msg)
+        elif priority < 150 and jobs_to_schedule > 100:
+            util.schedule_fail(msg)
 
     def schedule_suite(self):
         """
@@ -613,6 +629,10 @@ class Run(object):
 
         if jobs_to_schedule:
             self.write_rerun_memo()
+
+        # Before scheduling jobs, check the priority
+        if self.args.priority and jobs_to_schedule and not self.args.force_priority:
+            self.check_priority(len(jobs_to_schedule))
 
         self.schedule_jobs(jobs_missing_packages, jobs_to_schedule, name)
 


scripts/suite.py
```diff
@@ -168,6 +168,7 @@
                               with --rerun argument. This number can be found
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
+ --skip                       Skip the priority check.
```
kshtsk, line 171: The `--skip` option is ambiguous. It is better to rename it to `--check-priority` with defaults to False.

teuthology/suite/run.py
```diff
@@ -614,6 +622,9 @@ def schedule_suite(self):
         if jobs_to_schedule:
             self.write_rerun_memo()
 
+        if jobs_to_schedule and not self.args.skip:
```
kshtsk, line 625: Maybe add a comment why this check should go exactly here?

scripts/suite.py
jdurgin: I think the idea is to stop inappropriate use of high priority by default, so we'd want the check on by default. maybe ``--force-priority`` or something?

teuthology/suite/run.py
```diff
@@ -501,6 +501,14 @@ def schedule_jobs(self, jobs_missing_packages, jobs_to_schedule, name):
...
                 time.sleep(int(throttle))
 
+    def check_priority(self, jobs_to_schedule, name):
+        priority = self.args.priority
+        if priority < 50:
+            util.schedule_fail('This priority is used only in emergency. To force run, pass --skip', name)
+        elif priority < 75 and jobs_to_schedule > 25:
+            util.schedule_fail('To be used for running less than 25 jobs. To force run, pass --skip', name)
+        elif priority < 150 and jobs_to_schedule > 100:
+            util.schedule_fail('To be used for running less than 100 jobs. To force run, pass --skip', name)
```
kshtsk, line 511: What do you think about adding a reference to the message with detailed description how priorities should be used, like you've added to the commit message:

Or just print whole hint:
```Priority < 10: Use this if the sky is falling and some group of tests must be run ASAP.

10 <= Priority < 50: Use this if your tests are urgent and blocking other important development.

50 <= Priority < 75: Use this if you are testing a particular feature/fix and running fewer than about 25 jobs. This range can also be used for urgent release testing.

75 <= Priority < 100: Tech Leads will regularly schedule integration tests with this priority to verify pull requests against master.

100 <= Priority < 150: This priority is to be used for QE validation of point releases.

150 <= Priority < 200: Use this priority for 100 jobs or fewer of a particular feature/fix that you’d like results on in a day or so.

200 <= Priority < 1000: Use this priority for large test runs that can be done over the course of a week.```

File: teuthology/suite/run.py [hunk 4]
```diff
@@ -501,6 +501,14 @@ def schedule_jobs(self, jobs_missing_packages, jobs_to_schedule, name):
                 log.info("pause between jobs : --throttle " + str(throttle))
                 time.sleep(int(throttle))
 
+    def check_priority(self, jobs_to_schedule, name):
```
kshtsk, line 504: Not sure, why do we need a suite name here?

teuthology/suite/run.py
kshtsk: The priority should not be checked if `--priority` option is not used.

teuthology/suite/run.py
neha-ojha: I think we should also update https://docs.ceph.com/docs/master/dev/developer_guide/tests-integration-tests/#testing-priority to reflect this change.

scripts/suite.py
kshtsk: The `--force-priority` which allows to skip the check looks fine to me.

varshar16:
@kshtsk Thanks for the clarification. I have updated the PR. Please review it again.

teuthology/suite/run.py
```diff
@@ -614,6 +630,10 @@ def schedule_suite(self):
         if jobs_to_schedule:
             self.write_rerun_memo()
 
+        # Before scheduling jobs, check the priority
+        if jobs_to_schedule and not self.args.force_skip_priority:
```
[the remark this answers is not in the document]
varshar16: > @kshtsk Thanks, done.

scripts/suite.py
```diff
@@ -168,6 +168,7 @@
                               with --rerun argument. This number can be found
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
+ --force-skip-priority        Skip the priority check.
```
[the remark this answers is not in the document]
varshar16: Done.

teuthology/suite/run.py
varshar16: Right, I have modified the condition.

scripts/suite.py
varshar16: @kshtsk Thank you for the detailed explanation.  :)

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1551, opened 2020-08-25
kill: find targets for killing a job

kshtsk:
In some circumstances teuthology-kill does not unlock the nodes
if it is used to kill just a single job from a run, for example,
while a job in the middle of locking many targets the job_info
does not include targets yet, corresponding if someone kills
the job in the middles, the targets remains in locked state.
This commit address the issue.

### DIFF
--- a/teuthology/kill.py
+++ b/teuthology/kill.py
@@ -72,7 +72,10 @@ def kill_job(run_name, job_id, archive_base=None, owner=None):
                 "Please pass --owner <owner>.")
         owner = job_info['owner']
     kill_processes(run_name, [job_info.get('pid')])
-    targets = dict(targets=job_info.get('targets', {}))
+    # Because targets can be missing for some cases, for example, when all
+    # the necessary nodes ain't locked yet, we do not use job_info to get them,
+    # but use find_targets():
+    targets = find_targets(run_name, owner, job_id)
     nuke_targets(targets, owner)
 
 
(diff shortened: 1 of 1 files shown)


kshtsk:
@susebot run deploy

teuthology/kill.py
```diff
@@ -191,12 +194,12 @@ def find_pids(run_name):
     return run_pids
 
 
-def find_targets(run_name, owner):
+def find_targets(run_name, owner, job_id=None):
```
tchaikov, line 197:
```suggestion
def find_targets(owner, run_name, job_id=None):
```

tchaikov (approved):
just two nits.

teuthology/kill.py
kshtsk: why do you want to change the order?

teuthology/kill.py
```diff
@@ -191,12 +194,12 @@ def find_pids(run_name):
...
     lock_args = [
         'teuthology-lock',
         '--list-targets',
         '--desc-pattern',
-        '/' + run_name + '/',
+        '/' + run_name + '/' +  (job_id if job_id else ''),
```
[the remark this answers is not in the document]
kshtsk: That makes sense, I need to test it.

teuthology/kill.py
tchaikov: because `run_name` is semantically closer to `job_id`. it's like pass `find_time_slot(user, year, month)`.

kshtsk: good point, though let me not change the signature that much for now...

djgalloway:
I'm good if @tchaikov's good.

kshtsk:
@susbot run deploy

tchaikov (approved):

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

PR #1552, opened 2020-08-26
run_tasks: improve sentry reporting

jdurgin:
Correct the url generated, and include additional tags for filtering by suite/machine_type/os.

### DIFF
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -111,10 +111,11 @@ def run_tasks(tasks, ctx):
                 'task': taskname,
                 'owner': ctx.owner,
             }
-            if 'teuthology_branch' in config:
-                tags['teuthology_branch'] = config['teuthology_branch']
-            if 'branch' in config:
-                tags['branch'] = config['branch']
+            optional_tags = ('teuthology_branch', 'branch', 'suite',
+                             'machine_type', 'os_type', 'os_version')
+            for tag in optional_tags:
+                if tag in config:
+                    tags[tag] = config[tag]
 
             # Remove ssh keys from reported config
             if 'targets' in config:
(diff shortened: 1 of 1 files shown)


teuthology/run_tasks.py
```diff
@@ -111,10 +111,11 @@ def run_tasks(tasks, ctx):
...
             }
-            if 'teuthology_branch' in config:
-                tags['teuthology_branch'] = config['teuthology_branch']
-            if 'branch' in config:
-                tags['branch'] = config['branch']
+            optional_tags = ('teuthology_branch', 'branch', 'suite',
+                             'machine_type', 'os_type', 'os_version')
+            for tag in optional_tags:
+                if tag in config:
+                    tags[tag] = config[tag]
```
kshtsk, line 118: Why do not use standard teuthology deep_merge and invent something new?

teuthology/run_tasks.py
```diff
@@ -111,10 +111,11 @@ def run_tasks(tasks, ctx):
                 'task': taskname,
                 'owner': ctx.owner,
             }
-            if 'teuthology_branch' in config:
-                tags['teuthology_branch'] = config['teuthology_branch']
-            if 'branch' in config:
-                tags['branch'] = config['branch']
+            optional_tags = ('teuthology_branch', 'branch', 'suite',
+                             'machine_type', 'os_type', 'os_version')
+            for tag in optional_tags:
```
kshtsk, line 116: This change looks unrelated to the PR name.

teuthology/run_tasks.py
jdurgin: it's doing something different - we're only adding a subset of values from the config. deep_merge would include all of them.

teuthology/run_tasks.py
jdurgin: changed the title after I added a 2nd commit

teuthology/run_tasks.py
kshtsk: yeah... I see it now

badone (approved):

PR #1553, opened 2020-08-26
task/install: skip package removal by default

batrick:
Every teuthology job cleans up its package install after completion.
This was necessary cleanup when we didn't reimage boxes before the use
of FOG as we wanted a "clean" slate for the next job to acquire the
machine. Now this is just unnecessary work which takes up valuable
machine time. For one job I looked at, this takes about 2 minutes.

We should still test that there are no unexpected issues with removing
the packages but this can be delegated to a small subset of smoke tests.
That will be posted in another PR to ceph.git.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -8,6 +8,7 @@ from teuthology.orchestra import run
 from teuthology.orchestra import connection
 from teuthology.orchestra import console
 from teuthology.orchestra.opsys import OS
+import teuthology.provision
 from teuthology import misc
 from teuthology.exceptions import CommandFailedError
 from teuthology.misc import host_shortname
@@ -33,6 +34,7 @@ class Remote(object):
 
     # for unit tests to hook into
     _runner = staticmethod(run.run)
+    _reimage_types = None
 
     def __init__(self, name, ssh=None, shortname=None, console=None,
                  host_key=None, keep_alive=True):
@@ -52,6 +54,9 @@ class Remote(object):
         self.keep_alive = keep_alive
         self._console = console
         self.ssh = ssh
+
+        if self._reimage_types is None:
+            Remote._reimage_types = teuthology.provision.get_reimage_types()
 
     def connect(self, timeout=None, create_key=None, context='connect'):
         args = dict(user_at_host=self.name, host_key=self._host_key,
@@ -149,6 +154,10 @@ class Remote(object):
                 return None
             self._machine_type = remote_info.get("machine_type", None)
         return self._machine_type
+
+    @property
+    def is_reimageable(self):
+        return self.machine_type in self._reimage_types
 
     @property
     def shortname(self):
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -99,11 +99,13 @@ def remove_packages(ctx, config, pkgs):
         "deb": deb._remove,
         "rpm": rpm._remove,
     }
+    cleanup = config.get('cleanup', False)
     with parallel() as p:
         for remote in ctx.cluster.remotes.keys():
-            system_type = teuthology.get_system_type(remote)
-            p.spawn(remove_pkgs[
-                    system_type], ctx, config, remote, pkgs[system_type])
+            if not remote.is_reimageable or cleanup:
+                system_type = teuthology.get_system_type(remote)
+                p.spawn(remove_pkgs[
+                        system_type], ctx, config, remote, pkgs[system_type])
 
 
 def remove_sources(ctx, config):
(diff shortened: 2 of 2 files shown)


sebastian-philipp:
really looking forward to see less nose when analyzing teuthology logs

kshtsk:
@batrick what about doing or not doing the cleanup based on `reimageable` target is or it is not. We still have code in lock module which relies on that logic.

batrick:
> @batrick what about doing or not doing the cleanup based on `reimageable` target is or it is not. We still have code in lock module which relies on that logic.

Sure, I'll have a look.

tchaikov (approved):

batrick:
> @batrick what about doing or not doing the cleanup based on `reimageable` target is or it is not. We still have code in lock module which relies on that logic.

@kshtsk is there a simple way to check if a remote can be reimaged? I didn't see any code that checks it (in the lock module or otherwise).

batrick:
See also: https://github.com/ceph/ceph/pull/36827

kshtsk:
> > @batrick what about doing or not doing the cleanup based on `reimageable` target is or it is not. We still have code in lock module which relies on that logic.
> 
> @kshtsk is there a simple way to check if a remote can be reimaged? I didn't see any code that checks it (in the lock module or otherwise).

You need to check the remote's machine_type presence in _teuthology.provision.get_reimage_types()_

teuthology/orchestra/remote.py
```diff
@@ -8,6 +8,7 @@
 from teuthology.orchestra import connection
 from teuthology.orchestra import console
 from teuthology.orchestra.opsys import OS
+from teuthology.provision import get_reimage_types
```
kshtsk, line 11: Just do:
```
import teuthology.provision
```

batrick: done

batrick: reverted this change, it gave me a bizarre import error.

batrick:
https://github.com/ceph/ceph/pull/36827#issuecomment-681310115

batrick:
https://pulpito.ceph.com/pdonnell-2020-08-27_18:05:18-smoke-master-distro-basic-smithi/

^ removes packages as usual. New fs run showing install unwind skipping removal:

https://pulpito.ceph.com/pdonnell-2020-08-27_19:37:17-fs-master-distro-basic-smithi/

teuthology/task/install/__init__.py
```diff
@@ -218,7 +222,6 @@ def install(ctx, config):
         remove_packages(ctx, config, package_list)
         remove_sources(ctx, config)
 
-
```
kshtsk, line 221: This is extra

teuthology/orchestra/remote.py
```diff
@@ -150,6 +155,10 @@ def machine_type(self):
             self._machine_type = remote_info.get("machine_type", None)
         return self._machine_type
 
+    @property
+    def is_reimageable(self):
```
kshtsk, line 159: I wonder why do we need `is_` prefix?

teuthology/task/install/__init__.py
batrick: fixed

teuthology/orchestra/remote.py
batrick: because that's fairly conventional?

````
$ git grep def\ is_ 
teuthology/lock/query.py:def is_vm(name=None, status=None):
teuthology/ls.py:    def is_job_dir(parent, subdir):
teuthology/misc.py:def is_type(type_, cluster=None):
teuthology/misc.py:def is_in_dict(searchkey, searchval, d):
teuthology/orchestra/remote.py:    def is_reimageable(self):
teuthology/orchestra/remote.py:    def is_online(self):
teuthology/orchestra/remote.py:    def is_vm(self):
teuthology/provision/pelagos.py:    def is_task_active(self, task_url):
teuthology/prune.py:def is_old_enough(file_name, days):
teuthology/repo_utils.py:def is_fresh(path):
teuthology/task/hadoop.py:def is_hadoop_type(type_):
````

Comment (kshtsk) [reply teuthology/orchestra/remote.py:159]: ok, thanks

File: teuthology/orchestra/remote.py [hunk 4]
```diff
@@ -53,6 +55,9 @@ def __init__(self, name, ssh=None, shortname=None, console=None,
         self._console = console
         self.ssh = ssh
 
+        if self._reimage_types is None:
+            self._reimage_types = teuthology.provision.get_reimage_types()
+
```
kshtsk, line 60: If you wished to set static class parameter I think you need to use `Remote._reimage_types` instead of `self._reimage_types`.

kshtsk: For the code:
```
class A(object):
    _x = None
    def __init__(self, name):
        if self._x is None:
            print(f"Setting _x for {name}")
            self._x = "A"

class B(object):
    _x = None
    def __init__(self, name):
        if B._x is None:
            print(f"Setting _x for {name}")
            B._x = "B"

a1 = A('a1')
a2 = A('a2')
print(f'a1={a1._x}')
print(f'a2={a2._x}')

b1 = B('b1')
b2 = B('b2')
print(f'b1={b1._x}')
print(f'b2={b2._x}')
```

we have the output:

```
Setting _x for a1
Setting _x for a2
a1=A
a2=A
Setting _x for b1
b1=B
b2=B
```

batrick: ACK, done

batrick:
https://pulpito.ceph.com/pdonnell-2020-08-27_21:19:21-smoke-master-distro-basic-smithi/

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1558, opened 2020-09-14
install: add adding yum repo support

lxbsz:
For the ceph-iscsi test suite, we need to add extra ceph-iscsi and
tcmu-runner repos from shaman to install the packages.

Fixes: https://tracker.ceph.com/issues/47431

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -454,8 +454,16 @@ def task(ctx, config):
             priority: 1
             url: "http://location.repo"
 
-    Note: The 'repos' are supported for SUSE-based distros only, but patches
-    are welcome to add support for other distros.
+        repos:
+          - name: "project-name"
+            url: "https://1.chacra.ceph.com/r/project-name/{branch}/{sha1}/{os_type}/{os_version}/flavors/{flavor}/{arch}
+
+    Note: The 'repos' are supported for RHEL/SUSE-based distros only, but
+    patches are welcome to add support for other distros.
+
+    For RHEL-based distros, the "name" will be used to get the {branch} and
+    {sha1} for the "project-name" and "priority" will be ingored, and the dnf
+    config-manager will generate a new repo name based the "url".
 
 
     Overrides are project specific:
--- a/teuthology/task/install/rpm.py
+++ b/teuthology/task/install/rpm.py
@@ -48,6 +48,8 @@ def _remove(ctx, config, remote, rpm):
     if repos:
         if dist_release in ['opensuse', 'sle']:
             _zypper_removerepo(remote, repos)
+        elif dist_release in ['rhel', 'fedora', 'centos']:
+            _yum_removerepo(ctx, config, remote)
         else:
             raise Exception('Custom repos were specified for %s ' % remote_os +
                             'but these are currently not supported')
@@ -61,6 +63,58 @@ def _remove(ctx, config, remote, rpm):
                  "because the test machine will be destroyed or reimaged anyway")
     else:
         remote.run(args='sudo yum clean expire-cache')
+
+def _yum_get_url(ctx, config, remote, repo):
+    builder = _get_builder_project(ctx, remote, config)
+    builder._init_from_remote()
+
+    os_type = builder.os_type
+    # such as for rhel 8.0, the repo link will use "8" instead.
+    os_version = builder.os_version.replace("8.0", "8")
+    flavor = builder.flavor
+    if flavor == "basic":
+        flavor = "default"
+    arch = builder.arch
+
+    branch = config.get(f"{repo['name']}-branch", 'master')
+    sha1 = config.get(f"{repo['name']}-sha1", '')
+    if sha1 == '':
+        raise
+
+    return repo['url'].format(branch=branch, sha1=sha1, os_type=os_type,
+                              os_version=os_version, flavor=flavor, arch=arch)
+
+def _yum_addrepo(ctx, config, remote):
+    """
+    Add yum repos to the remote system.
+
+    :param remote: remote node where to add packages
+    :param repo_list: list of dictionaries with keys 'url'.
+    :return:
+    """
+    repo_list = config.get('repos')
+    for repo in repo_list:
+        # add and enable the repo
+        url = _yum_get_url(ctx, config, remote, repo)
+        remote.sh(f'sudo dnf config-manager --add-repo --refresh --no-gpgcheck {url}')
+
+def _yum_removerepo(ctx, config, remote):
+    """
+    Remove yum repos on the remote system.
+
+    :param remote: remote node where to remove packages from
+    :param repo_list: list of dictionaries with keys 'url'
+    :return:
+    """
+    repo_list = config.get('repos')
+    for repo in repo_list:
+        # removes prefixed "https://" or "http://" and replaces
+        # all '/' with '_'
+        url = _yum_get_url(ctx, config, remote, repo)
+        repo_name = url.replace("https://", "") \
+                       .replace("http://", "") \
+                       .replace('/', '_')
+        remote.sh(f'sudo rm -f /etc/yum.repos.d/{repo_name}.repo')
 
 
 def _zypper_addrepo(remote, repo_list):
@@ -177,6 +231,8 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
         if dist_release in ['opensuse', 'sle']:
             _zypper_wipe_all_repos(remote)
             _zypper_addrepo(remote, repos)
+        elif dist_release in ['rhel', 'fedora', 'centos']:
+            _yum_addrepo(ctx, config, remote)
         else:
             raise Exception('Custom repos were specified for %s ' % remote_os +
                             'but these are currently not supported')


teuthology/task/install/rpm.py
```diff
@@ -62,6 +64,45 @@ def _remove(ctx, config, remote, rpm):
...
 
+def _yum_addrepo(remote, repo_list):
+    """
+    Add yum repos to the remote system.
+
+    :param remote: remote node where to add packages
+    :param repo_list: list of dictionaries with keys 'name', 'url', etc.
+    :return:
+    """
+    for repo in repo_list:
```
dillaman, line 75: See `task/internal/redhat.py` for an example, but you can download/write the file locally, copy it to the remote, and then "sudo cp" it to its destination. It will look cleaner than this. Some of the code from `ShamanProject` can also probably be refactored / reused to handle pulling the repo file down.

teuthology/task/install/rpm.py
```diff
@@ -62,6 +64,45 @@ def _remove(ctx, config, remote, rpm):
...
+        args=['sudo', 'touch', filepath]
+        remote.run(args=args)
+        args=['sudo', 'echo', '[{}]'.format(repo['name']), run.Raw('>'), filepath]
+        remote.run(args=args)
+        args=['sudo', 'echo', 'name={}'.format(repo['name']), run.Raw('>>'), filepath]
+        remote.run(args=args)
+        args=['sudo', 'echo', 'enabled=1', run.Raw('>>'), filepath]
+        remote.run(args=args)
+        args=['sudo', 'echo', 'baseurl={}'.format(repo['url']), run.Raw('>>'), filepath]
+        remote.run(args=args)
```
kshtsk, line 86: You have python but using this shell on the remote node?
I would suggest to compose repo data and just write to a remote file, for example:
```
    yum_repo_text = (
         f"[{repo['name']}]\n"
         f"name={repo['name']}\n"
         f"enabled=1\n"
         f"baseurl={repo['url']}\n")
    if 'gpgcheck' in repo:
         yum_repo_text += (
              f"gpgcheck={repo['gpgcheck']}\n"
              f"gpgkey={repo['gpgkey']}\n")
    remote.write_file(yum_repo_file, yum_repo_text, sudo=True)
```

teuthology/task/install/rpm.py
kshtsk: @dillaman why this is so complicated?

kshtsk: In general the idea was to use commands like `zypper` or `yum`.

teuthology/task/install/rpm.py
kshtsk: but as I said earlier, since we are using `zypper addrepo` command for zypper, can we use similar `yum` command for repos manipulation?

teuthology/task/install/rpm.py
dillaman: The end goal is to be able to pull a dynamic repo using the shaman API (i.e. pull the latest master branch repo for "ceph-iscsi" for the EL8 environment).

kshtsk: @dillaman I don't understand how is this code related for shaman? The shaman repos are used in different place. This code is used when you provide test repo list in suite yaml or in override block.

teuthology/task/install/__init__.py
```diff
@@ -454,7 +454,12 @@ def task(ctx, config):
             priority: 1
             url: "http://location.repo"
 
-    Note: The 'repos' are supported for SUSE-based distros only, but patches
+          - name: "repo-alias"
+            gpgcheck = 1
+            gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
```
kshtsk, line 459: This is invalid for yaml syntax.

teuthology/task/install/rpm.py
tchaikov: agreed. `dnf config-manager --add-repo` and `dnf config-manager --setopt` should do the trick.

lxbsz: Yeah, sure. Will switch to `dnf config-manager --add-repo`.

teuthology/task/install/__init__.py
lxbsz: Yeah, the `dnf config-manager --add-repo` won't need this any more.

teuthology/task/install/rpm.py
tchaikov: @kshtsk it might not directly related to shaman. but if the only use case is to use the repos exposed by shaman RESTful service. it might be a lot simpler to reuse `ShamanProject` , as we can query shaman for a ready-to-use repo file, instead of composing a `dnf` command.

tchaikov: from the description in the tracket ticket (https://tracker.ceph.com/issues/47431):

> For the ceph-iscsi test suite, we need to add extra ceph-iscsi and tcmu-runner repos from shaman to install the packages.

i conclude that these repos are exposed by shaman. i think the code is confusing. as if we want to packages exposed by shaman, instead of looking for a "repo", what this method should accept `project`, `branch`, `sha1`, `dist`, `codename`. as shaman expects them in, for instance, following RESTful API:

> https://shaman.ceph.com/api/repos/${project}/${branch}/${sha1}/${dist}/${codename}/repo

teuthology/task/install/rpm.py
```diff
@@ -177,6 +214,8 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
         if dist_release in ['opensuse', 'sle']:
             _zypper_wipe_all_repos(remote)
             _zypper_addrepo(remote, repos)
+        if dist_release in ['rhel', 'fedora', 'centos']:
```
tchaikov, line 217:
```suggestion
        elif dist_release in ['rhel', 'fedora', 'centos']:
```

otherwise this breaks openSUSE and SLE support.

teuthology/task/install/rpm.py
```diff
@@ -62,6 +64,41 @@ def _remove(ctx, config, remote, rpm):
...
+        # remove prefixed "https://" or "http://"
+        repo_name = repo['url'].replace("https://", "")
+        repo_name = repo_name.replace("http://", "")
+
+        # then replace all '/' with '_'
+        repo_name = repo_name.replace('/', '_')
+
+        remote.run(args=[
+            'sudo', 'rm', '-f', '/etc/yum.repos.d/{0}.repo'.format(repo_name)
+            ])
```
tchaikov, line 100:
```suggestion
        remote.sh(f'sudo rm -f /etc/yum.repos.d/{repo_name}.repo')
```

teuthology/task/install/rpm.py
```diff
@@ -62,6 +64,41 @@ def _remove(ctx, config, remote, rpm):
...
+    :param remote: remote node where to add packages
+    :param repo_list: list of dictionaries with keys 'url'.
+    :return:
+    """
+    for repo in repo_list:
+        # add and enable the repo
+        remote.run(args=[
+            'sudo', 'dnf', 'config-manager', '--add-repo', '--refresh',
+            '--no-gpgcheck', repo['url'],
+            ])
```
tchaikov, line 80:
```suggestion
        remote.sh(f'sudo dnf config-manager --add-repo --refresh --no-gpgcheck {repo["url"]}')
```

tchaikov (changes requested):

teuthology/task/install/rpm.py
lxbsz: Will fix it.

teuthology/task/install/rpm.py
lxbsz: Will fix it.

teuthology/task/install/rpm.py
lxbsz: Will fix it.

teuthology/task/install/rpm.py
lxbsz: @tchaikov 

The link from `https://shaman.ceph.com/api/repos/XXX` won't work for the `dnf config-manager --add-repo https://shaman.ceph.com/api/repos/XXX.../repo`.

The `dnf config-manager` is trying to find an `example.repo` file from the URL, if there is no it will just create a repo and set the `baseurl` to the URL specified and then it will be like:

```
name=created by dnf config-manager from https://1.chacra.ceph.com/repos/XXX/repo
baseurl=https://shaman.ceph.com/api/repos/XXX/repo
enabled=1
```  

But the URL here couldn't find the packages we needed.

Maybe we should do:
such as for the repo in `https://1.chacra.ceph.com/repos/tcmu-runner/master/ebffeff02ac35ef3f09371c41038f5c83e993ad1/centos/8/flavors/default/repo`
```
    - install:
        repos:
          -  prefix: "https://1.chacra.ceph.com/r/"
             project: "tcmu-runner"
             branch: "mater"
             sha1: "XXXXX"
             dist: "centos"
             relase: "8"
             flavers: "default"
             archs: "x86_64, noarch, RPMS"
```

Then do `sudo dnf config-manager --add-repo ${url}/${project}/${branch}/${sha1}/${disto}/${release}/flavors/${flaver}/${arch}

teuthology/task/install/rpm.py
```diff
@@ -48,6 +48,8 @@ def _remove(ctx, config, remote, rpm):
     if repos:
         if dist_release in ['opensuse', 'sle']:
             _zypper_removerepo(remote, repos)
+        if dist_release in ['rhel', 'fedora', 'centos']:
```
tchaikov, line 51:
```suggestion
        elif dist_release in ['rhel', 'fedora', 'centos']:
```

teuthology/task/install/rpm.py
tchaikov: > The link from https://shaman.ceph.com/api/repos/XXX won't work for the dnf config-manager --add-repo https://shaman.ceph.com/api/repos/XXX.../repo.

i am not suggesting use RESTful API along with `dnf config-manager`. i think you missed my comment at https://github.com/ceph/teuthology/pull/1558/files#r490040959

teuthology/task/install/rpm.py
```diff
@@ -62,6 +64,35 @@ def _remove(ctx, config, remote, rpm):
...
+    :param repo_list: list of dictionaries with keys 'url'
+    :return:
+    """
+    for repo in repo_list:
+        # remove prefixed "https://" or "http://"
+        repo_name = repo['url'].replace("https://", "")
+        repo_name = repo_name.replace("http://", "")
+
+        # then replace all '/' with '_'
+        repo_name = repo_name.replace('/', '_')
```
kshtsk, line 93: (style) the code is error prone, it would be much better for readability to use chaining and right to variable once:
```
repo_name = repo['url'].replace("https://", "") \
                       .replace("http://", "") \
                       .replace('/', '_')

```

teuthology/task/install/rpm.py
kshtsk: I understand what the ticket is about but `repos` is opposite what you want, of course it can be used by the code which is use Shaman class, but the feature of the repos is exactly to override what is used by Shaman, for example, we downstream just use it with obs and by external scheduler when we override the repos.

kshtsk: > Then do `sudo dnf config-manager --add-repo ${url}/${project}/${branch}/${sha1}/${disto}/${release}/flavors/${flaver}/${arch}

instead of this, we can add a feature for `url` and `name` to render job properties, like:
```
install:
  repos:
    - name: "tcmu-runner-{branch}"
      url: "https://1.chacra.ceph.com/r/tcmu-runner/{branch}/{sha1}/{os_type}/{os_version}/flavors/{flavor}/{arch}
```

lxbsz:
@tchaikov @kshtsk 
I think we can discard this change. And for the tcmu-runner/open-iscsi packages, we can simply use the Shaman class and it will be simple:

 ```
   - install:
        project: ceph-iscsi
        branch: master
        sha1: ebffeff02ac35ef3f09371c41038f5c83e993ad1
```

tchaikov:
@lxbsz are you going to close this PR or want to keep it around in case we want to enable a random RPM repo which is not exposed by shaman?

lxbsz:
> @lxbsz are you going to close this PR or want to keep it around in case we want to enable a random RPM repo which is not exposed by shaman?

Let's keep it for the other none Shaman repos.

lxbsz:
No need this any more and will close it.

PR #1564, opened 2020-09-26
kill.py: Allow deleting runs where bootstrap is failing

djgalloway:
`<kyr> looks like because bootstrapping failed, the teuthology-kill was not able to find machine_type and owner from the run archive directory, the directory has been created but no actual job started somehow`

### DIFF
--- a/teuthology/kill.py
+++ b/teuthology/kill.py
@@ -47,8 +47,13 @@ def kill_run(run_name, archive_base=None, owner=None, machine_type=None,
         run_archive_dir = os.path.join(archive_base, run_name)
         if os.path.isdir(run_archive_dir):
             run_info = find_run_info(serializer, run_name)
-            machine_type = run_info['machine_type']
-            owner = run_info['owner']
+            if 'machine_type' in run_info:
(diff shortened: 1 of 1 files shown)


teuthology/kill.py
```diff
@@ -47,8 +47,13 @@ def kill_run(run_name, archive_base=None, owner=None, machine_type=None,
         run_archive_dir = os.path.join(archive_base, run_name)
         if os.path.isdir(run_archive_dir):
             run_info = find_run_info(serializer, run_name)
-            machine_type = run_info['machine_type']
-            owner = run_info['owner']
+            if 'machine_type' in run_info:
+                machine_type = run_info['machine_type']
+                owner = run_info['owner']
+            else:
```
ryg-, line 53: Please see also usage of `machine_type' in line 62. Should it be done in case of None value?
I suppose, in case job is already failed no operation should be done with queue.

kshtsk: Makes sense to me, the current PR is a quick fix for an issue and was a workaround.

kshtsk (approved):

PR #1569, opened 2020-10-10
packaging: add 'version_major' option support

lxbsz:
In case for the distros having the same major version, such as for
RHEL 8/8.1/8.2, etc, and there has only one repo using the major
version in the url for one specified package.

The git builder class will just try to get the os version from the
remote node and then use it to build the request uri. So it won't
work for the above case.

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -472,9 +472,13 @@ class GitbuilderProject(object):
         self.os_version = self.remote.os.version
         self.codename = self.remote.os.codename
         self.pkg_type = self.remote.system_type
+        _version = self.remote.os.version
+        _major_version = self.job_config.get('install_per_major_version', False)
+        if _major_version:
+            _version = _version.split('.')[0]
         self.distro = self._get_distro(
             distro=self.remote.os.name,
-            version=self.remote.os.version,
+            version=_version,
             codename=self.remote.os.codename,
         )
         # when we're initializing with a remote we most likely have
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -431,6 +431,14 @@ def task(ctx, config):
         branch: foo
         extra_packages: ['samba']
     - install:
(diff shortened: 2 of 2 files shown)


teuthology/packaging.py
```diff
@@ -455,6 +455,7 @@ def __init__(self, project, job_config, ctx=None, remote=None):
         # of roles instead, ctx is only used in _get_config_value_for_remote.
         self.ctx = ctx
         self.remote = remote
+        self.version_major = self.job_config.get('version_major', False)
```
tchaikov, line 458: who's reading this member variable?

lxbsz: It's in `_init_from_remote()`

tchaikov, line 458: how about `install_per_major_version`? IMHO `version_major` is a little bit confusing.

lxbsz: Yeah, looks better. Will fix it.

teuthology/packaging.py
```diff
@@ -472,9 +472,13 @@ def _init_from_remote(self):
         self.os_version = self.remote.os.version
         self.codename = self.remote.os.codename
         self.pkg_type = self.remote.system_type
+        _version = self.remote.os.version
+        _major_version = self.job_config.get('install_per_major_version', False)
```
dillaman, line 476: Just planing ahead, but I don't think this will work for SUSE [1] since it stores its version as `15.2`. Perhaps this strip the minor logic just needs to be conditional on `rhel`/`centos` (similar to how ubuntu/debian switches the version to code names like `bionic`).

[1] https://shaman.ceph.com/api/search/?status=ready&project=ceph&ref=master&distros=opensuse/15.2&sha1=latest

lxbsz: Yeah, right.
I will close the PR now, because the `_get_distro()` code has already done the minor version stripping for centos/rhel.

PR #1571, opened 2020-10-23
rpm: retry installing packages if mirrors are temporarily unreachable

lxbsz:
Locally I had hit the same issue when installing packages it will fail
with repo metadata downloading and some other failures. Mostly this can
recovery later.

Fixes: https://tracker.ceph.com/issues/47852

### DIFF
--- a/teuthology/task/install/rpm.py
+++ b/teuthology/task/install/rpm.py
@@ -145,15 +145,18 @@ def _downgrade_packages(ctx, remote, pkgs, pkg_version, config):
     remote.run(args='sudo yum -y downgrade {}'.format(' '.join(pkgs_opt)))
     return [pkg for pkg in pkgs if pkg not in downgrade_pkgs]
 
-def _retry_if_503_in_output(remote, args):
+def _retry_if_failures_are_recoverable(remote, args):
     # wait at most 5 minutes
     with safe_while(sleep=10, tries=30) as proceed:
         while proceed():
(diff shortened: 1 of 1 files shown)


teuthology/task/install/rpm.py
```diff
@@ -145,16 +145,19 @@ def _downgrade_packages(ctx, remote, pkgs, pkg_version, config):
     remote.run(args='sudo yum -y downgrade {}'.format(' '.join(pkgs_opt)))
     return [pkg for pkg in pkgs if pkg not in downgrade_pkgs]
 
-def _retry_if_503_in_output(remote, args):
+def _retry_if_failures_are_restorable(remote, args):
```
batrick, line 148:
```suggestion
def _retry_if_failures_are_recoverable(remote, args):
```

batrick (changes requested):

teuthology/task/install/rpm.py
lxbsz: Done, thanks.

kshtsk:
do we still want this? if not, maybe close?

lxbsz:
@batrick ping

batrick (approved):

batrick:
> do we still want this? if not, maybe close?

Yes

batrick:
@susebot run deploy

kshtsk (approved):

PR #1572, opened 2020-10-23
task/install/deb: retry installing the package if resource temporarily unavailable

lxbsz:
Fixes: https://tracker.ceph.com/issues/46878

### DIFF
--- a/teuthology/task/install/deb.py
+++ b/teuthology/task/install/deb.py
@@ -4,12 +4,32 @@ import os
 from io import StringIO
 
 from teuthology.orchestra import run
+from teuthology.contextutil import safe_while
 
 from teuthology.task.install.util import _get_builder_project, _get_local_dir
 
 
 log = logging.getLogger(__name__)
 
+def _retry_if_eagain_in_output(remote, args):
+    # wait at most 5 minutes
+    with safe_while(sleep=10, tries=30) as proceed:
+        while proceed():
+            stderr = StringIO()
+            try:
+                return remote.run(args=args, stderr=stderr)
+            except run.CommandFailedError:
+                if "could not get lock" in stderr.getvalue().lower():
+                    stdout = StringIO()
+                    args = ['sudo', 'fuser', '-v', '/var/lib/dpkg/lock-frontend']
+                    remote.run(args=args, stdout=stdout)
+                    log.info("The processes holding 'lock-frontend':\n{}".format(stdout.getvalue()))
+                    continue
+                else:
+                    raise
+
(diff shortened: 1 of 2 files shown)


teuthology/task/install/deb.py
```diff
@@ -10,6 +10,21 @@
...
+def _retry_if_eagain_in_output(remote, args):
+    # wait at most 5 minutes
+    with safe_while(sleep=10, tries=30) as proceed:
+        while proceed():
+            stdout = StringIO()
+            try:
+                return remote.run(args=args, stdout=stdout)
+            except run.CommandFailedError:
+                if "Resource temporarily unavailable" in stdout.getvalue().lower():
+                    continue
```
tchaikov, line 22: i am not sure if this would work. as test nodes are always locked for exclusive use. would be better if we can figure out the other user holding the frontend lock of dpkg. 

i think the first step is to catch `CommandFailedError` exception and run
```sh
sudo fuser -v /var/lib/dpkg/lock-frontend
```
if "Could not get lock" is in the error message, to understand the root cause.

also in newer versions of `apt-get` the error message might vary, see https://github.com/Debian/apt/blob/289ee74dd23cba7e08b08c6c3602bcf4bf8167bc/apt-pkg/contrib/fileutl.cc#L310-L313

so i'd suggest search "Could not get lock" instead.

and you might want to capture the stderr instead of stdout for the error message.

tchaikov (changes requested):

teuthology/task/install/deb.py
lxbsz: Makes sense and fixed it.

kshtsk:
@susebot run deploy

teuthology/task/install/deb.py
tchaikov: > i am not sure if this would work. as test nodes are always locked for exclusive use. would be better if we can figure out the other user holding the frontend lock of dpkg.

i still think we need to root cause the issue before using "retry" as the cure. it's like,

> well, have you tried turning it if off and on again?

lxbsz: Checked the teuthology logs again more carefully and I couldn't find any clue about who is holding the lock. We need this patch to reproduce it to list all the users of `lock-frontend` to dig it furture.

tchaikov (approved):
could you keep the tracker ticket open before root causing it?

lxbsz:
> could you keep the tracker ticket open before root causing it?

Sure. Thanks.

PR #1573, opened 2020-10-27
orchestra/console: raise RuntimeError when fail to power on

smithfarm:
Fixes: https://tracker.ceph.com/issues/45570

### DIFF
--- a/teuthology/orchestra/console.py
+++ b/teuthology/orchestra/console.py
@@ -239,7 +239,8 @@ class PhysicalConsole(RemoteConsole):
         if self.check_power('on'):
             log.info('Power on for {s} completed'.format(s=self.shortname))
         else:
-            log.error('Failed to power on {s}'.format(s=self.shortname))
+            err_msg = 'Failed to power on {s}'.format(s=self.shortname)
+            raise RuntimeError(err_msg)
 
     def power_off(self):
         """


kshtsk:
@susebot run deploy

kshtsk:
@smithfarm you have a typo

smithfarm:
> @smithfarm you have a typo

Thanks for noticing. Should be fixed now.

teuthology/orchestra/console.py
```diff
@@ -239,7 +239,8 @@ def power_on(self):
         if self.check_power('on'):
             log.info('Power on for {s} completed'.format(s=self.shortname))
         else:
-            log.error('Failed to power on {s}'.format(s=self.shortname))
+            err_msg = 'Failed to power on {s}'.format(s=self.shortname)
+            raise RuntimeError(err_msg)
```
kshtsk, line 243: Since we all took direction to py3 I wonder why just not do this:
```
else:
    raise RuntimeError(f"Failed to power on {self.shortname}")
```
And if we add an exception throwing, then this should be reflected in the inline docs for the function, like:
```
:raises: :class:`RuntimeError`: failed to power on
```
Also, I didn't investigate the code enough, but do we need the same in case of error in pair function *power_off*.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1577, opened 2020-11-02
Check shaman not only for repo but for build complete

dmick:
The package repo might be done, but the build must be totally
complete for the container to be present.  For a specific build
(in global variables, for now), demand that the build be complete
before claiming packages exist.  This allows suite jobs to
fail or --newest to continue searching when the container isn't
done.

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -970,6 +970,38 @@ class ShamanProject(GitbuilderProject):
             'repo',
         )
 
+    @property
+    def build_complete(self):
+        # use the repo search results to get a ref and a sha1; the
+        # input to teuthology-suite doesn't contain both
+        try:
+            self.assert_result()
+        except VersionNotFoundError:
+            return False
+
+        search_result = self._result.json()[0]
+
+        # now look for the build complete status
+        path = '/'.join(
+            ('builds/ceph', search_result['ref'], search_result['sha1'])
+        )
+        build_url = urljoin(self.query_url, path)
+
+        try:
+            resp = requests.get(build_url)
+            resp.raise_for_status()
+        except requests.HttpError:
+            return False
+        for build in resp.json():
+            if (
+                build['distro'] == search_result['distro'] and
+                build['distro_version'] == search_result['distro_version'] and
+                build['flavor'] == search_result['flavor'] and
+                build['distro_arch'] in search_result['archs']
+               ):
+                return build['status'] == 'completed'
+        return False
+
     def _get_repo(self):
         resp = requests.get(self.repo_url)
         resp.raise_for_status()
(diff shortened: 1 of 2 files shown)


teuthology/packaging.py
```diff
@@ -970,6 +970,39 @@ def repo_url(self):
...
+
+        search_result = self._result.json()[0]
+
+        # now look for the build complete status
+        build_url = '/'.join((
+            self.query_url.rstrip('/'),
+            'builds/ceph',
+            search_result['ref'],
+            search_result['sha1'],
+        ))
```
zmc, line 990: Check out `urljoin` - it's already imported here!

teuthology/packaging.py
```diff
@@ -970,6 +970,39 @@ def repo_url(self):
...
+        except requests.HttpError:
+            return False
+        for build in resp.json():
+            if (
+                build['distro'] == search_result['distro'] and
+                build['distro_version'] == search_result['distro_version'] and
+                build['flavor'] == search_result['flavor']
+               ):
+                return build['status'] == 'completed'
+        return False
```
zmc, line 1004: Will there ever be `>1` builds that satisfy the above conditions? Is architecture relevant?

teuthology/suite/util.py
```diff
@@ -271,6 +274,13 @@ def package_version_for_hash(hash, kernel_flavor='basic', distro='rhel',
             sha1=hash,
         ),
     )
+
+    if bp.distro == CONTAINER_DISTRO and bp.flavor == CONTAINER_FLAVOR:
+        log.info('container build %s, checking for build_complete' % bp.distro)
+        if not bp.build_complete:
+            log.info('build not complete')
+            return None
```
zmc, line 282: Because this behavior isn't configurable, I'm a little concerned it could break others' workflow. Or can we be sure everyone will want to block/fail on container builds?

teuthology/packaging.py
dmick: Yeah...it only seems to do two at a time, which seemed annoying.  If it's still better I can use it.

teuthology/packaging.py
dmick: I don't *think* so if I consider arch as well.  Can't hurt (I don't think; I'm pretty sure arch is available where it needs to be)

teuthology/suite/util.py
dmick: I debated it.  The worst it should do, in the "successful build" case, is fail and delay the testing for another 10 minutes (or choose the second-newest success with --newest).  I don't have strong opinions though, and it could be Yet Another Commandline Switch I suppose.

teuthology/packaging.py
dmick: Compromise to hopefully get maximum benefit with least repetition.

zmc (dismissed):

zmc (approved):

PR #1578, opened 2020-11-04
packaging: try noarch repo if arch repo doesn't exist

lxbsz:
There has some cases we need to specify the particalur option to
override the one parsed from the remote os.

### DIFF
--- a/docs/detailed_test_config.rst
+++ b/docs/detailed_test_config.rst
@@ -270,3 +270,15 @@ the location of the sandbox directory, the following option can be
 specified in ``$HOME/.teuthology.yaml``::
 
     test_path: <directory>
+
+Shaman options
+==============
+
+Shaman is a helper class which could be used to build the uri for specified
+packages based the 'shaman_host': 'shaman.ceph.com'.
+
+Options::
+
+    use_shaman: True # Enable to use Shaman, False as default
+    shaman:
+      force_noarch: True # Force to use "noarch" to build the uri
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -853,6 +853,9 @@ class ShamanProject(GitbuilderProject):
         super(ShamanProject, self).__init__(project, job_config, ctx, remote)
         self.query_url = 'https://%s/api/' % config.shaman_host
 
+        # Force to use the "noarch" instead to build the uri.
+        self.force_noarch = self.job_config.get("shaman", {}).get("force_noarch", False)
+
     def _get_base_url(self):
         self.assert_result()
         return self._result.json()[0]['url']
@@ -882,7 +885,8 @@ class ShamanProject(GitbuilderProject):
         req_obj['status'] = 'ready'
         req_obj['project'] = self.project
         req_obj['flavor'] = flavor
-        req_obj['distros'] = '%s/%s' % (self.distro, self.arch)
+        arch = "noarch" if self.force_noarch else self.arch
+        req_obj['distros'] = '%s/%s' % (self.distro, arch)
         ref_name, ref_val = list(self._choose_reference().items())[0]
         if ref_name == 'tag':
             req_obj['sha1'] = self._sha1 = self._tag_to_sha1()


lxbsz:
https://github.com/ceph/ceph/pull/37138, we need to override the remote's `arch` with `noarch` for the ceph-iscsi package.

kshtsk:
ok to test

kshtsk:
I am not sure I understand which cases this change is needed for.
This is kinda new functionality and it requires to be documented.

lxbsz:
> I am not sure I understand which cases this change is needed for.
> This is kinda new functionality and it requires to be documented.

For ceph-iscsi, the package was in the noarch/ directory, likes:

`https://shaman.ceph.com/api/search/?status=ready&project=ceph-iscsi&flavor=default&distros=centos%2F8%2Fnoarch&sha1=latest`

While the shaman will parse the `self.arch` from the remote os, which will be `x86_64` instead, likes::

`https://pulpito.ceph.com/xiubli-2020-10-28_05:55:13-rbd:iscsi:-wip-ceph-iscsi-qa-20201026-1259-distro-basic-smithi/5566982/`

So we need one we to tell the Shaman class to use the `noarch` instead to search the repo uri.

kshtsk:
@lxbsz still it is not clear why this should replace remote architecture values, maybe we should look into changing shaman module instead of workaround the core features?

Alternatively, instead of overriding the remote arch the could check existence of both remote arch and noarch distro, and use one of them or both.

lxbsz:
> @lxbsz still it is not clear why this should replace remote architecture values, maybe we should look into changing shaman module instead of workaround the core features?
> 
> Alternatively, instead of overriding the remote arch the could check existence of both remote arch and noarch distro, and use one of them or both.

Yeah, this alternative approach also sounds good.

> @lxbsz still it is not clear why this should replace remote architecture values, maybe we should look into changing shaman module instead of workaround the core features?
> 
> Alternatively, instead of overriding the remote arch the could check existence of both remote arch and noarch distro, and use one of them or both.

Yeah, this sounds reasonable and better. Let me try this. Thanks.

teuthology/packaging.py
```diff
@@ -936,7 +938,12 @@ def get_sha1(url):
 
     def assert_result(self):
         if len(self._result.json()) == 0:
-            raise VersionNotFoundError(self._result.url)
+            if not self.try_noarch and self.arch != "noarch":
+                self.try_noarch = True
+                self._result_obj = None
```
dillaman, line 943: Nit: it's not immediately clear to me how the query will restart. Will it just automatically retry if `_result_obj` is `None` and it receives a `VersionNotFoundError`?

lxbsz: The Line#940 will try to expand and assign the `self._result_obj` based the `self.arch`, which is from the remote OS. For our issue it will be the `x86_64` and the url will be:

`https://shaman.ceph.com/api/search/?status=ready&project=ceph-iscsi&flavor=default&distros=centos%2F8%2Fx86_64&sha1=latest`

But the `self._result.json()` length will be 0, before raising the Error it will try to read the `self._result` again in Line#945 bellow which will try to expand and assign it with new url based the `noarch` again when it see the `self._result_obj` is `None`. 

Only when both the `x86_64` and `noarch` fail will it raise the Error.

Because the `assert_result()` checking is in everywhere, so I put the above logic in it. Maybe we should move it to:

```
 861     @property
 862     def _result(self):
 863         if getattr(self, '_result_obj', None) is None:                                                                               
 864             self._result_obj = self._search()   
 865         return self._result_obj
```

???

lxbsz: @kshtsk @dillaman 
BTW, will it be better to add an opetion likes `try_noarch = True` to enable trying the alternate noarch uri, if the uri based something like the `x86_64` does not exist ? And as default the `try_noarch = False` or absent.

dillaman: Personally, I think it would be cleaner if we just added a `force_noarch: True` or just `noarch: True` to the config where needed instead of this hidden logic to always implicitly attempt to fallback to `noarch`.

lxbsz: Okay, sounds reasonable. Will add it. Thanks.

lxbsz: Updated it. Thanks.

kshtsk: Sorry for being late. But actually I do not understand why do we need an option for this, what is the bad to always check both repos?

dillaman: It wasn't checking both -- it was checking for x86_64 and failing that, then it checked for noarch. If it adds both repos if both are valid, I'm fine with that -- but this is also a use-case where we know specifically that we want to pull from noarch so why not just offer the override?

kshtsk: @dillaman thanks for the explanation, so if we want packaging module should override arch, then we need to provide this withing packaging namespace.
I would like to see also the docs updated to explain how this is supposed to be used from the user stand point since it is kinda new feature.

lxbsz: @kshtsk BTW, which doc file should we document it ? I didn't find an appropriate one.

kshtsk: I would say `docs/detailed_test_config.rst`.

kshtsk: Maybe add somewhere else afterwards.

teuthology/packaging.py
```diff
@@ -853,6 +853,9 @@ def __init__(self, project, job_config, ctx=None, remote=None):
         super(ShamanProject, self).__init__(project, job_config, ctx, remote)
         self.query_url = 'https://%s/api/' % config.shaman_host
 
+        # Force to use the "noarch" instead to build the uri.
+        self.force_noarch = self.job_config.get("force_noarch", False)
```
kshtsk, line 857: Since this is related to _Shaman_ maybe use _shaman_ scope.

teuthology/packaging.py
lxbsz: Okay, will document it. Thanks.

teuthology/packaging.py
lxbsz: Sounds good.

dillaman (dismissed):
lgtm

kshtsk:
@susebot run deploy

docs/detailed_test_config.rst
```diff
@@ -270,3 +270,15 @@ the location of the sandbox directory, the following option can be
 specified in ``$HOME/.teuthology.yaml``::
 
     test_path: <directory>
+
+Shaman options
+==============
+
+Shaman represents a project that is built by gitbuilder, which will
```
tchaikov, line 277: we don't use gitbuilder anymore. jenkins.ceph.com is used for building packages.

lxbsz: Fixed the doc, please take a look. Thanks.

lxbsz:
@tchaikov How about this change ? Thanks.

docs/detailed_test_config.rst
```diff
@@ -270,3 +270,15 @@ the location of the sandbox directory, the following option can be
...
+==============
+
+Shaman is a helper class which could be used to build the uri for specified
+packages based the 'shaman_host': 'shaman.ceph.com'.
+
+Options::
+
+    use_shaman: True # Enable to use Shaman, False as default
+    shaman:
+    - force_noarch: True # Force to use "noarch" to build the uri
```
kshtsk, line 284: According this line, the `shaman` includes a list of hashes, but we want `shaman` be a hash.

lxbsz: > According to this line, the `shaman` includes a list of hashes, but we want `shaman` be a hash.

@kshtsk 

Sorry, I didn't get it. Could you kindly give an example of it ?

Thanks

kshtsk: Okay, just illustrating:

```>>> print(yaml.safe_dump(dict(shaman=dict(force_noarch=True))))
shaman:
  force_noarch: true

>>> print(yaml.safe_dump(dict(shaman=list(dict(force_noarch=True)))))
shaman:
- force_noarch
```

kshtsk: But use the code:
```
        self.force_noarch = self.job_config.get("shaman", {}).get("force_noarch", False)
```
Which is supposed to work with first case.

lxbsz: @kshtsk 
Okay, good catch. Fixed it, thanks.

kshtsk:
@susebot run deploy

dillaman:
Any last minute objections to this PR or can I merge it?

kshtsk:
> Any last minute objections to this PR or can I merge it?

weld it

kshtsk (approved):

PR #1583, opened 2020-11-27
task.internal.archive: gzip archived file size > 128MB

tchaikov:
* misc: add an optional write_to argument to misc.pull_directory()
        so the caller can optionally specify the function to write
        to local file.
* task/internal: add an global option "log-compress-min-size" which
        defaults to "128MB". if the size of a file pulled from remote
        host is greater or equal to the specified size, it will be
        compressed with gzip with the extension of ".gz" before
        stored in the archive directory.

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -4,18 +4,23 @@ Note that there is no corresponding task defined for this module.  All of
 the calls are made from other modules, most notably teuthology/run.py
 """
 import contextlib
+import functools
+import gzip
 import logging
 import os
+import shutil
 import time
 import yaml
 import subprocess
+
+import humanfriendly
 
 import teuthology.lock.ops
 from teuthology import misc
 from teuthology.packaging import get_builder_project
 from teuthology import report
 from teuthology.config import config as teuth_config
-from teuthology.exceptions import VersionNotFoundError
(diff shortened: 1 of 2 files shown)


badone (approved):
Looks good, thanks Kefu.

One nit is the commit comment has "add an global option" which should be "add a global option"

tchaikov:
@badone  fixed and repushed.

badone (approved):

teuthology/task/internal/__init__.py
```diff
@@ -334,6 +339,20 @@ def fetch_binaries_for_coredumps(path, remote):
             remote.get_file(debug_path, coredump_path)
 
 
+def gzip_if_too_large(min_size_config, src, tarinfo, local_path):
+    try:
+        min_size = humanfriendly.parse_size(min_size_config)
+    except humanfriendly.InvalidSize:
+        msg = 'invalid "log_compress_min_size": {}'.format(min_size_config)
+        log.error(msg)
+        raise ConfigError(msg)
```
kshtsk, line 348: 1) The code of parsing the size is better to place close to reading the value from the config.
2) Also we need to update the doc describing the `log_compress_min_size` parameter usage and value format.

tchaikov:
@badone thanks for approving my PR. i don't think i have enough bandwidth continue working on it anymore. so closing it.

PR #1585, opened 2020-11-30
teuthology-suite: add Rocket.Chat notification

kshtsk:
Add Rocket.Chat notification for sleep before teardown.

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -109,6 +109,9 @@ Scheduler arguments:
                               When tests finish or time out, send an email
                               here. May also be specified in ~/.teuthology.yaml
                               as 'results_email'
+  --rocketchat <rocketchat>   Comma separated list of Rocket.Chat channels where
+                              to send a message when tests finished or time out.
+                              To be used with --sleep-before-teardown option.
   -N <num>, --num <num>       Number of times to run/queue the job
                               [default: 1]
   -l <jobs>, --limit <jobs>   Queue at most this many jobs
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -185,6 +185,7 @@ class TeuthologyConfig(YamlConfig):
                 'size': 1,
             },
         },
+        'rocketchat': None,
         'sleep_before_teardown': 0,
     }
 
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -4,6 +4,7 @@ import os
 import sys
 import time
 import types
+import yaml
 
 from copy import deepcopy
 from humanfriendly import format_timespan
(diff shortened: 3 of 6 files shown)


kshtsk:
@susebot run deploy

kshtsk:
@smithfarm could you please review?

scripts/suite.py
```diff
@@ -109,6 +109,9 @@
                               When tests finish or time out, send an email
                               here. May also be specified in ~/.teuthology.yaml
                               as 'results_email'
+  --rocket <rocket>           Comma separated list of rocketchat channels where
+                              to send a message when tests finished or time out.
+                              To be used with --sleep-before-teardown option.
```
smithfarm, line 114: Hm, I don't like `--rocket`. I would prefer `--rocketchat` or `--rocket-chat` because the software is called "Rocket.Chat" (not "Rocket"). The word "rocket" can mean several things (including a vegetable), but people not familiar with "Rocket.Chat" will simply be confused by an option called `--rocket`.

Therefore, please consider renaming the option to `--rocketchat` or `--rocket-chat`?

teuthology/config.py
```diff
@@ -185,6 +185,7 @@ class TeuthologyConfig(YamlConfig):
                 'size': 1,
             },
         },
+        'rocket': None,
```
smithfarm, line 188: Here again, the name "rocket" is too cryptic in my option. Is it a Soyuz rocket, or a Saturn V? Or a vegetable? To remove this ambiguity, I would prefer something like "rocket_chat" or "rocket_chat_notification_channels" here.

teuthology/run_tasks.py
```diff
@@ -203,6 +204,41 @@ def run_tasks(tasks, ctx):
             del exc_info
         timer.mark("tasks complete")
 
+
+def build_rocket_message(ctx, stack, sleep_time_sec, template_path=None):
```
smithfarm, line 208: I suggest as a less ambiguous name for the function: `build_rocket_chat_message`

teuthology/run_tasks.py
```diff
@@ -239,7 +275,57 @@ def build_email_body(ctx, stack, sleep_time_sec):
     )
     return (subject.strip(), body.strip())
 
+
+def rocket_message(ctx, message, channels):
```
smithfarm, line 279: I suggest as a less ambiguous name for this function: `rocket_chat_message`

teuthology/templates/rocket-sleep-before-teardown.jinja2
```diff
@@ -0,0 +1,6 @@
+The teuthology job [{{ job_id }}]({{ job_info }}) for suite *{{ suite_name }}* owned by '{{ owner }}' has fallen asleep with status '{{ status }}' at {{ sleep_date }} for __{{ sleep_time }}__ ({{ sleep_time_sec }} seconds).
+Open [teuthology.log]({{ job_logs }}teuthology.log) for details, or go to [all logs]({{ job_logs}}).
+
+Job Description: {{ job_desc }}
+Run Name: {{ run_name }}
+Task Stack: {{ task_stack }}
```
smithfarm, line 6: I suggest as a less ambiguous name for this file: `rocket-chat-sleep-before-teardown.jinja2` or simply `rocket-chat-notification.jinja2` (since this is the only type of Rocket.Chat notification in teuthology, there is no need to distinguish it from other types)

kshtsk: Thanks for the suggestion, I would prefer leave `sleep-before-teardown` suffix because of possible adding another template for run completion rocket chat notification. Didn't add this in this PR because I have no time for it yet.

scripts/suite.py
kshtsk: I go with `--rocketchat`, thanks for the suggestion.

teuthology/config.py
kshtsk: I would prefer Союз rocket in this case... Here and in other cases I have renamed it to rocketchat.

teuthology/run_tasks.py
kshtsk: Thanks. Renamed to `rocketchat_send_message`.

teuthology/templates/rocket-sleep-before-teardown.jinja2
kshtsk: Changed the `rocket-` to `rocketchat-`, Danke sehr.

smithfarm (approved):

kshtsk:
@susebot run deploy

PR #1586, opened 2020-12-02
task.internal.archive: gzip archived file size > 128MB

tchaikov:
* misc: add an optional write_to argument to misc.pull_directory()
        so the caller can optionally specify the function to write
        to local file.
* task/internal: add a global option "log-compress-min-size" which
        defaults to "128MB". if the size of a file pulled from remote
        host is greater or equal to the specified size, it will be
        compressed with gzip with the extension of ".gz" before
        stored in the archive directory.

### DIFF
--- a/docs/detailed_test_config.rst
+++ b/docs/detailed_test_config.rst
@@ -227,6 +227,21 @@ global option can be set to true to make the unlocking happen unconditionally.
 Troubleshooting
 ===============
 
+Postmortem Debugging
+--------------------
+
+After completion of a test, the ``archive`` subdirectory is archived under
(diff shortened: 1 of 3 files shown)


tchaikov:
tested at https://pulpito.ceph.com/kchai-2020-11-30_17:04:28-rados-wip-jan-testing-2020-11-30-1049-distro-basic-smithi/

badone (approved):
LGTM

tchaikov:
https://pulpito.ceph.com/kchai-2020-12-02_05:39:45-rados-wip-jan-testing-2020-11-30-1049-distro-basic-smithi/

PR #1587, opened 2020-12-03
install: install extra packages from overrides

alimaredia:
At the moment the only way to install extra
packages is through the config in the install
task. Installing extra packages could not be done
through the install tasks overrides.

This commit changes adds 'extra_pacakages'
from overrides to the rpm/deb package list to
install.

Signed-off-by: Ali Maredia <amaredia@redhat.com>

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -203,6 +203,13 @@ def install(ctx, config):
         debs += extra_pkgs
         rpms += extra_pkgs
 
+    overrides_extra_packages = ctx.config.get(
+        'overrides', {}).get('install', {}).get('extra_packages', [])
+    log.info('extra packages from overrides: {packages}'.format(packages=overrides_extra_packages))
+    if isinstance(overrides_extra_packages, dict):
+        debs += overrides_extra_packages.get('deb', [])
+        rpms += overrides_extra_packages.get('rpm', [])
+
     # When extras is in the config we want to purposely not install ceph.
     # This is typically used on jobs that use ceph-deploy to install ceph
     # or when we are testing ceph-deploy directly.  The packages being


jdurgin (approved):

teuthology/task/install/__init__.py
```diff
@@ -203,6 +203,13 @@ def install(ctx, config):
         debs += extra_pkgs
         rpms += extra_pkgs
 
+    overrides_extra_packages = ctx.config.get(
+        'overrides', {}).get('install', {}).get('extra_packages', [])
+    log.info('extra packages from overrides: {packages}'.format(packages=overrides_extra_packages))
+    if isinstance(overrides_extra_packages, dict):
+        debs += overrides_extra_packages.get('deb', [])
+        rpms += overrides_extra_packages.get('rpm', [])
+
```
kshtsk, line 212: This looks good, but that does not look like override algorithm, but extension of present extra_packages.

kshtsk: The `overrides` description in the docs:
```
* ``overrides``: override behavior. Typically, this includes sub-tasks being
  overridden. Overrides technically is not a task (there is no 'def task' in
  an overrides.py file), but from a user's standpoint can be described as
  behaving like one.
  Sub-tasks can nest further information.  For example, overrides
  of install tasks are project specific, so the following section of a yaml
  file would cause all ceph installations to default to using the jewel
  branch::

    overrides:
      install:
        ceph:
          branch: jewel
```

alimaredia: You are correct, an extension is what I'm going for. I'm not trying to override the current packages list, just trying to be able to add to it easier.

The need for this occurred when I was writing a new task (rgw-datacache) in this sub suite: https://github.com/ceph/ceph/tree/master/qa/suites/rgw/verify/tasks.

The .yaml I'm adding like this:
https://github.com/alimaredia/ceph/blob/wip-rgw-d3n-datacache-upstreaming/qa/suites/rgw/verify/tasks/rgw-datacache.yaml

I wanted an easy way to install the s3cmd package for whenever that task is run and not for every task run in verify which is what is happening in my other change here: https://github.com/alimaredia/ceph/blob/wip-rgw-d3n-datacache-upstreaming/qa/suites/rgw/verify/tasks/0-install.yaml#L6.

I considered installing the package by hand in `rgw_datacache.py` itself but I figured using the `install` task was the best approach.

alimaredia:
@tchaikov @kshtsk any update on my comments or this PR? As of now my new datacache task (https://github.com/alimaredia/ceph/blob/wip-rgw-d3n-datacache-upstreaming/qa/suites/rgw/verify/tasks/rgw-datacache.yaml) uses the overrides in the install task in the way the PR implements it and I'd like to know whether or not to keep it that way.

teuthology/task/install/__init__.py
kshtsk: @alimaredia I do not get why do you want install `s3cmd` via overrides block if it is already present in extra_pkgs?

When you are adding a new task and want to install extra dependencies why just not update the [0-install.yaml](https://github.com/alimaredia/ceph/blob/wip-rgw-d3n-datacache-upstreaming/qa/suites/rgw/verify/tasks/0-install.yaml)?

If you need some details, you are free to use comment which task needs the package:
```
tasks:
- install:
      extra_packages:
        deb:
             - s3cmd                  # s3
             - some-datacache-pkg     # rgw-datacache
        rpm:
             - s3cmd                  # s3
             - some-datacache-pkg     # rgw-datacache

```

kshtsk:
> @tchaikov @kshtsk any update on my comments or this PR? As of now my new datacache task (https://github.com/alimaredia/ceph/blob/wip-rgw-d3n-datacache-upstreaming/qa/suites/rgw/verify/tasks/rgw-datacache.yaml) uses the overrides in the install task in the way the PR implements it and I'd like to know whether or not to keep it that way.

As I said, this is incorrect to extend when we declare override.

PR #1588, opened 2020-12-09
install: pass the 'shaman' to Shaman class

lxbsz:

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -610,6 +610,8 @@ def task(ctx, config):
         )
         if repos:
             nested_config['repos'] = repos
+        if 'shaman' in config:
+            nested_config['shaman'] = config['shaman']
         with contextutil.nested(
             lambda: install(ctx=ctx, config=nested_config),
             lambda: ship_utilities(ctx=ctx, config=None),


lxbsz:
Fixing the error in https://github.com/ceph/ceph/pull/37138#issuecomment-741827256.

It didn't pass the
```
shaman:
   force_shaman: True
```
down to the Shanman Class.

teuthology/task/install/__init__.py
```diff
@@ -607,6 +607,7 @@ def task(ctx, config):
                 sha1=config.get('sha1'),
                 tag=config.get('tag'),
                 wait_for_package=config.get('wait_for_package', False),
+                shaman=config.get('shaman', {})
```
dillaman, line 610: Nit: perhaps treat this more like the `repos` key below:

```
if 'shaman' in config:
    nested_config['shaman'] = config['shaman']
```

lxbsz: Done.

dillaman (approved):
lgtm

kshtsk:
interesting....

kshtsk:
@susebot run deploy

PR #1590, opened 2020-12-13
teuthology: add a seprate task to ship the utilities

lxbsz:
The 'ship_utilities_early: True' will make the 'install' tasks skip
shipping the utilities and will add a separate task to ship them
earlier than all the other tasks. This could make sure that the
utilities will be unwinded at the last time in case that some other
tasks will use the utilitied when unwinding.

### DIFF
--- a/teuthology/task/install/util.py
+++ b/teuthology/task/install/util.py
@@ -51,18 +51,14 @@ def get_flavor(config):
                 flavor = 'gcov'
     return flavor
 
-
-@contextlib.contextmanager
-def ship_utilities(ctx, config):
+def _ship_utilities(ctx):
     """
     Write a copy of valgrind.supp to each of the remote sites.  Set executables
     used by Ceph in /usr/local/bin.  When finished (upon exit of the teuthology
     run), remove these files.
 
     :param ctx: Context
-    :param config: Configuration
     """
-    assert config is None
     testdir = teuthology.get_testdir(ctx)
     filenames = []
 
@@ -109,19 +105,49 @@ def ship_utilities(ctx, config):
                         dst,
                     ],
                 )
+    return filenames
 
-    try:
+def _remove_utilities(ctx, filenames):
+    """
+    Remove the shipped utilities.
+
+    :param ctx: Context
+    :param filenames: The utilities install paths
+    """
+    log.info('Removing shipped files: %s...', ' '.join(filenames))
+    if filenames == []:
+        return
+    run.wait(
+        ctx.cluster.run(
+            args=[
+                'sudo',
+                'rm',
+                '-f',
+                '--',
+            ] + list(filenames),
+            wait=False,
+        ),
+    )
+
+@contextlib.contextmanager
+def ship_utilities(ctx, config):
+    """
+    Ship utilities during the first call, and skip it in the following ones.
+    See also `_ship_utilities`.
+
+    :param ctx: Context
+    :param config: Configuration
+    """
+    assert config is None
+
+    do_ship_utilities = ctx.get('do_ship_utilities', True)
+    if do_ship_utilities:
+        ctx['do_ship_utilities'] = False
+        filenames = _ship_utilities(ctx)
+        try:
+            yield
+        finally:
+            _remove_utilities(ctx, filenames)
+    else:
+        log.info('Utilities already shipped, skip it...')
         yield
-    finally:
-        log.info('Removing shipped files: %s...', ' '.join(filenames))
-        run.wait(
-            ctx.cluster.run(
-                args=[
-                    'sudo',
-                    'rm',
-                    '-f',
-                    '--',
-                ] + list(filenames),
-                wait=False,
-            ),
-        )


lxbsz:
More detail about this PR please see: https://github.com/ceph/ceph/pull/37138#issuecomment-743994999

lxbsz:
Currently all the `install:` tasks must be placed before `ceph:` task in the `.yaml` file. This will make sure the `ceph:` task will be unwinded before the `install:` tasks. Because the `ceph:` task will use the `adjust-admin` tool, but the `install:` will remove it.

Maybe we should make the utilities shipping always before all the other tasks as default ? And then the utilities will be unwinded at the last time.

teuthology/run_tasks.py
```diff
@@ -81,13 +81,25 @@ def run_tasks(tasks, ctx):
...
+        if ctx.config.get('ship_utilities_early', False):
+            taskname = "ship_utilities"
+            log.info('Running task %s...', taskname)
+            timer.mark('%s enter' % taskname)
+            manager = run_one_task(taskname, ctx=ctx, config=None)
+            if hasattr(manager, '__enter__'):
+                stack.append((taskname, manager))
+                manager.__enter__()
+            skip_shipping_utilities = True
+
```
kshtsk, line 94: That's dirty hack... because this is common algorithm for running tasks. If you want to inject some job into the stack please go to 'teuthology/run.py get_initial_tasks'.

kshtsk: For sure the run_tasks.py should not be touched for this purpose.

kshtsk: Another question what was wrong when it was added in the install tasks. Since this tasks is pretty specific to the install task.

teuthology/task/ship_utilities.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Ship the utilities
+
+    ship_utilities_early: True
+    tasks:
+    - install:
+        project: ceph
+        branch: bar
+
+    The 'ship_utilities_early: True' will make the 'install' tasks skip
```
kshtsk, line 23: Instead if you want a separate tasks for ship_utilities, I suppose you to use `ship_utilities` instead of `ship_utilities_early` and add to the tasks list along with other internal tasks. Taking into account that we do not do shipping by default, the default value for ship_utilities should be False. 

But first of all what is the goal of this change, what do we want to have eventually? Like install utils before the `install` task or make it possible to ship_utils without running `install` task?

dillaman: @lxbsz If I am understanding the issue correctly, each `install:` task will run the `ship_utilities` task and therefore uninstall the utilities when each individual `install` task is unwound (even if the utilities are still needed by the next `install` task's unwind handler). In that case, could we not somehow just only apply `ship_utilities` to the first `install` task by storing some state in the shared context programatically?

lxbsz: > @lxbsz If I am understanding the issue correctly, each `install:` task will run the `ship_utilities` task and therefore uninstall the utilities when each individual `install` task is unwound (even if the utilities are still needed by the next `install` task's unwind handler). 

Yeah, it is. The `install:` task will ship the utilities implicitly as default. 

While the `ceph:` task is unwound, the ceph will use the `adjust-limit` utility, so this is why the `ceph:` task must be after all the `install:` tasks in the qa suite's `.yaml`, then `ceph:` task will be unwound in frist place in reverse order:
```
tasks:
- install:
- ceph:
```

Or we will hit:

```
2020-12-13T04:11:21.429 DEBUG:teuthology.run_tasks:Unwinding manager ceph
2020-12-13T04:11:21.441 INFO:tasks.ceph.ceph_manager.ceph:Canceling any pending splits or merges...
2020-12-13T04:11:21.449 DEBUG:teuthology.orchestra.run.smithi066:> sudo adjust-ulimits ceph-coverage /home/ubuntu/cephtest/archive/coverage timeout 120 ceph --cluster ceph osd dump --format=json
2020-12-13T04:11:21.493 INFO:teuthology.orchestra.run.smithi066.stderr:sudo: adjust-ulimits: command not found
2020-12-13T04:11:21.494 DEBUG:teuthology.orchestra.run:got remote process result: 1
2020-12-13T04:11:21.495 ERROR:teuthology.contextutil:Saw exception from nested tasks
```

But when installing the ceph-iscsi rpm package, it will depend on the ceph packages, that means in this case the tasks order must be:

```
tasks:
- ceph:
- install:
    packages:
      rpm:
      - ceph-iscsi
    shaman:
      force_noarch: True
    project: ceph-iscsi
    sha1: latest
```
Or we will hit:

```
2020-12-13T10:03:07.260 INFO:teuthology.orchestra.run.smithi189.stderr:Error:
2020-12-13T10:03:07.261 INFO:teuthology.orchestra.run.smithi189.stderr: Problem: conflicting requests
2020-12-13T10:03:07.261 INFO:teuthology.orchestra.run.smithi189.stderr:  - nothing provides python3-rbd >= 10.2.2 needed by ceph-iscsi-3.4-47.gf79bf8c.el8.noarch
2020-12-13T10:03:07.261 INFO:teuthology.orchestra.run.smithi189.stderr:  - nothing provides python3-rados >= 10.2.2 needed by ceph-iscsi-3.4-47.gf79bf8c.el8.noarch
2020-12-13T10:03:07.261 INFO:teuthology.orchestra.run.smithi189.stderr:  - nothing provides ceph-common >= 10.2.2 needed by ceph-iscsi-3.4-47.gf79bf8c.el8.noarch
2020-12-13T10:03:07.262 INFO:teuthology.orchestra.run.smithi189.stdout:(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
2020-12-13T10:03:07.304 DEBUG:teuthology.orchestra.run:got remote process result: 1
2020-12-13T10:03:07.305 ERROR:teuthology.contextutil:Saw exception from nested tasks
```

This two cases above are confilicting. And this confilicting is a potential bug for the current code, to fix it we must keep the utilities shipped until the last user is unwound.

> In that case, could we not somehow just only apply `ship_utilities` to the first `install` task by storing some state in the shared context programatically?

I am afraid this won't work, because in some tasks will make the `ship_utilities` as a nested task.

> Instead if you want a separate tasks for ship_utilities, I suppose you to use `ship_utilities` instead of `ship_utilities_early` and add to the tasks list along with other internal tasks. Taking into account that we do not do shipping by default, the default value for ship_utilities should be False.
> 

Yeah, I think this sounds reasonable, but we need a small improvement, that's when iterating and running the task list one by one, we will only ship the utilities for the first task whose `ship_utilities` is `True`, for all the others just ignore the option.

> But first of all what is the goal of this change, what do we want to have eventually? Like install utils before the `install` task or make it possible to ship_utils without running `install` task?

As described above, just want to make sure that when the tasks are unwound the utilites will be kept shipped until the last user is unwounded.

dillaman: > > In that case, could we not somehow just only apply `ship_utilities` to the first `install` task by storing some state in the shared context programatically?
> 
> I am afraid this won't work, because in some tasks will make the `ship_utilities` as a nested task.

Ok, can you help me understand this better? If you have a helper that adds `ship_utilities` as a nested task only if it hasn't already been added (as denoted by a state variable in the context), the first `install` task to be executed should therefore install it and also uninstall it on unwind, right?

teuthology/run_tasks.py
lxbsz: As described below, the current code has potential bug that if there have two `install:` tasks, they will ship the utilites but that not means they really need them. If both the two `install:` tasks really need the utilities, so when the second task is unwound in first place in reverse order it will remove the utilities, while the first task may still need it when being unwound.

teuthology/task/ship_utilities.py
lxbsz: In ceph code, such as:

```
qa/tasks/ceph_deploy.py:728:            lambda: install_fn.ship_utilities(ctx=ctx, config=None),
qa/tasks/ceph_deploy.py:912:        lambda: install_fn.ship_utilities(ctx=ctx, config=None),
```
Yeah, anyway we need to change this code too.

dillaman: Yup -- or make it so `ship_utilities` is the wrapper and becomes a no-op for setup and unwind if another task content was already run.

lxbsz: > Yeah, I think this sounds reasonable, but we need a small improvement, that's when iterating and running the task list one by one, we will only ship the utilities for the first task whose `ship_utilities` is `True`, for all the others just ignore the option.

@dillaman Does the approach mentioned above work for you ?

dillaman: If you can do it programmatically, why would we want a new variable?

lxbsz: Not only the `install:` tasks will ship the utilities, but also for some other non `install:` tasks, like the `ceph:`, etc. Shouldn't we need one way to know who need them ?

lxbsz: We need to make the imlicit nested `install.ship_utilities` to be explicit in the `yaml`.

lxbsz: Okay, maybe I misunderstand it, you mean do it in the `ship_utilies()` instead, right ?

dillaman: Yes. `ship_utilities` is already decorated as a context manger, so the first instance to run should test to see if it's the first. If yes, proceed as normal. Otherwise, no-op itself with just a `yield` that doesn't install nor cleanup anything.

lxbsz: Sounds cool, will try it. Thanks.

teuthology/task/install/util.py
```diff
@@ -125,3 +121,25 @@ def ship_utilities(ctx, config):
...
         )
+
+utilities_shipped = False
+
+@contextlib.contextmanager
+def ship_utilities(ctx, config):
+    """
+    Write a copy of valgrind.supp to each of the remote sites.  Set executables
+    used by Ceph in /usr/local/bin.  When finished (upon exit of the teuthology
+    run), remove these files.
```
kshtsk, line 132: I think no need to duplicate docs from _ship_utilitities, it is enough to refer this function and do some more explanation regarding idea of doing it once, for example:
```
Ship utilities during the first call, and skip it in the following ones.
See also `_ship_utilities`.
```

teuthology/task/install/util.py
```diff
@@ -125,3 +121,25 @@ def ship_utilities(ctx, config):
...
+
+    :param ctx: Context
+    :param config: Configuration
+    """
+    assert config is None
+
+    global utilities_shipped
+    if utilities_shipped:
```
kshtsk, line 140: minor, rename `utilities_shipped` to `do_ship_utilities` with defaults to True.
So code look like:
```
if do_ship_utilities:
   do_ship_utilities = False
   _ship_utilities(ctx, config)
else:
   log.info('Utilities already shipped, skip it...')
   yield
```

kshtsk:
The new changes looks good, see the comments.

teuthology/task/install/util.py
lxbsz: Will fix it.

teuthology/task/install/util.py
lxbsz: Looks good will modify it. Thanks.

teuthology/task/install/util.py
```diff
@@ -125,3 +121,24 @@ def ship_utilities(ctx, config):
...
+    """
+    Ship utilities during the first call, and skip it in the following ones.
+    See also `_ship_utilities`.
+
+    :param ctx: Context
+    :param config: Configuration
+    """
+    assert config is None
+
+    global do_ship_utilities
```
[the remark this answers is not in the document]
lxbsz: Okay, done.

kshtsk:
@susebot run deploy

kshtsk:
@lxbsz integration test is failed, see logs from susebot https://ceph-ci.suse.de/job/teuthology-deploy-ovh/348/artifact/logs/run/teuthology-job-2.html

kshtsk:
Let me guess, function subcall does not work as as expected.

dillaman:
> Let me guess, function subcall does not work as as expected.

Right, I think you would need to use `yield from _ship_utilities(ctx)` (which is Python 3) or just not attempt to break the function up since it was originally really a generator, not a function.

lxbsz:
How ahout:

```
+    if do_ship_utilities:
+        ctx['do_ship_utilities'] = False
+        filenames = _ship_utilities(ctx)
+        try:
+            yield
+        finally:
+            _remove_utilities(ctx, filenames)
+    else:
+        log.info('Utilities already shipped, skip it...')
         yield
```

teuthology/task/install/util.py
```diff
@@ -109,19 +105,43 @@ def ship_utilities(ctx, config):
                         dst,
                     ],
                 )
+    return filenames
 
-    try:
+def _remove_utilities(ctx, filenames):
```
dillaman, line 110: Nit: just to remain consistent, perhaps add docstring

lxbsz: Done, thanks.

kshtsk:
@susebot run deploy

dillaman (approved):

kshtsk (approved):

PR #1592, opened 2020-12-23
add docker-compose scripts for development setups

kshtsk:


kshtsk:
Hey @jdurgin, though this is follow up PR for the podman I've only added support for docker compose which can be used on MacOS as well, I suggest this ticket for the discussion of how this can be implemented for podman too.

dmick:
Lost...what's this got to do with podman?

jdurgin:
https://github.com/containers/podman-compose looks promising, I haven't tried it though.

Another approach is using a k8s-style yaml to define a pod and the containers in it: https://www.redhat.com/sysadmin/compose-podman-pods

These containers are simple enough that you could use a script containing the docker/podman commands directly as well, using the same arguments for both.

kshtsk:
> https://github.com/containers/podman-compose looks promising, I haven't tried it though.
> 
> Another approach is using a k8s-style yaml to define a pod and the containers in it: https://www.redhat.com/sysadmin/compose-podman-pods
> 
> These containers are simple enough that you could use a script containing the docker/podman commands directly as well, using the same arguments for both.

This is interesting but maybe what about native podman approach, do you know? What we need here is a possibility to local images for several services and then start/stop, using single command?

jdurgin:
> > https://github.com/containers/podman-compose looks promising, I haven't tried it though.
> > Another approach is using a k8s-style yaml to define a pod and the containers in it: https://www.redhat.com/sysadmin/compose-podman-pods
> > These containers are simple enough that you could use a script containing the docker/podman commands directly as well, using the same arguments for both.
> 
> This is interesting but maybe what about native podman approach, do you know? What we need here is a possibility to local images for several services and then start/stop, using single command?

I'm not a podman expert, but we can group containers into a pod and stop/start the whole pod with a single command. Perhaps this would be better? https://developers.redhat.com/blog/2019/01/15/podman-managing-containers-pods/

kshtsk:
sorry, guys, I was not able to work on this, so maybe make it a bit easier... maybe land this as docker-compose staff, and for real podman implementation we create another ticket....

kshtsk:
@jdurgin does it makes sense to add docker-compose for now to mainline??

kshtsk:
the main frustration is all podman stuff is kinda adoption of docker-compose scripts... and also I have no idea where to store images...

jdurgin:
yeah, landing this with just docker-compose is fine

kshtsk:
@jdurgin nice, I will double check if it still works for me and update some docks how it can be used, and get back to you for review

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
hey @jdurgin I've added some notes how to use it, and I think we can merge it now, and improve with follow up commits, take a look

jdurgin (approved):

jdurgin:
looks like a great start! I think a single script to do all the prereqs (kind of like bootstrap.sh) would help a lot

PR #1593, opened 2021-01-04
sentry: use new library and group CommandFailedErrors more finely

jdurgin:
By default sentry uses the stack trace / error type / rough error
message, which ends up with many failures from different workunits
grouped together. Include the actual command run, the exit status, and
the optional label to group these more accurately. This will group
failures of the same workunit together, for example.

Here are a couple test errors that are treated as unique due to differing labels - at the bottom of the page you can expand the grouping info, and you can see the instances under the 'events' towards the top:
https://sentry.ceph.com/organizations/ceph/issues/724/?project=2&query=is%3Aunresolved#grouping-info
https://sentry.ceph.com/organizations/ceph/issues/725/?project=2&query=is%3Aunresolved#grouping-info

An example of sentry grouping unrelated failures together is almost any workunit failure being treated as the same issue:
https://sentry.ceph.com/organizations/ceph/issues/10/events/?project=2

Sentry docs on fingerprint:
https://docs.sentry.io/platforms/python/guides/logging/data-management/event-grouping/sdk-fingerprinting/


kshtsk:
@jdurgin test failed:
```
No module named 'sentry_sdk'
```

jdurgin:
update requirements.txt as well now

kshtsk:
> updated requirements.txt as well now

I wonder if it is needed to update setup.py too

jdurgin:
> > updated requirements.txt as well now
> 
> I wonder if it is needed to update setup.py too

good catch, that was the last reference to raven. I'd thought setup.py used requirements.txt, but I must have been thinking of another project.

kshtsk:
@susebot run deploy

kshtsk (approved):
lgtm

PR #1594, opened 2021-01-08
requirements: add libvirt-python for ubuntu

clwluvw:
libvirt-python is needed to import libvirt in python for ubuntu

Signed-off-by: Seena Fallah <seenafallah@gmail.com>

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -19,6 +19,7 @@ humanfriendly
 backports.ssl-match-hostname
 beanstalkc3>=0.4.0
 httplib2
+libvirt-python
 ndg-httpsclient
 pyasn1
 pyopenssl>=0.13
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,102 +4,346 @@
 #
 #    pip-compile --output-file=requirements.txt
 #
-ansible==2.8.8            # via -r requirements.in
-apache-libcloud==2.8.0    # via -r requirements.in
-appdirs==1.4.4            # via openstacksdk, virtualenv
-argparse==1.4.0           # via -r requirements.in
-atomicwrites==1.4.0       # via pytest
-attrs==19.3.0             # via cmd2, pytest
-backports.ssl-match-hostname==3.5.0.1  # via -r requirements.in
-bcrypt==3.1.7             # via paramiko
-beanstalkc3==0.4.0        # via -r requirements.in
-boto3==1.9.161            # via -r requirements.in
-boto==2.46.1              # via -r requirements.in
-botocore==1.12.253        # via boto3, s3transfer
-certifi==2020.6.20        # via requests
-cffi==1.14.0              # via bcrypt, cryptography, pynacl
-chardet==3.0.4            # via requests
-click==7.1.2              # via pip-tools
-cliff==3.3.0              # via osc-lib, python-openstackclient
-cmd2==1.2.1               # via cliff
-colorama==0.4.3           # via cmd2
-configobj==5.0.6          # via -r requirements.in
-configparser==3.5.0       # via -r requirements.in
-cryptography==3.2         # via -r requirements.in, ansible, openstacksdk, paramiko, pyopenssl
-debtcollector==2.1.0      # via oslo.config, oslo.utils, python-keystoneclient
-decorator==4.4.2          # via dogpile.cache, openstacksdk
-distlib==0.3.1            # via virtualenv
-docopt==0.6.2             # via -r requirements.in
-docutils==0.15.2          # via botocore
-dogpile.cache==0.9.2      # via openstacksdk
-filelock==3.0.12          # via virtualenv
-gevent==1.4.0             # via -r requirements.in
-greenlet==0.4.16          # via gevent
-httplib2==0.18.0          # via -r requirements.in
-humanfriendly==8.1        # via -r requirements.in
-idna==2.8                 # via requests
-ipy==1.0                  # via -r requirements.in
-iso8601==0.1.12           # via keystoneauth1, openstacksdk, oslo.utils, python-novaclient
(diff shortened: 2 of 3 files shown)


kshtsk:
1) The requirements.txt file looks corrupted
2) why the libvirt-python is ubuntu specific?

clwluvw:
> 1. The requirements.txt file looks corrupted
I've made it with pip-compile like a command above the file and I don't know why it generates it like this! Maybe because of the version or..? Is there anything I should specify in my command to be generated as before?

> 2. why the libvirt-python is ubuntu specific?
I've tested it in ubuntu env and I thought it's just needed for ubuntu because I think it's working for others that using centOS :D

kshtsk:
> > 1. why the libvirt-python is ubuntu specific?
> 
> I've tested it in ubuntu env and I thought it's just needed for ubuntu because I think it's working for others that using centOS :D

Which teuthology code are you referring that you've tested with.
Basically the libvirt-python is not requirements because there is a system package supposed to be preinstalled. It is deployed by ceph-cm-ansible role AFAIK.

clwluvw:
> > > 1. why the libvirt-python is ubuntu specific?
> > 
> > 
> > I've tested it in ubuntu env and I thought it's just needed for ubuntu because I think it's working for others that using centOS :D
> 
> Which teuthology code are you referring that you've tested with.
> Basically the libvirt-python is not requirements because there is a system package supposed to be preinstalled. It is deployed by ceph-cm-ansible role AFAIK.

If I set `ceph_cm_ansible: False` it won't be installed.

kshtsk:
> > > > 1. why the libvirt-python is ubuntu specific?
> > > 
> > > 
> > > I've tested it in ubuntu env and I thought it's just needed for ubuntu because I think it's working for others that using centOS :D
> > 
> > 
> > Which teuthology code are you referring that you've tested with.
> > Basically the libvirt-python is not requirements because there is a system package supposed to be preinstalled. It is deployed by ceph-cm-ansible role AFAIK.
> 
> If I set `ceph_cm_ansible: False` it won't be installed.

No, it is not for testnodes. It is teuthology role. https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/vars/apt_systems.yml#L9

clwluvw:
> > > > > 1. why the libvirt-python is ubuntu specific?
> > > > 
> > > > 
> > > > I've tested it in ubuntu env and I thought it's just needed for ubuntu because I think it's working for others that using centOS :D
> > > 
> > > 
> > > Which teuthology code are you referring that you've tested with.
> > > Basically the libvirt-python is not requirements because there is a system package supposed to be preinstalled. It is deployed by ceph-cm-ansible role AFAIK.
> > 
> > 
> > If I set `ceph_cm_ansible: False` it won't be installed.
> 
> No, it is not for testnodes. It is for the role of teuthology itself. https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/vars/apt_systems.yml#L9

Oh now I see what you mean with:
> Which teuthology code

I'm running it teuthology-suite manually from this repo and so I just installed the bootstrap here.

kshtsk:
Now, I am not sure, maybe you better add this dep to the bootstrap script?

clwluvw:
> Now, I am not sure, maybe you better add this dep to the bootstrap script?

I see all the python packages are installing from `requirements.txt`! Why should I move it to the bootstrap script?

kshtsk:
> > Now, I am not sure, maybe you better add this dep to the bootstrap script?
> 
> I see all the python packages are installing from `requirements.txt`! Why should I move it to the bootstrap script?

Because it is supposed to be a system package, otherwise you need to fix the ceph-cm-ansible role to not include libvirt-python, however, it needs to check if ceph-cm-ansible still work fine.

clwluvw:
> > > Now, I am not sure, maybe you better add this dep to the bootstrap script?
> > 
> > 
> > I see all the python packages are installing from `requirements.txt`! Why should I move it to the bootstrap script?
> 
> Because it is supposed to be a system package, otherwise you need to fix the ceph-cm-ansible role to not include libvirt-python, however, it needs to check if ceph-cm-ansible still work fine.

I think removing it from ceph-cm-ansible and add it here would be better because ceph-cm-ansible is running bootstrap too: https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/tasks/setup_users.yml#L75

@kshtsk If it's okay let me know to submit a patch there too.

kshtsk:
> > > > Now, I am not sure, maybe you better add this dep to the bootstrap script?
> > > 
> > > 
> > > I see all the python packages are installing from `requirements.txt`! Why should I move it to the bootstrap script?
> > 
> > 
> > Because it is supposed to be a system package, otherwise you need to fix the ceph-cm-ansible role to not include libvirt-python, however, it needs to check if ceph-cm-ansible still work fine.
> 
> I think removing it from ceph-cm-ansible and add it here would be better because ceph-cm-ansible is running bootstrap too: https://github.com/ceph/ceph-cm-ansible/blob/master/roles/teuthology/tasks/setup_users.yml#L75
> 
> @kshtsk If it's okay let me know to submit a patch there too.

I guess there was a reason why python-libvirt has been installing via system packages, maybe to be the same version as system libvirt package.

clwluvw:
@kshtsk I've tested with the system package and it won't work again. I've also set the virtualenv config to use system packages but it again failed!

kshtsk:
> @kshtsk I've tested with the system package and it won't work again. I've also set the virtualenv config to use system packages but it again failed!

again, what exactly do you test? sepia teuthology server is running on ubuntu

kshtsk:
No response from contributor more than two weeks.
Thank you for contribution feel free to create another PR with better explanation.

PR #1596, opened 2021-01-12
task/install/redhat: Removing packages based on OS in cleanup task

SrinivasaBharath:
script modified to remove packages  based on OS in the cleanup task.While uninstalling packages, both rpm and deb packages are getting first and  in the next step checking the OS name and removing the packages.

As part of fix first checking the OS name after that getting the appropriate packages to remove.


kshtsk:
I wonder, does RedHat support deb packages?

SrinivasaBharath:
Teuthology Log-   http://pulpito.ceph.redhat.com/skanta-2021-01-12_07:32:23-rados:singleton-master-distro-basic-clara/

rakeshgm:
We support for some of the older versions

SrinivasaBharath:
Currently we are supporting the deb packages.

SrinivasaBharath:
To clean up the branch and to raise the PR  only with required changes.

PR #1597, opened 2021-01-13
exceptions: only use one of label or command for fingerprint

jdurgin:
Commands like those running workunits include the ceph sha1 being
tested, so they're not useful for grouping. This also lets us group
together other tests if we like, for example to map tests with small
differences in configuration to the same fingerprint for sentry.

Also use the plain command, it's already a string at this point
so there's no reason to add spaces between its characters.

### DIFF
--- a/teuthology/exceptions.py
+++ b/teuthology/exceptions.py
@@ -74,9 +74,8 @@ class CommandFailedError(Exception):
         Used by sentry instead of grouping by backtrace.
         """
         return [
-            ' '.join(self.command),
+            self.label or self.command,
             'exit status {}'.format(self.exitstatus),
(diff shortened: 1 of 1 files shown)


jdurgin:
We only use the label argument for remote.run() in a few places right now, namely s3tests, ragweed, workunit, and cephfs/test_acls.py.

It might make sense to expand to further tasks running tests with differences in parameters. @batrick @dillaman a bunch of the rbd and cephfs tasks come to mind. You might want more specific errors than the CommandFailed exceptions for those though, e.g. looking at the output for the particular error from the test.

neha-ojha (approved):

PR #1599, opened 2021-01-17
Use the same version of teuthology and ceph-qa-suite for a whole run

jdurgin:
Specify the commit of teuthology, not just the branch, when scheduling jobs.

On the server side, use per-commit checkouts to avoid any changes while a job is reading from them.

Keep checkout-by-branch possible, since it's useful on the client side where there's only one user, e.g. updating master each time in someone's homedir each time they schedule a suite.

This makes suites more reliable, and allows https://github.com/ceph/teuthology/pull/1546 to work.

Here's a run with this PR on top of the dispatcher:

https://pulpito.ceph.com/joshd-2021-01-17_09:08:06-smoke-octopus-distro-basic-mira/

and some with only this PR, and the current worker daemon:

https://pulpito.ceph.com/joshd-2021-01-17_07:38:36-smoke-master-distro-basic-mira/
https://pulpito.ceph.com/joshd-2021-01-17_08:00:39-smoke-master-distro-basic-mira/


teuthology/repo_utils.py
```diff
@@ -70,33 +70,39 @@ def ls_remote(url, ref):
...
-def enforce_repo_state(repo_url, dest_path, branch, remove_on_error=True):
+def enforce_repo_state(repo_url, dest_path, branch, commit=None, remove_on_error=True):
     """
     Use git to either clone or update a given repo, forcing it to switch to the
     specified branch.
 
     :param repo_url:  The full URL to the repo (not including the branch)
     :param dest_path: The full path to the destination directory
     :param branch:    The branch.
+    :param commit:    The commit.
```
kshtsk, line 81: Should be added info about default.

teuthology/repo_utils.py
```diff
@@ -70,33 +70,39 @@ def ls_remote(url, ref):
...
     :param remove:    Whether or not to remove dest_dir when an error occurs
```
kshtsk, line 82: Hm.... not related to this PR but this should be updating according to actual signature of the function.

teuthology/repo_utils.py
```diff
@@ -70,33 +70,39 @@ def ls_remote(url, ref):
...
     :raises:          BranchNotFoundError if the branch is not found;
                       GitError for other errors
     """
     validate_branch(branch)
     sentinel = os.path.join(dest_path, '.fetched')
+    repo_reset = os.path.join(dest_path, '.fetched_and_reset')
```
kshtsk, line 88: Maybe add more comments regarding this variable?

teuthology/repo_utils.py
```diff
@@ -262,12 +268,13 @@ def fetch_branch(repo_path, branch, shallow=True):
...
 
 
-def reset_repo(repo_url, dest_path, branch):
+def reset_repo(repo_url, dest_path, branch, commit=None):
     """
 
     :param repo_url:  The full URL to the repo (not including the branch)
     :param dest_path: The full path to the destination directory
     :param branch:    The branch.
+    :param commit:    The commit to checkout, if different from HEAD.
```
kshtsk, line 277: - Add the defaults info.
- Is it checked if the commit different from the HEAD indeed, if not, why this message here?

teuthology/repo_utils.py
```diff
@@ -323,9 +335,9 @@ def fetch_repo(url, branch, bootstrap=None, lock=True):
             try:
                 while proceed():
                     try:
-                        enforce_repo_state(url, dest_path, branch)
+                        enforce_repo_state(url, dest_path, branch, commit)
                         if bootstrap:
-                            bootstrap(dest_path)
+                            bootstrap(dest_path, commit)
```
kshtsk, line 340: The bootstrap has nothing to do with commit and only used to check it to proceed:

```python
if commit and os.path.exists(sentinel) or is_fresh(sentinel):
            log.info(
                "Skipping bootstrap as it was already done in the last %ss",
                FRESHNESS_INTERVAL,
            )
            return
```

We rather do this check outside of bootstrap method.

teuthology/repo_utils.py
```diff
@@ -368,31 +380,33 @@ def url_to_dirname(url):
...
+    :param commit: The commit to checkout
     :returns:      The destination path
     """
     url = config.ceph_git_base_url + 'teuthology.git'
-    return fetch_repo(url, branch, bootstrap_teuthology, lock)
+    return fetch_repo(url, branch, commit, bootstrap_teuthology, lock)
 
 
-def bootstrap_teuthology(dest_path):
+def bootstrap_teuthology(dest_path, commit):
```
kshtsk, line 407: It is useless to have commit for the bootstrap method (especially without branch context), so check commit is empty on the caller side.

teuthology/suite/placeholder.py
```diff
@@ -47,6 +47,7 @@ def _substitute(input_dict, values_dict):
     'branch': Placeholder('ceph_branch'),
     'sha1': Placeholder('ceph_hash'),
     'teuthology_branch': Placeholder('teuthology_branch'),
+    'teuthology_sha1': Placeholder('teuthology_hash'),
```
kshtsk, line 50: Can we use *teuthology_sha1* instead of *teuthology_hash*.

teuthology/suite/run.py
```diff
@@ -99,7 +99,7 @@ def create_initial_config(self):
         else:
             self.suite_repo_path = util.fetch_repos(
                 suite_branch, test_name=self.name)
-        teuthology_branch = self.choose_teuthology_branch()
+        teuthology_branch, teuthology_hash = self.choose_teuthology_branch()
```
kshtsk, line 102: Can we use *teuthology_sha1* instead of *teuthology_hash*?

teuthology/suite/run.py
```diff
@@ -113,6 +113,7 @@ def create_initial_config(self):
             ceph_hash=ceph_hash,
             ceph_repo=config.get_ceph_git_url(),
             teuthology_branch=teuthology_branch,
+            teuthology_hash=teuthology_hash,
```
kshtsk, line 116: Can we use *teuthology_sha1* instead of *teuthology_hash*.

teuthology/suite/run.py
```diff
@@ -201,7 +202,7 @@ def choose_ceph_version(self, ceph_hash):
 
     def choose_teuthology_branch(self):
         """Select teuthology branch, check if it is present in repo and
-        return the branch name value.
+        return the branch name and hash.
```
kshtsk, line 205: Maybe better to use
```
return tuple (branch, hash) where hash is a commit sha1 corresponding to the HEAD of the branch.
```

kshtsk:
Tests are failed due to build machine lacks of some system dependencies.

jdurgin:
good points, will fix up. also found we can fetch a specific sha1 without regard to branch, so will do that too.

jdurgin:
Addressed the comments, new run here: http://pulpito.ceph.com/joshd-2021-01-19_06:54:15-smoke-octopus-distro-basic-mira/

Checking out a specific sha1 that's not the tip of a branch directly needs a server-side git option that isn't commonly enabled, so I kept the cloning as-is.

djgalloway:
> @djgalloway https://jenkins.ceph.com/job/teuthology-pull-requests/1458/ the are some jenkins node issue with lacking system dependencies. Is it okay that teuthology PR job runs against 172.21.4.75+confusa13 and if so, can we update it with teuthology system dependencies.

We should probably just prevent that job from running on arm64 builders.  https://github.com/ceph/ceph-build/pull/1725

kshtsk:
retest this please

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1603, opened 2021-02-02
dispatcher: allow empty os_type for fake config

jdurgin:
This is the same default as reimaging uses,
though it's not too important in the supervisor.


djgalloway (approved):
I guess

PR #1605, opened 2021-02-04
dispatcher/supervisor: always unlock machines and save status

jdurgin:
If we can't connect to the machines anymore, we still need to clean
up.


dmick (approved):
LGTM (and today I learned about except Exception:)

PR #1612, opened 2021-02-10
ls: nicer ls output

liewegas:
- no error when teuthology.log is missing (provisioning)
- leave off pid

### DIFF
--- a/teuthology/ls.py
+++ b/teuthology/ls.py
@@ -56,23 +56,14 @@ def print_debug_info(job, job_dir, archive_dir):
     print('%s      ' % job, end='')
 
     try:
-        pidfile = os.path.join(job_dir, 'pid')
-        found = False
-        if os.path.isfile(pidfile):
-            pid = open(pidfile, 'r').read()
-            if os.path.isdir("/proc/%s" % pid):
-                cmdline = open('/proc/%s/cmdline' % pid,
-                               'r').read()
-                if cmdline.find(archive_dir) >= 0:
-                    print('(pid %s)' % pid, end='')
-                    found = True
-        if not found:
-            print('(no process or summary.yaml)', end='')
-        # tail
-        tail = os.popen(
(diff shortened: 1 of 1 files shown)


teuthology/ls.py
```diff
@@ -56,23 +56,13 @@ def print_debug_info(job, job_dir, archive_dir):
...
-            print('(no process or summary.yaml)', end='')
-        # tail
-        tail = os.popen(
-            'tail -1 %s/%s/teuthology.log' % (archive_dir, job)
-        ).read().rstrip()
-        print(tail, end='')
+        if os.path.exists(os.path.join(archive_dir, job, 'teuthology.log')):
+            tail = os.popen(
+                'tail -1 %s/%s/teuthology.log' % (archive_dir, job)
+            ).read().rstrip()
```
kshtsk, line 62: Maybe just reuse log path?
```
        log_path = os.path.join(archive_dir, job, 'teuthology.log')
        if os.path.exists(log_path):
            tail = os.popen(f'tail -1 {log_path}').read().rstrip()
```

teuthology/ls.py
```diff
@@ -56,23 +56,14 @@ def print_debug_info(job, job_dir, archive_dir):
...
-            print('(no process or summary.yaml)', end='')
-        # tail
-        tail = os.popen(
-            'tail -1 %s/%s/teuthology.log' % (archive_dir, job)
-        ).read().rstrip()
-        print(tail, end='')
+        log_path = os.path.join(archive_dir, job, 'teuthology.log')
+        if os.path.exists(log_path):
+            tail = os.popen(
+                'tail -1 %s' % log_path
```
kshtsk, line 62: nit: looks good, but what do  you have against f-strings? :-) since we switched to py3 only we encourage everyone to use it

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1613, opened 2021-02-11
nuke: only use keep_logs from the cli

jdurgin:
nuke() is called outside of the cli with a ctx that does not include
all the cli args. Use a default parameter for the functions instead of ctx.

### DIFF
--- a/teuthology/nuke/__init__.py
+++ b/teuthology/nuke/__init__.py
@@ -232,10 +232,10 @@ def main(args):
         else:
             subprocess.check_call(["kill", "-9", str(ctx.pid)])
 
-    nuke(ctx, ctx.unlock, ctx.synch_clocks, ctx.noipmi)
+    nuke(ctx, ctx.unlock, ctx.synch_clocks, ctx.noipmi, ctx.keep_logs)
 
 
-def nuke(ctx, should_unlock, sync_clocks=True, noipmi=False):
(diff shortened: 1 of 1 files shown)


jdurgin:
@susebot run deploy

neha-ojha (approved):
looks fine to me

teuthology/nuke/__init__.py
```diff
@@ -277,7 +277,7 @@ def nuke(ctx, should_unlock, sync_clocks=True, noipmi=False):
 def nuke_one(ctx, target, should_unlock, synch_clocks,
              check_locks, noipmi):
     ret = None
-    keep_logs = ctx.keep_logs
+    keep_logs = getattr(ctx, 'keep_logs', False)
```
kshtsk, line 280: Since keep_logs, is nuke cli specific, why just don't add this as an optional argument `keep_logs=False` similar to other parameters like `noimpi` and `sync_clocks`?

jdurgin: sure we can do that, don't want to end up with too many args for the function but one more doesn't hurt

jdurgin:
@susebot run deploy

neha-ojha:
@kshtsk @jdurgin It is becoming very hard to review failures without being able to look at failure reasons (see https://pulpito.ceph.com/nojha-2021-02-10_18:54:18-rados:-master-distro-basic-smithi/), can we please merge this PR soon?

neha-ojha (approved):

PR #1614, opened 2021-02-12
nuke: fix no_reboot only being present in the cli and add unit tests

jdurgin:
This is causing errors in many jobs now due to lack of ctx.no_reboot in internal nuke callers.


jdurgin:
@susebot run deploy

neha-ojha (approved):

PR #1618, opened 2021-02-18
misc: make valgrind behave with tcmalloc

liewegas:


jdurgin (approved):
I assume this is harmless when run without tcmalloc (e.g. on older ceph branches)

jdurgin:
@susebot run deploy

PR #1619, opened 2021-02-18
py2: backport the dispatcher and specific-version changes

jdurgin:
We still need the py2 branch for testing luminous and mimic clients against octopus. This requires the dispatcher --supervisor to be able to run from the py2 branch.

There were only a few trivial conflicts, and one new commit to fix the py3 format strings.


jdurgin:
@yuriw is testing it with https://pulpito.ceph.com/teuthology-2021-02-18_23:28:56-upgrade:client-upgrade-mimic-octopus-octopus-distro-basic-smithi/5893178/

yuriw (approved):
:+1:

jdurgin:
@susebot run deploy

kshtsk:
interesting... we are not supporting py2 branch deployment anymore, so there is no sense to run the dispatcher itself via py2 at all.

jdurgin:
agreed, just the supervisor mode is needed, which works in yuri's tests.

PR #1626, opened 2021-03-09
remote: move non-file transfer methods from Remote

rishabh-d-dave:
Move methods that operate on a remote machine without depending on
another remote machine from orchestra.remote.Remote to a different class
in same module. This enables vstart_runner to reuse these methods.

Related PR: https://github.com/ceph/ceph/pull/37667

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -26,8 +26,302 @@ import netaddr
 log = logging.getLogger(__name__)
 
 
-class Remote(object):
+class RemoteShell(object):
+    """
+    Contains methods to run miscellaneous shell commands on remote machines.
 
+    These methods were originally part of orchestra.remote.Remote. The reason
+    for moving these methods from Remote is that applications that use
+    teuthology for testing usually have programs that can run tests locally on
+    a single node machine for development work (for example, vstart_runner.py
+    in case of Ceph). These programs can import and reuse these methods
+    without having to deal SSH stuff. In short, this class serves a shared
+    interface.
+
+    To use these methods, inherit the class here and implement "run()" method in
+    the subclass.
+    """
+
+    def remove(self, path):
+        self.run(args=['rm', '-fr', path])
+
+    def mkdtemp(self, suffix=None, parentdir=None):
+        """
+        Create a temporary directory on remote machine and return it's path.
+        """
+        args = ['mktemp', '-d']
+
+        if suffix:
+            args.append('--suffix=%s' % suffix)
+        if parentdir:
+            args.append('--tmpdir=%s' % parentdir)
+
+        return self.sh(args).strip()
+
+    def mktemp(self, suffix=None, parentdir=None, data=None):
+        """
+        Make a remote temporary file.
+
+        :param suffix:      suffix for the temporary file
+        :param parentdir:   parent dir where temp file should be created
+        :param data:        write data to the file if provided
+
+        Returns: the path of the temp file created.
+        """
+        args = ['mktemp']
+        if suffix:
+            args.append('--suffix=%s' % suffix)
+        if parentdir:
+            args.append('--tmpdir=%s' % parentdir)
+
+        path = self.sh(args).strip()
+
+        if data:
+            self.write_file(path=path, data=data)
+
+        return path
+
+    def sh(self, script, **kwargs):
+        """
+        Shortcut for run method.
+
+        Usage:
+            my_name = remote.sh('whoami')
+            remote_date = remote.sh('date')
+        """
+        if 'stdout' not in kwargs:
+            kwargs['stdout'] = BytesIO()
+        if 'args' not in kwargs:
+            kwargs['args'] = script
+        proc = self.run(**kwargs)
+        out = proc.stdout.getvalue()
+        if isinstance(out, bytes):
+            return out.decode()
+        else:
+            return out
+
+    def sh_file(self, script, label="script", sudo=False, **kwargs):
+        """
+        Run shell script after copying its contents to a remote file
+
+        :param script:  string with script text, or file object
+        :param sudo:    run command with sudo if True,
+                        run as user name if string value (defaults to False)
+        :param label:   string value which will be part of file name
+        Returns: stdout
+        """
+        ftempl = '/tmp/teuthology-remote-$(date +%Y%m%d%H%M%S)-{}-XXXX'\
+                 .format(label)
+        script_file = self.sh("mktemp %s" % ftempl).strip()
+        self.sh("cat - | tee {script} ; chmod a+rx {script}"\
+            .format(script=script_file), stdin=script)
+        if sudo:
+            if isinstance(sudo, str):
+                command="sudo -u %s %s" % (sudo, script_file)
+            else:
+                command="sudo %s" % script_file
+        else:
+            command="%s" % script_file
+
+        return self.sh(command, **kwargs)
+
+    def chmod(self, file_path, permissions):
+        """
+        As super-user, set permissions on the remote file specified.
+        """
+        args = [
+            'sudo',
+            'chmod',
+            permissions,
+            file_path,
+            ]
+        self.run(
+            args=args,
+            )
+
+    def chcon(self, file_path, context):
+        """
+        Set the SELinux context of a given file.
+
+        VMs and non-RPM-based hosts will skip this operation because ours
+        currently have SELinux disabled.
+
+        :param file_path: The path to the file
+        :param context:   The SELinux context to be used
+        """
+        if self.os.package_type != 'rpm' or \
+                self.os.name in ['opensuse', 'sle']:
+            return
+        if teuthology.lock.query.is_vm(self.shortname):
+            return
+        self.run(args="sudo chcon {con} {path}".format(
+            con=context, path=file_path))
+
+    def copy_file(self, src, dst, sudo=False, mode=None, owner=None,
+                                              mkdir=False, append=False):
+        """
+        Copy data to remote file
+
+        :param src:     source file path on remote host
+        :param dst:     destination file path on remote host
+        :param sudo:    use sudo to write file, defaults False
+        :param mode:    set file mode bits if provided
+        :param owner:   set file owner if provided
+        :param mkdir:   ensure the destination directory exists, defaults
+                        False
+        :param append:  append data to the file, defaults False
+        """
+        dd = 'sudo dd' if sudo else 'dd'
+        args = dd + ' if=' + src + ' of=' + dst
+        if append:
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/remote.py
```diff
@@ -26,7 +26,275 @@
 log = logging.getLogger(__name__)
 
 
-class Remote(object):
+class LocalOps(object):
```
kshtsk, line 29: Maybe _Common_ is more suitable name for this kind of orchestration operation... 
Also, since you want to isolate this API maybe moving to separate module (remote) for better abstraction?

teuthology/orchestra/remote.py
```diff
@@ -26,7 +26,275 @@
...
+    """
+    This class serves a shared interface for teuthology's orchestra and
+    vstart_runner. It contains methods that operate on the given remote
```
kshtsk, line 32: I don't think vstart_runner should ever mentioned in teuthology code base, because it is a consumer software... we can build the interface universal but teuthology as a framework should not aware about whoever who use it.

kshtsk: But in general why not use pure ssh for vstart_runner, maybe it is metter of configuring to localhost for simplification of interface?

kshtsk:
Could you please elaborate why this is required, and how this enable to 3rd party software to reuse something from teuthology?

batrick (commented):
(Thought I finished this review yesterday but github must have left it incomplete.)

teuthology/orchestra/remote.py
batrick: @kshtsk teuthology is used as a library in the Ceph code base but which parts of it are used is not always clear. I think we should at least emphasize this class is inherited from externally.

teuthology/orchestra/remote.py
batrick: A separate module makes sense to me. I also agree the name could be improved. Let's call it `RemoteOrchestration`?

rishabh-d-dave: I've been thinking about a better name too. `Common` feels too generic and `RemoteOrchestration` doesn't specify why methods like `put_file()` and `get_file()` are in Remote and not here. Perhaps... `NonTransferOps`? I am not sure myself.

rishabh-d-dave:
> Could you please elaborate why this is required, and how this enable to 3rd party software to reuse something from teuthology?

Do you mean elaborate in the commit message?

rishabh-d-dave:
@kshtsk 
> Could you please elaborate why this is required, and how this enable to 3rd party software to reuse something from teuthology?

Do you mean elaborate in the commit message?

teuthology/orchestra/remote.py
```diff
@@ -150,175 +418,71 @@ def hostname(self):
 
     @property
     def machine_type(self):
-        if not getattr(self, '_machine_type', None):
```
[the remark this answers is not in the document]
rishabh-d-dave: Nope. I need to take a look...

teuthology/orchestra/remote.py
rishabh-d-dave: @batrick 
>  think we should at least emphasize this class is inherited from externally.

Do you mean, "emphasize that this class is inherited externally?" I am confused and don't understand this last part.

From a general perspective, I find myself agreeing with @kshtsk here...

batrick: > @batrick
> 
> > think we should at least emphasize this class is inherited from externally.
> 
> Do you mean, "emphasize that this class is inherited externally?" I am confused and don't understand this last part.

It means this class is used as an interface in code outside of the teuthology.git code base.

> From a general perspective, I find myself agreeing with @kshtsk here...

It's not always clear in teuthology's code which methods/classes are interacted with externally by ceph qa/ tree. A curious code reader might assume pulling these methods out of Remote was unnecessary. A comment should explain why it's done.

teuthology/orchestra/remote.py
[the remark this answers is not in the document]
rishabh-d-dave: Got fixed recent changes.

teuthology/orchestra/remote.py
rishabh-d-dave: Okay. Please let me know if the current comment is enough.

teuthology/orchestra/remote.py
rishabh-d-dave: Renamed.

teuthology/orchestra/remote.py
```diff
@@ -86,8 +87,11 @@ def reconnect(self, timeout=None, socket_timeout=None, sleep_time=30):
             self.ssh.close()
         if not timeout:
             return self._reconnect(timeout=socket_timeout)
-        start_time = time.time()
-        elapsed_time = lambda: time.time() - start_time
+        start_time = time_time()
+
+        def elapsed_time():
+            return time_time() - start_time
+
```
kshtsk, line 94: Not related.

teuthology/orchestra/remote.py
```diff
@@ -135,10 +139,10 @@ def _set_iface_and_cidr(self):
         regexp = 'inet.? %s' % self.ip_address
         for line in ip_addr_show.split('\n'):
             line = line.strip()
-            if re.match(regexp, line):
+            if re_match(regexp, line):
```
kshtsk, line 142: Why?

teuthology/orchestra/remote.py
```diff
@@ -135,10 +139,10 @@ def _set_iface_and_cidr(self):
...
                 items = line.split()
                 self._interface = items[-1]
-                self._cidr = str(netaddr.IPNetwork(items[1]).cidr)
+                self._cidr = str(netaddr_IPNetwork(items[1]).cidr)
```
kshtsk, line 145: Please revert.

kshtsk:
Overall patch fbadec9  `orchestra: fix flake8 issues in remote.py and remote_ops.py  …` doesn't do anything useful, the actual flake8 failures.

```
teuthology/orchestra/remote.py:16:1: F401 'teuthology.provision' imported but unused
teuthology/orchestra/remote.py:61:37: F821 undefined name 'get_reimage_types'
```

teuthology/orchestra/remote.py
kshtsk: It is a bit strange to have RemoteOrchestration and overkill, it is in the module, orchestra.remote already.

rishabh-d-dave:
@kshtsk I moved those commits to #1629 before your finished reviewing (I suppose) since those weren't related to this PR.

rishabh-d-dave:
@kshtsk @batrick PTAL.

teuthology/orchestra/remote_ops.py
```diff
@@ -0,0 +1,295 @@
+"""
+Contains methods to run standard operations (like creating/deleting files) on
```
batrick, line 2:
```suggestion
Contains methods to run miscellaneous shell commands on
```

teuthology/orchestra/remote_ops.py
```diff
@@ -0,0 +1,295 @@
...
+from io import BytesIO
+from io import StringIO
+from os.path import dirname as os_path_dirname
+
+from teuthology.exceptions import CommandFailedError
+from teuthology.lock.query import is_vm
+from teuthology.orchestra.opsys import OS
+
+
+class NonTransferRemoteOps(object):
```
batrick, line 26: The name still irks me. These are all just shell commands. Something like `RemoteShellOps` would be appropriate.

batrick (changes requested):

teuthology/orchestra/remote_ops.py
kshtsk: Now, I think, we should move that code back to Remote :-D seriously, instead maybe it is better to moveout just few methods Remote.sh to RemoteShell as a base class for remote?

kshtsk: Well... Looks like we need just base Runner class which can be overridden by RemoteRunner and LocalRunner (I don't know if this is an appropriate name for vstart use case)

rishabh-d-dave: I've changed `NonTransferRemoteOps` to `RemoteShell` and `remote_ops.py` to `remote_shell.py` in latest push.

rishabh-d-dave: BTW, I still feel that moving `RemoteShell` to a separate module is a bit of an overkill. It's best to keep it in `remote.py`.

rishabh-d-dave: @kshtsk 
> Well... Looks like we need just base Runner class which can be overridden by RemoteRunner and LocalRunner (I don't know if this is an appropriate name for vstart use case)

Programs like `vstart_runner.py` don't need those SSH bits that current `remote.Remote` contains .

rishabh-d-dave:
@batrick @kshtsk PTAL.

teuthology/orchestra/remote_ops.py
kshtsk: > BTW, I still feel that moving `RemoteShell` to a separate module is a bit of an overkill. It's best to keep it in `remote.py`.

Agree.

What about to have two classes RemoteShell which will include only ssh run and sh, and have LocalShell, both based on AbstractShell. The Remote class should use either of the RemoteShell or LocalShell to construct.

rishabh-d-dave: @kshtsk Just separating `sh()` and `run()` won't separate methods that depend on SSH from methods that don't which is the goal of this PR. Or do you mean to make this a further change on the top of the current patch?

Perhaps, I am not clear about what you want to say. In that case, can we chat about this on IRC? What's you nickname? Mine is rishabh.

rishabh-d-dave: FYI: I've moved contents remote_shell.py as it is to remote.py.

rishabh-d-dave:
@kshtsk PTAL.

batrick:
needs rebase, changes look fine (but the diff is still weird, who knows why)

rishabh-d-dave:
@batrick I've rebased the PR. I am not exactly sure what causes the issue with the diff but it has something to do with moving of `os()` and `arch()`; see https://github.com/rishabh-d-dave/teuthology/commits/orch-remote-2-2.

rishabh-d-dave:
@kshtsk PTAL.

rishabh-d-dave:
@batrick @kshtsk Please take a look.

batrick (approved):

batrick:
@susebot run deploy

rishabh-d-dave:
@batrick Thank you for the review and for merging the PR.

PR #1630, opened 2021-03-17
tasks/exec: Document "all"

sebastian-philipp:
as this is a source of confusion.

### DIFF
--- a/teuthology/task/exec.py
+++ b/teuthology/task/exec.py
@@ -23,6 +23,14 @@ def task(ctx, config):
     It stops and fails with the first command that does not return on success. It means
     that if the first command fails, the second won't run at all.
 
+    Execute commands for all roles. Note that ``all`` is propbably running the
+    commands multiple times per host. 
+
+        tasks:
+        - exec:
+          all:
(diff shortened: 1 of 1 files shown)


teuthology/task/exec.py
```diff
@@ -23,6 +23,14 @@ def task(ctx, config):
     It stops and fails with the first command that does not return on success. It means
     that if the first command fails, the second won't run at all.
 
+    Execute commands for all roles. Note that ``all`` is propbably running the
```
kshtsk, line 26: probable? maybe you meant possible? or even it definitely will if there are several nodes per role

not sure that this behavior is expected, btw

liewegas: there were two users, both incorrect. both are fixed in master; backport branches still (mis)use it.

i would suggest removing it, or replacing 'all' with two alternatives, 'all-hosts' and 'all-roles'.

kshtsk: I like `all-hosts` and `all-roles`, maybe 'all' make a synonym to 'all-hosts'?

PR #1634, opened 2021-03-26
orchestra/remote: extend mktemp() to accept data 

rishabh-d-dave:


batrick (approved):

teuthology/orchestra/remote.py
```diff
@@ -599,6 +598,22 @@ def sudo_write_file(self, path, data, **kwargs):
         """
         self.write_file(path, data, sudo=True, **kwargs)
 
+    def write_temp_file(self, **kwargs):
```
batrick, line 601:
```suggestion
    def write_temp_file(self, suffix=None, parentdir=None, **kwargs):
```

teuthology/orchestra/remote.py
```diff
@@ -599,6 +598,22 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        Reason for excluding sudo is that files in /tmp can not be written by
+        any user (including the root user) besides the file owner.
+
+        Takes arguments of write_file() and of mktemp().
+        """
+        kwargs['sudo'] = False
+
+        if not kwargs.get('path'):
+            kwargs['path'] = self.mktemp(kwargs.pop('suffix',  None),
+                                         kwargs.pop('parentdir', None))
```
batrick, line 613:
```suggestion
            kwargs['path'] = self.mktemp(suffix=suffix, parentdir=parentdir)
```

teuthology/orchestra/remote.py
```diff
@@ -599,6 +598,22 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        self.write_file(**kwargs)
```
batrick, line 614: Above fixes are because these two parameters are not valid for this call.

batrick (changes requested):

teuthology/orchestra/remote.py
```diff
@@ -599,6 +598,22 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        This is a convenience method that call write_file() but without sudo.
+        Reason for excluding sudo is that files in /tmp can not be written by
+        any user (including the root user) besides the file owner.
+
+        Takes arguments of write_file() and of mktemp().
+        """
+        kwargs['sudo'] = False
+
+        if not kwargs.get('path'):
+            kwargs['path'] = self.mktemp(kwargs.pop('suffix',  None),
```
kshtsk, line 612: Instead of adding another method, could you please extend `mktemp` method with `data` argument, and if it not None, call remote_write.

kshtsk (changes requested):
I suggest to just use mktemp with data paremeter, then adding another one write_file method.

teuthology/orchestra/remote.py
rishabh-d-dave: Looking at the code snippet I am confused. Which call? Call to `write_file()`? `suffix` and `parentdir` will not be contained by `kwargs` at this point because `pop()` will remove them from kwargs.

teuthology/orchestra/remote.py
rishabh-d-dave: Thanks for the review. I don't understand why accepting them in kwargs is not good enough. Also, an disadvantage of this style of parameters is that callers of `write_temp_file()` can end up passing multiple values for a parameter which is an unnecessary bug. Just `**kwargs` is better IMHO.

teuthology/orchestra/remote.py
rishabh-d-dave: Thanks for the review.

> Instead of adding another method, could you please extend `mktemp` method with `data` argument,

Okay, I get this much...

> and if it not None, call remote_write.

but I don't get this part. By remote_write do you mean `remote.Remote.write_file()`?

Also, just adding `data` as parameter to `Remote.mktemp` won't be enough in case someone wants to wanted to append instead of overwrite, set owner, perm modes, etc. too on the file. So, will it be clean and simple if mktemp accepted all arguments of `write_file()` besides `suffix` and `parentdir`?

I like the idea of extending mktemp(), I am just wondering if there is a case that extended mktemp() can't cover but two separate methods can.

teuthology/orchestra/remote.py
batrick: If `kwargs['path'] is None` then `kwargs['suffix']` and `kwargs['parentdir']` will be left set, causing an error in the call to `self.write_file`.

teuthology/orchestra/remote.py
batrick: > can end up passing multiple values for a parameter which is an unnecessary bug.

example?

teuthology/orchestra/remote.py
kshtsk: there is no sense at all in 'append' feature when creating a temporary file, because the whole idea of the _mktemp_ is always to create a new file with distinct name. The _mktemp_ method is supposed to be used in similar way as system `mktemp`, which is not likely used to have another owner than invoking user and mode is supposed to be 0600. If we really have use cases when we need other modes and owner, maybe we can extend mktemp, with additional parameters, but I don't think we need this.

teuthology/orchestra/remote.py
rishabh-d-dave:
```
>>> def somefunc(a, b, c, d, e):
...     pass
... 
>>> somefunc(1, 2, 3, 4,**kw)
TypeError: somefunc() got multiple values for argument 'c'
>>> 
```
Not mixing leads to a code that is more sturdy, lesser chance for this sort of bug.

rishabh-d-dave:
Removed write_temp_file() and upgraded mktemp() instead in last push.

teuthology/orchestra/remote.py
rishabh-d-dave: Extended mktemp() and deleted write_temp_file().

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,28 @@ def mkdtemp(self, suffix=None, parentdir=None):
 
         return self.sh(args).strip()
 
-    def mktemp(self, suffix=None, parentdir=None):
+    def mktemp(self, suffix=None, parentdir=None, path=None, data=None,
```
kshtsk, line 232: I don't understand why do you need path argument? The mktemp should care about the path, if user wants to use path, the write_file should be used?
Also, could you please explain why owner is required?

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,28 @@ def mkdtemp(self, suffix=None, parentdir=None):
...
+               owner=None, mode=None):
         """
         Make a remote temporary file
 
         Returns: the path of the temp file created.
```
kshtsk, line 237: The documentation must be updated with argument description.

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,28 @@ def mkdtemp(self, suffix=None, parentdir=None):
...
             args.append('--suffix=%s' % suffix)
         if parentdir:
             args.append('--tmpdir=%s' % parentdir)
 
-        return self.sh(args).strip()
+        if not path:
+            path = self.sh(args).strip()
+
+        if data:
+            # sudo is set to False since root user can't write files in /tmp
```
kshtsk, line 249: this is strange clause, root user can write to any directory in fact

kshtsk: And btw, if the file is created by other user, than current, you must use sudo, but the whole idea of temporary files is to use the same user you operate.

rishabh-d-dave: No, it can not anymore. Try writing files using sudo on kernel version post 4.19, I guess. Try ubuntu 20. This is what the entire is about. Please take a look at the commit message/PR description and at the ticket linked.

kshtsk: Ah... I see it is because of the new kernel feature, but anyways, why would you explicitly declare False, if it is already a default value?

kshtsk: As for https://tracker.ceph.com/issues/49466 this looks like a test bug, not the teuthology. The client code should care about tracking users or just always use default and not use sudo_write_file when it is not needed.

kshtsk: Also, if we want to write_file as a specific user, maybe it is a good idea to add user name and group support for _sudo_ argument for write_file.

kshtsk:
The bug referred in the commit message is related to the code:
```
    def create_keyring_file(self, remote, keyring):
        keyring_path = remote.run(args=['mktemp'], stdout=StringIO()).\
            stdout.getvalue().strip()
        sudo_write_file(remote, keyring_path, keyring)

        # required when triggered using vstart_runner.py.
        remote.run(args=['chmod', '644', keyring_path])

        return keyring_path
```
Instead of teuthology api.

kshtsk:
Why just instead of:
```
sudo_write_file(remote, keyring_path, keyring)
```
write:
```
remote.write_file(keyring_path, keyring)
```

teuthology/orchestra/remote.py
rishabh-d-dave: > Ah... I see it is because of the new kernel feature, but anyways, why would you explicitly declare False, if it is already a default value?

So that the reader know it's intentional and required even if the reader isn't aware of context.

rishabh-d-dave:
> The bug referred in the commit message is related to the code:
> 
> ```
>     def create_keyring_file(self, remote, keyring):
>         keyring_path = remote.run(args=['mktemp'], stdout=StringIO()).\
>             stdout.getvalue().strip()
>         sudo_write_file(remote, keyring_path, keyring)
> 
>         # required when triggered using vstart_runner.py.
>         remote.run(args=['chmod', '644', keyring_path])
> 
>         return keyring_path
> ```
> 
> Instead of teuthology api.

I never said the bug is in teuthology API. About the code referred to in the commit message , it's already being fixed. PTAL at https://github.com/ceph/ceph/pull/40431.

kshtsk:
So why this change is required? I don't think we need it.

rishabh-d-dave:
> ```
> sudo_write_file(remote, keyring_path, keyring)
> ```
>
> write:
>
> ```
> remote.write_file(keyring_path, keyring)
> ```

No, that won't fix the issue. What is need is -
```
keyring_path = remote.mktemp()
remote.write_file(keyring_path, keyring, sudo=False)
```

But what is much much better than this is -
```
keyring_path = remote.mktemp(data=keyring)
```

This is simpler and more powerful because I get to create and write temporary file in a single call. But, much more importantly, it closes the possibility of having a test failure because someone forgot to set `sudo` to `False` in possibly one of many calls to `write_file()`.

teuthology/orchestra/remote.py
kshtsk: It is self explanatory, if you create a random file and use write_file api, it is with user without sudo.

kshtsk:
> > The bug referred in the commit message is related to the code:
> > ```
> >     def create_keyring_file(self, remote, keyring):
> >         keyring_path = remote.run(args=['mktemp'], stdout=StringIO()).\
> >             stdout.getvalue().strip()
> >         sudo_write_file(remote, keyring_path, keyring)
> > 
> >         # required when triggered using vstart_runner.py.
> >         remote.run(args=['chmod', '644', keyring_path])
> > 
> >         return keyring_path
> > ```
> > 
> > 
> > Instead of teuthology api.
> 
> I never said the bug is in teuthology API. About the code referred to in the commit message , it's already being fixed. PTAL at [ceph/ceph#40431](https://github.com/ceph/ceph/pull/40431).

You said this "Helps fix https://tracker.ceph.com/issues/49466" which is not true.

kshtsk:
> > ```
> > sudo_write_file(remote, keyring_path, keyring)
> > ```
> > 
> > 
> > write:
> > ```
> > remote.write_file(keyring_path, keyring)
> > ```
> 
> 
> No, that won't fix the issue. What is need is -
>
> 
> ```
> keyring_path = remote.mktemp()

It is the same as: `keyring_path = remote.run(args=['mktemp'], stdout=StringIO()).\
            stdout.getvalue().strip()`

> remote.write_file(keyring_path, keyring, sudo=False)
> ```
 
As I said, teuthology api sudo=False is default, you should not point it exclusively, so there is no difference.
The remote.mktemp()
> 
> But what is much much better than this is -
> 
> 
> ```
> keyring_path = remote.mktemp(data=keyring)
> ```

It depends on your needs, I agree it is more handy, as I mentioned earlier.

> 
> This is simpler and more powerful because I get to create and write temporary file in a single call. But, much more importantly, it closes the possibility of having a test failure because someone forgot to set `sudo` to `False` in possibly one of many calls to `write_file()`.

Again, we have standard api write_file, and we change it to `sudo=True` unlikely, because it is odd.

rishabh-d-dave:
@kshtsk 
> You said this "Helps fix https://tracker.ceph.com/issues/49466" which is not true.

This code `keyring_path = remote.mktemp(data=keyring)` cannot work without this PR. So this PR is helping out the fix for the tracker issue.

It's best not to just fix the bug but to also fix the root cause of the bug, so that it can't occur again.

kshtsk:
Summarising:
1) This PR does not help or resolve the new feature of kernel to protect files in `/tmp` and it becomes more like extending of _mktemp_ usability, so it is needed only to add `data` argument and use `write_file` not `sudo_write_file`. No need to use 'sudo=False', default operations should be sudo less as much as possible.
2) The commit message does not correspond the idea of the PR, and have to be changed appropriately.

kshtsk:
> @kshtsk
> 
> > You said this "Helps fix https://tracker.ceph.com/issues/49466" which is not true.
> 
> This code `keyring_path = remote.mktemp(data=keyring)` cannot work without this PR. So this PR is helping out the fix for the tracker issue.

But you still can use standard API to fix the issue, so no need to mention it here, so it is not necessary to have it.

> It's best not to just fix the bug but to also fix the root cause of the bug, so that it can't occur again.

You are not fixing the root cause of the bug with this PR. The root cause was incorrect usage of sudo_write_file in test.

teuthology/orchestra/remote.py
rishabh-d-dave: > It is self explanatory, if you create a random file and use write_file api, it is with user without sudo.

If it was, I wouldn't have written this in the first place - https://github.com/ceph/teuthology/pull/1634#discussion_r603966225. :)

kshtsk: I am sorry :-) I don't get it, what can be read differently when we have sudo_write_file and write_file? Why would we have sudo_write_file if write_file would have sudo=True?

rishabh-d-dave:
@batrick I've removed every mention of why sudo must not be used from this PR in last update

teuthology/orchestra/remote.py
```diff
@@ -561,7 +572,6 @@ def read_file(self, path, sudo=False, stdout=None,
 
         return proc.stdout.getvalue()
 
-
```
kshtsk, line 564: Unchange this line.

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,31 @@ def mkdtemp(self, suffix=None, parentdir=None):
 
         return self.sh(args).strip()
 
-    def mktemp(self, suffix=None, parentdir=None):
+    def mktemp(self, suffix=None, parentdir=None, data=None, owner=None,
```
kshtsk, line 232: We only need data, neither `owner`, nor `mode`.

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,28 @@ def mkdtemp(self, suffix=None, parentdir=None):
...
         return self.sh(args).strip()
 
-    def mktemp(self, suffix=None, parentdir=None):
+    def mktemp(self, suffix=None, parentdir=None, data=None):
         """
-        Make a remote temporary file
+        Make a remote temporary file.
+
+        :param suffix:      suffix for the temporary file
+        :param parentdir:   parent dir where temp files should be created
```
kshtsk, line 237: files => file

teuthology/orchestra/remote.py
```diff
@@ -229,20 +229,28 @@ def mkdtemp(self, suffix=None, parentdir=None):
...
+        :param data:        write data on the file if provided
```
kshtsk, line 238: write data *to* the file

kshtsk:
@susebot run deploy

kshtsk (approved):
Looks good to go now, thanks for the patience, and sorry that it took that much time.

kshtsk:
@batrick I think your change request is outdated.

batrick (approved):

rishabh-d-dave:
@kshtsk @batrick Thanks!

PR #1635, opened 2021-03-27
fetch more rerun parameters from last run

ideepika:
we only fetch seed and subset param's from previous reruns right now,
   this adds fetching ceph_branch, sha1, suite_branch, suite_repo as
   well from last ran job.

Signed-off-by: Deepika Upadhyay <dupadhya@redhat.com>


kshtsk:
I could image why we parse the log run log, but I would suggest to get rid of parsing the log and start generate some status file for a run in json/yaml format which can be easily parsed, with just a single line.

kshtsk:
Any updates on this, if it is not actual already, just close maybe?

ideepika:
sorry, could not update it recently, will do this week

ideepika:
> sorry, could not update it recently, will do this week

can work on a fix and open a new PR, closing for now, thanks!

PR #1639, opened 2021-04-06
orchestra/run: add parameter shell to run()

rishabh-d-dave:
Add parameter "shell" to teuthology.orchestra.run.run() with its default
value set to "True" to allow compatiblity with vstart_runner.py and to
also allow programs like vstart_runner.py to exercise greater
flexibility. This should not affect code in teuthology since "shell" is
set to "True" by default for every command executed.

Fixes: https://tracker.ceph.com/issues/50178

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -394,8 +394,8 @@ def run(
     quiet=False,
     timeout=None,
     cwd=None,
-    # omit_sudo is used by vstart_runner.py
-    omit_sudo=False
+    omit_sudo=False,
+    shell=True
 ):
     """
     Run a command remotely.  If any of 'args' contains shell metacharacters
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/run.py
```diff
@@ -395,7 +395,8 @@ def run(
     timeout=None,
     cwd=None,
     # omit_sudo is used by vstart_runner.py
-    omit_sudo=False
+    omit_sudo=False,
+    shell=True
```
batrick, line 399: Needs docs below.

I wonder if it'd be better to do:

``**kwargs``

instead so we don't need to keep them in sync all the time?

rishabh-d-dave: > Needs docs below.
> 

Added docs.

> I wonder if it'd be better to do:
> 
> `**kwargs`
> 
> instead so we don't need to keep them in sync all the time?

I personally feel accepting arguments in the current fashion keeps the API easier to get a grip on and also keeps track of parameters that are acceptable.

kshtsk: We should stop doing this, like adding useless parameters, and I was hopping we can get rid of `omit_sudo` as far as we added file operations api with _sudo_ arguments. The vstart_runner.py should not *hack* teuthology, but it should use it in the correct way.

kshtsk: As argument, the usin of shell=True is a feature already, otherwise multiline script commands will fail once trying to run in this mode.

kshtsk:
Sorry guys, we must close this PR.

batrick:
@rishabh-d-dave is this PR still fixing the issue cited? I haven't seen that ticket reappear since April.

rishabh-d-dave:
Closing the PR since an alternative solution has been found.

PR #1647, opened 2021-05-14
task/internal/syslog: avoid failing runs when ceph daemon logs go to syslog misc.log

liewegas:
Avoid the problem of tests failing by not checking for magic strings in misc.log--these were intended for kern.log checks.

It is still a bit of a mystery why there is a difference in behavior when ceph daemons log to stdout/stderr (output *never* makes its way to misc.log) vs ceph daemons logging to the journald socket (https://github.com/ceph/ceph/pull/40640) (logs *sometimes* end up in misc.log).

### DIFF
--- a/teuthology/task/internal/syslog.py
+++ b/teuthology/task/internal/syslog.py
@@ -99,7 +99,7 @@ def syslog(ctx, config):
                 [
                     'egrep', '--binary-files=text',
                     '\\bBUG\\b|\\bINFO\\b|\\bDEADLOCK\\b',
-                    run.Raw('{adir}/syslog/*.log'.format(adir=archive_dir)),
+                    run.Raw(f'{archive_dir}/syslog/kern.log'),
                     run.Raw('|'),
                     'grep', '-v', 'task .* blocked for more than .* seconds',
                     run.Raw('|'),


sebastian-philipp (approved):

liewegas:
@susebot run deploy

teuthology/task/internal/syslog.py
```diff
@@ -54,6 +61,11 @@ def syslog(ctx, config):
                 data=conf_fp,
             )
             conf_fp.seek(0)
+            misc.sudo_write_file(
+                remote=rem,
+                path='/etc/systemd/journald.conf',
+                data=journald_conf.encode(),
+            )
```
kshtsk, line 68: the misc.sudo_write_file is deprecated, please use remote method like this:
```
rem.write_file('/etc/systemd/journald.conf', journald_conf.encode(), sudo=True)
```
or

```
rem.sudo_write_file('/etc/systemd/journald.conf', journald_conf.encode())
```

kshtsk (changes requested):
Could you please avoid using of misc.sudo_write_file, it is deprecated.

teuthology/task/internal/syslog.py
```diff
@@ -99,7 +116,7 @@ def syslog(ctx, config):
                 [
                     'egrep', '--binary-files=text',
                     '\\bBUG\\b|\\bINFO\\b|\\bDEADLOCK\\b',
-                    run.Raw('{adir}/syslog/*.log'.format(adir=archive_dir)),
+                    run.Raw('{adir}/syslog/kern.log'.format(adir=archive_dir)),
```
kshtsk, line 119: instead of format, we use `f''` string, since we are python3 only.

teuthology/task/internal/syslog.py
```diff
@@ -33,6 +33,13 @@ def syslog(ctx, config):
         )
     )
 
+    # in systemd 237, the default is to forward to wall and syslog
```
huww98, line 36: This is incorrect. On Ubuntu, the default is to forward to wall and **syslog** because Ubuntu developers patched systemd. It is not related to systemd version.

huww98:
> The first patch ought to resolve this, except that there is some weirdness where the option doesn't seem to change the loggin behavior, even when systemd-journald is restarted. I can't figure out what is going on.

On some systems (Fedora and CentOS, I think), rsyslog can pull logs from journald, and this behaviour is not affected by `ForwardToSyslog` config.

teuthology/task/internal/syslog.py
```diff
@@ -48,12 +55,19 @@ def syslog(ctx, config):
...
                 rem.chcon(log_path, log_context)
-            misc.sudo_write_file(
+            misc.write_file(
                 remote=rem,
                 path=CONF,
                 data=conf_fp,
+                sudo=True,
             )
             conf_fp.seek(0)
+            misc.write_file(
```
kshtsk, line 65: this is still uses `misc`

liewegas:
I dropped the first patch; this now just ignores misc.log.

kshtsk:
@susebot run deploy

kshtsk (approved):

PR #1652, opened 2021-06-09
exceptions: change error message for CommandFailedError

rishabh-d-dave:
For the convenience of user, print the command that led to
CommandFailedError in error message as a str than as a list or a tuple.
This makes it more readable and enables the user to copy and use the
command without any hassles.


teuthology/exceptions.py
```diff
@@ -51,7 +51,12 @@ class CommandFailedError(Exception):
     Exception thrown on command failure
     """
     def __init__(self, command, exitstatus, node=None, label=None):
-        self.command = command
```
kshtsk, line 54: From one side this might seem useful for a user to copy and paste command,
but from another point of view, it is bad for debugging, it will be not possible to figure out what
exactly command was before it got to the Exception. Also, the next line joining command with spaces is absolutely not the same as it makes it quoted inside run methods, which makes it hard to reproduce issue.

rishabh-d-dave: > From one side this might seem useful for a user to copy and paste command,

> but from another point of view, it is bad for debugging, it will be not possible to figure out what
exactly command was before it got to the Exception.

I don't see how would figuring out become difficult since parameter would be printed as it is (at least individually). For example, `['ls', 'dir1']` would become `ls dir1`. I have grepped plenty times and either don't make much of a difference. 

> Also, the next line joining command with spaces is absolutely not the same as it makes it quoted inside run methods, which makes it hard to reproduce issue.

Same as what? 

Umm... are you referring to [this line](https://github.com/ceph/teuthology/blob/master/teuthology/orchestra/run.py#L65) when you say "quoted inside run methods"? I think there's no nice way around that but printing as string definitely takes away effort of deleting a dozen inverted commas every time.

kshtsk: Yes, I mean the quotation thing.
Also, if you want to get the exact command, it is already printed in the logs, you just need to find the place where debug is printing the command out before the execution.

teuthology/exceptions.py
```diff
@@ -51,7 +51,12 @@ class CommandFailedError(Exception):
...
+        if isinstance(command, (list, tuple)):
+            self.command = ' '.join(command)
```
batrick, line 55: I think the right place for this transformation would be in the `__str__` method. It would improve readability, yes.

rishabh-d-dave: True, that'll be much better.

teuthology/exceptions.py
```diff
@@ -50,21 +54,30 @@ class CommandFailedError(Exception):
     """
     Exception thrown on command failure
     """
-    def __init__(self, command, exitstatus, node=None, label=None):
-        self.command = command
+    def __init__(self, cmd, exitstatus, node=None, label=None):
+        self.cmd = cmd
```
kshtsk, line 58: Why renaming `command` to `cmd`? 
- This change is not necessary.
- Breaks the convention of naming arguments.

teuthology/exceptions.py
```diff
@@ -50,21 +54,30 @@ class CommandFailedError(Exception):
...
     def __str__(self):
+        if isinstance(self.cmd, (list, tuple)):
+            self.cmd = [str(Raw) if isinstance(x, Raw) else x \
+                            for x in self.cmd]
+            self.cmd = ' '.join(self.cmd)
+        elif isinstance(self.cmd, str):
+            self.cmd = sub(r'Raw\([\'"](.*)[\'"]\)', r'\1', self.cmd)
+        else:
+            raise RuntimeError('variable "command" is not str, list or tuple')
+
```
kshtsk, line 72: Do not override the original command, just return the str representation computed in local variable.

teuthology/orchestra/run.py
```diff
@@ -9,13 +9,13 @@
...
 import gevent.event
 import socket
-import pipes
 import logging
 import shutil
 
 from teuthology.contextutil import safe_while
 from teuthology.exceptions import (CommandCrashedError, CommandFailedError,
                                    ConnectionLostError)
+from teuthology.orchestra.run_helper import Raw, quote
```
kshtsk, line 18: Instead of `run_helper` I suggest more appropriate name for the module, `util` or `tool` or `common` or for example

teuthology/exceptions.py
```diff
@@ -50,21 +54,30 @@ class CommandFailedError(Exception):
...
         self.node = node
         self.label = label
 
     def __str__(self):
+        if isinstance(self.cmd, (list, tuple)):
+            self.cmd = [str(Raw) if isinstance(x, Raw) else x \
+                            for x in self.cmd]
+            self.cmd = ' '.join(self.cmd)
+        elif isinstance(self.cmd, str):
+            self.cmd = sub(r'Raw\([\'"](.*)[\'"]\)', r'\1', self.cmd)
```
kshtsk, line 69: Why this branch is needed?

teuthology/exceptions.py
rishabh-d-dave: > Why renaming command to cmd?

Lesser effort to type, and easier to keep statements under 79 characters which in turn saves us from hard to read split-over-several-line statements .

teuthology/orchestra/run.py
rishabh-d-dave: I find run_helper appropriate since it contains classes that help run.py.

BTW I moving rest of misc. classes from run.py to run_helper.py too.

teuthology/exceptions.py
rishabh-d-dave: Transforms str `'param1 param2 Raw('&&") param3'` to str `'param1 param2 && param3'`.

teuthology/exceptions.py
rishabh-d-dave: Okay, that would be better, I guess.

teuthology/exceptions.py
kshtsk: why this is needed to transform I do not understand

teuthology/orchestra/run.py
kshtsk: that looks weird, if you really want to have run based classes it would have more appropriate to have this layout:

teuthology/orchestra/run/__init__.py
teuthology/orchestra/run/helper.py

instead of underscore separated.

teuthology/exceptions.py
kshtsk: or in which situation this is happening...

rishabh-d-dave:
@kshtsk @batrick 
`teuthology.orchestra.run.spawn_asyncresult()` is a dead function, there are no calls to it in teuthology codebase. Better to get rid of it?

kshtsk:
> @kshtsk @batrick
> `teuthology.orchestra.run.spawn_asyncresult()` is a dead function, there are no calls to it in teuthology codebase. Better to get rid of it?

I don't know, Isn't it used in any named branch of ceph `qa/tasks`?

rishabh-d-dave:
@kshtsk 
> > @kshtsk @batrick
> > `teuthology.orchestra.run.spawn_asyncresult()` is a dead function, there are no calls to it in teuthology codebase. Better to get rid of it?
> 
> I don't know, Isn't it used in any named branch of ceph `qa/tasks`?

No call there too.

```
$ git pull --no-rebase && grep -rnI spawn_asyncresult qa/
Already up to date.
```

teuthology/exceptions.py
rishabh-d-dave: Same reason for which i am converting `list` to `str`.

kshtsk: well... I don't understand why we do not use quote method here from the helper?

teuthology/orchestra/run.py
rishabh-d-dave: I personally feel this a bit of overkill but I don't really mind.

@batrick Any opinion?

teuthology/exceptions.py
rishabh-d-dave: How would that do the job?

```
$ python
Python 3.8.9 (default, Apr  6 2021, 00:00:00) 
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from teuthology.orchestra.run_helper import Raw, quote
>>> a = quote('a b Raw("&&") c')
>>> a
'a b Raw("&&") c'
>>> 
```

kshtsk: Yep, but how is it possible to have the Raw in the string? From my point of view if you have Raw in the str it means that you've made a typo and this command is not going to be run successfully.

PR #1655, opened 2021-06-10
teuthology/orchestra: store temporaries in tmp subdir

batrick:
Due to protections turned on in Ubuntu 20.04 from [1], files created by
other users (e.g. ubuntu) can not be manipulated by other users, even
root.

Instead of using /tmp as the parent directory for `mktemp` files, use a
sub-directory under it instead.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=30aba6656f61ed44cba445a3c0d38b296fa9e8f5

Fixes: https://tracker.ceph.com/issues/51169

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -154,6 +154,16 @@ def get_testdir(ctx=None):
     )
 
 
+def get_tmpdir():
+    """
+    :returns: A temporary directory for use by tests on the test node.
+    """
+    return config.get(
+        'test_tmpdir',
+        '/tmp/cephtest'
+    )
+
+
 def get_test_user(ctx=None):
     """
     :param ctx: Unused; accepted for compatibility
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -53,8 +53,10 @@ class RemoteShell(object):
 
         if suffix:
             args.append('--suffix=%s' % suffix)
-        if parentdir:
-            args.append('--tmpdir=%s' % parentdir)
+        if parentdir is None:
+            parentdir = misc.get_tmpdir()
+            self.sh(f"mkdir -p {parentdir}")
+        args.append('--tmpdir=%s' % parentdir)
 
         return self.sh(args).strip()
 
@@ -69,10 +71,13 @@ class RemoteShell(object):
         Returns: the path of the temp file created.
         """
         args = ['mktemp']
+
         if suffix:
             args.append('--suffix=%s' % suffix)
-        if parentdir:
-            args.append('--tmpdir=%s' % parentdir)
+        if parentdir is None:
+            parentdir = misc.get_tmpdir()
+            self.sh(f"mkdir -p {parentdir}")
+        args.append('--tmpdir=%s' % parentdir)
 
         path = self.sh(args).strip()
 


batrick:
Will test this on https://pulpito.ceph.com/?branch=wip-pdonnell-testing-20210610.025127

teuthology/misc.py
```diff
@@ -154,6 +154,18 @@ def get_testdir(ctx=None):
...
+def get_tmpdir():
+    """
+    :returns: A test directory
+    """
+    if 'tmp_path' in config:
+        return config['tmp_path']
+    return config.get(
+        'tmp_path',
+        '/tmp/cephtest'
+    )
```
tchaikov, line 166:
```suggestion
    return config.get(
        'tmp_path',
        '/tmp/cephtest'
    )
```

batrick: Thanks, done. Bad copy paste!

tchaikov (approved):

tchaikov:
@susebot run deploy

teuthology/misc.py
```diff
@@ -154,6 +154,16 @@ def get_testdir(ctx=None):
...
+def get_tmpdir():
+    """
+    :returns: A test directory
+    """
+    return config.get(
+        'tmp_path',
+        '/tmp/cephtest'
+    )
+
+
```
kshtsk, line 166: Why this method exist at all?

kshtsk:
This change overall is bad, it is trying to workaround some problem and changes the logic of the idea of the mkdtemp method, it should not included values. Also it will make it impossible to use systems default temp directory at all.

teuthology/misc.py
kshtsk: If the purpose to have change default temporary directory for remotes, this method should go to orchestra remote and return by default None, and the `/tmp/cephtest` value should be set in teuthology config then.

teuthology/orchestra/remote.py
```diff
@@ -53,8 +53,10 @@ def mkdtemp(self, suffix=None, parentdir=None):
 
         if suffix:
             args.append('--suffix=%s' % suffix)
-        if parentdir:
-            args.append('--tmpdir=%s' % parentdir)
+        if parentdir is None:
+            parentdir = misc.get_tmpdir()
+            self.sh(f"mkdir -p {parentdir}")
+        args.append('--tmpdir=%s' % parentdir)
```
kshtsk, line 59: Here and later the fix is breaking the overall logic of the function. If we want to manipulate the default parent dir for remote temp files, it should definitely named correspondingly in config, like `remote_temp_parentdir` and read directly from the config, not from misc.

Given that by default remote_temp_parentdir is None the following code should be sufficient:
```
parentdir = parentdir or config.get('remote_temp_parentdir', None)
if parentdir:
     args.append('--tmpdir=%s' % parentdir)
```

teuthology/orchestra/remote.py
```diff
@@ -69,10 +71,13 @@ def mktemp(self, suffix=None, parentdir=None, data=None):
...
         args = ['mktemp']
+
         if suffix:
             args.append('--suffix=%s' % suffix)
-        if parentdir:
-            args.append('--tmpdir=%s' % parentdir)
+        if parentdir is None:
+            parentdir = misc.get_tmpdir()
+            self.sh(f"mkdir -p {parentdir}")
+        args.append('--tmpdir=%s' % parentdir)
```
kshtsk, line 80: see the comment above

kshtsk (changes requested):
See the comments above.

teuthology/misc.py
batrick: It exists here because that seemed to be the most coherent place for it. e.g. `get_testdir`, `get_archive_dir`, etc.

teuthology/orchestra/remote.py
batrick: > If we want to manipulate the default parent dir for remote temp files, it should definitely named correspondingly in config, like remote_temp_parentdir and read directly from the config, not from misc.

Is that not what `tmp_path` in the misc function is fetching?

kshtsk: yeah, the `tmp_path` is just not a good name, and also, since it is related to a _remote_ context it should not be in a misc, in general the `misc` looks like a garbage, and everything should be placed in right module instead of storing to misc.

kshtsk: I would declare misc as deprecated at all.

batrick: `tmpdir` maybe then? 

> since it is related to a remote context it should not be in a misc, in general the misc looks like a garbage, and everything should be placed in right module instead of storing to misc.

That's refactoring. I'm not going to blaze a trail to making the code more sensical in a bug fix.

teuthology/misc.py
kshtsk: it is not the most, as I said, this is not a temp path for teuthology itself, this is used for remote nodes only. But in general I don't see any advantage of having this proxy method at all.

batrick: Uh, these `get_` methods are all for paths on the test nodes themselves. I don't see how it's "for teuthology itself".

kshtsk: Also in other comments, the `get_tmpdir` does not say to user anything, is this local temp dir where teuthology is running? Or is it only related to some specific test? Why there is a hardcoded name like `/tmp/cephtest`? Why not `/tmp/teuthology`? Teuthology as a test framework and common modules should be designed and developed as not hardly bounded to ceph, it is hard to believe, but it could run not only ceph specific stuff.

teuthology/orchestra/remote.py
kshtsk: but you're not fixing, you are introducing new logic, the mktemp is exactly to use system mktemp, if you to fix test, go to test modules and pass parentdir argument.

teuthology/misc.py
kshtsk: > Uh, these `get_` methods are all for paths on the test nodes themselves. I don't see how it's "for teuthology itself".

I'm not following.

teuthology/orchestra/remote.py
kshtsk: in other words this change is wasting API function, instead of fixing the real test logic, exactly test should manage the situation when it creates temp file in wrong place, this is why the parentdir argument exists.

kshtsk:
So, instead of this change, why not to fix the tests to use parentdir argument instead?

teuthology/orchestra/remote.py
batrick: You're asking me to make widespread changes in the qa/ code to fix a problem created by Ubuntu 20.04 instead of fixing it one place in teuthology? I considered that but it's too much code to fix.

kshtsk: First of all I ask not to not the base API complicated and difficult to maintain. But yes, if the test is written incorrectly for Ubuntu, fix it. If some one making a mess in client code, why not fix it, instead of doing something strange in the framework?

kshtsk: is this only code that makes it failing:
```
        for (file_size, file_name) in zip(file_sizes, file_names):
            temp_file = self.mount_a.client_remote.mktemp(file_name)
            self.mount_a.run_shell(f"fallocate -l {file_size} {temp_file}")
            self.mount_a.run_shell(f'mv {temp_file} ./')
```

kshtsk: also just imaging with this code you introduce several issues:
- hiding the real problem instead of resolving, reading the code it is not clear why the system mktemp cannot be used in general. And the code of the test is not written correctly and people do not understand why it works when it should not.
- this method will 100% fail if the /tmp/cephtest does not exists, why it must be created before the method called?
- user reading the doc for remote.mktemp does not understand it should have created /tmp/cephtest on remote host

teuthology/orchestra/remote.py
rishabh-d-dave, line 80: IMHO, it would be simpler to modify `mktemp()` to set `parentdir`'s default value to `config.get(test_tmpdir)` and add an assert check for `parentdir` that `/tmp/` is not the parent dir for the temp files.

kshtsk: it is simpler, but now, when I looked into the details of the cause of this PR even appeared I think it is just incorrect... please stop breaking framework with workarounds instead of fixing wrong algorithms in the tests, if kernel has the feature not allowing to create files in directory own by someone else even for root, then just stop doing this, and do make another bottlenecks

batrick:
> it is simpler, but now, when I looked into the details of the cause of this PR even appeared I think it is just incorrect... please stop breaking framework with workarounds instead of fixing wrong algorithms in the tests, if kernel has the feature not allowing to create files in directory own by someone else even for root, then just stop doing this, and do make another bottlenecks

This is incorrect. The problem is rampant use of `sudo` in all of teuthology's helper functions. It's too widespread to fix teuthology to stop using sudo unnecessarily. It would break tests everywhere. The only viable thing to do here is avoid using `/tmp` to store temporary files.

kshtsk:
> > it is simpler, but now, when I looked into the details of the cause of this PR even appeared I think it is just incorrect... please stop breaking framework with workarounds instead of fixing wrong algorithms in the tests, if kernel has the feature not allowing to create files in directory own by someone else even for root, then just stop doing this, and do make another bottlenecks
> 
> This is incorrect. The problem is rampant use of `sudo` in all of teuthology's helper functions. It's too widespread to fix teuthology to stop using sudo unnecessarily. It would break tests everywhere. The only viable thing to do here is avoid using `/tmp` to store temporary files.

Not sure I understand your point, what exactly is incorrect? The bug you are referring had bug when took a look at it. Teuthology by default does not use sudo, it is False by default. Please example?

PR #1657, opened 2021-07-01
teuthology/suite/run.py, scripts/suite.py: disallow scheduling too many jobs

neha-ojha:
Add check_num_jobs() to prevent users from accidentally scheduling too many jobs, like
in rfriedma-2021-06-26_19:32:15-rados-wip-ronenf-scrubs-config-distro-basic-smithi.
JOBS_TO_SCHEDULE_THRESHOLD, set to 500 (most runs have fewer jobs than this),
will disallow users from scheduling more than 500 jobs. Users can schedule
more than 500 jobs by disabling this check using the --disable-num-jobs-check flag.

Testing:

- Default behavior

```
2021-07-09 21:52:27,306.306 INFO:teuthology.suite.util:Memo: ./virtualenv/bin/teuthology-schedule --name nojha-2021-07-09_21:52:08-rados-wip-51354-3-distro-basic-smithi --worker smithi --dry-run --priority 60 -v --first-in-suite --subset 1/5000 --seed 7987
  File "./virtualenv/bin/teuthology-suite", line 11, in <module>
    load_entry_point('teuthology', 'console_scripts', 'teuthology-suite')()
  File "/home/nojha/py3_teuthology/scripts/suite.py", line 194, in main
    return teuthology.suite.main(args)
  File "/home/nojha/py3_teuthology/teuthology/suite/__init__.py", line 146, in main
    run.prepare_and_schedule()
  File "/home/nojha/py3_teuthology/teuthology/suite/run.py", line 398, in prepare_and_schedule
    num_jobs = self.schedule_suite()
  File "/home/nojha/py3_teuthology/teuthology/suite/run.py", line 654, in schedule_suite
    self.check_num_jobs(len(jobs_to_schedule))
  File "/home/nojha/py3_teuthology/teuthology/suite/run.py", line 534, in check_num_jobs
    util.schedule_fail(msg)
  File "/home/nojha/py3_teuthology/teuthology/suite/util.py", line 76, in schedule_fail
    raise ScheduleFailError(message, name)
teuthology.exceptions.ScheduleFailError: Scheduling failed: Unable to schedule 637 jobs, too many jobs.

Note: If you still want to go ahead, use --disable-num-jobs-check
```
- With --disable-num-jobs-check
```
./virtualenv/bin/teuthology-suite -v -s rados -p 60 --subset 1/5000 -c wip-51354-3   --force-priority  -m smithi --disable-num-jobs-check  --dry-run 
2021-07-09 21:56:44,410.410 INFO:teuthology.suite.run:Suite rados in /home/nojha/src/git.ceph.com_ceph-c_wip-51354-3/qa/suites/rados scheduled 637 jobs.
```

Signed-off-by: Neha Ojha <nojha@redhat.com>

### DIFF
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -172,6 +172,11 @@ Scheduler arguments:
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
+ --disable-num-jobs-check     Skip the number of jobs check. By default,
+                              teuthology will not allow you to schedule more
+                              than JOBS_TO_SCHEDULE_THRESHOLD=500 jobs, because
+                              it is too high. Use this if you need to schedule
+                              more than 500 jobs.
 
 """.format(
     default_machine_type=config.default_machine_type,
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -29,6 +29,7 @@ log = logging.getLogger(__name__)
 class Run(object):
     WAIT_MAX_JOB_TIME = 30 * 60
     WAIT_PAUSE = 5 * 60
+    JOBS_TO_SCHEDULE_THRESHOLD = 500
     __slots__ = (
         'args', 'name', 'base_config', 'suite_repo_path', 'base_yaml_paths',
         'base_args', 'package_versions', 'kernel_dict', 'config_input',
@@ -525,6 +526,13 @@ Note: To force run, use --force-priority'''
         elif priority < 150 and jobs_to_schedule > 100:
             util.schedule_fail(msg)
 
+    def check_num_jobs(self, jobs_to_schedule):
+        msg=f'''Unable to schedule {jobs_to_schedule} jobs, too many jobs.
+
+Note: If you still want to go ahead, use --disable-num-jobs-check'''
+        if jobs_to_schedule > Run.JOBS_TO_SCHEDULE_THRESHOLD:
+            util.schedule_fail(msg)
+
     def schedule_suite(self):
         """
         Schedule the suite-run. Returns the number of jobs scheduled.
@@ -640,6 +648,10 @@ Note: To force run, use --force-priority'''
         # Before scheduling jobs, check the priority
         if self.args.priority and jobs_to_schedule and not self.args.force_priority:
             self.check_priority(len(jobs_to_schedule))
+
+        # Before scheduling jobs, check number of jobs to avoid scheduling 1000s
+        if jobs_to_schedule and not self.args.disable_num_jobs_check:
+            self.check_num_jobs(len(jobs_to_schedule))
 
         self.schedule_jobs(jobs_missing_packages, jobs_to_schedule, name)
 


yuriw (approved):
500 is generous :)

neha-ojha:
@susebot run deploy

kshtsk:
oops the teuthology test job cannot be completed because it needs centos 8.3, but we only have 8.1 image, so I guess we need to add centos 8.3 image

kshtsk:
@susebot run deploy

kshtsk:
I've added centos-8.3 image, will see how it goes

kshtsk:
@djgalloway seems like ceph-cm-ansbile roles testnode and user lacks lsb release dependency:
```
2021-07-07T13:25:16.698 DEBUG:teuthology.orchestra.run.target-pr-1657-003:> sudo lsb_release -is
2021-07-07T13:25:16.772 DEBUG:teuthology.orchestra.run:got remote process result: 1
2021-07-07T13:25:16.774 INFO:teuthology.orchestra.run.target-pr-1657-003.stderr:sudo: lsb_release: command not found
```

kshtsk:
@susebot run deploy

djgalloway:
> @djgalloway seems like ceph-cm-ansbile roles testnode and user lack lsb release dependency:

Hm.  I think because we usually install via [Cobbler](https://github.com/ceph/ceph-cm-ansible/blob/master/roles/cobbler/templates/snippets/cephlab_packages_rhel#L20)

Fixing now.

djgalloway:
https://github.com/ceph/ceph-cm-ansible/pull/637

kshtsk:
> [ceph/ceph-cm-ansible#637](https://github.com/ceph/ceph-cm-ansible/pull/637)

nice, thanks a lot

teuthology/suite/run.py
```diff
@@ -29,6 +29,7 @@
 class Run(object):
     WAIT_MAX_JOB_TIME = 30 * 60
     WAIT_PAUSE = 5 * 60
+    MAX_JOBS = 500
```
idryomov, line 32: Nit: `MAX_JOBS` is a bit misleading IMO -- it's not a hard limit but may be interpreted as such.  I would just hard-code `500` in `check_num_jobs()`, similar to `check_priority()`.

kshtsk: maybe DEFAULT_JOB_LIMIT is better for this "const"?

neha-ojha: I think DEFAULT_JOB_LIMIT makes it more clear, updated the commit.

kshtsk:
@neha-ojha just wondering, maybe rename --force-job to --ignore-job-limit or something to be more self-descriptive.

kshtsk:
hm... we already have some job limit functionality but it only relates to sleep before teardown feature:
```
  -l <jobs>, --limit <jobs>   Queue at most this many jobs
                              [default: 0]
```
interesting if it will confuse anyone with this job limit
```
  --sleep-before-teardown <seconds>
                              Number of seconds to sleep before teardown.
                              Use with care, as this applies to all jobs in the
                              run. This option is used along with --limit one.
                              If the --limit ommitted then it's forced to 1.
                              If the --limit is greater than 4, then user must
                              confirm it interactively to avoid massive lock
                              of resources, however --non-interactive option
                              can be used to skip user input.
                              [default: 0]
```

scripts/suite.py
```diff
@@ -172,6 +172,8 @@
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
+ --ignore-job-limit           Skip the number of jobs check. Use this to schedule
+                              more than DEFAULT_JOB_LIMIT=500 jobs.
```
kshtsk, line 176: Maybe rename to --ignore-job-limit, also it would be better to say couple of words what is the default limit and that it is used, here. Or even have --limit-job options defaults to 500, and we need to go with maximum available we can use parameter equals 0 or '-1'.
I agree this is a bit weird to discuss this a posteriori.

kshtsk: ah... what do you think about using that `  -l <jobs>, --limit <jobs>   Queue at most this many jobs` option instead?

neha-ojha:
> @neha-ojha just wondering, maybe rename --force-job to --ignore-job-limit or something to be more self-descriptive.

I changed the flag to `ignore-job-limit` and made the description more clear. How about now?

scripts/suite.py
kshtsk: but if we change the default job limit to 500 then functionality of sleep before teardown will be broken, the code following code should be fixed: 
https://github.com/ceph/teuthology/blob/master/teuthology/suite/run.py#L563-L573

kshtsk: like:
```
if job_limit == 0 or job_limit == DEFAULT_JOB_LIMIT:
```

kshtsk: also, if we use --limit options, we do not need to check jobs number to schedule because it will not be generated more than the default using existing code as far as I can see.

kshtsk:
@neha-ojha see another comments regarding existing `--limit` option

scripts/suite.py
idryomov: This is different -- it is used as a cut off for the number of (often similar) jobs to be scheduled.  For example, if a `workload/foo` facet has 10 different variations and you only want to schedule one of them, `--filter workload/foo --limit 1` would do that.

neha-ojha: The --limit option is highly dangerous for a large suite like rados, and not encouraged at all.

idryomov:
FWIW I think `--force-jobs` was better.  It matched `--force-priority` and didn't introduce potential confusion with `--limit`.  The only nit I had was the constant ;)  But I'll leave it to @kshtsk.

neha-ojha:
> FWIW I think `--force-jobs` was better. It matched `--force-priority` and didn't introduce potential confusion with `--limit`. The only nit I had was the constant ;) But I'll leave it to @kshtsk.

that was the initial idea behind `force-jobs`, though, `ignore-job-limit` is more self-explanatory, along with the extra description

neha-ojha:
> @neha-ojha see another comments regarding existing `--limit` option

I have responded :) I'd rather encourage people to use subset to contain the number of jobs and TBH, ignore-job-limit shouldn't be needed once a developer knows enough about a particular suite. Also, I have seen scheduled nightly runs  with > 500 jobs, due to improper subset selection/bugs, this would prevent such runs.

neha-ojha:
> > @neha-ojha see another comments regarding existing `--limit` option
> 
> I have responded :) I'd rather encourage people to use subset to contain the number of jobs and TBH, ignore-job-limit shouldn't be needed once a developer knows enough about a particular suite. Also, I have seen scheduled nightly runs with > 500 jobs, due to improper subset selection/bugs, this would prevent such runs.

here's an example https://pulpito.ceph.com/teuthology-2021-01-17_02:30:02-rados-nautilus-distro-basic-smithi/

kshtsk:
@neha-ojha oh... I see the difference, so we do not want to limit jobs, we just need to alert user about suite too large, sorry, the word 'limit' misguides really. So we need to avoid the 'limit' name then.
Let me think, what about, rename it to `--disable-jobs-assert` and JOBS_ASSERT_NUMBER=500 correspondingly?

scripts/suite.py
kshtsk: yeah, thanks, I agree it is a mess, the feature is in alerting not limiting

idryomov:
This is turning into a bit of bike-shed ;)  I would stay consistent with `--force-priority` and `check_priority():`
- `--force-jobs` or `--force-num-jobs` for the option name
- "Skip the number of jobs check. Use this to schedule more than 500 jobs." for the description
- avoid inventing the name for the constant and just hard-code 500 in `check_num_jobs()`

neha-ojha:
> @neha-ojha @idryomov oh... I see the difference, so we do not want to limit jobs, we just need to alert user about suite too large, sorry, the word 'limit' misguides really. So we need to avoid the 'limit' name then.
> Let me think, what about, rename it to `--disable-jobs-assert` and JOBS_ASSERT_NUMBER=500 correspondingly?

It is not "about suite too large", it is about how many jobs will be scheduled by issuing a teuthology-suite command with options like subset, filter etc.

Ideally, doing a --dry-run should tell users how many jobs they are about to schedule, but there are a few problems

- not everybody does a dry-run
- there are no guardrails at the moment to tell users what is too many (or too little) number of jobs
- scheduled nightly runs have both the above problems

So technically, this PR just introduces max number of jobs, a user will be allowed to schedule, by default, hence providing some sort of guardrail and avoid wastage of machine time because of user ignorance. Therefore, the original name, `MAX_JOBS` in the first version. 

In the example above https://github.com/ceph/teuthology/pull/1657#issue-682088773, one can easily schedule 637 jobs, without this patch and I believe it is not necessary for any suite in Ceph to run so many jobs.

I really want the functionality, don't care too much about what the flags are called, most suites other than than rados won't even need it ;)  

How about `MAX_JOBS_THRESHOLD` and `--force-num-jobs`?

kshtsk:
> This is turning into a bit of a bike-shed ;) I would stay consistent with `--force-priority` and `check_priority():`
> 
> * `--force-jobs` or `--force-num-jobs` for the option name
> * "Skip the number of jobs check. Use this to schedule more than 500 jobs." for the description
> * avoid inventing the name for the constant and just hard-code 500 in `check_num_jobs()`

well the `--force-jobs` or `--force-num-jobs` made me confused already, and still for me it is not clear what to *force* in sense of *jobs* or what to force about number jobs?

kshtsk:
> > @neha-ojha @idryomov oh... I see the difference, so we do not want to limit jobs, we just need to alert user about suite too large, sorry, the word 'limit' misguides really. So we need to avoid the 'limit' name then.
> > Let me think, what about, rename it to `--disable-jobs-assert` and JOBS_ASSERT_NUMBER=500 correspondingly?
> 
> It is not "about suite too large", it is about how many jobs will be scheduled by issuing a teuthology-suite command with options like subset, filter etc.

but from my point of view I meant exactly equivalence between "about how many jobs" and "about suite too large" when use teuthology-suite for scheduling.

> Ideally, doing a --dry-run should tell users how many jobs they are about to schedule, but there are a few problems
> 
> * not everybody does a dry-run
> * there are no guardrails at the moment to tell users what is too many (or too little) number of jobs
> * scheduled nightly runs have both the above problems
> 
> So technically, this PR just introduces max number of jobs, a user will be allowed to schedule, by default, hence providing some sort of guardrail and avoid wastage of machine time because of user ignorance. Therefore, the original name, `MAX_JOBS` in the first version.

I understand your point. This is why I think "alert" or "assert" are suitable.

> In the example above [#1657 (comment)](https://github.com/ceph/teuthology/pull/1657#issue-682088773), one can easily schedule 637 jobs, without this patch and I believe it is not necessary for any suite in Ceph to run so many jobs.
> 
> I really want the functionality, don't care too much about what the flags are called, most suites other than rados won't even need it ;)
> 
> How about `MAX_JOBS_THRESHOLD` and `--force-num-jobs`?

This is confusing as for me, because it is not clear what to force to, I believe you guys know details of implementation and you think it is natural but it is not.

kshtsk:
Well guys, leaving it up to you, do not want to block this tiny-mighty PR ;-)

scripts/suite.py
```diff
@@ -172,6 +172,11 @@
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
+ --disable-num-jobs-check     Skip the number of jobs check. By default,
+                              teuthology will not allow you to schedule more
+                              than JOBS_TO_SCHEDULE_THRESHOLD=500 jobs, because
```
kshtsk, line 177: you can pass the threshold via format name, like: `more than {jobs_threashold) jobs`,
giving it to format:
```
from teuthology.suite.run import Run
.
.
.
""".format(
    default_machine_type=config.default_machine_type,
    default_results_timeout=config.results_timeout,
    default_ceph_repo=defaults('--ceph-repo',
                            config.get_ceph_git_url()),
    default_suite_repo=defaults('--suite-repo',
                            config.get_ceph_qa_suite_git_url()),
    default_ceph_branch=defaults('--ceph-branch', 'master'),
    jobs_threshold=Run.JOBS_TO_SCHEDULE_THRESHOLD,
)

```

teuthology/suite/run.py
```diff
@@ -29,6 +29,7 @@
 class Run(object):
     WAIT_MAX_JOB_TIME = 30 * 60
     WAIT_PAUSE = 5 * 60
+    JOBS_TO_SCHEDULE_THRESHOLD = 500
```
kshtsk, line 32: JOB_THRESHOLD would be concise I guess :-)

scripts/suite.py
```diff
@@ -172,6 +172,11 @@
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
+ --disable-num-jobs-check     Skip the number of jobs check. By default,
```
kshtsk, line 175: god, I didn't want this :-D maybe `--disable-threshold`?

teuthology/suite/run.py
neha-ojha: I don't see an issue with a more explicit name, which is exactly what you wanted earlier :)

kshtsk (approved):

teuthology/suite/run.py
kshtsk: I prefer some kind of balance between explicit and cryptic, like concise :-D cheers

kshtsk: @neha-ojha anyways, seems like my comments are getting out of constructive way, go ahead and merge if the functionality is fine, I'll raise another follow up PR with improvement :-)

neha-ojha:
@susebot run deploy

teuthology/suite/run.py
neha-ojha: @kshtsk The tests are passing https://github.com/ceph/teuthology/pull/1657#issuecomment-878474666 and my local testing looks good as well https://github.com/ceph/teuthology/pull/1657#issue-682088773. I am going to go ahead and merge this version, please feel free to update the flag/variable names as per your liking. Thanks!

PR #1658, opened 2021-07-14
orchestra/remote: fix RemoteShell.mktemp()

rishabh-d-dave:
* Allow mktemp() writing empty files
* Change the default value of parameter "data" of the method "mktemp()" from None to empty string.

See commit message for rationale behind these changes.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -76,7 +76,7 @@ class RemoteShell(object):
 
         path = self.sh(args).strip()
 
-        if data:
+        if data is not None:
             self.write_file(path=path, data=data)
 
         return path
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -58,7 +58,7 @@ class RemoteShell(object):
 
         return self.sh(args).strip()
 
-    def mktemp(self, suffix=None, parentdir=None, data=None):
+    def mktemp(self, suffix=None, parentdir=None, data=''):
         """
         Make a remote temporary file.
 
@@ -68,16 +68,17 @@ class RemoteShell(object):
 
         Returns: the path of the temp file created.
         """
+        if not isinstance(data, str):
+            raise RuntimeError('value of parameter data is not str type')
+
         args = ['mktemp']
         if suffix:
             args.append('--suffix=%s' % suffix)
         if parentdir:
             args.append('--tmpdir=%s' % parentdir)
-
         path = self.sh(args).strip()
 
-        if data is not None:
-            self.write_file(path=path, data=data)
+        self.write_file(path=path, data=data)
 
         return path
 


kshtsk:
It is by design not to allow write data to tempfile if the empty data. There is no a bug.

kshtsk:
I am not following what are you fixing here.

kshtsk:
Really, if you want to make sure you've written an empty string to a file, just call write_file method with it exclusively.

kshtsk:
and I forgot to mention that, data is an optional parameter, and this PR is breaking the ability to make temp file without extra write_file operation, if there is a real need to throw an exception on empty data it should be handled on client-side, so please do not merge it.

rishabh-d-dave:
@kshtsk 
The idea of PR is to allow creating an empty temporary file in a single call. It can't be done right now without two subsequent calls to `mktemp()` and `write_file()`.

If there's any significance to skipping the call to `write_file()` in `mktemp()` (which I don't think there is), we add following code to `mktemp()` -

```
if data is not None:
    self.write_file(path=path, data=data)
```

kshtsk:
> @kshtsk
> The idea of PR is to allow creating an empty temporary file in a single call. It can't be done right now without two subsequent calls to `mktemp()` and `write_file()`.

I am not following what you mean, the mktemp is EXACTLY creating empty temporary file when it used in standard way, without extra arguments, either you provide _data_ argument with '' line or not providing it all. You do not need to call `write_file` to have the temporary file created empty.

Current version of mktemp is creating temporary file without calling write_file if you are not providing `data` argument, exactly like it was used before we added this argument as optional.

> 
> If there's any significance to skipping the call to `write_file()` in `mktemp()` (which I don't think there is), we add following code to `mktemp()` -

There is a big sense in "skipping", because we do not change the logic of original method. 
From the beginning it should not write any data, because mktemp should provide a single operation of using system _mktemp_ to give an idea of user what is going to be done on remote host. 
Read the function description carefully:
```
:param data: write data to the file if provided
```

You might remember we add `write_file` to make it possible to put data as an option, I would not approve this PR if you wanted to make it mandatory.

Me as a user would like to have possibilty to still call mktemp without calling write_file. User should control data agrument outside.

If you really want write_file mandatory, it is better to do on the client side, so there were no hidden logic, so the code can be self documented, what is wrong with this?
```
   t = remote_host.mktemp()
   remote_host.write_file(t, data)
```

If you really want this mandatory and force **write_file** sub-call, you could create a helper function for this purpose, for example, **mktemp_write**, or you can 

> 
> ```
> if data is not None:
>     self.write_file(path=path, data=data)
> ```

kshtsk:
Hm.... wait, you want to force write_data call when data=''? Ok, then it makes sense to add _if data is not None_ check instead of 'if data'. In this case you should not change default for data from None to empty str. 

The commit message should be updated for better description what was wrong.

teuthology/orchestra/remote.py
```diff
@@ -68,16 +68,17 @@ def mktemp(self, suffix=None, parentdir=None, data=None):
 
         Returns: the path of the temp file created.
         """
+        if not isinstance(data, str):
+            raise RuntimeError('value of parameter data is not str type')
+
```
kshtsk, line 73: I would probably like to left the possibility to write binary data too.

kshtsk (changes requested):
1) Fix 'if data' to check None to allow empty str written.
2) Still allow other data types but str.
3) Revert data default value to None.
4) Squash and rework commit message to reflect better the idea of fix.

kshtsk:
@rishabh-d-dave ping?

kshtsk:
Closing for now, since no response for more than a year.

PR #1660, opened 2021-07-15
suite: rename disable-num-jobs-check to job-threshold

kshtsk:
Rename --disable-num-jobs-check to --job-threshold:
- for shorter recallable name;
- to allow change threshold value via parameter;
- to allow define default threshold value in teuthology config.

Use `--job-threshold 0` to disable job threshold check.

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -242,3 +242,7 @@ Here is a sample configuration with many of the options set and documented::
     pelagos:
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Do not allow more than that many jobs in a single run by default.
+    # To disable this check use 0.
+    job_threshold: 500
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -172,11 +172,9 @@ Scheduler arguments:
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
- --disable-num-jobs-check     Skip the number of jobs check. By default,
-                              teuthology will not allow you to schedule more
-                              than JOBS_TO_SCHEDULE_THRESHOLD=500 jobs, because
-                              it is too high. Use this if you need to schedule
-                              more than 500 jobs.
+ --job-threshold <threshold>  Do not allow to schedule the run if the number
+                              of jobs exceeds <threshold>. Use 0 to allow
+                              any number [default: {default_job_threshold}].
 
 """.format(
     default_machine_type=config.default_machine_type,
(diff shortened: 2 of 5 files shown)


kshtsk:
@susebot run deploy

scripts/suite.py
```diff
@@ -172,11 +172,9 @@
...
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
- --disable-num-jobs-check     Skip the number of jobs check. By default,
-                              teuthology will not allow you to schedule more
-                              than JOBS_TO_SCHEDULE_THRESHOLD=500 jobs, because
-                              it is too high. Use this if you need to schedule
-                              more than 500 jobs.
+ --job-threshold <value>      Do not allow to schedule the suite if the number
+                              of jobs exceeds the given <value>. Use 0 to skip
+                              the check [default: {default_job_threshold}].
```
idryomov, line 177: Nit: The number of jobs in a suite is constant (in the hundreds of thousands for rados, I think).  What is being scheduled and where the number of jobs is variable is a run:
```suggestion
 --job-threshold <threshold>  Do not allow to schedule the run if the number
                              of jobs exceeds <threshold>. Use 0 for unlimited
                              [default: {default_job_threshold}].
```

docs/siteconfig.rst
```diff
@@ -242,3 +242,7 @@ Here is a sample configuration with many of the options set and documented::
     pelagos:
       endpoint: http://head.ses.suse.de:5000/
       machine_types: ['type1', 'type2', 'type3']
+
+    # Do not allow user to schedule that many jobs by default.
```
idryomov, line 246: Mentioning the user could be interpreted as some sort of per-user quota or limit.  Stress that it is per-run:
```suggestion
    # Do not allow more than that many jobs in a single run by default.
```

idryomov (commented):
Looks good to me.  I left a couple of language suggestions for your and Neha's consideration.

scripts/suite.py
kshtsk: That makes sense as for run, but we should avoid the 'unlimited' or 'limit' key words because there is --limit options which limits the number of jobs to schedule, i.e. it anyways schedule job but with less jobs.

idryomov:
Please squash these "Update ..." commits into the original commit.

idryomov (approved):

scripts/suite.py
```diff
@@ -172,11 +172,9 @@
                               in the output of teuthology-suite command. -1
                               for a random seed [default: -1].
  --force-priority             Skip the priority check.
- --disable-num-jobs-check     Skip the number of jobs check. By default,
-                              teuthology will not allow you to schedule more
-                              than JOBS_TO_SCHEDULE_THRESHOLD=500 jobs, because
-                              it is too high. Use this if you need to schedule
-                              more than 500 jobs.
+ --job-threshold <threshold>  Do not allow to schedule the run if the number
```
neha-ojha, line 175: nit: Do not allow scheduling a run

neha-ojha (approved):

PR #1661, opened 2021-07-25
.readthedocs: add .readthedocs.yml

tchaikov:
to address the missing document of
https://docs.ceph.com/projects/teuthology/en/latest/commands/list.html,

they require that teuthology cli tools accessible when building sphinx
document.


tchaikov:
before this change

- https://docs.ceph.com/projects/teuthology/en/latest/commands/teuthology.html

after this change

- https://ceph--1661.org.readthedocs.build/projects/teuthology/en/1661/commands/teuthology.html

kshtsk (approved):

PR #1664, opened 2021-08-02
DNM: orch/run: refactor and rename run.quote()

rishabh-d-dave:
* Refactor to add extra checks and simply the current definition of method.
* Rename since method name and docstring doesn't describe anything.

TODO: update codebase at ceph's qa/ as soon as the PR merges.
DNM, wait for https://github.com/ceph/teuthology/pull/1663.

### DIFF
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -62,7 +62,7 @@ class RemoteProcess(object):
         self.client = client
         self.args = args
         if isinstance(args, list):
-            self.command = quote(args)
+            self.command = convert_args_list_to_str(args)
         else:
             self.command = args
 
@@ -240,9 +240,11 @@ class Raw(object):
         return self.value == value
 
 
-def quote(args):
+def convert_args_list_to_str(args):
     """
-    Internal quote wrapper.
+    Convert list args to str args. While doing so, quote str instances to
+    disable special interpretation of characters and substitute Raw instances
+    with their actual values.
     """
     if isinstance(args, str):
         return args
@@ -430,9 +432,11 @@ def run(
         if transport:
             (host, port) = transport.getpeername()[0:2]
         else:
-            raise ConnectionLostError(command=quote(args), node=name)
+            raise ConnectionLostError(command=convert_args_list_to_str(args),
+                                      node=name)
     except socket.error:
-        raise ConnectionLostError(command=quote(args), node=name)
+        raise ConnectionLostError(command=convert_args_list_to_str(args),
+                                  node=name)
 
     if name is None:
         name = host
--- a/teuthology/orchestra/run.py
+++ b/teuthology/orchestra/run.py
@@ -244,19 +244,14 @@ def quote(args):
     """
     Internal quote wrapper.
     """
-    def _quote(args):
-        """
-        Handle quoted string, testing for raw charaters.
-        """
-        for a in args:
-            if isinstance(a, Raw):
-                yield a.value
-            else:
-                yield pipes.quote(a)
-    if isinstance(args, list):
-        return ' '.join(_quote(args))
-    else:
+    if isinstance(args, str):
         return args
+
+    assert isinstance(args, list)
+
+    args = [pipes.quote(a) if isinstance(a, str) else a.value for a in args]
+
+    return ' '.join(args)
 
 
 def copy_to_log(f, logger, loglevel=logging.INFO, capture=None, quiet=False):
--- a/teuthology/orchestra/test/test_run.py
+++ b/teuthology/orchestra/test/test_run.py
@@ -266,16 +266,16 @@ class TestRun(object):
 
 class TestQuote(object):
     def test_quote_simple(self):
-        got = run.quote(['a b', ' c', 'd e '])
+        got = run.convert_args_str_to_list(['a b', ' c', 'd e '])
         assert got == "'a b' ' c' 'd e '"
 
     def test_quote_and_quote(self):
-        got = run.quote(['echo', 'this && is embedded', '&&',
+        got = run.convert_args_str_to_list(['echo', 'this && is embedded', '&&',
                          'that was standalone'])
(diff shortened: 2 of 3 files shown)


teuthology/orchestra/run.py
```diff
@@ -240,24 +240,24 @@ def __eq__(self, value):
...
 
+    assert isinstance(args, list)
+
+    for a in args:
+        if isinstance(a, Raw):
+            a = a.value
+        else:
+            a = pipes.quote(a)
+
+    return ' '.join(_quote(args))
```
tchaikov, line 259: where is `_quote()` defined ?

rishabh-d-dave: I was supposed to remove the call to `_quote()` here. Thanks for pointing out!

teuthology/orchestra/run.py
```diff
@@ -240,24 +240,24 @@ def __eq__(self, value):
...
+    if isinstance(args, str):
         return args
 
+    assert isinstance(args, list)
+
+    for a in args:
+        if isinstance(a, Raw):
+            a = a.value
+        else:
+            a = pipes.quote(a)
```
tchaikov, line 257: this does not work:
```py
In [1]: a = ["a", "b"]

In [2]: for i in a:
   ...:     i = "c"
   ...:

In [3]: a
Out[3]: ['a', 'b']
```

please note, `a` defined by value not a reference.

tchaikov (changes requested):

kshtsk:
Why should we rename it? I skipped something?

kshtsk:
I just do not understand why should we rename the method just because documentation for it is not ideal? If you want to improve method description just change it.

rishabh-d-dave:
> Why should we rename it? I skipped something?

@kshtsk Someone who isn't well-acquainted with the teuthology.orchestra.run wouldn't know the context and just "quote" wouldn't tell much. More descriptive name would not only tell more about the method in brief and be easier to remember but it would also enable new users to reuse that method.

kshtsk:
> > Why should we rename it? I skipped something?
> 
> @kshtsk Someone who isn't well-acquainted with the teuthology.orchestra.run wouldn't know the context and just "quote" wouldn't tell much. More descriptive name would not only tell more about the method in brief and be easier to remember but it would also enable new users to reuse that method.

I don't agree, short name _quote_ is self explanatory, if newbie not be so lazy, he/she goes to the source code and read or use pydoc to read more about usage. So no need to rename.

teuthology/orchestra/run.py
```diff
@@ -240,24 +240,21 @@ def __eq__(self, value):
...
-                yield pipes.quote(a)
-    if isinstance(args, list):
-        return ' '.join(_quote(args))
-    else:
+    if isinstance(args, str):
         return args
 
+    assert isinstance(args, list)
+
+    args = [pipes.quote(a) if isinstance(a, str) else a.value for a in args]
```
kshtsk, line 254: Can we avoid mutability? In general is bad practice to reuse existing variables?
For the case, I guess it would be enough to return immediately without overriding args:
```
return ' '.join(pipes.quote(a) if isinstance(a, str) else a.value for a in args)
```

teuthology/orchestra/run.py
```diff
@@ -240,24 +240,21 @@ def __eq__(self, value):
         return self.value == value
 
 
-def quote(args):
+def convert_args_list_to_str(args):
```
kshtsk, line 243: I don't think this name is much better than, just 'quote', I had no difficulties to understand what is it, though, shorter name can probably suits better: args2str or args_to_str/args_as_str, or `quoted_args`. Or if we want to be even more certain about function (sarcasm) we need to rename to **quote_args_if_required_and_return_as_space_joined_str** :-D.

teuthology/orchestra/run.py
```diff
@@ -240,24 +240,21 @@ def __eq__(self, value):
...
     """
-    Internal quote wrapper.
+    Convert list args to str args. While doing so, quote str instances to
+    disable special interpretation of characters and substitute Raw instances
+    with their actual values.
```
kshtsk, line 247: This description is not precise enough. I suggest to rework it, for example, you do not say a word about the case if args, is not a list.

kshtsk:
Man, sometimes I have a feeling you want to make a PR with as many as possible of number of lines changes, this is not a good approach imho :-D

teuthology/orchestra/run.py
tchaikov: i concur with Kyr. IMHO, the readability is not improved with this change.  i took more time parsing this oneliner. and `quote()` is a better name, i think.

tchaikov (changes requested):

teuthology/orchestra/run.py
tchaikov: i think `quote()` is an internal method as the previous doc string explains. its only consumer in production resides in this very same file. the other 4 consumers are tests. what's the motivation to improve it?

teuthology/orchestra/run.py
rishabh-d-dave: > i concur with Kyr. IMHO, the readability is not improved with this change. i took more time parsing this oneliner. and quote() is a better name, i think.

Okay, in that case it's best to close the PR.

teuthology/orchestra/run.py
rishabh-d-dave: > i think quote() is an internal method as the previous doc string explains. its only consumer in production resides in this very same file. the other 4 consumers are tests. what's the motivation to improve it?

It has been used in vstart_runner.py as well - https://github.com/ceph/ceph/blob/master/qa/tasks/vstart_runner.py#L441.

Motivation: currently, run.quote() is an teuthology specific extension of pipes.quote() (since it deals with Raw instances and list args as well) and there's no hint of it in the name or docstring. If that wasn't the case, it would be easier to know and therefore use it.

kshtsk:
Closing after couple of years unattended.

PR #1650, opened 2021-05-27
Removing Beanstalk from teuthology

amathuria:
The following changes support the removal of Beanstalk from Teuthology.
In place of Beanstalk, we will now be using Paddles for queue management in Teuthology.
This PR has the corresponding changes for the paddles PR: https://github.com/ceph/paddles/pull/94/files.

The changes include:
 1. Removing all beanstalk related code
 2. Teuthology scheduler and dispatcher using Paddles queue for scheduling and dispatching jobs
 3. Adding support for Paddles queue management
 4. Additional functionality of being able to change the priority of Teuthology jobs in the queued state in the teuthology-queue command

Signed-off-by: Aishwarya Mathuria <amathuri@redhat.com>

PR #1665, opened 2021-08-04
Added dockerfile + docker-compose script dev setup

kamoltat:
The purpose of this PR is to allow developers to set up a dev environment for Teuthology. This includes setting up the containers for Postgres, Paddles, Pulpito, Beanstalk, Teuthology. For now, this dev-setup will rely on sepia lab cluster
for test nodes.

The changes include:

1. Adding a Docker file for Teuthology
2. README.md file for the setup guide
3. Adding a docker-compose.yml for building containers
4. start.sh script for automating the processing of cloning repositories, building Docker Images and executing docker-compose.

This PR is also pending for:

- https://github.com/ceph/teuthology/pull/1650
   -  Delete first_in_suite_job and last_in_suite_job
   - clean_config() in dispatcher/.__init__.py skip key `status`, not include in .yaml files
- https://github.com/ceph/paddles/pull/94

to be merged, as currently, we use these branches
as images for paddles and pulpito.


docs/docker-compose/start.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+git clone https://github.com/ceph/paddles.git
+cd paddles
+git fetch origin pull/94/head:wip-amathuria-removing-beanstalkd
+git checkout wip-amathuria-removing-beanstalkd
+cd ../
+git clone https://github.com/ceph/pulpito.git
+git clone https://github.com/ceph/teuthology.git
+cd teuthology
```
zmc, line 9: What's the purpose of cloning another copy of teuthology?

docs/docker-compose/start.sh
```diff
@@ -0,0 +1,12 @@
...
+git fetch origin pull/1650/head:wip-amathuria-replace-beanstalkd-paddles
+git checkout wip-amathuria-replace-beanstalkd-paddles
+docker-compose up
```
zmc, line 12:
```suggestion
docker-compose up --build
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    git qemu-utils python3-dev libssl-dev \
+    python3-pip python3-virtualenv vim \
+    libev-dev libvirt-dev libffi-dev \
+    libyaml-dev lsb-release && apt-get \ 
+    clean all
+RUN echo -e 'lock_server: http://paddles:8080\n\
+results_server: http://paddles:8080\n\
+queue_host: 0.0.0.0\n\
+queue_port: 11300\n\
+teuthology_path: ./' >> ~/.teuthology.yaml
```
zmc, line 16: Might be worth storing this as a file in this directory and using `COPY` or `ADD` to include it in the image

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN mkdir archive_dir
+RUN mkdir log
+CMD  ./bootstrap && ./virtualenv/bin/teuthology-suite \
+-v --ceph-repo https://github.com/ceph/ceph.git \
+--suite-repo https://github.com/ceph/ceph.git \
+-c master -m smithi --subset 9000/100000 --limit 1 \
+--suite rados:mgr -k distro --filter 'tasks/progress' \
+--suite-branch master -p 75 --force-priority -n 100 \
+&& tail -f /dev/null
```
zmc, line 25: This is long enough that it makes me wonder if we'd want to use an entry-point script like some other projects do

zmc (changes requested):
I'm having problems getting this to work. Oddly enough, if I run `docker compose up --build` from the _project root_, everything starts up (but no jobs are scheduled). If I run `start.sh` or just `docker-compose build` from `docs/docker-compose`, I end up with this error message:
`failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount136320793/Dockerfile: no such file or directory`

docs/docker-compose/start.sh
zmc: @kamoltat this got things working for me. there are tons of errors after things come up, but that may be unrelated. for reference, they seem to be related to:

```
postgres_1   | 2021-08-11 21:51:54.922 UTC [114] ERROR:  relation "job_id_sequence" does not exist at character 16
```

kamoltat: @amathuria This is the error I was telling you about, think we might need to change `job_id_sequence` to what is saved in psql: `jobs_id_seq`.

amathuria: Ah okay I see. I'm trying this out now so I'll let you know how we can fix it.

amathuria: It's fixed now.

kamoltat: thank you!

Dockerfile
zmc: Especially since I've just realized that if you want to change the teuthology command that gets run, you have to rebuild the container image.

docs/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+        - postgres
+    healthcheck:
+        test: ["CMD", "curl", "-f", "http://0.0.0.0:8080"]
+        timeout: 5s
+        interval: 10s
+        retries: 2
+    ports: 
+        - 8080:8080
+  pulpito:
+    build: ./pulpito
```
zmc, line 39:
```suggestion
    image: quay.io/ceph-infra/pulpito
```

docs/docker-compose/start.sh
kamoltat: Needed to use #1650 version of technology to run dispatcher without beanstalk.

Dockerfile
kamoltat: Yea just adding jobs to the queue for initial testing

zmc (approved):
Thanks for the updates @kamoltat!

This gets us a lot closer to being able to test changes to the full teuthology stack locally. As you know I've got some work in progress on top of this, so I'll merge this PR as-is now and we can follow-up on it next week.

PR #1673, opened 2021-09-13
Fix dictionary update in try_push_job_info()

kamoltat:
the function `try_push_job_info()` is not
updating `job_info` dictionary properly since
we want to update `job_info` with `extra_info`,
however, in lines 498 and 499 we are assigning
`job_info` to a copy of `extra_info` and updating
`job_info` with `job_config` which is incorrect.
Instead, we should assign `job_info` with
a copy of `job_config` and update `job_info` with
`extra_info`

### DIFF
--- a/teuthology/report.py
+++ b/teuthology/report.py
@@ -495,8 +495,8 @@ def try_push_job_info(job_config, extra_info=None):
     job_id = job_config['job_id']
 
     if extra_info is not None:
-        job_info = extra_info.copy()
-        job_info.update(job_config)
+        job_info = job_config.copy()
+        job_info.update(extra_info)
     else:
         job_info = job_config
 


kamoltat:
@zmc Just want to confirm with you that this makes sense since you initially wrote this part a long time ago and there weren't really any issues until I tested out my docker-compose script.

zmc:
@kamoltat Conceptually you're definitely right here; `extra_info` should override `job_config`. There's only one place where this is used, however, and that's _potentially_ updating the `failure_reason` in case of a job timeout. I'm honestly unsure which behavior we want there.

Out of curiosity, what problem did this cause with the compose deployment?

kamoltat:
> @kamoltat Conceptually you're definitely right here; `extra_info` should override `job_config`. There's only one place where this is used, however, and that's _potentially_ updating the `failure_reason` in case of a job timeout. I'm honestly unsure which behavior we want there.
> 
> Out of curiosity, what problem did this cause with the compose deployment?

@zmc it is not used in the current master of technology but with #1650, it is used in https://github.com/ceph/teuthology/pull/1650/files#diff-518b3a856fe844adcc19187dade87bd07738e36d20a528eef4834be57b0ba4d0R116

zmc:
@kamoltat that makes sense, thanks! Would you mind rebasing and repushing just to get the unit tests to run?

zmc:
@kamoltat tests failed because of a dependency issue on fudge (which is old and bad) so I rewrote the tests that used it. Once #1677 is merged, a rebase should clear things up.

zmc (approved):

PR #1676, opened 2021-09-15
Add GitHub Actions definition for CI

zmc:


zmc:
Second commit takes CI time down from 29m to 4m.

kshtsk (approved):
Looks great, however, do we really need to run tests on push?

zmc:
Unit tests? I'd argue yes - we have decent unit test coverage and can prevent regressions this way

kshtsk:
> Unit tests? I'd argue yes - we have decent unit test coverage and can prevent regressions this way

No, I mean, these checks will run for PR, and later when they merged to master (on push). I was asking if want this, I don't object, just asking

zmc:
Oh! I just misunderstood. I agree, we don't need to run tests on push. I'll remove that. Thanks!

kshtsk (approved):

PR #1677, opened 2021-09-16
tests: Drop fudge

zmc:
Use mock instead.


zmc:
fudge somehow started causing CI to fail. It's old and awful so let's just stop using it.

kshtsk (approved):

PR #1678, opened 2021-09-22
task/ansible: move ansible logs to ansible.log

kamoltat:
Currently ansible tasks are logging to
teuthology.log. We want to separate this to
it's own log files for better readability.

### DIFF
--- a/teuthology/task/ansible.py
+++ b/teuthology/task/ansible.py
@@ -15,8 +15,8 @@ from teuthology.repo_utils import fetch_repo
 
 from teuthology.task import Task
 
-log = logging.getLogger(__name__)
-
+log = logging.Logger(__name__)
+log.setLevel(logging.INFO)
 
 class LoggerFile(object):
     """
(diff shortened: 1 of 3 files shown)


ideepika:
@kamoltat can you investigate the CI failures locally, I think the ansible tests might need some changes.

you can do that using `./virtualenv/bin/tox` from root dir

teuthology/task/ansible.py
```diff
@@ -278,7 +280,7 @@ def execute_playbook(self, _logfile=None):
         command = ' '.join(args)
         log.debug("Running %s", command)
 
-        out_log = self.log.getChild('out')
+        out_log = log
```
ideepika, line 283: maybe we can directly use log?

ideepika (approved):
looks good overall! @kamoltat have you also tested it on interactive mode?

kamoltat:
Tested this PR with a real ceph job scheduled using local teuth set up. The log separation works fine, despite the fact that there were some issues with Ansible on my local teuth setup. But, I am pretty certain that the issue doesn't affect this change or vice-versa.  Just to make sure, requesting @zmc for review

kamoltat:
Update: I resolved the ansible issue, which was to do with my local setup for ansible inventory, preventing some of the ansible tasks from completing. Tested it again, and this works just fine.

zmc (approved):

PR #1680, opened 2021-09-30
Enhance aarch64 platform compatibility

Zhiwei-Dai:
modified:   teuthology/packaging.py

Signed-off-by: Dai Zhiwei <daizhiwei3@huawei.com>

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -20,6 +20,7 @@ from teuthology.exceptions import (VersionNotFoundError, CommitNotFoundError,
 from teuthology.misc import sudo_write_file
 from teuthology.orchestra.opsys import OS, DEFAULT_OS_VERSION
 from teuthology.orchestra.run import Raw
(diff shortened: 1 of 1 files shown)


jdurgin (dismissed):

liewegas:
@kshtsk ping?

kshtsk:
I don't understand why this is required? Could someone give some explanation?

teuthology/packaging.py
```diff
@@ -486,7 +487,7 @@ def _init_from_config(self):
         """
         Initializes the class from a teuthology job config
         """
-        self.arch = self.job_config.get('arch', 'x86_64')
+        self.arch = util.get_arch(self.job_config.get("machine_type"))
```
kshtsk, line 490: This looks not good for me, because the util.get_arch makes a request to the lock server, which weird for constructor method from `job_config`. We should not be bound on lock server here at all.

PR #1681, opened 2021-09-30
orch/run: Making teuthology a better detective

kamoltat:
Currently when a unittest in RemoteProcess
fails it will throw a CommandFailure with
no information regarding which unittest
failed.

This PR will enable the CommandFailure to scan
for unittest failures in teuthology.log
and output them to paddles.

Currently only supports these unittests:
- gtest c++
- python nose

### DIFF
--- a/teuthology/exceptions.py
+++ b/teuthology/exceptions.py
@@ -50,11 +50,14 @@ class CommandFailedError(Exception):
     """
     Exception thrown on command failure
     """
-    def __init__(self, command, exitstatus, node=None, label=None):
+    def __init__(
+        self, command, exitstatus, node=None, label=None, unit_test_error=None
+        ):
         self.command = command
         self.exitstatus = exitstatus
         self.node = node
         self.label = label
+        self.unit_test_error = unit_test_error
 
     def __str__(self):
         prefix = "Command failed"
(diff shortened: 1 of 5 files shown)


kamoltat:
https://pulpito.ceph.com/ksirivad-2021-12-01_22:29:48-rgw:sts-wip-skoduri-dbstore-mp-distro-basic-smithi/

zmc (changes requested):
I really like the idea of this feature, but I'm concerned about the implementation; teuthology logfiles can get into the multi-gigabyte range, and as-is this would cause each line to be evaluated against two regexes each time any command fails to execute, even if the caller catches the exception.

An easy way to mitigate this would be to make the feature opt-in. This could be done by adding a flag to `run()` that controls whether to scan after a given command fails. It could also be done by creating a helper function specifically for running unit tests that runs the tests, catches any `CommandFailedError`, and in the case of a unit-test failure raises its _own_ unit-test-specific exception.

A niche advantage to the second approach is that a helper could remember the last line it scanned each time it sees an exception, and avoid having to re-scan lines on subsequent exceptions.

kshtsk:
Parsing logs during core exception handling is in general not a good idea. Especially if we do that for any failed command, even if we know that we are not running unittest and we expect the command to be failed. If you really want to parse logs from unittest running remotely, it is better to do with separate utility wrapper runner and the decision should be taken on the caller side.
As an idea, for example, maybe there can be a parsers class for each unittest type, and there can be a list of parsers passed to the run method as a parameter to check the stream.

kshtsk (changes requested):
See the comment above.

kamoltat:
Work being done in https://github.com/ceph/teuthology/pull/1768 instead

PR #1683, opened 2021-10-07
tasks/kernel: add 'hwe' config flag for ubuntu distro hwe kernel

liewegas:
The hwe kernel supports nvme_loop, but the non-hwe kernel does not.

I don't want to futz with the 'distro' moniker (although that is another
valid approach) because only some tests need hwe, and I can imagine a
situation where we want to run tests on both kernels.  Note that this
flag doesn't rule out adding support for something like
'-k distro-hwe' later.


kshtsk:
@susebot run deploy

PR #1687, opened 2021-10-15
dispatcher/supervisor: Mark machine down if reimage error >= 10x

kamoltat:
Introduce a new feature where we mark the machine down
if we found more than 10 consecutive reimaging failure
when performing teuthology tests.

### DIFF
--- a/teuthology/dispatcher/supervisor.py
+++ b/teuthology/dispatcher/supervisor.py
@@ -3,7 +3,9 @@ import os
 import subprocess
 import time
 import yaml
+import requests
 
+from urllib.parse import urljoin
 from datetime import datetime
 
 import teuthology
@@ -151,6 +153,47 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
     if 'targets' in job_config:
         unlock_targets(job_config)
 
+def failure_is_reimage(failure_reason):
+    if not failure_reason:
+        return False
+    reimage_failure = "Error reimaging machines:"
+    if reimage_failure in failure_reason:
+        return True
+    else:
+        return False
+
+def check_for_reimage_failures_and_mark_down(targets, count=10):
+    # Grab paddles history of jobs in the machine
+    # and count the number of reimaging errors
+    # if it fails N times then mark the machine down
+    base_url = teuth_config.results_server
+    for k, _ in targets.items():
+        machine = k.split('@')[-1]
+        url = urljoin(
+                base_url,
+                '/nodes/{0}/jobs/?count={1}'.format(
+                machine, count)
+        )
+        resp = requests.get(url)
+        jobs = resp.json()
+        if len(jobs) < count:
+            continue
+        reimage_failures = list(filter(
+            lambda j: failure_is_reimage(j['failure_reason']),
+            jobs
+        ))
+        if len(reimage_failures) < count:
+            continue
+        # Mark machine down
+        machine_name = shortname(k)
+        teuthology.lock.ops.update_lock(
+	    machine_name,
+	    description='reimage failed {0} times'.format(count),
+	    status='down',
+	)
+        log.error(
+            'Reimage failed {0} times ... marking machine down'.format(count)
+        )
 
 def reimage(job_config):
     # Reimage the targets specified in job config
@@ -166,6 +209,8 @@ def reimage(job_config):
         # Reimage failures should map to the 'dead' status instead of 'fail'
         report.try_push_job_info(ctx.config, dict(status='dead', failure_reason='Error reimaging machines: ' + str(e)))
         nuke(ctx, True)
+        # Machine that fails to reimage after 10 times will be marked down
+        check_for_reimage_failures_and_mark_down(targets)
         raise
     ctx.config['targets'] = reimaged
     # change the status to running after the reimaging process
--- a/teuthology/dispatcher/test/test_reimage_error_mark_machine_down.py
+++ b/teuthology/dispatcher/test/test_reimage_error_mark_machine_down.py
@@ -1,0 +1,104 @@
+from teuthology.dispatcher import supervisor
+from unittest.mock import patch
+
+class TestCheckReImageFailureMarkDown(object):
+    def setup(self):
+        self.the_function = supervisor.check_for_reimage_failures_and_mark_down
+
+    def create_n_out_of_10_reimage_failed_jobs(self, n):
+        ret_list = []
+        for i in range(n):
+            obj1 = {
+              "failure_reason":"Error reimaging machines: Manually raised error"
+              }
+            ret_list.append(obj1)
+        for j in range(10-n):
+            obj2 = {"failure_reason":"Error something else: dummy"}
+            ret_list.append(obj2)
+        return ret_list
+
+    @patch('teuthology.dispatcher.supervisor.shortname')
+    @patch('teuthology.lock.ops.update_lock')
+    @patch('teuthology.dispatcher.supervisor.requests')
+    @patch('teuthology.dispatcher.supervisor.urljoin')
+    @patch('teuthology.dispatcher.supervisor.teuth_config')
+    def test_one_machine_ten_reimage_failed_jobs(
+        self,
+        m_t_config,
+        m_urljoin,
+        m_requests,
+        mark_down,
+        shortname
+        ):
+        targets = {'fakeos@rmachine061.front.sepia.ceph.com': 'ssh-ed25519'}
+        m_requests.get.return_value.json.return_value = \
+            self.create_n_out_of_10_reimage_failed_jobs(10)
+        shortname.return_value = 'rmachine061'
+        self.the_function(targets)
+        assert mark_down.called
+
+    @patch('teuthology.dispatcher.supervisor.shortname')
+    @patch('teuthology.lock.ops.update_lock')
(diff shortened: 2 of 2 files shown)


teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,45 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
     if 'targets' in job_config:
         unlock_targets(job_config)
 
+def mark_machine_down(machine_name, count):
```
zmc, line 156: I think the name of this function is too generic considering the implementation is very specific to reimage failures. I think it's okay to keep this inside the main function

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,45 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+    teuthology.lock.ops.update_lock(machine_name,
+                                    description='reimage failed {0} times'.format(count),
+                                    status='down',
+                                    )
+    log.exception(
+        'Reimage failed {0} times ... marking machine down'.format(count)
+    )
```
zmc, line 163: `log.exception` isn't really appropriate here; I think `log.error` is pretty sensible

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,45 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+def failure_is_reimage(failure_reason):
+    if not failure_reason:
+        return False
+    reimage_failure = "Error reimaging machines:"
+    if reimage_failure in failure_reason:
+        return True
+    else:
+        return False
+
+def check_for_failures_and_mark_down(targets, count=10):
```
zmc, line 174: Even though it's really long, I think `check_for_reimage_failures_and_mark_down` is a little better

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,45 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+                '/nodes/{0}/jobs/?count={1}'.format(
+                machine, count)
+        )
+        resp = requests.get(url)
+        jobs = resp.json()
+        if len(jobs) < count:
+            return
+        for job in jobs:
+            if not failure_is_reimage(job['failure_reason']):
+                break
```
zmc, line 192: This `break` will only step out of this inner loop, so healthy nodes would get reimaged too

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,45 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+        machine = k.split('@')[-1]
+        url = urljoin(
+                base_url,
+                '/nodes/{0}/jobs/?count={1}'.format(
+                machine, count)
+        )
+        resp = requests.get(url)
+        jobs = resp.json()
+        if len(jobs) < count:
+            return
```
zmc, line 189: This will skip the remaining nodes

teuthology/dispatcher/supervisor.py
zmc: It might make more sense to do something like:
```
reimage_failures = list(filter(
    lambda j: failure_is_reimage(j['failure_reason']),
    jobs
))
if len(reimage_failures) < count:
    break
```

teuthology/test/test_reimage_error_mark_machine_down.py
```diff
@@ -0,0 +1,75 @@
+from teuthology.dispatcher import supervisor
```
zmc, line 1: Let's actually put this file in `teuthology/dispatcher/test/`

teuthology/test/test_reimage_error_mark_machine_down.py
```diff
@@ -0,0 +1,75 @@
...
+from unittest.mock import patch
+import unittest
```
zmc, line 3: I'm curious about the decision to use `unittest` over `py.test` here

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,46 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+        if len(jobs) < count:
+            continue
+        reimage_failures = list(filter(
+            lambda j: failure_is_reimage(j['failure_reason']),
+            jobs
+        ))
+        if len(reimage_failures) < count:
+            continue
+        # Mark machine down
+        machine_name = machine.split('.')[0]
```
zmc, line 188: This shouldn't be necessary

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,46 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+        teuthology.lock.ops.update_lock(machine_name,
+                                        description='reimage failed {0} times'.format(count),
+                                        status='down',
+                                       )
```
zmc, line 192:
```suggestion
        teuthology.lock.ops.update_lock(
            machine_name,
            description='reimage failed {0} times'.format(count),
            status='down',
        )
```

teuthology/test/test_reimage_error_mark_machine_down.py
```diff
@@ -0,0 +1,75 @@
...
+        with patch("teuthology.lock.ops.update_lock") as mark_down:
+            targets = {'fakeos@rmachine061.front.sepia.ceph.com': 'ssh-ed25519',
+                       'fakeos@rmachine179.back.sepia.ceph.com': 'ssh-ed45333'
+                      }
+            m_requests.get.return_value.json.side_effect = [
+                        self.create_n_out_of_10_reimage_failed_jobs(0),
+                        self.create_n_out_of_10_reimage_failed_jobs(10)
+                        ]
+            self.the_function(targets)
+            self.assertEqual(mark_down.call_count, 1)
```
zmc, line 75: For this test we should also verify the _correct_ machine was marked down. With `py.test` it's as simple as:
`assert mark_down.call_args_list[0][0][0].startswith('rmachine179')`

zmc (changes requested):
I really appreciate the tests! This is nearly done; just some style and test feedback.

teuthology/test/test_reimage_error_mark_machine_down.py
```diff
@@ -0,0 +1,75 @@
...
+              "failure_reason":"Error reimaging machines: Manually raised error"
+              }
+            ret_list.append(obj1)
+        for j in range(10-n):
+            obj2 = {"failure_reason":"Error something else: dummy"}
+            ret_list.append(obj2)
+        return ret_list
+
+    @patch('teuthology.dispatcher.supervisor.requests')
+    @patch('teuthology.dispatcher.supervisor.urljoin')
```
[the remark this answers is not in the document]
kamoltat: we need the urljoin because the bsae_url in teuth_config is mocked and we the urljoin() will throw an error if we don't mock it

teuthology/test/test_reimage_error_mark_machine_down.py
```diff
@@ -0,0 +1,75 @@
...
+    @patch('teuthology.dispatcher.supervisor.teuth_config')
```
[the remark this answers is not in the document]
kamoltat: we are patching teuth_config since it cannot be guaranteed that we even will have a teuth_config when running this test

teuthology/dispatcher/supervisor.py
kamoltat: Think it is necessary since we are comparing the count of the failures to determine if we should mark the machines down or not as well as we do the split because we need only the machine_name to execute the markdown command.

teuthology/dispatcher/supervisor.py
kamoltat: done

teuthology/test/test_reimage_error_mark_machine_down.py
kamoltat: From what I see from other unit tests in teuthology, a majority of them use a mix of pytest and unittest. Especially, patch from unittest.mock. I just find unittest documentation easier to follow and it is easier to mock functions. I've tested the unittest with pytest command as well and it works.

teuthology/dispatcher/supervisor.py
```diff
@@ -151,6 +153,47 @@ def run_job(job_config, teuth_bin_path, archive_dir, verbose):
...
+def failure_is_reimage(failure_reason):
+    if not failure_reason:
+        return False
+    reimage_failure = "Error reimaging machines:"
+    if reimage_failure in failure_reason:
+        return True
+    else:
+        return False
+
+def check_for_reimage_failures_and_mark_down(targets, count=10):
```
kshtsk, line 165: looks like a poem

teuthology/test/test_reimage_error_mark_machine_down.py
kamoltat: Changed to use pytest, only using patching from unittest.mock

teuthology/dispatcher/supervisor.py
kamoltat: Changed to using `decanonicalize_hostname` in `teuthology.misc` for getting the `machine_name`

zmc (approved):

PR #1692, opened 2021-11-08
task/{kernel,clock}: parallelize per-role work

liewegas:

### DIFF
--- a/teuthology/task/clock.py
+++ b/teuthology/task/clock.py
@@ -30,8 +30,9 @@ def task(ctx, config):
     """
 
     log.info('Syncing clocks and checking initial clock skew...')
-    for rem in ctx.cluster.remotes.keys():
-        rem.run(
+
+    run.wait(
+        ctx.cluster.run(
             args = [
                 'sudo', 'systemctl', 'stop', 'ntp.service', run.Raw('||'),
                 'sudo', 'systemctl', 'stop', 'ntpd.service', run.Raw('||'),
@@ -50,22 +51,26 @@ def task(ctx, config):
                 'true'
             ],
             timeout = 360,
+            wait=False,
         )
+    )
 
     try:
         yield
 
     finally:
         log.info('Checking final clock skew...')
-        for rem in ctx.cluster.remotes.keys():
-            rem.run(
+        run.wait(
+            ctx.cluster.run(
                 args=[
                     'PATH=/usr/bin:/usr/sbin', 'ntpq', '-p', run.Raw('||'),
                     'PATH=/usr/bin:/usr/sbin', 'chronyc', 'sources',
                     run.Raw('||'),
                     'true'
-                    ],
-                )
+                ],
+                wait=False,
+            )
+        )
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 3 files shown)


teuthology/task/kernel.py
```diff
@@ -1230,105 +1333,12 @@ def task(ctx, config):
...
-        # enable or disable kdb if specified, otherwise do not touch
-        if role_config.get('kdb') is not None:
-            kdb[role] = role_config.get('kdb')
-
-    if need_install:
-        install_firmware(ctx, need_install)
-        download_kernel(ctx, need_install)
-        install_and_reboot(ctx, need_install, config)
-        wait_for_reboot(ctx, need_version, timeout, config)
+        t = threading.Thread(target=process_role, args=(ctx, config, timeout,
```
jdurgin, line 1338: you can use the parallel() construction to simplify this a little, and use gevent rather than native threads, e..g. https://github.com/ceph/teuthology/blob/master/teuthology/reimage.py#L54 - looks good though!

kshtsk:
@susebot run deploy

teuthology/task/kernel.py
```diff
@@ -1130,6 +1131,108 @@ def get_sha1_from_pkg_name(path):
     return sha1
 
 
+def process_role(ctx, config, timeout, role, role_config):
```
kshtsk, line 1134: I wonder if it make sense to leave the code of process after the task one so it is better seen what is changed.

teuthology/task/internal/syslog.py
```diff
@@ -169,5 +169,19 @@ def syslog(ctx, config):
...
+
+        log.info('Gathering journactl -b0...')
+        run.wait(
+            ctx.cluster.run(
+                args=[
+                    'sudo', 'journalctl', '-b0',
+                    run.Raw('|'),
+                    'gzip', '-9',
+                    run.Raw('>'),
+                    a'{archive_dir}/syslog/journalctl-b0.gz',
```
kshtsk, line 183: syntax error here

kshtsk: Can we stop using of positional argument in favour for script:
```
ctx.cluster.run(
   args=f'sudo jounrnalctl -b0 | gzip -9 > {archive_dir}/syslog/journalctl-b0.gz', wait=False)
```

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
Looks like system smoke test is failing, wait another result from susebot to make sure we have the issue with parallelization.

kshtsk:
Take a look at the log https://ceph-ci.suse.de/job/teuthology-deploy-ovh/465/artifact/logs/run/teuthology-job-2.html
Seems like parallelization makes the run failing.

kshtsk:
@susebot run deploy

kshtsk:
I suggest that `wait_for_reboot` is problematic, since it makes call to misc.reconnect, which happens for one role while another are still running.

kshtsk:
The misc.reconnect() should be used with `remotes` parameter or instead remote.reconnect `remote.reconnect()` should be used exclusively. I would prefer the latter.
Also the misc.reconnect() method should be moved from misc to `cluster` and refactored.

kshtsk (changes requested):
The `misc.reconnect()` should be used with remotes parameter or instead remote.reconnect `remote.reconnect()` should be used exclusively. I would prefer the latter.
Also it would great to. move the `misc.reconnect()` method from misc to cluster for context based usage.

kshtsk:
@susebot run deploy

teuthology/task/kernel.py
```diff
@@ -694,14 +694,14 @@ def wait_for_reboot(ctx, need_install, timeout, config, distro=False):
...
     while need_install:
-        teuthology.reconnect(ctx, timeout)
         for client in list(need_install.keys()):
             if 'distro' in str(need_install[client]):
                  distro = True
             log.info('Checking client {client} for new kernel version...'.format(client=client))
             try:
                 if distro:
                     (remote,) = ctx.cluster.only(client).remotes.keys()
+                    remote.reconnect(timeout=timeout)
```
kshtsk, line 704: Running reconnect for distro only does not look the same it was before.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

teuthology/task/clock.py
```diff
@@ -30,8 +30,9 @@ def task(ctx, config):
     """
 
     log.info('Syncing clocks and checking initial clock skew...')
-    for rem in ctx.cluster.remotes.keys():
-        rem.run(
+
+    run.wait(
+        ctx.cluster.run(
```
kshtsk, line 35: Suggestion: maybe instead of run.wait combination use the #1695 and have:
```
ctx.cluster.run(
  args=[...],
  timeout = 360,
  sync=True
)
```
Or we can make the ctx.cluster.run process remotes synchronous and wait by default.

kshtsk (approved):

PR #1649, opened 2021-05-21
Restrict build_complete check to x86_64

badone:
Without this restriction a failed arm64 build will result in the
container build reporting failure.

Fixes: https://tracker.ceph.com/issues/50922

Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

PR #1693, opened 2021-11-09
teuthology/packaging.py: fix build_complete: search for requested arch

dmick:
The workaround from https://github.com/ceph/teuthology/pull/1649 was
necessary because my original algorithm was faulty: when searching
through all the builds for a ref/sha1, one must match the arch
requested by the call to build_complete (in the Builder object);
that arch's presence in the shaman api/search result is not enough
of a match, as it can contain multiple arches in multiple states
of build success.  Only a failure *on the requested arch* should be
considered a "requested build not complete".

(note: this will still currently fail a request for a build whose
repo is complete but container build failed, as "build complete"
currently conflates those two statuses.  Teuthology does not
contain the information whether a build is being requested for
packages, containers, or both.)

Also add testing for build_complete().

Fixes: https://tracker.ceph.com/issues/53205

### DIFF
--- a/teuthology/packaging.py
+++ b/teuthology/packaging.py
@@ -977,6 +977,9 @@ class ShamanProject(GitbuilderProject):
         except VersionNotFoundError:
             return False
 
+        # self._result has status, project, flavor, distros, arch, and sha1
+        # restrictions, so the only reason for multiples should be "multiple
+        # builds of the same sha1 etc."; the first entry is the newest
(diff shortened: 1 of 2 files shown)


badone:
This looks good to me Dan but I'll defer to Zack to give the final OK as discussed.

liewegas (approved):

zmc (approved):
This looks great. Thanks for the thorough test!

PR #1694, opened 2021-11-12
tasks/internal/syslog: gather journalctl -b0

liewegas:


liewegas:
@susebot run deploy

PR #1695, opened 2021-11-13
orchestra/cluster: make run parallel and wait

kshtsk:
Make cluster.run able to execute a command in parallel on all
nodes and wait for completion.

In order to run a command in parallel on all nodes of the
cluster, instead of combination:
```
  run.wait(cluster.run(wait=False, args=...))
```
this patch makes it possible to do the same just by:
```
  cluster.run(parallel=True, args=...)
```
the `wait` arguments defaults to True in this case.

Also, the behavior of the method is not changed
when `wait` is set to False (like in the first example),
so it is supposed to be backward compatible.

### DIFF
--- a/teuthology/orchestra/cluster.py
+++ b/teuthology/orchestra/cluster.py
@@ -3,7 +3,7 @@ Cluster definition
 part of context, Cluster is used to save connection information.
 """
 import teuthology.misc
-
+from teuthology.orchestra import run
 
 class Cluster(object):
     """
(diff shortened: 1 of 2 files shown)


liewegas:
``sync`` doesn't seem like the right arg name to me.  How about ``parallel``?

kshtsk:
@susebot run deploy

kshtsk:
> `sync` doesn't seem like the right arg name to me. How about `parallel`?

@liewegas the change is made, the parameter changed to _parallel_.

kshtsk:
ping @liewegas @jdurgin

kshtsk:
ping @liewegas @jdurgin

kshtsk:
guys could you please review this again, no feedback for months

jdurgin (approved):

jdurgin:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

kshtsk:
ok, we are back now

PR #1696, opened 2021-11-16
teuthology/suite/matrix.py: Add item to assert msg

sebastian-philipp:
Give folks at least some hint about where things might
be broken:

```
AssertionError: Sum requires non-empty _submats: 0-distro
```

Signed-off-by: Sebastian Wagner <sewagner@redhat.com>


jdurgin (approved):
better than the existing lack of clues!

PR #1697, opened 2021-11-18
teuthology/beanstalk: workaround yaml load deprecation

kshtsk:
There is a deprecation error when beanstalk parser called:
```
  YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe
```
Since beanstalkc3 does not want to apply a fix for that
we can override yaml parser to suppress the error message
from the log.

Fixes: https://tracker.ceph.com/issues/44225


dmick:
s/deprication/deprecation/

jdurgin (approved):

PR #1698, opened 2021-11-18
orchestra/remote: throw exception on reconnect failed

kshtsk:
In run method if a reconnect required there is a remote.reconnect
method called which is supposed to return True or False.
On failure it is the best to raise an exception early,
to make it clear what was the original error instead of
waiting until somewhere later teuthology fails with
some undeclared variables.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -505,7 +505,8 @@ class Remote(RemoteShell):
         if not self.ssh or \
            not self.ssh.get_transport() or \
            not self.ssh.get_transport().is_active():
-            self.reconnect()
+            if not self.reconnect():
+                raise Exception(f'Cannot connect to remote host {self.shortname}')
         r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
         r.remote = self
         return r


teuthology/orchestra/remote.py
```diff
@@ -505,7 +505,8 @@ def run(self, **kwargs):
         if not self.ssh or \
            not self.ssh.get_transport() or \
            not self.ssh.get_transport().is_active():
-            self.reconnect()
+            if not self.reconnect():
+                raise Exception(f'Cannot connect to remote host {self.shortname}')
```
amathuria, line 509: Should this be 'Cannot reconnect to remote host'?

kshtsk: I think it should not, because the self.reconnect() is just an utility function and it is used in general for both initial connect and reconnect after connection got lost and transport is not available. And as in this case when run is executed for the first time it would be weird to see in log that it is going to reconnect during the first time.

amathuria: Makes sense

amathuria (approved):

teuthology/orchestra/remote.py
zmc: What about raising a [ConnectionLostError](https://github.com/ceph/teuthology/blob/master/teuthology/exceptions.py#L111) here?

kshtsk: > What about raising a [ConnectionLostError](https://github.com/ceph/teuthology/blob/master/teuthology/exceptions.py#L111) here?

@zmc see the comment above, it is not about connection lost only, but for establishing initial connection either.

zmc: I did see that, but if we're going to raise an exception we should give a way for callers to catch that specific exception. I think it should either be `ConnectionLostError` or a sub- or super-class of it. Or we could rename that exception to just `ConnectionError` I suppose

kshtsk:
@susebot run deploy

kshtsk:
> 

one fail and one pass, the fail is none for missing route:
```
2022-04-25T21:03:42.837 DEBUG:teuthology.orchestra.run.target-pr-1698-003:> route
2022-04-25T21:03:42.865 DEBUG:teuthology.orchestra.run:got remote process result: 127
2022-04-25T21:03:42.867 INFO:teuthology.orchestra.run.target-pr-1698-003.stderr:bash: route: command not found
```
which is unrelated issue.

teuthology/orchestra/remote.py
kshtsk: @zmc sorry I've merged already, but following your comment, in current code organisation and usage of the run method the differentiation of all cases will cause more just a single line and more hours for testing the code correctness, the goal of this patch is to fix the "silence". Also I am not sure it is really needed for caller of run, to see what exact it error it was, start connection or lost connection or what ever connection, what is the benefit?

PR #1699, opened 2021-11-18
misc/sh: do not strip output from left

kshtsk:
We want to see output of commands with nice formatting,
for example when there are yaml data


jdurgin (approved):
nice, didn't realize we were doing this

PR #1700, opened 2021-11-23
Stop using LogRecord.asctime (not stable); move to pytest 6.2.5

dmick:
See commit messages for details.  Pytest 6.2.5 exposed the use of asctime, which may not be present on a LogRecord.  Luckily it's much simpler to use crtime, and then we can update pytest.


zmc (approved):
Thanks for this, @dmick!

PR #1703, opened 2022-01-12
kill: Check if sudo works before using it

zmc:
Also, use sudo's -n flag to avoid the password prompt and resulting
delay.

Fixes: https://tracker.ceph.com/issues/53853

### DIFF
--- a/teuthology/kill.py
+++ b/teuthology/kill.py
@@ -179,12 +179,18 @@ def kill_processes(run_name, pids=None):
         log.info("No teuthology processes running")
     else:
         log.info("Killing Pids: " + str(to_kill))
+        may_need_sudo = \
+            psutil.Process(int(pid)).username() != getpass.getuser()
+        if may_need_sudo:
+            sudo_works = subprocess.Popen(['sudo', '-n', 'true']).wait() == 0
+            if not sudo_works:
+                log.debug("Passwordless sudo not configured; not using sudo")
+        use_sudo = may_need_sudo and sudo_works
         for pid in to_kill:
             args = ['kill', str(pid)]
             # Don't attempt to use sudo if it's not necessary
-            proc_user = psutil.Process(int(pid)).username()
-            if proc_user != getpass.getuser():
-                args.insert(0, 'sudo')
+            if use_sudo:
+                args = ['sudo', '-n'] + args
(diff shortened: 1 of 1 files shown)


kamoltat (approved):
nice and simple, just need to test before merging

teuthology/kill.py
```diff
@@ -177,8 +177,8 @@ def kill_processes(run_name, pids=None):
             # Don't attempt to use sudo if it's not necessary
             proc_user = psutil.Process(int(pid)).username()
             if proc_user != getpass.getuser():
-                args.insert(0, 'sudo')
-            subprocess.call(args)
+                args = ['sudo', '-n'] + args
+            subprocess.check_call(args)
```
kshtsk, line 181: Since we change to check_call, it means the kill command can be interrupted in the middle of sequential calls, like in killing several job, not even trying to kill the next job in a list. Do we really want this?

jdurgin: It looks like it would fail if one of the processes already exited - could we check for ability to run sudo before this loop instead?

zmc: Thanks for the feedback; I've re-implemented this, doing a check for `sudo`'s functionality before attempting to use it - and logging to inform the user of the choice we make if we can't use it.

jdurgin (approved):

teuthology/kill.py
```diff
@@ -172,12 +172,19 @@ def kill_processes(run_name, pids=None):
...
+            psutil.Process(int(pid)).username() != getpass.getuser()
+        if may_need_sudo:
+            sudo_works = subprocess.Popen(['sudo', '-n', 'true']).wait() == 0
+            if not sudo_works:
+                log.debug("Passwordless sudo not configured; not using sudo")
+        use_sudo = may_need_sudo and sudo_works
         for pid in to_kill:
             args = ['kill', str(pid)]
             # Don't attempt to use sudo if it's not necessary
             proc_user = psutil.Process(int(pid)).username()
```
jdurgin, line 185: flake8 reports proc_user is unused now

zmc: What I get for not re-running tests after fixing up that last commit. It's passing now!

jdurgin (approved):

PR #1704, opened 2022-01-14
teuthology/suite: create nested matrix subsets

batrick:
This is a draft for nested subset functionality in the qa suite. The general idea is to allow the `%` convolution operator to also subset the resulting matrix. This is done by specifying a number of divisions for the subset in the `%` file. Like so:

https://github.com/ceph/ceph/pull/44486/files#diff-c360eb6cfc66f8d7cfd253c15923075019e53b187ae790d803c11538de865a6c

This PR maps a matrix index range of `[0, Subset.size())` to the matrix it is taking a subset of, `[0, Matrix.size())`. To get full coverage, a random number is used to specify "which" subset to use. Contrast with the `--subset` argument to `teuthology-suite` which lets you specify which subset.

Rationale for this PR is https://github.com/ceph/ceph/pull/44486#issuecomment-1008997798 . I've already tested this change to teuthology with that PR and it works as advertised.

TODO:
[x] Write tests
[x] Create `teuthology-suite` argument to turn off nested subsets.


athanatos (approved):

teuthology/suite/__init__.py
```diff
@@ -2,6 +2,12 @@
 # by generating combinations of facets found in
 # https://github.com/ceph/ceph-qa-suite.git
 
+def open(*args, **kwargs):
+    """
+    wrapper of open for test hooks
+    """
+    return __builtins__.open(*args, **kwargs)
+
```
kshtsk, line 10: This is not needed, drop it.

teuthology/suite/build_matrix.py
```diff
@@ -113,7 +115,7 @@ def _build_matrix(path, mincyclicity=0, item=''):
         elif '%' in files:
             # convolve items
             files.remove('%')
-            with open(os.path.join(path, '%')) as f:
+            with teuthology.suite.open(os.path.join(path, '%')) as f:
```
kshtsk, line 118: Revert this line.

batrick: Tests fail without this change.

kshtsk:
This patches is missing documentation explaining the feature and how to use, more detailed in than in the PR description and in the better place for it.

batrick:
WIP doc: https://github.com/ceph/ceph/pull/44967

batrick:
@susebot run deploy

kshtsk:
> WIP doc: [ceph/ceph#44967](https://github.com/ceph/ceph/pull/44967)

Can we include generalized documentation (not hardly bounded to ceph) on this feature here in PR and in teuthology repo instead of ceph. It is more sense to have it here, and if from ceph pages need more explanation there should be link to the teuthology docs provided only.

batrick:
There's already no docs in teuthology.git for even `--subset`. I agree the docs should be here but I don't see a point adding only nested subsets in this PR.

scripts/results.py
```diff
@@ -1,19 +1,20 @@
...
+  -h, --help            show this help message and exit
+  -v, --verbose         be more verbose
+  --dry-run             Instead of sending the email, just print it
+  --email EMAIL         address to email test failures to
+  --timeout TIMEOUT     how many seconds to wait for all tests to finish
+                        [default: 0]
+  --archive-dir DIR     path under which results for the suite are stored
+  --name NAME           name of the suite
+  --subset SUBSET       subset passed to teuthology-suite
+  --seed SEED           random seed used in teuthology-suite
```
kshtsk, line 16: Why this is needed?

batrick: test_help failed without adjusting this whitespace.

teuthology/suite/build_matrix.py
```diff
@@ -7,7 +7,7 @@
 log = logging.getLogger(__name__)
 
 
-def build_matrix(path, subset=None, seed=None):
+def build_matrix(path, subset=None, no_nested_subset=False, seed=None):
```
kshtsk, line 10: Could you please add inline docs here about no_nested_subset?

scripts/results.py
batrick: The `--no-nested-subset` was longer than other options in the listing.

teuthology/suite/build_matrix.py
batrick: `nested_subset` should be the default so a negative option makes the most sense, to me.

teuthology/suite/build_matrix.py
```diff
@@ -53,31 +53,24 @@ def build_matrix(path, subset=None, seed=None):
             'Subset=%s/%s' %
             (str(subset[0]), str(subset[1]))
         )
+    if no_nested_subset:
+        log.info("no_nested_subset")
```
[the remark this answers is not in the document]
batrick: The surrounding config prints are also at `info`. I think it's helpful when scheduling to see the subset by default.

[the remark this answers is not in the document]
kshtsk: Yes, and it should go to debug either.

kshtsk:
@susebot run deploy

kshtsk:
@susebot run deploy

batrick:
Any objections to merging this? Don't want it to become stale. Needs reviewed by some of the teuthology maintainers.

teuthology/suite/build_matrix.py
kshtsk: Thanks for adding the description. But I didn't understand why `nested_subset=True` does not work for you?

kshtsk: Also, I wonder why haven't you put the new option at the end, so you do not need to change the code in hypothetic cases where build_matrix calls are not using it.

batrick: > Thanks for adding the description. But I didn't understand why nested_subset=True does not work for you?

`nested_subset` should by default be `True`. Boolean options on the CLI are generally indicated by a simple switch, i.e. `--foo` without arguments. So, if we want it to be true by default, then the option should be `--no-foo`. This is not uncommon in UNIX, e.g. see `--no-seek` for GNU `tar` or `--no-clobber` in GNU `mv`.

batrick: > Also, I wonder why haven't you put the new option at the end, so you do not need to change the code in hypothetic cases where build_matrix calls are not using it.

I fixed everywhere `build_matrix` is called. I felt it was sensible to group `subset` and `no_nested_subset` together in the argument list.

kshtsk: > > Thanks for adding the description. But I didn't understand why nested_subset=True does not work for you?
> 
> `nested_subset` should by default be `True`. Boolean options on the CLI are generally indicated by a simple switch, i.e. `--foo` without arguments. So, if we want it to be true by default, then the option should be `--no-foo`. This is not uncommon in UNIX, e.g. see `--no-seek` for GNU `tar` or `--no-clobber` in GNU `mv`.

I understand your preferences for gnu switches, but the matter is to make function parameters, so no sense to use `no_` prefix everywhere, i.e. the switch can stay named `--no-nested-subset` but in function it is better to change to strict logic for nested subset.

zmc:
@batrick I apologize for not having seen this sooner - I will give it a review tomorrow

vshankar:
> @batrick I apologize for not having seen this sooner - I will give it a review tomorrow

my apologies for not pinging you earlier on this (@batrick did point me to this pr).

zmc (approved):
Really happy to see this improvement! Implementation and tests look good to me. I do share @kshtsk's dismay about docs for this not living in this repo, but I'm happy that some exist at all and I appreciate what you've done there.

vshankar:
@zmc anything that holding back merge?

kshtsk:
> @zmc anything that holding back merge?
like rename `no_nested_subset` to `nested_subset` for the functions parameter (cli not required) and move it to the final position in method signatures, so you do not need to change api significantly?

vshankar:
> > @zmc anything that holding back merge?
> 
> like rename `no_nested_subset` to `nested_subset` for the function's parameter (cli not required) and move it to the final position in method signatures, so you do not need to change api significantly?

Am I correct in understanding that this is more of a minor code cleanup and cosmetic change? If yes, we can probably fix such things later, especially, since @batrick is on PTO until May'ish and we do not want to end up with this PR getting marked as stale.

batrick:
(trivial rebase)

zmc (approved):

PR #1711, opened 2022-02-14
Cluster: Add filter() function

zmc:


teuthology/orchestra/cluster.py
```diff
@@ -144,3 +144,13 @@ def exclude(self, *roles):
             if remote not in matches.remotes:
                 c.add(remote, has_roles)
         return c
+
+    def filter(self, func):
+        """
+        Return a cluster whose remotes are filtered by `func`.
```
kshtsk, line 150: Maybe giving here a usage example is a good idea.

zmc: Added

PR #1712, opened 2022-02-15
Remote: Add is_container property

zmc:

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -680,6 +680,15 @@ class Remote(RemoteShell):
         return self._is_vm
 
     @property
+    def is_container(self):
+        if not hasattr(self, '_is_container'):
+            self._is_container = not bool(self.run(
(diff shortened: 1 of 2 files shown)


teuthology/orchestra/remote.py
```diff
@@ -679,6 +679,15 @@ def is_vm(self):
             self._is_vm = teuthology.lock.query.is_vm(self.name)
         return self._is_vm
 
+    @property
+    def is_container(self):
+        if not hasattr(self, '_is_container'):
+            self._is_container = not bool(self.run(
+                args="ls /.dockerenv || ls /run/.containerenv",
```
kshtsk, line 686: ls?

zmc: Just checking for the existence of either file.

kshtsk: maybe `test` is better?

zmc: Sure, updatedd

jdurgin (approved):

PR #1713, opened 2022-02-15
task.ansible: Don't direct task's logging to file

zmc:
We only want output generated by ansible itself in ansible.log


kamoltat (approved):
make sense, LGTM.

PR #1717, opened 2022-02-24
selinux.py: Ignore both contexts for NetworkManager_dispatcher_t

djgalloway:


guits:
@djgalloway does that mean it will ignore it when it's `tcontext` or `scontext`?

djgalloway:
> @djgalloway does that mean it will ignore it when it's `tcontext` or `scontext`?

Yeah, the task just checks for these strings in the logs and ignores the warnings if present

guits (approved):

PR #1719, opened 2022-03-01
task/internal: open gzip file if core is compressed

kamoltat:
Added a case where if the core is compressed, then
we will try to open it as gzip and read the content
inside.

Also, add try and except when we are parsing
through the content to prevent the program from
crashing.

Create a unit test to test the function,
`fetch_binaries_for_coredumps()`

Fixes: https://tracker.ceph.com/issues/53206

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -12,6 +12,7 @@ import shutil
 import time
 import yaml
 import subprocess
+import tempfile
 
 import humanfriendly
 
@@ -318,7 +319,28 @@ def fetch_binaries_for_coredumps(path, remote):
             # 1422917770.7450.core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, \
             # from 'radosgw --rgw-socket-path /home/ubuntu/cephtest/apache/tmp.client.0/fastcgi_soc'
             log.info(f' core looks like: {dump_out}')
-            dump_program = dump_out.split("from '")[1].split(' ')[0]
+
+            if 'gzip' in dump_out:
+                try:
+                    log.info("core is compressed, try accessing gzip file ...")
+                    with gzip.open(dump_path, 'rb') as f_in, \
+                         tempfile.NamedTemporaryFile(mode='w+b') as f_out:
+                         shutil.copyfileobj(f_in, f_out)
+                         dump_info = subprocess.Popen(['file', f_out.name],
+                                                        stdout=subprocess.PIPE)
+                         dump_out = dump_info.communicate()[0].decode()
+                         log.info(f' core looks like: {dump_out}')
+                except Exception as e:
+                    log.info('Something went wrong while opening the compressed file')
+                    log.error(e)
+                    continue
+            try:
+                dump_program = dump_out.split("from '")[1].split(' ')[0]
+                log.info(f' dump_program: {dump_program}')
+            except Exception as e:
+                log.info("core doesn't have the desired format, moving on ...")
+                log.error(e)
+                continue
 
             # Find path on remote server:
             remote_path = remote.sh(['which', dump_program]).rstrip()
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -13,7 +13,7 @@ import time
 import yaml
 import subprocess
 import tempfile
-
+import re
 import humanfriendly
 
 import teuthology.lock.ops
@@ -335,7 +335,7 @@ def fetch_binaries_for_coredumps(path, remote):
                     log.error(e)
                     continue
             try:
-                dump_program = dump_out.split("from '")[1].split(' ')[0]
+                dump_program = re.findall("from '([^']+)'", dump_out)[0]
                 log.info(f' dump_program: {dump_program}')
             except Exception as e:
                 log.info("core doesn't have the desired format, moving on ...")
--- a/teuthology/task/tests/test_fetch_coredumps.py
+++ b/teuthology/task/tests/test_fetch_coredumps.py
@@ -1,0 +1,116 @@
+from teuthology.task.internal import fetch_binaries_for_coredumps
+from unittest.mock import patch, Mock
+import gzip
+import os
+
+class TestFetchCoreDumps(object):
+    class MockDecode(object):
+        def __init__(self, ret):
+            self.ret = ret
+            pass
+
+        def decode(self):
+            return self.ret
+
+    class MockPopen(object):
+        def __init__(self, ret):
+            self.ret = ret
+
+        def communicate(self, input=None):
+            return [TestFetchCoreDumps.MockDecode(self.ret)]
+
+    def setup(self):
+        self.the_function = fetch_binaries_for_coredumps
+        with gzip.open('file.gz', 'wb') as f:
+            f.write(b'Hello world!')
+        self.core_dump_path = "file.gz"
+        self.m_remote = Mock()
+        self.uncompressed_correct = self.MockPopen(
+            "ELF 64-bit LSB core file,"\
+            " x86-64, version 1 (SYSV), SVR4-style, from 'ceph_test_rados_api_io',"\
+            " real uid: 1194, effective uid: 1194, real gid: 1194,"\
+            " effective gid: 1194, execfn: '/usr/bin/ceph_test_rados_api_io', platform: 'x86_64'"
+        )
+        self.uncompressed_incorrect = self.MockPopen("ASCII text")
+        self.compressed_correct = self.MockPopen(
+            "gzip compressed data, was "\
+            "'correct.format.core', last modified: Wed Jun 29"\
+            " 19:55:29 2022, from Unix, original size modulo 2^32 3167080"
+        )
+
+        self.compressed_incorrect = self.MockPopen(
+            "gzip compressed data, was "\
+            "'incorrect.format.core', last modified: Wed Jun 29"\
+            " 19:56:56 2022, from Unix, original size modulo 2^32 11"
+        )
+
+    # Core is not compressed and file is in the correct format
+    @patch('teuthology.task.internal.subprocess.Popen')
+    @patch('teuthology.task.internal.os')
+    def test_uncompressed_correct_format(self, m_os, m_subproc_popen):
+        m_subproc_popen.side_effect = [
+                self.uncompressed_correct,
+                Exception("We shouldn't be hitting this!")
+            ]
+        m_os.path.join.return_value = self.core_dump_path
+        m_os.path.sep = self.core_dump_path
+        m_os.path.isdir.return_value = True
+        m_os.path.dirname.return_value = self.core_dump_path
+        m_os.path.exists.return_value = True
+        m_os.listdir.return_value = [self.core_dump_path]
+        self.the_function(None, self.m_remote)
+        assert self.m_remote.get_file.called
+
+    # Core is not compressed and file is in the wrong format
+    @patch('teuthology.task.internal.subprocess.Popen')
+    @patch('teuthology.task.internal.os')
+    def test_uncompressed_incorrect_format(self, m_os, m_subproc_popen):
+        m_subproc_popen.side_effect = [
+                self.uncompressed_incorrect,
+                Exception("We shouldn't be hitting this!")
+            ]
+        m_os.path.join.return_value = self.core_dump_path
+        m_os.path.sep = self.core_dump_path
+        m_os.path.isdir.return_value = True
+        m_os.path.dirname.return_value = self.core_dump_path
+        m_os.path.exists.return_value = True
+        m_os.listdir.return_value = [self.core_dump_path]
+        self.the_function(None, self.m_remote)
+        assert self.m_remote.get_file.called == False
+
+    # Core is compressed and file is in the correct format
+    @patch('teuthology.task.internal.subprocess.Popen')
+    @patch('teuthology.task.internal.os')
+    def test_compressed_correct_format(self, m_os, m_subproc_popen):
+        m_subproc_popen.side_effect = [
+                self.compressed_correct,
+                self.uncompressed_correct
+            ]
+        m_os.path.join.return_value = self.core_dump_path
+        m_os.path.sep = self.core_dump_path
+        m_os.path.isdir.return_value = True
+        m_os.path.dirname.return_value = self.core_dump_path
+        m_os.path.exists.return_value = True
+        m_os.listdir.return_value = [self.core_dump_path]
+        self.the_function(None, self.m_remote)
+        assert self.m_remote.get_file.called
+
+    # Core is compressed and file is in the wrong format
+    @patch('teuthology.task.internal.subprocess.Popen')
+    @patch('teuthology.task.internal.os')
+    def test_compressed_incorrect_format(self, m_os, m_subproc_popen):
+        m_subproc_popen.side_effect = [
+                self.compressed_incorrect,
+                self.uncompressed_incorrect
+            ]
+        m_os.path.join.return_value = self.core_dump_path
+        m_os.path.sep = self.core_dump_path
+        m_os.path.isdir.return_value = True
+        m_os.path.dirname.return_value = self.core_dump_path
+        m_os.path.exists.return_value = True
+        m_os.listdir.return_value = [self.core_dump_path]
+        self.the_function(None, self.m_remote)
+        assert self.m_remote.get_file.called == False
+
+    def teardown(self):
+        os.remove(self.core_dump_path)


teuthology/task/internal/__init__.py
```diff
@@ -318,7 +318,11 @@ def fetch_binaries_for_coredumps(path, remote):
             # 1422917770.7450.core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, \
             # from 'radosgw --rgw-socket-path /home/ubuntu/cephtest/apache/tmp.client.0/fastcgi_soc'
             log.info(f' core looks like: {dump_out}')
-            dump_program = dump_out.split("from '")[1].split(' ')[0]
+            try:
+                dump_program = dump_out.split("from '")[1].split(' ')[0]
+            except:
+                log.info("core doesn't have the desired format, moving on ...")
```
zmc, line 324: Should we go ahead and do something like `log.debug(f"dump_out=")` as well?

teuthology/task/tests/test_fetch_coredumps.py
```diff
@@ -0,0 +1,85 @@
...
+        m_subprocess.Popen.return_value = TestFetchCoreDumps.MockPopen(example_output[0])
+        self.the_function(None, m_remote)
+        assert m_remote.get_file.called
+    
+    @patch('teuthology.task.tests.test_fetch_coredumps.TestFetchCoreDumps.MockRemote')
+    @patch('teuthology.task.internal.subprocess')
+    @patch('teuthology.task.internal.os')
+    def test_correct_format(self, m_os, m_subprocess, m_remote):
+        example_output = [
+            "01010101010101001010101010101010110101011"
```
zmc, line 79: Didn't notice the tests until after I made the previous comment about logging - thanks for adding these!!

Is that a realistic non-matching output? If so, I guess we might not want to log it after all...

kamoltat: no haha, realistic ones are shown in the mentioned tracker. Will change it to the tracker one

teuthology/task/internal/__init__.py
kamoltat: wouldn't be the same as ` log.info(f' core looks like: {dump_out}')` in the line above? or do you mean changing log.info to log.debug

zmc: oh I just plain didn't see that line - disregard 😄

zmc (dismissed):
Thanks!

kamoltat:
@zmc sorry I dismissed your review by accident

zmc (approved):

ljflores (approved):
LGTM.

teuthology/task/tests/test_fetch_coredumps.py
```diff
@@ -0,0 +1,89 @@
...
+    @patch('teuthology.task.internal.subprocess')
+    @patch('teuthology.task.internal.os')
+    def test_incorrect_format(self, m_os, m_subprocess, m_remote):
+        example_output = [
+            '/home/teuthworker/archive/'
+            'dis-2022-02-23_21:10:50-rbd-wip-rbd-quincy-batch-4-distro'
+            '-default-smithi/6703486/remote/smithi133/coredump/'
+            '1645664031.59278.core.gz: gzip compressed data, was '
+            '"1645664031.59278.core", last modified: Thu Feb 24'
+            '00:55:10 2022, max compression'
```
idryomov, line 83: We need to preserve the respective binary even if the core is compressed.  It looks like the fix would be to temporarily decompress it to that `file` sees the actual core file and not a gzip archive.  Perhaps `zcat core.gz | file -`?

idryomov (changes requested):
It seems like this just hides the bug, leaving `fetch_binaries_for_coredumps` non-functional.  This code is intended to preserve ceph binaries that correspond to core files, so that core files can be easily inspected with `gdb` after the job is nuked.  If the there is a mismatch between `file` output and what is expected by `split`, we need to fix that, not swallow the exception.

kamoltat:
jenkins retest this please

kamoltat:
@idryomov  Just added some change via your suggestion, let me know what you think!

teuthology/task/internal/__init__.py
```diff
@@ -318,7 +318,21 @@ def fetch_binaries_for_coredumps(path, remote):
...
             # from 'radosgw --rgw-socket-path /home/ubuntu/cephtest/apache/tmp.client.0/fastcgi_soc'
             log.info(f' core looks like: {dump_out}')
-            dump_program = dump_out.split("from '")[1].split(' ')[0]
+            if 'gzip' in dump_out:
+                try:
+                    log.info("core is compressed, try accessing gzip file ...")
+                    with gzip.open(dump_path, 'r') as f:
+                        file_content = f.read().decode()
+                    log.info(f' file content looks like: {file_content}')
+                    dump_out = file_content
```
idryomov, line 327: `dump_out` is supposed to be the result of running `file` on a core, right?  `dump_program` is then parsed out from that.

However, here, if the core is compressed, `dump_out` gets the decompressed core contents...  If you want to do it this way, you need to run `file` twice.

idryomov: Actually, `file` seems to be able to handle compression natively:
```
     -z, --uncompress
             Try to look inside compressed files.

     -Z, --uncompress-noreport
             Try to look inside compressed files, but report information about the contents only not the
             compression.
```
So `file -z -Z` should do the trick and require no new code.

idryomov (changes requested):

teuthology/task/internal/__init__.py
kamoltat: thank you @idryomov for your comments,
So it tested out `file -z -Z` and I think it doesn't really give you the content of what is inside the compressed file, rather it tells you what kind of text is inside the file, usually it is `ASCII text`

kamoltat: actually, nvm I think I wasn't familiar with how core formats are when you use `file` on it. Now I get it, thank you.

kamoltat:
@idryomov so I tested out `file -z` and it doesn't really show the program that generates the core dump. Therefore, we still have to decompress the file, store it on a temporary file and use `file` again to get the program that generates the file.
```
(bash) $ file -z correct.format.core.gz
correct.format.core.gz: ELF 64-bit LSB core file, x86-64, version 1 (SYSV) (gzip compressed data, was "correct.format.core", last modified: Wed Jun 29 19:55:29 2022, from Unix)
(bash) $ file correct.format.core
correct.format.core: ELF 64-bit LSB core file, x86-64, version 1 (SYSV), SVR4-style, from '-bash', real uid: 1194, effective uid: 1194, real gid: 1194, effective gid: 1194, execfn: '/bin/bash', platform: 'x86_64'
```

kamoltat:
@idryomov let me know what you think of the change I made

idryomov:
> @idryomov so I tested out `file -z` and it doesn't really show the program that generates the core dump. Therefore, we still have to decompress the file, store it on a temporary file and use `file` again to get the program that generates the file.

I suggested `file -z -Z`, not `file -z`.

kamoltat:
> > @idryomov so I tested out `file -z` and it doesn't really show the program that generates the core dump. Therefore, we still have to decompress the file, store it on a temporary file and use `file` again to get the program that generates the file.
> 
> I suggested `file -z -Z`, not `file -z`.

I tried `file -z -Z`, it still doesn't show `from `.

```
root@f641596c9263:/teuthology/teuthology/task/tests/coredump# ls        
correct.format.core  correct.format.core.gz  incorrect.format.core  incorrect.format.core.gz
root@f641596c9263:/teuthology/teuthology/task/tests/coredump# file -z -Z correct.format.core.gz 
correct.format.core.gz: ELF 64-bit LSB core file, x86-64, version 1 (SYSV)
root@f641596c9263:/teuthology/teuthology/task/tests/coredump# file correct.format.core
correct.format.core: ELF 64-bit LSB core file, x86-64, version 1 (SYSV), SVR4-style, from '-bash', real uid: 1194, effective uid: 1194, real gid: 1194, effective gid: 1194, execfn: '/bin/bash', platform: 'x86_64'
root@f641596c9263:/teuthology/teuthology/task/tests/coredump# 
```

idryomov:
Argh, so it doesn't work as advertised...

zmc (changes requested):
I really don't want to include huge files in the repo just for testing; we should explore other ways to unit-test this.

zmc:
An idea for avoiding shipping the core dump files in this repo:

1. Abstract the `subprocess` calls to `/usr/bin/file` into a function
2. mock the function in the unit test

To do this with a function that is called multiple times, I like to use [the `side_effect` attribute](https://github.com/ceph/teuthology/blob/main/teuthology/test/test_run.py#L142). From [the mock docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock): 
"If side_effect is an iterable then each call to the mock will return the next value from the iterable."

If you need to inspect the calls themselves after they happen, you can use [`call_args_list`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list).

kamoltat:
> An idea for avoiding shipping the core dump files in this repo:
> 
> 1. Abstract the `subprocess` calls to `/usr/bin/file` into a function
> 2. mock the function in the unit test
> 
> To do this with a function that is called multiple times, I like to use [the `side_effect` attribute](https://github.com/ceph/teuthology/blob/main/teuthology/test/test_run.py#L142). From [the mock docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock): "If side_effect is an iterable then each call to the mock will return the next value from the iterable."
> 
> If you need to inspect the calls themselves after they happen, you can use [`call_args_list`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list).

Thanks Zack for the suggestion, will revise.

kamoltat:
@zmc Just made the change, let me know how it is, thank you.

teuthology/task/tests/test_fetch_coredumps.py
```diff
@@ -0,0 +1,139 @@
...
+import os
+
+class TestFetchCoreDumps(object):
+    class MockDecode(object):
+        def __init__(self, ret):
+            self.ret = ret
+            pass
+
+        def decode(self):
+            return self.ret
```
zmc, line 13: Why not use [Mock](https://docs.python.org/3/library/unittest.mock.html#the-mock-class) and [MagicMock](https://docs.python.org/3/library/unittest.mock.html#magicmock-and-magic-method-support)? That way you could, among other things, define separate return values and/or exceptions for each method.

Is there a benefit to all these custom classes?

teuthology/task/tests/test_fetch_coredumps.py
```diff
@@ -0,0 +1,139 @@
...
+                return [compressed_incorrect, uncompressed_incorrect]
+        else:
+            if correct:
+                return [uncompressed_correct, uncompressed_correct]
+            else: 
+                return [uncompressed_incorrect, uncompressed_incorrect]
+
+    # Core is not compressed and file is in the correct format
+    @patch('teuthology.task.internal.subprocess.Popen')
+    @patch('teuthology.task.tests.test_fetch_coredumps.TestFetchCoreDumps.MockRemote')
```
zmc, line 96: I think I'm more confused now; won't `MockRemote` just become a `Mock` here?

teuthology/task/tests/test_fetch_coredumps.py
```diff
@@ -0,0 +1,139 @@
...
+    @patch('teuthology.task.internal.os')
+    def test_uncompressed_correct_format(self, m_os, m_remote, m_subproc_popen):
+        m_subproc_popen.side_effect = self.side_effect_func(False, True)
```
zmc, line 99: `side_effect_func()` makes readability a little funny; why not e.g.:
```
m_subproc_popen.side_effect = [self.uncompressed_correct, self.uncompressed_correct]
```

kamoltat: like

teuthology/task/tests/test_fetch_coredumps.py
kamoltat: Yea there is definitely some cleaning up that I forgot to do. Thanks for catching it.

teuthology/task/tests/test_fetch_coredumps.py
kamoltat: No benefits, I agree that custom classes are not the best way, more like I was exploring different ways to Mock objects.

kamoltat:
Recent core dump-related failures in:
/a/yuriw-2022-07-22_03:30:40-rados-wip-yuri3-testing-2022-07-21-1604-distro-default-smithi/6943763/

Indicates that we did not do a good job at extracting the program from the dump. Will add a commit that will improve this.

```
core looks like: /home/teuthworker/archive/yuriw-2022-07-22_03:30:40-rados-wip-yuri3-testing-2022-07-21-1604-distro-default-smithi/6943763/remote/smithi083/coredump/1658515951.94790.core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, from 'ceph_test_rados_api_io', real uid: 1000, effective uid: 1000, real gid: 1265, effective gid: 1265, execfn: '/usr/bin/ceph_test_rados_api_io', platform: 'x86_64'

2022-07-22T19:11:45.927 DEBUG:teuthology.orchestra.run.smithi083:> which 'ceph_test_rados_api_io'"'"','
2022-07-22T19:11:45.953 DEBUG:teuthology.orchestra.run:got remote process result: 1
2022-07-22T19:11:45.954 INFO:teuthology.orchestra.run.smithi083.stderr:/usr/bin/which: no ceph_test_rados_api_io', in (/home/ubuntu/.local/bin:/home/ubuntu/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin)
2022-07-22T19:11:45.956 ERROR:teuthology.run_tasks:Manager failed: internal.archive
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/run_tasks.py", line 188, in run_tasks
    suppress = manager.__exit__(*exc_info)
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/task/internal/__init__.py", line 399, in archive
    fetch_binaries_for_coredumps(path, rem)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/task/internal/__init__.py", line 324, in fetch_binaries_for_coredumps
    remote_path = remote.sh(['which', dump_program]).rstrip()
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/orchestra/remote.py", line 96, in sh
    proc = self.run(**kwargs)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/orchestra/remote.py", line 510, in run
    r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/orchestra/run.py", line 455, in run
    r.wait()
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/orchestra/run.py", line 161, in wait
    self._raise_for_status()
  File "/home/teuthworker/src/git.ceph.com_git_teuthology_8d598431210977f8caccec83230b4bfec7bd5d3f/teuthology/orchestra/run.py", line 183, in _raise_for_status
    node=self.hostname, label=self.label
teuthology.exceptions.CommandFailedError: Command failed on smithi083 with status 1: 'which \'ceph_test_rados_api_io\'"\'"\',\''
```

kamoltat:
@zmc Just pushed the new changes, I kept some of the custom classes because I think it is more readable than Mocking all the attributes and functions that are associated with Popen. But other than that, I improved the way we are extracting the program name in core dumps, using regex rather than python split() and etc.

zmc (approved):
Thanks for making those changes!

teuthology/task/internal/__init__.py
```diff
@@ -318,7 +319,28 @@ def fetch_binaries_for_coredumps(path, remote):
...
+                         dump_info = subprocess.Popen(['file', f_out.name],
+                                                        stdout=subprocess.PIPE)
+                         dump_out = dump_info.communicate()[0].decode()
+                         log.info(f' core looks like: {dump_out}')
+                except Exception as e:
+                    log.info('Something went wrong while opening the compressed file')
+                    log.error(e)
+                    continue
+            try:
+                dump_program = re.findall("from \'([^']*)\'", dump_out)[0]
```
idryomov, line 338: Am I understanding correctly that split -> regex change is the only non-test change since my last review?  It looks fine but:
- I don't understand the need for `\` before `'` in both cases
- I would change `*` to `+` to harden the pattern against texts like `from ''`

kamoltat: Done!

idryomov (approved):
LGTM!  As before, I would suggest creating a "bad" build (e.g. ceph-osd crashing with some low probability some time after being started or some test binary always crashing -- the latter may be easier) and see if the right binary is pulled into the test archive.

kamoltat:
> LGTM! As before, I would suggest creating a "bad" build (e.g. ceph-osd crashing with some low probability some time after being started or some test binary always crashing -- the latter may be easier) and see if the right binary is pulled into the test archive.

[idryomov](https://github.com/idryomov)
That sounds good, thanks for your help.

ljflores:
@kamoltat one thing you can try to verify this PR is run this test (must be on ubuntu) http://pulpito.front.sepia.ceph.com/lflores-2022-08-16_18:51:57-rados:singleton-nomsgr-wip-yuri4-testing-2022-08-15-0951-distro-default-smithi/6975747/.

The binary is not getting collected, just like what you recorded in https://tracker.ceph.com/issues/53206.

PR #1720, opened 2022-03-04
Revive the 'test' task, used by the 'teuthology' suite

zmc:
I'm unsure if we ever really used this task or its companion suite to do anything meaningful, but it needed some love to get working again.

This, in combination with our ongoing docker-compose work, should let us run a pretty thorough set of self-tests for teuthology and friends much more easily than before. I believe we'd be able to run these for every pull request using github actions, for example.

This does not yet enable the tests which require network access, though I do think we'll probably want to enable them in the future.

As of today, here's a one-line summary of the result:
```
588 passed, 72 skipped, 7 warnings in 14.69 seconds
```


dmick (dismissed):
lgtm

zmc:
@dmick thanks for the review - I noticed a dependency issue and had to re-push; care to re-approve?

dmick (approved):
still lgtm

PR #1721, opened 2022-03-04
github actions: Run docs build during CI

zmc:
If we do this, we can drop the Jenkins job. The openstack tests are
currently all skipped, so we can just leave them out for now.


zmc:
Hopefully this is obvious, but I'm referring to the `teuthology-pull-requests` job and not the `teuthology-docs` job, which has apparently been broken since December '21 😬

zmc:
This'll help #1719 when it is ready.

kamoltat (approved):
LGTM

PR #1722, opened 2022-03-04
Skip various operations on containerized testnodes

zmc:


zmc:
testing here https://pulpito.ceph.com/zack-2022-03-10_20:07:11-smoke-master-distro-default-smithi/

teuthology/task/clock.py
```diff
@@ -8,6 +8,9 @@
 
 log = logging.getLogger(__name__)
 
+def filter_cluster(cluster):
```
dmick, line 11: usage of this function would be more obvious in context if it referred to 'container' somehow

zmc:
Tested against bare metal here: https://pulpito.ceph.com/zack-2022-03-11_00:23:10-smoke-master-distro-default-smithi/

teuthology/task/clock.py
zmc: True - or, actually, I don't even know why I thought it would be important to abstract this into a separate function.

dmick: maybe you thought about making it take arguments to specify how to filter.  but it's probably overkill

zmc: Looks like I abstracted it because the same logic is used a few times throughout the task. Added a docstring.

dmick: I still think there's value in the name expressing more of its purpose so that the call site is more self-documenting, but it's a pretty small nit.

zmc: You think something like `filter_cluster_no_containers` would be better?

dmick: yeah, or filter_out_containers(), or remove_containers() (the cluster is the arg, so)

zmc: Done!

dmick (approved):

PR #1723, opened 2022-03-07
fixes for running non-scheduled jobs 

jdurgin:
Fetching tasks automatically did not work for non-scheduled jobs because it never actually imported them, like scheduled jobs (which have suite_path set) do a few lines earlier.

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -88,7 +88,7 @@ def fetch_tasks_if_needed(job_config):
     suite_sha1 = job_config.get('suite_sha1')
     suite_path = os.path.normpath(os.path.join(
         fetch_qa_suite(suite_branch, commit=suite_sha1),
-        job_config.get('suite_relpath', ''),
+        job_config.get('suite_relpath', 'qa'),
     ))
     sys.path.insert(1, suite_path)
     return suite_path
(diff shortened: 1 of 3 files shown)


teuthology/run.py
```diff
@@ -88,9 +88,10 @@ def fetch_tasks_if_needed(job_config):
     suite_sha1 = job_config.get('suite_sha1')
     suite_path = os.path.normpath(os.path.join(
         fetch_qa_suite(suite_branch, commit=suite_sha1),
-        job_config.get('suite_relpath', ''),
+        job_config.get('suite_relpath', 'qa'),
     ))
     sys.path.insert(1, suite_path)
+    import tasks
```
zmc, line 94: This line is causing a test failure. I don't think we need it, since the actual importing of individual tasks is handled by the `run_tasks` module. Was it required in your testing?

zmc (changes requested):

teuthology/run.py
jdurgin: you're right, I thought it was necessary, but it was a ModuleNotFound error just from having the wrong path -- not including 'qa' -- by default

zmc:
Scheduled a test here https://pulpito.ceph.com/zack-2022-03-11_22:16:39-orch:cephadm:smoke-master-distro-default-smithi/

zmc (approved):

PR #1725, opened 2022-03-18
task.clock: Fix call to filter_out_containers()

zmc:


zmc:
@dmick we both missed this when I renamed the function...

dmick (approved):
d'oh

PR #1730, opened 2022-03-28
docs/docker-compose: Omit teuthology-suite in DockerFile

kamoltat:
Get rid of the auto schedule of dummy job by
the Teuthology DockerFile as we run docker-compose.
This is because some users might not have access
to all contents of .teuthology.yaml in the sepia lab.

Also, clean up some stuff in README.md.


kamoltat:
This PR is no longer needed after https://github.com/ceph/teuthology/pull/1727 was merged

PR #1733, opened 2022-03-29
update packaging.py

coderSHeE:
made changes on string function and on some operators :


coderSHeE:
yep :)

tchaikov:
ok to test

tchaikov (changes requested):
these changes seem just wrong. and they are not companied with any explanations.

dmick:
Please use more descriptive titles; the fact that it's a change to packaging.py is obvious, but what does the change accomplish?  "update" is too general.

PR #1736, opened 2022-04-05
Added test for teuthology.orchestra.run:RemoteProcess._raise_for_status

UnknownCoder-uc:
@amathuria @kamoltat @zmc 

Added test for teuthology.orchestra.run:RemoteProcess._raise_for_status


zmc (changes requested):
- Why three commits?
- Commit messages are not useful at all
- The added test doesn't match the style of our existing tests

amathuria (changes requested):
I would suggest you go through our existing tests and see how the pytest and mock modules are used

PR #1739, opened 2022-04-13
docker-compose: "wait mode"

zmc:


docs/docker-compose/teuthology/teuthology.sh
```diff
@@ -14,23 +14,26 @@ else
...
+        --suite teuthology:no-ceph \
+        --filter-out "libcephfs,kclient,stream,centos,rhel" \
+        -d ubuntu -D 20.04 \
+        --suite-branch master \
+        --subset 9000/100000 \
+        -p 75 \
+        --seed 349 \
+        --force-priority \
+        $CUSTOM_CONF
+    DISPATCHER_EXIT_FLAG='--exit-on-empty-queue'
```
kamoltat, line 34: I don't think this should be here since for the Teuthology_wait we the dispatcher to keep running even with the empty queue, but with the other mode we want to exit the empty queue.

kamoltat (changes requested):
Overall looks good! just 1 comment regarding the wait mode

docs/docker-compose/teuthology/teuthology.sh
kamoltat: nvm I misread the syntax for -z and -n if statements, this looks good

kamoltat (dismissed):

kamoltat (approved):

PR #1741, opened 2022-04-14
teuthology.sh: Run tests using current branch

zmc:


docs/docker-compose/teuthology/teuthology.sh
```diff
@@ -15,6 +17,7 @@ else
 fi
 export MACHINE_TYPE=${MACHINE_TYPE:-testnode}
 teuthology-suite -v \
+    --teuthology-branch $TEUTHOLOGY_BRANCH \
```
kamoltat, line 20: Should also let user know in the wait for manual teuth-schedule feature that they need to include this

zmc: Yeah, when we add that feature we'll have to remember to document that

PR #1737, opened 2022-04-05
Packaging overhaul

zmc:

PR #1743, opened 2022-04-14
Revised packaging updates

zmc:
This PR contains many of the changes from #1737, but leaves out the problematic commits, which were related to virtualenv management. Those commits may or may not be included in a subsequent PR.

It has successfully run a job from the smoke suite:
https://pulpito.ceph.com/zack-2022-04-14_21:52:50-smoke-master-distro-default-smithi/


zmc:
The Sepia lab was idle so I tried running some jobs using a smithi dispatcher running this branch. I also scheduled some to run using master, so that I could see the behavior of managing multiple teuthology checkouts.

Jobs that used teuthology's master branch:
https://pulpito.ceph.com/zack-2022-04-19_21:52:26-smoke-master-distro-default-smithi/
https://pulpito.ceph.com/zack-2022-04-19_22:57:40-teuthology:no-ceph-master-distro-default-smithi/

Jobs. that used this branch:
https://pulpito.ceph.com/zack-2022-04-19_21:50:19-smoke-master-distro-default-smithi/
https://pulpito.ceph.com/zack-2022-04-19_21:50:49-teuthology:no-ceph-master-distro-default-smithi/

The unit test failures in the `teuthology:no-ceph` runs are present in both branches and are addressed by https://github.com/ceph/teuthology/pull/1747

andrewschoen (approved):
LGTM

kamoltat (approved):
Had to look up pyproject.toml, because I never used it before, but from my humble opinion this looks good.

PR #1745, opened 2022-04-17
GSOC | Unit Test added for _raise_for_status function under run.py

muskan399:
I've tested this test in my local and it raises `CommandFailedError` when check_status is true, returncode is != 0

Please review this.
@zmc @kamoltat @amathuria


muskan399:
@zmc @kamoltat @amathuria
Do something needs to be done from my end.
There is a job failure while running the pipeline.
https://github.com/ceph/teuthology/runs/6057949717?check_suite_focus=true

PR #1749, opened 2022-04-23
setup.cfg: add nose to "options.extras_require/test"

tchaikov:
some tests are using nose python package. since teuthology is
responsible for preparing the deps to run the tests. let's add
nose back to setup.cfg.

the "nose" package was included in requirements.txt before
21372fd9b2b87bb6417929f10097fba7db9cdca9. in this change, it,
it is brought back.

Signed-off-by: Kefu Chai <tchaikov@gmail.com>


idryomov:
I was going to say that we first saw `nose` import failures earlier than two days ago, but it looks like the initial attempt was merged over a week ago in https://github.com/ceph/teuthology/pull/1737 and later reverted in https://github.com/ceph/teuthology/pull/1740 so I guess that explains it.

setup.cfg
```diff
@@ -103,6 +103,7 @@ test =
     cryptography>=2.7
     ipy
     mock
+    nose
```
idryomov, line 106: Here is the original list:
```
            'boto >= 2.0b4',       # for qa/tasks/radosgw_*.py
            'cryptography >= 2.7',  # for qa/tasks/mgr/dashboard/test_rgw.py
            'nose', # for qa/tasks/rgw_multisite_tests.py',
            'pip-tools',
            'pytest',           # for tox.ini
            'requests',         # for qa/tasks/mgr/dashboard/helper.py
            'tox',
            # For bucket notification testing in multisite
            'xmltodict',
            'boto3',
            'PyJWT',            # for qa/tasks/mgr/dashboard/test_auth.py
            'ipy',              # for qa/tasks/cephfs/mount.py
            'toml',             # for qa/tasks/cephadm.py
```
It looks like `nose` and `requests` got lost.  Does `requests` need to be re-added as well?

idryomov: The comments in the original list are very helpful.  It is really unfortunate that `setup.cfg` syntax doesn't support line-end comments...

tchaikov: @idryomov `requests` is listed in `install_requires`. since we always "install" teuthology for using it, i think it'd be available when the tests are trying to import it. see https://github.com/ceph/teuthology/blob/1b11219b9af33884aae5eb66700541df4b9620ed/setup.cfg#L63

idryomov: In that case the same goes for `boto>=2.0b4`, `boto3`, `cryptography>=2.7`, `ipy`, `pip-tools`, `toml` and `xmltodict`.  Does the actual teuthology (i.e. non-test-task) code use all of them?  If so, should they be removed from `options.extras_require` as redundant?

tchaikov: @idryomov hi Ilya, after a second thought, i went the other way -- to remove them from `options.install_requires` . just to highlight that they are dependencies for tests not for the teuthology test driver.

in addition to the change to dedup the requires, i also took the opportunity to drop the "six" package dependency.

idryomov: This looks sensible to me but I feel a bit uneasy about the regenerated `requirements.txt` since a number of versions got changed.  I'm going to leave the approval to @zmc.

tchaikov: yeah, that was intentional. as @zmc used python3.9 while i used python 3.6 to regenerate `requirements.txt`, because  i assumed that most of us were using "teuthology" in our lab to run the teuthology workers. and "teuthology" runs ubuntu 18.0.4, which has python3.6
```console
kchai@teuthology:~$ python3 --version
Python 3.6.9
```

ljflores (approved):

zmc (approved):
Thank you for the fix @tchaikov! And apologies for the confused cause by the regression @cbodley. I am not sure how nose was dropped in https://github.com/ceph/teuthology/pull/1743/commits/21372fd9b2b87bb6417929f10097fba7db9cdca9.

This does make me wonder if it would be worth giving tasks a way to specify their own requirements...

PR #1751, opened 2022-04-27
bootstrap_teuthology: Call decode() on stderr

zmc:


tchaikov (approved):
i'd pass `universal_newlines=True` or `text=True`. but let's merge this change as it addresses an urgent issue.

PR #1753, opened 2022-04-27
docker-compose: Speed incremental builds by caching dependencies

zmc:
Previously, we'd invalidate the build cache any time the source tree changed; that meant reinstalling every python dependency. This change splits dependency installation from the rest of the bootstrap process, so that we can avoid reinstalling deps unless `requirements.txt` or `bootstrap` changes.


amathuria (approved):
LGTM!

djgalloway (approved):

PR #1754, opened 2022-04-28
Add new GitHub Action to test deps on upgrade

zmc:


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          ref: master
+          path: teuthology
+      - name: virtualenv
+        run: |
+          pip install --user virtualenv
+          virtualenv ./virtualenv
+          cd ./virtualenv/lib/python*
+          touch no-global-site-packages.txt
+        working-directory: ./teuthology
```
zmc, line 33: The above step is temporarily needed until we remove or fix `bootstrap`'s behavior of removing virtualenvs that use system site-packages, as github actions seems to only want to create that type of virtualenv.

zmc:
@tchaikov including you on this one because of our brief discussion about ansible versions; I initially created this to test an upgrade to v5, but this new test failed on py36 (which is what `teuthology.front` uses currently). So it looks like we may want 4.2.0 until we can upgrade that VM to a newer OS.

tchaikov (approved):

PR #1762, opened 2022-06-06
qa: correct line processing logic

batrick:
Most runs don't use --no-nested-subset and for those that used --subset,
the if conditions would correctly pickup "seed" (when it mattered).
However, when --subset was not specified in the original run, the "seed"
was not correctly picked up. Therefore, the result of inserting the "if
no_nested_subset is None:" before the "elif seed is None:" caused it to
never read the seed for most folks teuthology runs.

### DIFF
--- a/teuthology/report.py
+++ b/teuthology/report.py
@@ -413,10 +413,8 @@ class ResultsReporter(object):
                     subset = self._parse_log_line(line, 'subset:')
                 if no_nested_subset is None:
                     no_nested_subset = self._parse_log_line(line, 'no_nested_subset:')
-                elif seed is None:
+                if seed is None:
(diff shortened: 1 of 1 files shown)


batrick:
Accidentally closed this, see #1763

batrick:
nevermind, found the reopen button!

zmc (approved):
This is a good fix but it doesn't change issue 55815, which is reproducible when passing `--seed`

zmc:
I was wrong - this fixes the main use-case! There appears to be a separate issue affecting _some_ runs that cause incorrect behavior when the archive is unavailable, even when passing correct values, but I'm very glad to have this fixed. Thanks!!

PR #1765, opened 2022-06-08
nuke.nuke: Rework lock-checking logic

zmc:
Previously, we would call list_locks(), then iterate over the response,
each time iterating over the list of targets. If list_locks()
encountered an error and returned an empty response, we'd never actually
verify what we intended to. Instead, we should specifically query for
each target. This is far safer and faster.


zmc:
This is for https://tracker.ceph.com/issues/45512

kamoltat (approved):
LGTM

PR #1768, opened 2022-06-13
orch/run: Make teuthology a better detective

VallariAg:
Fixes https://tracker.ceph.com/issues/54995

## Description

To improve error logs for unit tests, this PR will:
- Scan teuthology log files for unit-test failures
- Keep a flag which points to the last byte read by the ErrorScanner. Do not read above this flag index to avoid re-reading.
 (checkout static member `ErrorScanner.__flag__`)
- Raise new `UnitTestError` with error message found in log files.
- Enabled for just nose and gtests, so far. 

### Results:
gtest failure: https://pulpito.ceph.com/vallariag-2022-06-22_16:34:18-rbd:basic-gtest-vallariag-test-distro-default-smithi/
nose failure: https://pulpito.ceph.com/vallariag-2022-06-20_13:16:35-smoke-quincy-release-distro-default-smithi/

## How to enable this feature for a new unit-test?

Explained with an example to enable this for unit test of type "xyz". And when a test fails for "xyz" it throws an error like "XYZ-FAILED: ABC.abc test failed". 

#### Changes in teuthology:
Add regex (of the failure) in [`ERROR_PATTERNS`](https://github.com/VallariAg/teuthology/blob/wip-vallariag-improve-log/teuthology/orchestra/run.py#L245) dictionary in `/teuthology/orchestra/run.py`. You can add multiple regex strings in case the test prints different kinds of failure messages.
```python
 ERROR_PATTERNS = {
        # other tests
        "xyz":  [re.compile(r"\[\s\sXYZ-FAILED\s\s\]")],
   }
```
#### Changes in ceph qa section:

TL;DR: checkout [changes of this PR](https://github.com/ceph/ceph/pull/47335/files) to see all the changes needed.

1. **Add parameter [`scan_tests_errors`](https://github.com/VallariAg/teuthology/blob/wip-vallariag-improve-log/teuthology/orchestra/run.py#L513)** where the test file calls teuthology.orchestra.run() method.
The parameter is a list of the unit-test name to match the key in ERROR_PATTERNS,  example 
  ```
    remote.run(
        # other args
         scan_tests_errors=["xyz"]`
    )
  ```
  Real example: checkout changes required for s3tests - [link](https://github.com/ceph/ceph/pull/47335/files#diff-03d9d4ca9118fab79c09621596ea11a8803ee363a3341df799e8a9f562d22ef6).

2. **Make the feature opt-in** by adding a `scan_logs: True` key-pair in job's yaml file and reading the value of the key before making the teuthology.orchestra.run call. [Real example](https://github.com/ceph/ceph/pull/47335/files#diff-03d9d4ca9118fab79c09621596ea11a8803ee363a3341df799e8a9f562d22ef6) 

----

TODO before merging:
- Test for large log files: 
  1. Use a generator function to replicate the reading of log file (with for loop or sleep).
  2. Generate a huge file and then time the scanning with [perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter) or [timeit](https://docs.python.org/3/library/timeit.html)
- Scan [XML files](https://github.com/google/googletest/blob/main/docs/advanced.md#generating-an-xml-report) instead: configure the tests to generate XML files and then scan them.


zmc:
ok to test

teuthology/orchestra/run.py
```diff
@@ -221,6 +233,46 @@ def __repr__(self):
...
+        if not logfile or not self.test_names:
+            return None
+        
+        ERROR_PATTERN = {
+            "nose": r"ERROR:\s",
+            "gtest":  r"\[\s\sFAILED\s\s\]",
+        }
+        logs = []
+        with open(logfile, 'r') as f:
+            logs = f.readlines()
```
zmc, line 255: Rather than read the entire file at once - which could use large amounts of memory - let's look into using a [generator](https://wiki.python.org/moin/Generators) to read and process one line (or several lines) at a time. That could be integrated with the `for` loop below.

VallariAg: Great point! For lazy reading of files, we can simply loop through the file object created with `open('textfile')` like:
```python
for line in open("file.txt"):
     print(line)
```
This will not load the entire file in memory, instead read it line by line. 

For `__flag__` implementation, we can use [seek() & tell()](https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects) methods, to do something like: 
(reading files top to bottom here)
```python
__flag__ = 3   # last read byte

f = open('sample.txt', 'r')
f.seek(__flag__ + 1) # seek to next unread byte

for line in f:
    # search regex matches in `line`

__flag__ = f.tell()

f.close()
```
Thoughts about this approach?

teuthology/orchestra/run.py
```diff
@@ -34,30 +34,33 @@ class RemoteProcess(object):
...
         'remote',
         'label',
+        'scan_tests_errors',
         ]
 
     deadlock_warning = "Using PIPE for %s without wait=False would deadlock"
 
     def __init__(self, client, args, check_status=True, hostname=None,
-                 label=None, timeout=None, wait=True, logger=None, cwd=None):
+                 label=None, timeout=None, wait=True, logger=None, cwd=None, scan_tests_errors=[]):
```
jdurgin, line 43: the default for scan_tests_for_errors should be None rather than an empty list -- due to the way python initializes parameters, using [] or {} as defaults ends up making the variable persistent across invocations of the function, which is usually not what you want.

More detail: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

teuthology/orchestra/run.py
```diff
@@ -221,6 +233,73 @@ def __repr__(self):
...
+                    return line[error.start():].strip()
+        return None
+
+    def is_prev_detected_error(self, line) -> bool:
+        for detected_pattern in self.ERROR_PATTERNS['prev_detected']:
+            if detected_pattern.search(line):
+                return True
+        return False
+
+    def scan(self, test_names=[]):
```
jdurgin, line 268: same issue with a mutable default argument here

teuthology/orchestra/run.py
VallariAg: Interesting, thanks for explaining that! Fixed it.

teuthology/orchestra/run.py
VallariAg: After exploring the following ways to read logs, I compared them with [timeit](https://docs.python.org/3/library/timeit.html) and [tracemalloc](https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_traced_memory) on a 0.5GB file.
1. `f.readline()` 
2.  [islice()](https://stackoverflow.com/a/6335876)
3.  f.readlines(num_of_lines)
4.  `for line in f`
5.  [mmap](https://docs.python.org/3/library/mmap.html)

Looks like for-in loop on a file object (option 4 above) is the most memory and time efficient for huge log files. 
Further, re.compile improved the performance a lot.

teuthology/orchestra/run.py
```diff
@@ -178,6 +182,14 @@ def _raise_for_status(self):
                 # signal; sadly SSH does not tell us which signal
                 raise CommandCrashedError(command=self.command)
             if self.returncode != 0:
+                if self.scan_tests_errors:
+                    test, error = ErrorScanner().scan(test_names=self.scan_tests_errors)
```
zmc, line 186: It might even be a good idea to wrap this statement in a `try: / except Exception` clause for fault-tolerance.

teuthology/orchestra/run.py
```diff
@@ -221,6 +233,73 @@ def __repr__(self):
             name=self.hostname,
             )
 
+class ErrorScanner(object):
+    """
+    Scan for unit tests errors in teuthology.log 
+    """
+    __flag__ = 0
+    def __init__(self):
+        self.ERROR_PATTERNS = {
```
zmc, line 242: I'm curious why `__flag__` is a class var but `ERROR_PATTERNS` is an instance var. I think it would make more sense if it were the other way around, since:

1. `ERROR_PATTERNS` doesn't appear to ever be modified, and contains `re.compile()` statements. Using `re.compile()` is a good choice, and it is not at all expensive to run a few times, but there's no real reason to do this multiple times in a given process.
2. `__flag__` does actually get modified, and multiple `ErrorScanner` instances will conflict when modifying it.

teuthology/orchestra/run.py
```diff
@@ -221,6 +233,73 @@ def __repr__(self):
...
+    def scan(self, test_names=None):
+        logfile = self._logfile
+        if not logfile or not test_names:
+            return None, None
+
+        error_test = None
+        error_msg = None
+
+        log.debug("__flag__ pointer at: {}".format(ErrorScanner.__flag__))
+        f = open(logfile, 'r')
```
zmc, line 277: Instead of calling `open()` here and `close()` separately later on, we should use a `with` statement. A short explanation for that is with this current implementation, any exception encountered between `open()` and `close()` will mean that `close()` never gets called. If that's confusing or if you're curious to learn more, check out this [very in-depth writeup](https://realpython.com/python-with-statement/#managing-resources-in-python)

zmc (commented):
This is really coming along! In addition to the comments I wrote, I wanted to talk about the attached logfiles briefly. I like the idea of using real logfiles to unit-test this feature, but I also want to balance this with the fact that the teuthology repo is fetched and re-fetched a _lot_, so adding large files will have an impact that adds up over time. I'd propose trimming down the files to only include maybe a few dozen lines that don't match the regexes we're looking for.

teuthology/orchestra/run.py
VallariAg: Since each job has it's own log file, `__flag__` should be shared between all processes in a job which is why I declared it as a class var. 
Suppose if we make `__flag__` an instance variable and create an `ErrorScanner` object in `RemoteProcess`, then flag would be bound to a process's scope and reinitialize it to 0 every time. This would defeat the purpose of tracking flag.
To test that flag as a class var is enclosed in each job's scope, I made [these changes](https://github.com/VallariAg/teuthology/commit/fa7f5bd40d60239867ee58c3a50f5026aafd559e) and the [logs](https://pulpito.ceph.com/vallariag-2022-06-25_13:05:19-rbd:basic-main-distro-default-smithi/) proved that changes to flag are persistent between processes and do not affect other jobs.  

`ERROR_PATTERNS` should indeed be a class var, makes sense!

teuthology/orchestra/run.py
VallariAg: I see, thank you for sharing this! :)

teuthology/orchestra/run.py
```diff
@@ -221,6 +237,70 @@ def __repr__(self):
...
+
+        for pattern in test_patterns:
+            error = pattern.search(line)
+            if error:
+                is_old_error = self._is_prev_detected_error(line)
+                if not is_old_error:
+                    return line[error.start():].strip()
+        return None
+
+    def _is_prev_detected_error(self, line) -> bool:
```
amathuria, line 265: Just to make sure I understood this correctly, you are using prev_detected errors to filter out all the UnitTestErrors in the log file?

VallariAg: Yes, because UnitTestErrors exceptions have the error message - which includes the error pattern (like "ERROR:") - so this is to ensure we avoid reading these exceptions as original failures.

VallariAg:
Quick question before I open a PR on the main repo - should I enable scanning of logs in all job yaml files for unit tests (hence making it enabled by default and could be disabled by removing `scan_logs` from yaml files)? Checkout [this](https://github.com/VallariAg/ceph/commit/0a7a6b4ad6ca4201c81cd7ec8c55da5f7880d2b2) implementation.

kamoltat:
> Quick question before I open a PR on the main repo - should I enable scanning of logs in all job yaml files for unit tests (hence making it enabled by default and could be disabled by removing `scan_logs` from yaml files)? Checkout [this](https://github.com/VallariAg/ceph/commit/0a7a6b4ad6ca4201c81cd7ec8c55da5f7880d2b2) implementation.

Note for history, addressed this offline, we decided to go with an opt-in feature for the test author to enable in their .yaml file.

kshtsk:
do you plan to continue work on this task?

teuthology/orchestra/run.py
kshtsk: Could you please give an example why this is needed?

VallariAg:
Hey @kshtsk, sorry for not posting an update on this PR. 
We decided to go for another solution for this task - create xml summary of the unit tests and then scan those xml files for failures. Here is the PR https://github.com/ceph/teuthology/pull/1792 which implemented it. 

Let me know if there's any another use-case where reading through teuthology logs could be needed. If not, we can probably close this PR.

kshtsk:
@VallariAg thanks, I gonna close it for now, I guess it would be easier to create another PR with some idea from this one once the use cases appear.

VallariAg:
that sounds good, thank you!

PR #1772, opened 2022-06-22
Revert "Merge pull request #1624 from batrick/get-valgrind-args"

batrick:
This reverts commit 40fff4b2911949847a302adfe2f4714cfaccb58e, reversing
changes made to 93d2b653facc0e5a7a532b39d9ff91a044f2c4a9.

PR #1624, opened 2021-03-03
misc: remove get_valgrind_args

batrick:
It is now part of [1].

[1] https://github.com/ceph/ceph/pull/39724
Signed-off-by: Patrick Donnelly <pdonnell@redhat.com>

PR #1773, opened 2022-06-23
Revert "Merge PR #1772 into master"

batrick:
Turtles.

Merge when Octopus is EOL. For reals.

See #1772 and #1624 for history.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1010,64 +1010,6 @@ def deep_merge(a: DeepMerge, b: DeepMerge) -> DeepMerge:
     return b
 
 
-def get_valgrind_args(testdir, name, preamble, v, exit_on_first_error=True):
-    """
-    Build a command line for running valgrind.
-
-    testdir - test results directory
-    name - name of daemon (for naming hte log file)
-    preamble - stuff we should run before valgrind
-    v - valgrind arguments
-    """
-    if v is None:
-        return preamble
-    if not isinstance(v, list):
-        v = [v]
-
-    # https://tracker.ceph.com/issues/44362
-    preamble.extend([
-        'env', 'OPENSSL_ia32cap=~0x1000000000000000',
-    ])
-
-    val_path = '/var/log/ceph/valgrind'
-    if '--tool=memcheck' in v or '--tool=helgrind' in v:
-        extra_args = [
-            'valgrind',
-            '--trace-children=no',
-            '--child-silent-after-fork=yes',
-            '--soname-synonyms=somalloc=*tcmalloc*',
-            '--num-callers=50',
-            '--suppressions={tdir}/valgrind.supp'.format(tdir=testdir),
-            '--xml=yes',
-            '--xml-file={vdir}/{n}.log'.format(vdir=val_path, n=name),
-            '--time-stamp=yes',
-            '--vgdb=yes',
-        ]
-    else:
-        extra_args = [
-            'valgrind',
-            '--trace-children=no',
-            '--child-silent-after-fork=yes',
-            '--soname-synonyms=somalloc=*tcmalloc*',
-            '--suppressions={tdir}/valgrind.supp'.format(tdir=testdir),
-            '--log-file={vdir}/{n}.log'.format(vdir=val_path, n=name),
-            '--time-stamp=yes',
-            '--vgdb=yes',
-        ]
-    if exit_on_first_error:
-        extra_args.extend([
-            # at least Valgrind 3.14 is required
-            '--exit-on-first-error=yes',
-            '--error-exitcode=42',
-        ])
-    args = [
-        'cd', testdir,
-        run.Raw('&&'),
-    ] + preamble + extra_args + v
-    log.debug('running %s under valgrind with args %s', name, args)
-    return args
-
-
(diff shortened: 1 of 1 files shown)


idryomov (commented):
> See #1772 and #1624 for history.

I clicked through the links in the description of the PR, but I think it would be nicer if the history was actually explained in the commit message because right now it's a revert of a revert and commit 31a356e mentioned there isn't relevant today at all.

idryomov:
Also, while octopus is EOL, we are still testing upgrades from it to reef.  Can this somehow interfere there (might be a dumb question but this had to be reverted once already so asking just to be sure)?

idryomov:
> You mean quincy I assume?

Yes, apologies for the braino.

kshtsk:
@batrick ping, do we still need this PR?

batrick:
> @batrick ping, do we still need this PR?

It's dead code to remove but needs to stick around for a while until we stop testing upgrade paths from octopus.

batrick:
> Octopus is EOL already, Pacific too, and Quincy as well.

PTAL

kshtsk:
@zmc looks integration/test is broken for all PRs, is it related to the latest merges to paddles?

idryomov (approved):
I'm still of the opinion that

> it would be nicer if the history was actually explained in the commit message because right now it's a revert of a revert and commit [31a356e](https://github.com/ceph/teuthology/commit/31a356e0c737f1781079045bb8dc366982be8713) mentioned there isn't relevant

batrick:
> I'm still of the opinion that
> 
> > it would be nicer if the history was actually explained in the commit message because right now it's a revert of a revert and commit [31a356e](https://github.com/ceph/teuthology/commit/31a356e0c737f1781079045bb8dc366982be8713) mentioned there isn't relevant

The mention of 31a356e0c737f1781079045bb8dc366982be8713 is boiler-plate from `git revert`. Do you still want that removed?

I'll improve the commit message about this history.

idryomov:
> The mention of [31a356e](https://github.com/ceph/teuthology/commit/31a356e0c737f1781079045bb8dc366982be8713) is boiler-plate from `git revert`. Do you still want that removed?

No, what you did LGTM!

zmc:
> @zmc looks integration/test is broken for all PRs, is it related to the latest merges to paddles?

It is looking like this is what broke the integration test here, and ceph-devstack's tests: https://github.com/ceph/ceph-cm-ansible/pull/734/files#diff-805018280dfbabfa8cd9ced1e8640ac7e83aea1ba0d58ae799cecf77dabbee60R2-R6

@djgalloway - can this role be made optional?

djgalloway:
> > @zmc looks integration/test is broken for all PRs, is it related to the latest merges to paddles?
> 
> It is looking like this is what broke the integration test here, and ceph-devstack's tests: https://github.com/ceph/ceph-cm-ansible/pull/734/files#diff-805018280dfbabfa8cd9ced1e8640ac7e83aea1ba0d58ae799cecf77dabbee60R2-R6
> 
> @djgalloway - can this role be made optional?

Of course.  I didn't realize were doing full legit integration testing for teuthology.  Sorry about that.

Where is the log of the actual failure though?  I don't see the ansible output in the GH Action log.

zmc:
> > > @zmc looks integration/test is broken for all PRs, is it related to the latest merges to paddles?
> > 
> > 
> > It is looking like this is what broke the integration test here, and ceph-devstack's tests: [ceph/ceph-cm-ansible#734 (files)](https://github.com/ceph/ceph-cm-ansible/pull/734/files#diff-805018280dfbabfa8cd9ced1e8640ac7e83aea1ba0d58ae799cecf77dabbee60R2-R6)
> > @djgalloway - can this role be made optional?
> 
> Of course. I didn't realize were doing full legit integration testing for teuthology. Sorry about that.
> 
> Where is the log of the actual failure though? I don't see the ansible output in the GH Action log.

It's ok, I don't think this affected e.g. the sepia lab. But it'd probably be good to have something set up to avoid things like this in the future. Essentially we need to be able to run the ceph-cm-ansible testnode playbook while skipping anything that requires a secrets repo.

teuthology's docker-compose tooling doesn't give us a way to view logs after the fact (maybe something @kamoltat could look at?), and as of 685607d0f96f172c2992b9859e1f6024952f1e3f we put ansible output in `ansible.log`. I could see maybe using a volume mount for the archive directory and using [GH actions artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#downloading-or-deleting-artifacts) to expose those logfiles.

kshtsk:
@zmc I like the idea to run ansible with some roles disable, because it covers partly the cephadm job itself, however, as an option it can be disabled by `ceph_cm_ansible: false` in overrides.yaml, if we want to safe some time on ansible, maybe it would be the chose. I wonder, maybe we can copy or reuse teuthology integration workflow in ceph_cm_ansible repo to avoid such errors?

djgalloway:
I've disabled the new role by default.  It can optionally be run at the group_vars (or host_vars) level: https://github.com/ceph/ceph-cm-ansible/pull/773

kshtsk (approved):

PR #1778, opened 2022-07-13
suite: Do not send emails with --dry-run at schedule_fail

VallariAg:
Do not send emails with `--dry-run` on schedule failures. 

Signed-off-by: Vallari Agrawal <val.agl002@gmail.com>

### DIFF
--- a/teuthology/suite/test/test_init.py
+++ b/teuthology/suite/test/test_init.py
@@ -11,6 +11,7 @@ from teuthology.config import config
 import pytest
 import time
 
+from teuthology.exceptions import ScheduleFailError
 
 def get_fake_time_and_sleep():
     # Below we set m_time.side_effect, but we also set m_time.return_value.
@@ -179,6 +180,35 @@ class TestSuiteMain(object):
                     '--throttle', throttle,
                     '--machine-type', machine_type,
                 ])
+
+    @patch('teuthology.suite.util.smtplib.SMTP')
+    def test_machine_type_multi_error(self, m_smtp):
+        config.results_email = "example@example.com"
+        with pytest.raises(ScheduleFailError) as exc:
+            main([
+                    '--ceph', 'main',
+                    '--suite', 'suite_name',
+                    '--throttle', '3',
+                    '--machine-type', 'multi',
+                    '--dry-run',
+            ])
+        assert str(exc.value) == "Scheduling failed: 'multi' is not a valid machine_type. \
+Maybe you want 'gibba,smithi,mira' or similar"
+        m_smtp.assert_not_called()
+
+    @patch('teuthology.suite.util.smtplib.SMTP')
+    def test_machine_type_none_error(self, m_smtp):
+        config.result_email = 'example@example.com'
+        with pytest.raises(ScheduleFailError) as exc:
+            main([
+                    '--ceph', 'main',
+                    '--suite', 'suite_name',
+                    '--throttle', '3',
+                    '--machine-type', 'None',
+                    '--dry-run',
+            ])
+        assert str(exc.value) == "Scheduling failed: Must specify a machine_type"
+        m_smtp.assert_not_called()
 
     def test_schedule_suite_noverify(self):
         suite_name = 'noop'
--- a/teuthology/suite/test/test_run_.py
+++ b/teuthology/suite/test/test_run_.py
@@ -141,6 +141,22 @@ class TestRun(object):
         with pytest.raises(ScheduleFailError):
             self.klass(self.args)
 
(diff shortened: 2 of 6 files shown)


kamoltat:
Overall looks good, just need to add a unit test.

kamoltat (changes requested):

teuthology/suite/util.py
```diff
@@ -51,17 +51,18 @@ def fetch_repos(branch, test_name):
                 fetch_teuthology('main')
         suite_repo_path = fetch_qa_suite(branch)
     except BranchNotFoundError as exc:
-        schedule_fail(message=str(exc), name=test_name)
+        schedule_fail(message=str(exc), name=test_name, dry_run=dry_run)
     return suite_repo_path
 
 
-def schedule_fail(message, name=''):
+def schedule_fail(message, name='', dry_run=None):
```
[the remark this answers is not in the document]
VallariAg: Yep, working on it! :+1:

teuthology/suite/test/test_init.py
```diff
@@ -179,6 +180,29 @@ def fake_false(*args, **kwargs):
...
+    def test_machine_type_none_error(self):
+        with pytest.raises(ScheduleFailError) as exc:
+            main([
+                    '--ceph', 'main',
+                    '--suite', 'suite_name',
+                    '--throttle', '3',
+                    '--machine-type', 'None',
+                    '--dry-run'
+            ])
+        assert str(exc.value) == "Scheduling failed: Must specify a machine_type"
```
kamoltat, line 205: As well as unit testing these specific exceptions and making sure that the exception output is correct. We also want to make sure that with `--dry-run`, the function `util.schedule_fail()`
is not hitting `smtp.sendmail()` in `util.schedule_fail()`. We can do this by importing `teuthology.suite.util.schedule_fail`, patching it, and add something like `assert m_schedule_fail.smtp.sendmail.called == False`. You might have to mock more stuff but this should get you started.

teuthology/suite/test/test_run_.py
```diff
@@ -141,6 +141,17 @@ def test_sha1_nonexistent(
...
+    def test_teuthology_branch_nonexistent(
+        self,
+        m_git_ls_remote
+    ):
+        config.teuthology_path = None
+        m_git_ls_remote.return_value = None
+        self.args.teuthology_branch = 'no_branch'
+        with pytest.raises(ScheduleFailError):
+            self.klass(self.args)
+
```
kamoltat, line 154: same thing here

teuthology/suite/test/test_util.py
```diff
@@ -50,6 +51,17 @@ class TestUtil(object):
...
+    def test_schedule_fail(self):
+        with pytest.raises(ScheduleFailError) as exc:
+            util.schedule_fail(message="error msg")
+        assert str(exc.value) == "Scheduling failed: error msg"
+
+    def test_fetch_repo_no_branch(self):
+        with pytest.raises(ScheduleFailError) as exc:
+            util.fetch_repos("no-branch", "test1", None)
+        assert str(exc.value) == "Scheduling test1 failed: \
+Branch 'no-branch' not found in repo: https://github.com/ceph/ceph-ci.git!"
```
kamoltat, line 63: same thing here, but might not have to mock as much stuff.

kamoltat (changes requested):
Great job with implementing the unit test and making use of existing code to your advantage. Just left some comments for you to make a few changes, after that this should be good.

kamoltat:
Saw your new push, looks good! I should have addressed this in my last comment but I think we also want a unit-test case where ``--dry-run`` is not present in the command and we are expecting to hit ``smtp.sendmail``.  I know it's tedious but a good unit test makes the code more bullet-proof 😊

VallariAg:
> Saw your new push, looks good! I should have addressed this in my last comment but I think we also want a unit-test case where --dry-run is not present in the command and we are expecting to hit smtp.sendmail. I know it's tedious but a good unit test makes the code more bullet-proof 

Sounds good! I'll add that as well!

kamoltat:
@VallariAg

I grep -nr "schedule_fail" and I found that:

you're missing ``dry_run=self.args.dry_run`` in: https://github.com/ceph/teuthology/blob/526dc3720490f71840421e83bcfad1d9f4defae8/teuthology/suite/run.py#L649

VallariAg:
> I grep -nr "schedule_fail" and I found that: 
> you're missing `dry_run=self.args.dry_run` in

My bad, added the parameter now!

kamoltat (approved):
LGTM

PR #1776, opened 2022-06-28
teuthology: add pre/post fragment merge hooks

batrick:
Documentation: https://github.com/batrick/teuthology/blob/yaml-preprocessor/docs/fragment_merging.rst
Ceph PR using this: https://github.com/ceph/ceph/pull/46882

Todo:
[x] Write tests
[x] Documentation

PR #1779, opened 2022-07-22
Fix scheduling regression

zmc:
I expect these initial two commits to fail CI; I believe #1776 regressed somehow, but haven't fully debugged it yet.

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -583,12 +583,12 @@ Note: If you still want to go ahead, use --job-threshold 0'''
                                seed=self.args.seed)
         generated = len(configs)
         log.info(f'Suite {suite_name} in {suite_path} generated {generated} jobs (not yet filtered or merged)')
-        configs = config_merge(configs,
+        configs = list(config_merge(configs,
             filter_in=self.args.filter_in,
             filter_out=self.args.filter_out,
             filter_all=self.args.filter_all,
             filter_fragments=self.args.filter_fragments,
-            suite_name=suite_name)
+            suite_name=suite_name))
 
         if self.args.dry_run:
             log.debug("Base job config:\n%s" % self.base_config)
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -248,23 +248,27 @@ class Run(object):
                         log.warning(
                             'The teuthology branch config is empty, skipping')
         if not teuthology_branch:
-            teuthology_branch = config.get('teuthology_branch', 'main')
+            teuthology_branch = config.get('teuthology_branch')
 
-        if config.teuthology_path is None:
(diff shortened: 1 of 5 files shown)


zmc:
I really want to get the CI fixed, and have made a lot of progress squashing bugs that seem to only pop up in GitHub Actions, but I've got to stop for now. Currently we're failing because Ubuntu can't find `python-apt`, which I've just verified does in fact exist...

zmc:
Finally got a pass. TIL that `docker-compose build` does not update base images, even if they use `:latest`.

teuthology/suite/run.py
```diff
@@ -583,12 +587,12 @@ def schedule_suite(self):
                                seed=self.args.seed)
         generated = len(configs)
         log.info(f'Suite {suite_name} in {suite_path} generated {generated} jobs (not yet filtered or merged)')
-        configs = config_merge(configs,
+        configs = list(config_merge(configs,
```
batrick, line 590: >    The var being left as a generator led to all jobs being filtered out of    some suites, rendering them broken.

! Can you show me an example?

batrick (commented):
rest looks okay to me; but I'm not that familiar with some of that code

teuthology/suite/run.py
zmc: Try this:
`teuthology-suite -v --teuthology-branch main --ceph-repo https://github.com/ceph/ceph.git --suite-repo https://github.com/ceph/ceph.git -c main -m testnode -n 100 --suite teuthology:no-ceph --suite-branch main --dry-run`

batrick: :open_mouth: is there some esoteric python gotcha I tripped on or is this just as baffling to you?

zmc: I didn't step through everything as thoroughly as I could have so I don't know exactly how the generator is being exhausted. But, generators in python can only be iterated over once:
```
>>> l = [n for n in range(5)]  # just a regular list
>>> list(l)
[0, 1, 2, 3, 4]
>>> list(l)
[0, 1, 2, 3, 4]
>>> g = (n for n in range(5))  # this one is a generator
>>> g
<generator object <genexpr> at 0x10464f7b0>
>>> list(g)
[0, 1, 2, 3, 4]
>>> list(g)
[]  # empty!
```

batrick: Yes I'm aware of that behavior. I don't have an explanation other than a bug in Python. The `configs` generator is only iterated once AFAICT.

batrick: Oh, I see it's because of `-n`. There is a while loop around `collect_jobs`.

Guess we can't use a generator!

batrick (approved):

PR #1784, opened 2022-08-10
misc._ssh_keyscan(): Drop '-t rsa'

zmc:
It appears to be unnecessary, and causes Ubuntu 22.04 to respond with
zero keys.


djgalloway:
Confirmed this fixes.  Ubuntu 22.04 returns a `ecdsa-sha2-nistp256` key

PR #1786, opened 2022-08-19
orchestra/remote: add method to check if given path is mounted

rishabh-d-dave:
Add a method mount() that checks if the given path is mounted or
unmounted without using "mount" of "findmnt" commands.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -300,6 +300,21 @@ class RemoteShell(object):
         """
         self.write_file(path, data, sudo=True, **kwargs)
 
+    def is_mounted(self, path):
+        """
+        Check if the given path is mounted on the remote machine.
+
+        This method checks the contents of "/proc/self/mounts" instead of
+        using "mount" or "findmnt" command since these commands hang when a
+        CephFS client is blocked and its mount point on the remote machine
+        is left unhandled/unmounted.
+
+        :param path: path on remote host
+        """
+        # XXX: matching newline too is crucial so that "/mnt" does not match
+        # "/mnt/cephfs" if it's present in the output.
+        return f'{path}\n' in self.sh("cat /proc/self/mounts | awk '{print $2}'")
+
     @property
     def os(self):
         if not hasattr(self, '_os'):


rishabh-d-dave:
@batrick 

The phrase I wrote here](https://github.com/ceph/ceph/pull/45036/commits/f96031b9202b2169e2d81746d14daa702864857c#diff-a31fb9fe7c629a895231a8bd8f0f3f5874dc251130c571fb0ac40a6911c554ceR172-R173) at your suggesstion is need for [this PR too](https://github.com/ceph/ceph/pull/47533/files#diff-5ae0a15493e0d168d3d428b62e30eb9d15004daa0328bdf7eee15c07d4eee5d1R148-R149).

This code phrase in general will be useful at other similar instances too, so it'll be great to have a convenience method for it.

batrick:
> @batrick
> 
> The phrase I wrote [here](https://github.com/ceph/ceph/pull/45036/commits/f96031b9202b2169e2d81746d14daa702864857c#diff-a31fb9fe7c629a895231a8bd8f0f3f5874dc251130c571fb0ac40a6911c554ceR172-R173) at your suggesstion is need for [this PR too](https://github.com/ceph/ceph/pull/47533/files#diff-5ae0a15493e0d168d3d428b62e30eb9d15004daa0328bdf7eee15c07d4eee5d1R148-R149).
> 
> This code phrase in general will be useful at other similar instances too, so it'll be great to have a convenience method for it.

Sure, but why should this be in teuthology instead of `ceph.git/qa`?

teuthology/orchestra/remote.py
```diff
@@ -217,6 +217,17 @@ def move_file(self, src, dst, sudo=False, mode=None, owner=None,
...
+    def is_mounted(self, path):
+        """
+        Check if the given path is mounted on the remote machine.
+
+        This method checks the contents of "/proc/self/mounts" instead of
+        using "mount" or "findmnt" command since these commands hang when a
+        CephFS client is blocked and it's mount point on the remote machine
+        is left unhandled/unmounted.
+        """
+        return path in self.read_file('/proc/self/mounts', stdout=StringIO())
```
zmc, line 229: I think the idea of this is fine, but the implementation isn't quite there. As is this will return `True` if `path` is e.g. `"ext4"`, `""`, `"\n"`, etc...

zmc: Also: Why `/proc/self/mounts/` as opposed to `/proc/mounts`?

zmc (changes requested):
I like using `/proc/` instead of `mount`, but I think it'd be worth making this a bit more robust, and also adding a unit test.

teuthology/orchestra/remote.py
rishabh-d-dave: Fixed.

rishabh-d-dave:
@batrick 

> > @batrick
> > The phrase I wrote [here](https://github.com/ceph/ceph/pull/45036/commits/f96031b9202b2169e2d81746d14daa702864857c#diff-a31fb9fe7c629a895231a8bd8f0f3f5874dc251130c571fb0ac40a6911c554ceR172-R173) at your suggesstion is need for [this PR too](https://github.com/ceph/ceph/pull/47533/files#diff-5ae0a15493e0d168d3d428b62e30eb9d15004daa0328bdf7eee15c07d4eee5d1R148-R149).
> > This code phrase in general will be useful at other similar instances too, so it'll be great to have a convenience method for it.
> 
> Sure, but why should this be in teuthology instead of `ceph.git/qa`?

The exact same phrase will be handy to check if path is mounted even when FS is not CephFS. So, having a helper method will be useful. And since this method is more about the remote machine, class Remote seems to be a good home for this code-phrase.

teuthology/orchestra/remote.py
```diff
@@ -300,6 +300,23 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        This method checks the contents of "/proc/self/mounts" instead of
+        using "mount" or "findmnt" command since these commands hang when a
+        CephFS client is blocked and its mount point on the remote machine
+        is left unhandled/unmounted.
+
+        :param path: path on remote host
+
+        :raises: :class:`RuntimeError`: unexpected error occurred
+        """
+        if not os.path.isdir(path):
```
batrick, line 316: This runs on the test machine not the remote?

teuthology/orchestra/remote.py
```diff
@@ -300,6 +300,23 @@ def sudo_write_file(self, path, data, **kwargs):
...
+            raise RuntimeError(f'path {path} is not a directory')
+        return path in self.read_file('/proc/self/mounts', stdout=StringIO())
```
batrick, line 318: `/proc/self/mounts` format is well-defined. I would prefer you actually check the second field of each line rather than grep the whole file.

teuthology/orchestra/remote.py
batrick: Circling back: I don't really think you need this check at all.

batrick (changes requested):

teuthology/orchestra/remote.py
rishabh-d-dave: That would either require me to run `cat <path> | awk '{print $2}` myself. Isn't this way simpler and therefore better? 


What about `        return f'{ path}' in self.read_file('/proc/self/mounts', stdout=StringIO())` instead?

rishabh-d-dave: What about `        return f' { path} ' in self.read_file('/proc/self/mounts', stdout=StringIO())` instead?

This would avoid first and last field but not rest.

rishabh-d-dave: Using regex `"[A-Za-z0-9]* /proc/self/mounts .*"` can also get this done...

teuthology/orchestra/remote.py
rishabh-d-dave: > This runs on the test machine not the remote?

In CephFS QA code, we write `self.mount_a.client_remote.read_file()`, so `teuthology.orchestra.remote.Remote` is being used.

rishabh-d-dave: > Circling back: I don't really think you need this check at all.

Initially I thought the same, after giving [this review comment](https://github.com/ceph/teuthology/pull/1786#discussion_r950554743) some time, I thought extra error-check won't hurt.

teuthology/orchestra/remote.py
batrick: You can run the regex in python on the string `read_file` returns?

teuthology/orchestra/remote.py
batrick: I doubt that. Try it!

rishabh-d-dave: Doubt and try what? If you can be more specific, I'll know what to look out for. :)

If `path` holds the value `ext4`, without this check it might return True.

teuthology/orchestra/remote.py
rishabh-d-dave: I've been staring at this comment too for quite some time, I don't get this one too. Aren't regex run on strings in as well as beyond Python?

teuthology/orchestra/remote.py
batrick: > > This runs on the test machine not the remote?
> 
> In CephFS QA code, we write `self.mount_a.client_remote.read_file()`, so `teuthology.orchestra.remote.Remote` is being used.

This method is run for the Remote class but that doesn't change that `os.path.isdir` will be called from the testing machine. So it's looking at the file system mounted there and not on the remote.

teuthology/orchestra/remote.py
batrick: You are restricting yourself to resolving this by running a single shell command on the remote. You don't need to do that! Just read the file using `self.read_file` and then use Python's regex library to parse it.

teuthology/orchestra/remote.py
rishabh-d-dave: Thank you.

I was under the impression that you meant "check if path is dir" was unnecessary.

rishabh-d-dave:
@batrick 
Won't it be more convenient if `teuthology.orchestra.remote.Remote.read_file()` returns `StringIO` instead of `BytesIO` type? Link to `read_file()` - https://github.com/ceph/teuthology/blob/main/teuthology/orchestra/remote.py#L220

teuthology/orchestra/remote.py
```diff
@@ -300,6 +300,23 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        using "mount" or "findmnt" command since these commands hang when a
+        CephFS client is blocked and its mount point on the remote machine
+        is left unhandled/unmounted.
+
+        :param path: path on remote host
+
+        :raises: :class:`RuntimeError`: unexpected error occurred
+        """
+        self.sh(f'test -d {path}')
+        return re.search(f'\n[A-Za-z0-9-]* {path} ',
```
batrick, line 317: `path` may contain characters interpreted by `re.search`. You need to capture the path (second field) then compare to `path`.

teuthology/orchestra/remote.py
```diff
@@ -300,6 +300,23 @@ def sudo_write_file(self, path, data, **kwargs):
...
+        This method checks the contents of "/proc/self/mounts" instead of
+        using "mount" or "findmnt" command since these commands hang when a
+        CephFS client is blocked and its mount point on the remote machine
+        is left unhandled/unmounted.
+
+        :param path: path on remote host
+
+        :raises: :class:`RuntimeError`: unexpected error occurred
+        """
+        self.sh(f'test -d {path}')
```
batrick, line 316: This may hang due to a `stat` on `path` if the mount is broken. I still don't see why this is necessary if you're parsing `/proc/self/mounts`.

batrick (changes requested):

batrick:
> @batrick Won't it be more convenient if `teuthology.orchestra.remote.Remote.read_file()` returns `StringIO` instead of `BytesIO` type? Link to `read_file()` - https://github.com/ceph/teuthology/blob/main/teuthology/orchestra/remote.py#L220. This is where the default is set inside the method - https://github.com/ceph/teuthology/blob/main/teuthology/orchestra/remote.py#L241.

No. Because we may use `read_file` (or already do) on binary files like gzip archives.

batrick (approved):

zmc (approved):

PR #1787, opened 2022-08-24
boostrap: Add support for CentOSStream

kamoltat:
Currently, we cannot do bootstrap for CentOSStream
so we are adding support for that.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -63,6 +63,26 @@ Linux)
             fi
         fi
         ;;
+    CentOSStream)
+        deps=(python3-pip python39-devel mariadb-devel libev-devel libvirt-devel libffi-devel)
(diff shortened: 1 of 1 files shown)


bootstrap
```diff
@@ -63,6 +63,26 @@ Linux)
             fi
         fi
         ;;
+    CentOSStream)
+        deps=(python3-pip python36-devel mariadb-devel libev-devel libvirt-devel libffi-devel)
```
zmc, line 67: Should/can we use python39 here?

kamoltat: yes we can.

kamoltat:
done

zmc (approved):

PR #1788, opened 2022-09-01
lock: remove --list --brief example

ktdreyer:
The `--list` and `--brief` arguments are exclusive.

```
teuthology-lock: error: argument --brief: not allowed with argument --list
```


zmc (approved):
Thanks Ken!

PR #1789, opened 2022-09-02
fog: suggest image names on error

ktdreyer:
Improve the error message when the user specifies a missing OS version for a machine type.

Old error message:

```
RuntimeError: Could not find an image for rhel 7.9
```

New error message:

```
RuntimeError: Fog has no cali_rhel_7.9 image. Available cali images: ['cali_rhel_8.6', 'cali_rhel_9.0']
```

### DIFF
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -156,9 +156,20 @@ class FOG(object):
         obj = resp.json()
         if not obj['count']:
             raise RuntimeError(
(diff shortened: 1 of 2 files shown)


dmick:
This seems awesome; also awesome would be a way to invoke this to get a list of appropriate image names for a machine type without having to try one that doesn't exist

zmc (approved):
🍻 thanks for including a test!

PR #1792, opened 2022-10-09
orch/run: Add unit test xml scanner

VallariAg:
This PR adds following changes:
1. add `run_unit_test` to Remote
2. create util/scanner.py: add `Scanner`, `UnitTestScanner`, `ValgrindScanner`
3. new exception: `UnitTestError`
4. add new [lxml](https://lxml.de/) dependency in setup.cfg

reason: to throw better exception messages for failures in unit tests and valgrind errors.


teuthology/orchestra/run.py
```diff
@@ -221,6 +238,63 @@ def __repr__(self):
...
+        xml_files = xml_files.split('\n')
+        for file in xml_files:
+            error = parse_xml(file, client)
+            if error:
+                return error
+    elif os.path.splitext(xmlfile_path)[1] == ".xml": # xml file
+        error = parse_xml(xmlfile_path, client)
+        return error
+
+def parse_xml(xmlfile: str, client):
```
kamoltat, line 260: may be put this function above  ``find_unittest_error``

teuthology/orchestra/run.py
```diff
@@ -221,6 +238,63 @@ def __repr__(self):
             name=self.hostname,
             )
 
+def find_unittest_error(xmlfile_path: str, client):
+    """
+    Parse the xml file.
+    If xmlfile_path is dir, parse all xml files.
+    """
+    if not xmlfile_path:
+        return "No XML file path was passed to process!"
```
kamoltat, line 247: I think you might throw an exception here instead of returning a string since the caller of the function will receive a string and teuthology will use that as unitTestError message

teuthology/orchestra/run.py
```diff
@@ -221,6 +238,63 @@ def __repr__(self):
...
+        xml_files = stdout.read()
+        xml_files = xml_files.split('\n')
+        for file in xml_files:
+            error = parse_xml(file, client)
+            if error:
+                return error
+    elif os.path.splitext(xmlfile_path)[1] == ".xml": # xml file
+        error = parse_xml(xmlfile_path, client)
+        return error
+
```
kamoltat, line 259: nit: for better readability instead of returning error in every if, elif condition, maybe we can just return ``error`` at the very end. Declare error = None at the top.

kshtsk:
Why do you think this feature should be embedded in orchestra/run?

VallariAg:
> Why do you think this feature should be embedded in orchestra/run?

@kshtsk 
Since the unit tests are executed when the [execution-command](https://github.com/VallariAg/ceph/blob/72cfd1ff89a4d04c17ee5c5193fcd46f21bf6d1f/qa/tasks/workunit.py#L425) is passed by ceph qa to orchestra's run function, so this feature scans xml when that execution fails. The flow of code for unit test execution does not go outside orchestra/run so that's why I included this in there.    
I'm working on more changes in which I've wrapping these function in a class.
I didn't define the class somewhere else and import it here, since orchestra/run is the only place the scanner is used so it makes sense to just define it in run.py itself.

kshtsk:
> 

The run method is pretty basic method to run a remote command, it should not be overloaded with extra functionality, no specific parsing code should be there. If you want some kind of parsing of specific xml, you might add another method making subcall of run. For example, if you need some shortcut for running test which can produce java specific test, probably it makes sense to have a dedicated function that gives a clue to a user about what is it, for example, _remote.nosetest_. As an example, take a look at remote.sh method to have an idea of how the _run_ can be reused.

VallariAg:
@kshtsk @zmc 
Thank you for explaining that! :)
I'll add an extra method to remote for this feature and import scanner functions from utils.

VallariAg:
Outputs:
1. unit test using `run_unit_test` and `UnitTestScanner` with [vallariag/ceph@xml-scanner](https://github.com/VallariAg/ceph/commits/xml-scanner) ceph branch: 
   - (s3 test): https://pulpito.ceph.com/vallariag-2023-09-19_18:37:01-smoke-main-distro-default-smithi/
   - (workunit test): https://pulpito.ceph.com/vallariag-2023-09-21_11:10:03-rados:monthrash-wip-vallariag-testing-distro-default-smithi/

2. valgrind error scans using `ValgrindScanner` using [vallariag/ceph@test-valgrind-xml](https://github.com/VallariAg/ceph/commits/test-valgrind-xml) ceph branch:
   https://pulpito.ceph.com/vallariag-2023-09-19_20:10:42-rados-main-distro-default-smithi/

teuthology/util/scanner.py
```diff
@@ -0,0 +1,148 @@
...
+from lxml import etree
+
+log = logging.getLogger(__name__)
+
+
+class Scanner():
+    def __init__(self, remote=None) -> None:
+        self.summary_data = []
+        self.remote = remote
+
```
kamoltat, line 13: This is nit, but I think the order of the functions should be:
1. _parse
2. scan_file
3. scan_all_files

Just for better readability

teuthology/util/scanner.py
```diff
@@ -0,0 +1,148 @@
...
+    def scan_all_files(self, path_regex: str):
```
kamoltat, line 14: maybe include expected return types as well? since you included expected argument types?

kamoltat (changes requested):
I went through the logic and it looks good, just have very minor formatting requests.

zmc (dismissed):
Looking great!

I'm avoiding merging immediately because I won't have much time to observe today; I can merge tomorrow, or feel free to do so yourself!

kamoltat (dismissed):

zmc (dismissed):

zmc (approved):

PR #1794, opened 2022-11-08
config: Update githelper URL

zmc:


dmick:
Sure.  we can change it again if we need to.

dmick (approved):

PR #1795, opened 2022-11-14
provision/fog.py: log exceptions while looping for ssh

dmick:
Make sure the user knows of potential configuration problems, etc.

Fixes: https://tracker.ceph.com/issues/58015

### DIFF
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -278,7 +278,12 @@ class FOG(object):
                     NoValidConnectionsError,
                     MaxWhileTries,
                     EOFError,
-                ):
+                ) as e:
+                    # log this, because otherwise lots of failures just
+                    # keep retrying without any notification (like, say,
+                    # a mismatched host key in ~/.ssh/known_hosts, or
+                    # something)
+                    log.warning(e)
                     pass
         sentinel_file = config.fog.get('sentinel_file', None)
         if sentinel_file:


dmick:
forcepushed change to "log.exception" as suggested in the bug

teuthology/provision/fog.py
```diff
@@ -278,7 +278,11 @@ def _wait_for_ready(self):
                     NoValidConnectionsError,
                     MaxWhileTries,
                     EOFError,
-                ):
+                ) as e:
+                    # log this, because otherwise lots of failures just keep retrying without
+                    # any notification (like, say, a mismatched host key in ~/.ssh/known_hosts, or
+                    # something)
+                    log.exception(e)
```
zmc, line 285: `log.exception` logs the traceback automatically - the arg should be a string like "exception encountered while __"

teuthology/provision/fog.py
```diff
@@ -278,7 +278,11 @@ def _wait_for_ready(self):
...
                     pass
```
zmc, line 286:
```suggestion
```

zmc (changes requested):

teuthology/provision/fog.py
zmc, line 285:
```suggestion
                    log.error(e)
```
probably don't want to log the full tb _and_ the exception

dmick: yes, true.  I don't even think I want error; I think warning is more apropos.  (there are expected timeouts while the host comes up.)

dmick:
jenkins retest this please

zmc (approved):

PR #1796, opened 2022-11-23
task: allow internal.coredump find sysctl

kshtsk:
This fix address error on SLE platform:

  sudo: sysctl: command not found


kshtsk:
Closing, since this is problem of missing secure_path in sudoers.

PR #1798, opened 2022-12-02
Drop Python 3.6; Allow choosing Python version at runtime

zmc:


dmick (approved):
It looks right to me.

dmick (dismissed):
This all makes sense to me.

dmick:
Would love a rationale for why we need to pin postgres, but it's new code

dmick (dismissed):

dmick (approved):
If there's something that will cause problems, I can't see it.

PR #1802, opened 2022-12-08
suite: Speed up 'newest' feature

zmc:
For a 'newest' value of N, we were calling find_git_parent N times. find_git_parent was invoking githelper's 'refresh' endpoint each time, which in my testing took 20-30 seconds. So, for a sha1 that needed 20 backtracks, I was seeing teuthology-suite take eight minutes to complete. With this change, it took 30 seconds for the same teuthology-suite invocation.

We were also not including the trailing slash on the 'refresh' endpoint, resulting in a 308 redirect each time - doubling the request count. That's fixed as well.

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -644,16 +644,16 @@ Note: If you still want to go ahead, use --job-threshold 0'''
         # if not, do it once
         backtrack = 0
         limit = self.args.newest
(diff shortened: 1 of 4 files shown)


zmc:
@dmick ping, would be good to be able to merge this one

dmick (approved):
Yup.

PR #1803, opened 2022-12-20
fog: make timeouts configurable

jdurgin:
This will help with the sepia lab, being able to increase these
temporarily to handle a new fog server that is sometimes exceeding the
hardcoded timeouts.


zmc:
@jdurgin Would you mind rebasing this? There will be a conflict because of #1847 which I made before knowing about this, but it looks like this PR needs to be repushed to get the github actions run anyway. Thanks!

jdurgin:
@zmc rebased

zmc (approved):

PR #1805, opened 2023-01-17
bootstrap: Correct python3 package names for EL8/9

zmc:
`python3-devel` is only provided by `python36-devel` for whatever reason, and 3.6 is not new enough.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -59,7 +59,7 @@ Linux)
         fi
         ;;
     RedHatEnterpriseWorkstation|RedHatEnterpriseServer|RedHatEnterprise|CentOS)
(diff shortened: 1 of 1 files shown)


batrick:
\o/

batrick (dismissed):

dmick (approved):
I can't make sense of the dnf module structure, but it seems to be the case at least on RHEL8.4 that, for a module that hasn't been explicitly disabled, as many of the modules are (show neither [e] nor [x] as status), "dnf install" will ... just work, and enable the module for you too.  I don't know if that makes this fragile or not depending on module state, but I guess we can fix that when it comes up.

PR #1809, opened 2023-01-30
tasks/selinux: fix escaping on ignore_known_denials

athanatos:
a106217a swapped this line to use a raw string.  This left the initial and final ```\'``` in the output resulting in a grep -av line which didn't work as intended:

```
sudo grep -a 'avc: .*denied' /var/log/audit/audit.log | grep -av \'\(comm="dmidecode"\|chronyd.service\|name="cephtest"\|scontext=system_u:system_r:nrpe_t:s0\|scontext=system_u:system_r:pcp_pmlogger_t\|scontext=system_u:system_r:pcp_pmcd_t:s0\|comm="rhsmd"\|scontext=system_u:system_r:syslogd_t:s0\|tcontext=system_u:system_r:nrpe_t:s0\|comm="updatedb"\|comm="smartd"\|comm="rhsmcertd-worke"\|comm="setroubleshootd"\|comm="rpm"\|tcontext=system_u:object_r:container_runtime_exec_t:s0\|comm="ksmtuned"\|comm="sssd"\|comm="sss_cache"\|context=system_u:system_r:NetworkManager_dispatcher_t:s0\)\'

| grep -av \'
```

Instead, use r""" """ and don't escape the initial and trailing '.

Fixes: https://tracker.ceph.com/issues/58610
Introduced: a106217a990efa3b5af540d3cef712047f7b1d2a
Signed-off-by: Samuel Just <sjust@redhat.com>

### DIFF
--- a/teuthology/task/selinux.py
+++ b/teuthology/task/selinux.py
@@ -141,7 +141,7 @@ class SELinux(Task):
         se_allowlist = self.config.get('allowlist', [])
         if se_allowlist:
             known_denials.extend(se_allowlist)
-        ignore_known_denials = r'\'\(' + str.join(r'\|', known_denials) + r'\)\''
+        ignore_known_denials = r"""'\(' + str.join(r'\|', known_denials) + r'\)'"""
         for remote in self.cluster.remotes.keys():
             proc = remote.run(
                 args=['sudo', 'grep', '-a', 'avc: .*denied',


athanatos:
Testing in https://pulpito.ceph.com/sjust-2023-01-30_18:40:19-crimson-rados-wip-sjust-58486-distro-default-smithi/

ljflores (approved):
We'll see how the fix goes in testing, but I approve the concept.

athanatos:
(Lol, github treats \' as ', so the PR description pulled from the commit message is nonsense)

zmc:
Thank you so much @athanatos - for penance I will go write a unit test to cover this

zmc (approved):

ljflores:
> Thank you so much @athanatos - for penance I will go write a unit test to cover this

This ended up being a lot simpler to fix than fixing all the SElinux denials at least!

ljflores:
@athanatos for future reference, how did you test this PR without merging it? I've been wondering how to do that for teuthology.

teuthology/task/selinux.py
```diff
@@ -141,7 +141,7 @@ def get_denials(self):
         se_allowlist = self.config.get('allowlist', [])
         if se_allowlist:
             known_denials.extend(se_allowlist)
-        ignore_known_denials = r'\'\(' + str.join(r'\|', known_denials) + r'\)\''
+        ignore_known_denials = r"""'\(' + str.join(r'\|', known_denials) + r'\)'"""
```
zmc, line 144: This results in:
```2023-01-30T19:17:22.557 INFO:teuthology.orchestra.run.smithi170.stderr:bash: -c: line 0: `sudo grep -a 'avc: .*denied' /var/log/audit/audit.log | grep -av '\(' + str.join(r'\|', known_denials) + r'\)''```
notice we're grepping for the literal `"known_denials"`, not the actual list of strings

zmc (changes requested):

teuthology/task/selinux.py
athanatos: Hah, will fix.

athanatos:
@ljflores teuthology-suite has an option for the teuthology branch to use.  -t I think

ljflores:
> @ljflores teuthology-suite has an option for the teuthology branch to use. -t I think

Ah, good to know!

teuthology/task/selinux.py
zmc: I'm also testing a different approach at https://github.com/ceph/teuthology/pull/1810, which eliminates the need to escape.

athanatos: Ah, feel free to close this one if you prefer that one.  Either way, let's run a small suite prior to merge though.

zmc: I'm rerunning the same suite you'd scheduled; we'll see if it passes. I'd thought of that approach because it'd be far easier to unit-test, but to me it sort of feels like there won't be much to test in that case as we wouldn't be generating a confusing-looking escaped string.

athanatos: Yep, works for me.  https://github.com/ceph/teuthology/pull/1810 seems cleaner.

ljflores: Update the Tracker with whatever PR ends up being used

zmc:
Closing in favor of #1810

PR #1811, opened 2023-02-01
misc.get_system_type: Use Remote.os

zmc:
... instead of lsb_release, as newer distros don't ship it.

Fixes: #58491


cbodley (approved):
looks great if it works! a 'grep lsb_release' found some matches in other files, but hopefully they won't matter

PR #1812, opened 2023-02-03
orchestra.connection: Look for 'ssh_key' in config

zmc:
On systems where SSH is configured using ~/.ssh/config.d/, we don't have a great way of locating and processing the full configuration, so we might not find the correct key to use. Allow the user to specify one key to use for all test nodes if they wish.

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -226,6 +226,10 @@ Here is a sample configuration with many of the options set and documented::
     conserver_master: conserver.front.sepia.ceph.com
     conserver_port: 3109
 
+    # Optionally use a specific SSH private key to connect to test nodes.
+    # Takes precedence over any entries in ~/.ssh/config.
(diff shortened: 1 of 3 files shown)


docs/siteconfig.rst
```diff
@@ -226,6 +226,10 @@ Here is a sample configuration with many of the options set and documented::
     conserver_master: conserver.front.sepia.ceph.com
     conserver_port: 3109
 
+    # Optionally use a specific SSH private key to connect to test nodes.
+    # Takes precendence over any entries in ~/.ssh/config.
```
dmick, line 230: precedence

zmc:
@dmick ping

dmick:
I don't know when ~/.ssh/config.d might exist, but I also can't see a problem in another way to specify a key

dmick (approved):

PR #1816, opened 2023-02-28
Improve tolerance of paddles faults

zmc:


dmick (approved):
I guess.  Endless loops always worry me, but...desperate times.

zmc:
> I guess. Endless loops always worry me, but...desperate times.

It's a daemon!

teuthology/lock/query.py
```diff
@@ -51,7 +51,7 @@ def is_vm(name=None, status=None):
     return status.get('is_vm', False)
 
 
-def list_locks(keyed_by_name=False, **kwargs):
+def list_locks(keyed_by_name=False, tries=10, **kwargs):
```
kshtsk, line 54: @zmc ...so, please, tell me why there is the _tries_ parameter if it is unused? Also giving an endless loop makes me frustrating when I run `teuthology_suite` to without lock server configured, and given that timeout is zero... I am experiencing trouble to find point where the error occurred.

zmc: The `tries` parameter is not unused.

I don't understand the issue you're describing, but this is really not the place to report or discuss it. A tracker ticket perhaps?

kshtsk: @zmc tracker ticket looks overkill for me for sharing some thoughts, full of formalism and bureaucracy :-)

kshtsk: > The `tries` parameter is not unused.

Look carefully, there is no usage of `tries` value in `list_locks`.

teuthology/lock/query.py
```diff
@@ -63,7 +63,11 @@ def list_locks(keyed_by_name=False, **kwargs):
             kwargs['machine_type'] = kwargs['machine_type'].replace(',','|')
         uri += '?' + urlencode(kwargs)
     with safe_while(
-            sleep=1, increment=0.5, action='list_locks') as proceed:
+            sleep=1,
+            increment=0.5,
+            tries=-1,
```
kshtsk, line 68: It's hardcoded here to be infinite.

PR #1817, opened 2023-03-02
Add teuthology-exporter, a Prometheus exporter

zmc:


dmick (approved):
All looks fine to me.  Should also update docs somewhere to note its existence, capabilities, and port number.

dmick (approved):

zmc:
Agreed, docs will be useful. I'm sure there will be follow-up PRs, so I'll merge this in order to start picking up job results.

teuthology/beanstalk.py
```diff
@@ -6,8 +6,9 @@
 import sys
 from collections import OrderedDict
 
+import teuthology.report as report
```
kamoltat, line 9: Sorry but what's the purpose of doing this? why can't we just ``from teuthology import report``

teuthology/dispatcher/__init__.py
```diff
@@ -7,18 +6,20 @@
 import yaml
 
 from datetime import datetime
+from typing import Dict, List
+
+import teuthology.dispatcher.supervisor as supervisor
+import teuthology.lock.ops as lock_ops
+import teuthology.nuke as nuke
```
kamoltat, line 13: Again there must a good reason you did this as oppose to ``from teuthology.nuke import nuke``, please explain.

teuthology/kill.py
```diff
@@ -84,6 +86,10 @@ def kill_job(run_name, job_id, archive_base=None, owner=None, skip_nuke=False):
                 "Please pass --owner <owner>.")
         owner = job_info['owner']
     kill_processes(run_name, [job_info.get('pid')])
+    if 'machine_type' in job_info:
```
kamoltat, line 89: maybe ``if job_info.get('machine_type')`` ?

teuthology/kill.py
```diff
@@ -84,6 +86,10 @@ def kill_job(run_name, job_id, archive_base=None, owner=None, skip_nuke=False):
...
+        teuthology.exporter.JobResults().record(job_info["machine_type"], job_info["status"])
```
kamoltat, line 90: what if  ``job_info["status"]`` is None, wanna also add `` and job_info.get('status')`` to the if statement?

teuthology/kill.py
```diff
@@ -84,6 +86,10 @@ def kill_job(run_name, job_id, archive_base=None, owner=None, skip_nuke=False):
...
+    else:
+        log.warn(f"Job {job_id} has no machine_type; cannot report via Prometheus")
```
kamoltat, line 92: wanna add the case where it has no `status` as well?

teuthology/report.py
```diff
@@ -579,6 +583,8 @@ def try_mark_run_dead(run_name):
             try:
                 log.info("Marking job {job_id} as dead".format(job_id=job_id))
                 reporter.report_job(run_name, job['job_id'], dead=True)
+                if "machine_type" in job:
```
kamoltat, line 586: same as comment above

kamoltat (commented):
Thank you for adding this PR, left some comments. Other than that everything looks great!

teuthology/beanstalk.py
zmc: You were right about this; my changes were more heavy-handed than they needed to be! The follow-up PR reverts the unnecessary bits.

teuthology/dispatcher/__init__.py
zmc: You're correct here as well.

teuthology/kill.py
zmc: I think if it is present in the dict but has a value of `None`, we may want to have that reported via prometheus; it would indicate something very odd went wrong.

teuthology/kill.py
zmc: hm actually I think we might want to always use the literal `'dead'` here, since we've just killed the process itself!

teuthology/report.py
zmc: it would be exceptionally strange (read: impossible?) for this to not be a real value, since paddles won't accept a job with no machine type

ivancich:
I'm still getting a circular import error when trying to invoke `teuthology-kill`. I'm using:
commit 7f7f7c33a372789d0252bbbcd5798300b0f044d0 (HEAD -> main, origin/main)

  File "/home/ivancich/src/teuthology/virtualenv/bin/teuthology-kill", line 5, in <module>
    from scripts.kill import main
  File "/home/ivancich/src/teuthology/scripts/kill.py", line 4, in <module>
    import teuthology.kill
  File "/home/ivancich/src/teuthology/teuthology/kill.py", line 12, in <module>
    import teuthology.exporter
  File "/home/ivancich/src/teuthology/teuthology/exporter.py", line 10, in <module>
    import teuthology.dispatcher
  File "/home/ivancich/src/teuthology/teuthology/dispatcher/__init__.py", line 11, in <module>
    import teuthology.dispatcher.supervisor as supervisor
  File "/home/ivancich/src/teuthology/teuthology/dispatcher/supervisor.py", line 22, in <module>
    from teuthology.kill import kill_job
ImportError: cannot import name 'kill_job' from partially initialized module 'teuthology.kill' (most likely due to a circular import) (/home/ivancich/src/teuthology/teuthology/kill.py)

PR #1819, opened 2023-03-09
Add timing instrumentation to the exporter

zmc:
This adds data for time spent while:
* locking nodes
* reimaging nodes
* running the bootstrap script
* running a job
* running each phase of each task in a job

### DIFF
--- a/teuthology/dispatcher/__init__.py
+++ b/teuthology/dispatcher/__init__.py
@@ -14,6 +14,7 @@ from teuthology import (
     install_except_hook,
     # modules
     beanstalk,
+    exporter,
     nuke,
     report,
     repo_utils,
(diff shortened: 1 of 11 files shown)


zmc:
@dmick this PR adds docs too :)

teuthology/dispatcher/__init__.py
```diff
@@ -224,13 +225,16 @@ def match(proc):
...
-        tries=-1,
-        reimage=False,
-    )
+    machine_type = job_config["machine_type"]
+    count = len(job_config['roles'])
+    with exporter.NodeLockingTime.labels(machine_type, count).time():
+        lock_ops.block_and_lock_machines(
+            fake_ctx,
+            count,
+            job_config['machine_type'],
```
dmick, line 234: since you now have machine_type, could replace this

zmc: oh thank you, I definitely intended to

teuthology/kill.py
```diff
@@ -87,7 +87,10 @@ def kill_job(run_name, job_id, archive_base=None, owner=None, skip_nuke=False):
         owner = job_info['owner']
     kill_processes(run_name, [job_info.get('pid')])
     if 'machine_type' in job_info:
-        teuthology.exporter.JobResults().record(job_info["machine_type"], job_info["status"])
+        teuthology.exporter.JobResults().record(
+            job_info["machine_type"],
+            job_info.get("status", "dead")
```
zmc, line 92: @kamoltat you were right about needing a change here

dmick (approved):
All looks good and right to me.

PR #1821, opened 2023-03-13
suite.placeholder: Speed up ansible.cephlab

zmc:
By skipping the tags "user,pubkeys,nagios,nrpe" we can save ~3min per job, assuming users won't need to ssh in to a testnode while a job is running - which is probably very close to 100% of jobs.


zmc:
Testing here: https://pulpito.ceph.com/zack-2023-03-13_23:47:48-smoke-main-distro-default-smithi/

zmc:
The original run is still enqueued; I scheduled this with a better priority and all jobs passed:
https://pulpito.ceph.com/zack-2023-03-22_22:52:17-smoke-main-distro-default-smithi/

zmc:
Tested here: https://pulpito.ceph.com/zack-2023-03-22_22:52:17-smoke-main-distro-default-smithi/
<details><summary>ansible timing (skipping nothing - current main branch behavior)</summary>
<p>

```
Thursday 23 March 2023  05:16:06 +0000 (0:00:00.427)       0:07:42.674 ******** 
=============================================================================== 
testnode : Install Amazon::S3. ---------------------------------------- 106.82s
users : Create all admin users with sudo access. ----------------------- 59.39s
users : Update authorized_keys using the keys repo --------------------- 46.26s
testnode : Ensure perl-doc and cpanminus is installed on apt systems. -- 23.61s

container-host : Install container packages ---------------------------- 14.29s
testnode : Upgrade packages --------------------------------------------- 5.86s
users : Remove revoked users -------------------------------------------- 4.63s

testnode : Create logical volume(s) ------------------------------------- 3.24s
users : Update authorized_keys for each user with literal keys ---------- 2.69s

Gathering Facts --------------------------------------------------------- 2.57s

container-host : Restart docker service --------------------------------- 2.43s
testnode : Update apt cache. -------------------------------------------- 2.14s

testnode : Install apt keys --------------------------------------------- 1.81s
container-host : Install registries-conf-ctl ---------------------------- 1.80s

testnode : List any leftover Ceph artifacts from previous jobs ---------- 1.26s

testnode : Write /scratch_devs ------------------------------------------ 1.24s
Gathering Facts --------------------------------------------------------- 1.22s

testnode : Install packages via pip ------------------------------------- 1.18s
Gathering Facts --------------------------------------------------------- 1.14s

ansible-managed : Create the cephlab_sudo sudoers.d file. --------------- 1.08s
```
</p>
</details>

kshtsk:
so, what is the decision?

kshtsk:
took another look, looks like it is good speed improvement, just wondering if there is a sense to include `user,pubkeys` by user request, like a teuthology-suite options, so it can override this options... or instead of doing this via ansible there is more sense to implement this is a provisioning feature
from my experience, when we run teuthology tests downstream, or "hypothetically" if the one might want to run test on customer side, they for sure don't want to give access to their nodes for anyone from ceph/keys

kshtsk:
> assuming users won't need to ssh in to a testnode while a job is running

it would probably great to allow user to ssh while job is running, however, it is better to allow this only owner of the job and probably admins

PR #1822, opened 2023-03-16
kill.kill_job: Tweak job status reporting

zmc:
This fixes a regression in `teuthology-kill`


dmick (approved):
Trying to enumerate all the paths that might set job_info makes my head hurt, but this seems harmless.

kamoltat (approved):
LGTM

PR #1826, opened 2023-03-23
sueprvisor: Do not instrument certain job times

zmc:
This should only really include first/last-in-suite jobs.


kamoltat (approved):
LGTM!

PR #1829, opened 2023-03-28
lock.ops.reimage_machines: Drop incorrect log msg

zmc:
This message was being logged when the reimage started, not finished.


zmc:
This commit got merged in #1849

PR #1831, opened 2023-03-30
teuthology-suite: --seed & --subset now also stored in teuthology.log, config.yaml and orig.config.yaml

kamoltat:
Notify the user if `results.log` is missing when
they issue a rerun.

1. Edited teuthology-suite doc to
inform the user about how `--rerun` by
default parse `--seed`, `--subset` and
`--no-nested-subset` from `results.log`
by default.

2. In addition to being stored in `results.log`
`--seed` and `--subset` are now also stored in:
`teuthlogy.log`, `config.yaml` and `orig.config.yaml`.
Therefore, in the case where `results.log` is missing, the user can explicitly pass `--seed` and `--subset`.

Fixes: https://tracker.ceph.com/issues/59118

### DIFF
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -166,8 +166,11 @@ def get_rerun_conf(conf):
     try:
         subset, no_nested_subset, seed = reporter.get_rerun_conf(conf.rerun)
     except IOError:
+        log.error('Error accessing results.log, file might be missing.')
+        log.warning('Using default/specified values for --seed, --subset and --no-nested-subset')
         return conf.subset, conf.no_nested_subset, conf.seed
     if seed is None:
+        log.warning('Missing seed in results.log, using default/specified values for --seed')
         return conf.subset, conf.no_nested_subset, conf.seed
     if conf.seed < 0:
         log.info('Using stored seed=%s', seed)
@@ -185,6 +188,7 @@ def get_rerun_conf(conf):
                   stored_subset=subset)
     if conf.no_nested_subset is True:
         log.info('Nested subsets disabled')
+        no_nested_subset = conf.no_nested_subset
     return subset, no_nested_subset, seed
 
 
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -121,7 +121,8 @@ Scheduler arguments:
                               contain each facet at least once) and schedule
                               piece <index>.  Scheduling 0/<outof>, 1/<outof>,
                               2/<outof> ... <outof>-1/<outof> will schedule all
-                              jobs in the suite (many more than once).
+                              jobs in the suite (many more than once). If specified,
+                              this value can be found in results.log.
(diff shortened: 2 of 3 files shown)


teuthology/suite/__init__.py
```diff
@@ -185,6 +188,7 @@ def get_rerun_conf(conf):
                   stored_subset=subset)
     if conf.no_nested_subset is True:
         log.info('Nested subsets disabled')
+        no_nested_subset = conf.no_nested_subset
```
kamoltat, line 191: no_nested_subset should get replaced with a new value when we explicitly pass the no_nested_subset flag.

dmick:
Please fix the spelling in commit msg and PR title before merging.  (for searches, if nothing else.)

kamoltat:
> Please fix the spelling in commit msg and PR title before merging. (for searches, if nothing else.)

whoops thanks!

zmc:
This is a good change! I'm curious what you think about the idea of actually refusing to schedule if we can't access results.log and `--seed` is not passed.

kamoltat:
> This is a good change! I'm curious what you think about the idea of actually refusing to schedule if we can't access results.log and `--seed` is not passed.

@zmc So I have thought about that and I've decided to still allow users to schedule reruns even without results.log because there are some runs that doesn't contain any jobs that are picked at random and also this leaves a work around where the user can just schedule a job with --seed and just remember the seed number that they've used and schedule a rerun by explicitly inputting --seed in the case where results.log is missing.

kamoltat:
Tested this PR by scheduling dummy jobs, you can look for `seed` & `subset` values here:

Ran the job with the command:

`./virtualenv/bin/teuthology-suite -v -m smithi -c main --suite-repo https://github.com/ceph/ceph-ci.git --suite-branch main --subset 300/100000 -s dummy --limit 1 -p 50 --force-priority -n 100`

- ksirivad-2023-04-03_18:32:51-dummy-main-distro-default-smithi/7230403/teuthology.log
- ksirivad-2023-04-03_18:32:51-dummy-main-distro-default-smithi/7230403/config.yaml
- ksirivad-2023-04-03_18:32:51-dummy-main-distro-default-smithi/7230403/orig.config.yaml
- ksirivad-2023-04-03_18:32:51-dummy-main-distro-default-smithi/results.log

zmc (approved):
@kamoltat Your explanation re: allowing `--seed` when we can't find the value makes sense - thanks for the explanation!

I'm not at all opposed to storing the seed in [orig.]config.yaml, but to make `--seed` work really well with `--rerun` we're going to want it in paddles too. At that point it won't really matter what files contain it.

If a paddles change doesn't seem reasonable in the short term, I wonder if we should be dropping a very small yaml file in the root of the run's archive dir containing the value when the first job _starts_, rather than waiting until the whole run finishes to put it in a logfile which then needs to be parsed.

In any case, this is ready to merge!

PR #1835, opened 2023-04-06
teuthology: fix the problem that git_build_url defaults to github

wolaidajiangyou:

The build_git_url function creates a github connection by default when the project name does not belong to one of ceph,ceph-cm-ansible,ceph-qa-suite. Modify the function to achieve that if project name is not one of the above three, create the ceph_git_base_url connection in the teutology.yaml file.

### DIFF
--- a/teuthology/repo_utils.py
+++ b/teuthology/repo_utils.py
@@ -51,9 +51,9 @@ def build_git_url(project, project_owner='ceph'):
     elif project == 'ceph':
         base = config.get_ceph_git_url()
     else:
-        base = 'https://github.com/{project_owner}/{project}'
+	base = '{ceph_git_base_url}/{project_owner}/{project}'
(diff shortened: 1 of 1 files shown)


teuthology/repo_utils.py
```diff
@@ -51,9 +51,9 @@ def build_git_url(project, project_owner='ceph'):
     elif project == 'ceph':
         base = config.get_ceph_git_url()
     else:
-        base = 'https://github.com/{project_owner}/{project}'
+	base = '{ceph_git_base_url}/{project_owner}/{project}'
```
kshtsk, line 54: According to https://github.com/ceph/teuthology/blob/cbf9f04aba99fabd05defc976000254a396fab38/docs/siteconfig.rst?plain=1#L66

the ceph_git_base_url is already a combination of host and project owner, so this will not work as expected.

PR #1836, opened 2023-04-06
teuthology:add the specified project_owner function

wolaidajiangyou:
    Signed-off-by: Yang Honggang <yanghonggang_yewu@cmss.chinamobile.com>

add the specified project_owner function,
project_owner is the name of the directory above the source code


kshtsk:
Don't you mind to elaborate on why this variables needed and where they are going to be used and for.

wolaidajiangyou:
> Don't you mind to elaborate on why this variables needed and where they are going to be used and for.

For example, for the following URL: https://github.com/ceph/teuthology, its projects_owner is "ceph", but for the teuthology source code under other users, the project_owner is not "ceph" but his own username. Not adding the project_owner configuration item will cause the teuthology code to not be updated normally when the user specifies the teuthology source code of his own private warehouse.

teuthology/suite/util.py
```diff
@@ -201,6 +201,7 @@ def git_validate_sha1(project, sha1, project_owner='ceph'):
     we have specific urls to HEAD for github and git.ceph.com/gitweb
     for now
     '''
+    project_owner=config.project_owner
```
kshtsk, line 204: This make `project_owner` argument useless.

Instead, if we really want to add this change, it might look like:
```
def git_validate_sha1(project, sha1, project_owner=None):
    url = build_git_url(project, project_owner or config.project_owner)
```

kshtsk:
Have you tested this change and do you have plans to finish it?

PR #1837, opened 2023-04-06
teuthology: add the function of finding the commit correctly when the…

wolaidajiangyou:
… code is not in github

add the function of finding the commit correctly when the code is not in github

### DIFF
--- a/teuthology/suite/util.py
+++ b/teuthology/suite/util.py
@@ -209,6 +209,8 @@ def git_validate_sha1(project, sha1, project_owner='ceph'):
         # kinda specific to knowing git.ceph.com is gitweb
         url = ('http://git.ceph.com/?p=%s.git;a=blob_plain;f=.gitignore;hb=%s'
                % (project, sha1))
+    elif config.ceph_git_base_url in url:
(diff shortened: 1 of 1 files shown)


teuthology/suite/util.py
```diff
@@ -209,6 +209,8 @@ def git_validate_sha1(project, sha1, project_owner='ceph'):
         # kinda specific to knowing git.ceph.com is gitweb
         url = ('http://git.ceph.com/?p=%s.git;a=blob_plain;f=.gitignore;hb=%s'
                % (project, sha1))
+    elif config.ceph_git_base_url in url:
+        url = '/'.join((url, 'commit', sha1))
```
kshtsk, line 213: The code of the method is making request based on the url of git repo in different services, in particular it uses REST api of GitHub and GitWeb based on the repo url. I can't understand why variable url should have the format which is suggested within the change, and which use cases it can be utilized.

kshtsk:
I'm afraid the PR can't be merged as is, and I really don't think the PR has well developed idea how to implement the request. Since it remained unchanged for a long time closing for now. Feel free to create another PR with better description and ready to review changes.

PR #1838, opened 2023-04-06
teuthology: add the function of specifying teuthology url

wolaidajiangyou:
add the function of specifying teuthology url

### DIFF
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -171,6 +171,7 @@ class TeuthologyConfig(YamlConfig):
         'koji_task_url': 'https://kojipkgs.fedoraproject.org/work/',
         'baseurl_template': 'http://{host}/{proj}-{pkg_type}-{dist}-{arch}-{flavor}/{uri}',
         'use_shaman': True,
+	'teuthology_git_url': 'https://github.com/ceph/teuthology.git',
         'shaman_host': 'shaman.ceph.com',
         'teuthology_path': None,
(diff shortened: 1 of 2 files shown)


teuthology/config.py
```diff
@@ -171,6 +171,7 @@ class TeuthologyConfig(YamlConfig):
         'koji_task_url': 'https://kojipkgs.fedoraproject.org/work/',
         'baseurl_template': 'http://{host}/{proj}-{pkg_type}-{dist}-{arch}-{flavor}/{uri}',
         'use_shaman': True,
+	'teuthology_git_url': 'https://github.com/ceph/teuthology.git',
```
kshtsk, line 174: This should be None by default, same as ceph_git_url and others.

teuthology/config.py
```diff
@@ -209,6 +210,9 @@ def get_ceph_qa_suite_git_url(self):
     def get_ceph_git_url(self):
         return (self.ceph_git_url or
                 self.ceph_git_base_url + 'ceph-ci.git')
+    def get_teuthology_git_url(self):
+        return self.teuthology_git_url
```
kshtsk, line 214: this might be the same as the other similar methods, like:
```
        return (self.teuthology_git_url or
                 self.ceph_git_base_url + 'teuthology.git')
```

kshtsk:
No response from the contributor for a long time, but I like the idea.
Closing this PR in favor of #2075 
Thank you

PR #1839, opened 2023-04-06
remove(ntpd): remove the ntpd operation

wolaidajiangyou:
Signed-off-by: Sun Junnan sunjunnan_yewu@cmss.chinamobile.com

The ntp service conflicts with the ntpdate-debian command.
(https://ubuntu.com/server/docs/network-ntp) 
Stop the ntp service to execute the ntpdate-debian command to synchronize the system time.
Then hwclock systohc writes the synchronized system time into the rtc hardware clock .
Finally restore the ntp service.

### DIFF
--- a/teuthology/nuke/actions.py
+++ b/teuthology/nuke/actions.py
@@ -404,20 +404,13 @@ def synch_clocks(remotes):
     for remote in remotes:
         remote.run(
             args=[
-                'sudo', 'systemctl', 'stop', 'ntp.service', run.Raw('||'),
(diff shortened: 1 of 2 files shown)


dmick:
The commit message is not at all what the code does?....

kshtsk:
This patch is removing the code which is used by other distros. I don't think it can be merged as is, also... would be great if there be a reference to log with the issue it resolves.

kshtsk:
no updates for a long time, feel free to reopen if you have time to rework the PR

PR #1842, opened 2023-04-30
task/install/packages.yaml: Update packages.yaml from ceph/main 

Matan-B:
From [get_package_list()](https://github.com/ceph/teuthology/blob/main/teuthology/task/install/__init__.py#LL133C5-L133C21), In the case where:
```
    # If packages.yaml isn't found in the suite_path, potentially use
    # teuthology's
```
Teuthology's `packages.yaml` will be used and since some unsupported packages (like `cephfs_java`) is listed in it - we will fail to install it in each job. (See https://github.com/ceph/ceph/pull/49711 for  `cephfs_java` removal discussion).
```
2023-04-30T12:53:36.334 DEBUG:teuthology.orchestra.run.smithi162:> sudo yum -y install cephfs-java
2023-04-30T12:53:37.090 INFO:teuthology.orchestra.run.smithi162.stdout:Last metadata expiration check: 0:21:53 ago on Sun 30 Apr 2023 12:31:44 PM UTC.
2023-04-30T12:53:38.205 INFO:teuthology.orchestra.run.smithi162.stdout:No match for argument: cephfs-java
2023-04-30T12:53:38.254 INFO:teuthology.orchestra.run.smithi162.stderr:Error: Unable to find a match: cephfs-java
2023-04-30T12:53:38.278 DEBUG:teuthology.orchestra.run:got remote process result: 1
```
From: https://pulpito.ceph.com/matan-2023-04-30_11:05:47-crimson-rados-main-distro-crimson-smithi/

Note: I'm not sure why the run attached here fell back to Teuthology's `packages.yaml` yet. However, this case shouldn't cause the whole run to fail. 

Signed-off-by: Matan Breizman <mbreizma@redhat.com>

### DIFF
--- a/teuthology/task/install/__init__.py
+++ b/teuthology/task/install/__init__.py
@@ -145,10 +145,12 @@ def get_package_list(ctx, config):
             yaml_path = suite_packages_path
     # If packages.yaml isn't found in the suite_path, potentially use
     # teuthology's
-    yaml_path = yaml_path or os.path.join(
-        os.path.dirname(__file__),
-        'packages.yaml',
-    )
+    if not yaml_path:
+        log.info("Falling back to Teuthology's packages.yaml")
+        yaml_path = os.path.join(
+            os.path.dirname(__file__),
+            'packages.yaml',
+        )
     default_packages = yaml.safe_load(open(yaml_path))
     default_debs = default_packages.get(project, dict()).get('deb', [])
     default_rpms = default_packages.get(project, dict()).get('rpm', [])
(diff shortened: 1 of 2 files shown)


kamoltat (approved):
LGTM @Matan-B thanks for doing this. Can you please use this branch to reschedule some of the failed jobs that required cephfs-java, as a sanity check? Thank you

Matan-B:
> LGTM @Matan-B thanks for doing this. Can you please use this branch to reschedule some of the failed jobs that required cephfs-java, as a sanity check? Thank you

The jobs didn't require cephfs-java, no job should require it (no longer in use). I'm not sure why the attached run fell back to teu's packages.yaml since I was using default qa dir.

PR #1844, opened 2023-05-17
task: install supports enable_coprs array

cbodley:
enable the installation of packages in fedora copr repositories

this should unblock testing against centos9 by using the `ceph/el9` copr for missing python packages

in ceph, https://github.com/ceph/ceph/pull/50441 adds a `centos_9.stream.yaml` file, and i plan to add an override there:
```yaml
overrides:
  install:
    ceph:
      enable_coprs: [ceph/el9]
```

### DIFF
--- a/teuthology/task/install/rpm.py
+++ b/teuthology/task/install/rpm.py
@@ -173,6 +173,13 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
     :param rpm: list of packages names to install
     :param config: the config dict
     """
+
+    enable_coprs = config.get('enable_coprs', [])
+    if len(enable_coprs):
+        remote.run(args=['sudo', 'dnf', '-y', 'install', 'dnf-command(copr)'])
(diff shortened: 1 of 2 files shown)


ktdreyer (approved):

teuthology/task/install/rpm.py
```diff
@@ -173,6 +173,13 @@ def _update_package_list_and_install(ctx, remote, rpm, config):
     :param rpm: list of packages names to install
     :param config: the config dict
     """
+
+    enable_coprs = config.get('enable_coprs', [])
+    if len(enable_coprs):
```
zmc, line 178: This could also simply be `if enable_coprs:` since an empty list is falsey - no need to change this, of course.

zmc (approved):
Looks great!

teuthology/task/install/rpm.py
cbodley: good to know, thanks

PR #1845, opened 2023-05-18
suite: fix typo osd pool default size

NitzanMordhai:
The 'osd pool default size' config had typo, we didn't took it and actually ignored it until this fix


NitzanMordhai:
All, we need to be careful with that suggestion, it means that we are going to fix something that will affect all other jobs that doesn't override that pool default size, if you have any comment please let me know!

NitzanMordhai:
I removed it completely, we don't want other jobs to start failing because of it

zmc:
Hi Nitzan,

I don't quite understand this PR; it looks like I'm seeing half a conversation. We've been using this setting for close to ten years: https://github.com/ceph/teuthology/commit/1b4368b3cf093c4692e128b01a57a0c950cfb68f

NitzanMordhai:
> Hi Nitzan,
> 
> I don't quite understand this PR; it looks like I'm seeing half a conversation. We've been using this setting for close to ten years: [1b4368b](https://github.com/ceph/teuthology/commit/1b4368b3cf093c4692e128b01a57a0c950cfb68f)

There is non setting that called 'osd default pool size' there is 'osd pool default size', it is confusing when you are looking on logs and seeing it and if you are not looking close enough you may think that you will use size =2 but actually you are using the something else.
if i'll fix it, we will start using 2 as default, which may cause other suites to fail (after 10 years that they use the real default...)

kamoltat:
@NitzanMordhai  Okay this is my understanding, so basically the command never gets triggered because of  the typo, rather than correcting the typo, you want to remove it completely because the cost of correcting it will affect other suite in a unpredictable way. However, you cannot leave it be because the command is misleading towards developers and users.

kamoltat:
@NitzanMordhai 
Also to my understanding, the command should be `osd pool default size: 2` not `osd default pool size: 2` Correct?

jdurgin (approved):
nice noticing this! so this makes it purely a cleanup

zmc:
@NitzanMordhai Excellent, thank you for the explanation (and apologies this fell off my radar)!

It's kind of funny that this sat there for so long. I'm noticing this parameter was for ceph-deploy; hopefully other tasks are doing more positive validation of their settings 😉

NitzanMordhai:
> Also to my understanding, the command should be `osd pool default size: 2` not `osd default pool size: 2` Correct?

Correct, we probably should consider some fix in the config set area in case we are setting wrong key and not just wrong values.

PR #1846, opened 2023-05-18
teuthology/task/install: add stdin-killer helper

batrick:
This helper tool runs commands which may or may not take data on stdin.
    Like "daemon-helper", if stdin signals EOF, stdin-killer will kill the
    command but only as a last resort. It forwards EOF to the command by
    closing the command's stdin (pipe) and then waiting a configurable
    amount of time for the command to gracefully exit.
    
    Additionally, if stdout or stderr are hung up -- i.e. the ssh parent
    process has terminated -- then stdin-killer also detects this and
    initiates the graceful shutdown of the command. This is something
    daemon-helper does not do.
    
    In general, this tool is a superior replacement of the daemon-helper
    tool because you can write to the command's stdin normally.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,6 +62,7 @@ python_requires = >=3.6
 scripts =
     teuthology/task/install/bin/adjust-ulimits
     teuthology/task/install/bin/daemon-helper
+    teuthology/task/install/bin/stdin-killer
 
 [options.entry_points]
 console_scripts =
@@ -128,6 +129,7 @@ teuthology.suite =
 teuthology.task.install =
     bin/adjust-ulimits
     bin/daemon-helper
+    bin/stdin-killer
 teuthology.task.internal =
     edit_sudoers.sh
 
--- a/teuthology/task/install/bin/stdin-killer
+++ b/teuthology/task/install/bin/stdin-killer
@@ -1,0 +1,251 @@
+#!/bin/python3
+
+# Forward stdin to a subcommand. If EOF is read from stdin or
+# stdin/stdout/stderr are closed or hungup, then give the command "timeout"
+# seconds to complete before it is killed.
+#
+# The command is run in a separate process group. This is mostly to simplify
+# killing the set of processes (if well-behaving). You can configure that with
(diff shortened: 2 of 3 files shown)


batrick:
For now I only intend to use this in CephFS QA testing. So potential fallout from merging this should be close to minimal.

zmc:
https://pulpito.ceph.com/pdonnell-2023-05-23_18:20:18-fs-wip-pdonnell-testing-20230523.134409-distro-default-smithi/

batrick:
added some comments and made the `setpgrp` behavior configurable

zmc (commented):
Looks great, no obvious bugs. I do wonder if it might be sensible to log (and/or make configurable) that SIGKILL is being sent.

Also, would you mind running `stdin-killer` through `black`?

batrick:
> Looks great, no obvious bugs. I do wonder if it might be sensible to log (and/or make configurable) that SIGKILL is being sent.

I'm confused, doesn't it?

> Also, would you mind running `stdin-killer` through `black`?

Done.

zmc:
> > Looks great, no obvious bugs. I do wonder if it might be sensible to log (and/or make configurable) that SIGKILL is being sent.
> 
> I'm confused, doesn't it?

Now I'm confused :)
I'm thinking of this: https://github.com/ceph/teuthology/pull/1846/files#diff-41aca9dcd85c8bfa893d3ce8489dc2fd81573a7e2c49e6b6c3535cf93e3726beR243-R245

You mention that you send `NS.signal` to the wrapped command, and to `stdin-killer` itself. After that signal is sent, `SIGKILL` is also sent afterward - but I don't see a log message or a mention in the help text of the command that that will happen. I guess my concern is someone will use e.g. `--signal SIGTERM`, see unexpected behavior after the `SIGKILL`, and not understand what happened internally.

> > Also, would you mind running `stdin-killer` through `black`?

Thanks!

batrick:
I understand now. I'll add a lot message for that fallback.

batrick:
done

zmc (approved):

PR #1847, opened 2023-05-23
fog: Increase timeout in wait_for_deploy_task()

zmc:
When too many reimaging ops are running concurrently, we're seeing timeouts. This isn't a true fix, but should help things until we've got one.


dmick (approved):
Clearly we have to root-cause, but, sure, bleeding

dmick:
(and yes I'm aware of https://github.com/ceph/teuthology/pull/1803)

dmick (approved):
Test works now.

PR #1849, opened 2023-05-23
exporter: Instrument node reimaging success/fail

zmc:


zmc:
[This job](https://pulpito.ceph.com/zack-2023-05-23_21:40:45-smoke-main-distro-default-smithi/) reimaged three nodes, running this branch, and caused this to show up on the exporter's endpoint - even though the exporter itself is still running `main`!
```
# HELP teuthology_reimaging_results_total Teuthology Reimaging Results
# TYPE teuthology_reimaging_results_total counter
teuthology_reimaging_results_total{machine_type="smithi",status="success"} 3.0
```

dmick:
I don't understand how the exporter could have gained a new metric without having code that knows it, and that makes me think I don't understand how this code works.  Let's discuss this Wednesday.

zmc:
We are using multiprocess mode, which uses the filesystem: https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn

dmick (approved):
Makes sense to me.  One would certainly like to see this metric hover around 0.

PR #1851, opened 2023-05-26
provision: Verify reimaged machine OS

zmc:


zmc:
https://pulpito.ceph.com/zack-2023-05-26_19:52:57-smoke-main-distro-default-smithi/

dmick (approved):
Sad about the stream marking, but, not worth fixing for this, agreed

dmick:
Merge at will; just wanna do it during the week

PR #1853, opened 2023-06-12
provision: Avoid a possible AttributeError

zmc:
This fixes a regression introuced in #1851. It hadn't been noticed because it only affects manual locking.


kamoltat (approved):
LGTM Just missing the PR description

PR #1854, opened 2023-06-12
bootstrap: Tolerate a missing lsb_release

zmc:
This fixes the lack of support for CentOS 9.Stream

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -36,7 +36,13 @@ VENV=${VENV:-"virtualenv"}
 
 case "$(uname -s)" in
 Linux)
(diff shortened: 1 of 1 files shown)


dmick:
while I don't think this is wrong given the circumstances, it just makes me hold my head:

https://access.redhat.com/solutions/6973382

[dmick ceph]$ grep -r lsb_release | wc -l
582

dmick (approved):
See comment.  <shakes head, approves>

zmc:
Yeah, it's a bit of a strange thing that happened. But, fortunately, even ubuntu 18.04 has `/etc/os-release`.

PR #1857, opened 2023-06-20
restore py36 support

batrick:
Upstream still builds on RHEL 8 and its derivatives. Teuthology should continue to build/function with the standard py3.6. Importantly, this avoids build problems with cmake discovering python3.9 during rebuilds which causes build failures like:

    $ cmake --build build/ --verbose
    [0/1] /usr/bin/cmake --regenerate-during-build -S/home/pdonnell/ceph -B/home/pdonnell/scratch/build
    -- Building with ccache: /usr/bin/ccache, CCACHE_DIR=
    -- allocator selected: tcmalloc
    CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
      Could NOT find Python3: Found unsuitable version "3.9.13", but required is
      exact version "3.6" (found /usr/bin/python3, found components: Interpreter)
    Call Stack (most recent call first):
      /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:592 (_FPHSA_FAILURE_MESSAGE)
      cmake/modules/FindPython/Support.cmake:1181 (find_package_handle_standard_args)
      cmake/modules/FindPython3.cmake:181 (include)
      CMakeLists.txt:517 (find_package)

    -- Configuring incomplete, errors occurred!
    See also "/home/pdonnell/scratch/build/CMakeFiles/CMakeOutput.log".
    See also "/home/pdonnell/scratch/build/CMakeFiles/CMakeError.log".
    FAILED: build.ninja
    /usr/bin/cmake --regenerate-during-build -S/home/pdonnell/ceph -B/home/pdonnell/scratch/build
    ninja: error: rebuilding 'build.ninja': subcommand failed

### DIFF
--- a/teuthology/repo_utils.py
+++ b/teuthology/repo_utils.py
@@ -5,8 +5,6 @@ import shutil
 import subprocess
 import time
 
-import teuthology.exporter as exporter
-
 from teuthology import misc
 from teuthology.util.flock import FileLock
 from teuthology.config import config
@@ -440,6 +438,9 @@ def fetch_teuthology(branch, commit=None, lock=True):
 
 
 def bootstrap_teuthology(dest_path):
+    # avoid circular imports
+    import teuthology.exporter as exporter
+
     with exporter.BootstrapTime.time():
         log.info("Bootstrapping %s", dest_path)
         # This magic makes the bootstrap script not attempt to clobber an
--- a/bootstrap
+++ b/bootstrap
@@ -67,7 +67,7 @@ Linux)
         fi
         ;;
     RedHatEnterpriseWorkstation|RedHatEnterpriseServer|RedHatEnterprise|CentOS)
-        deps=(python39-pip python39-devel mariadb-devel libev-devel libvirt-devel libffi-devel)
+        deps=(python36-pip python36-devel mariadb-devel libev-devel libvirt-devel libffi-devel)
         for package in ${deps[@]}; do
           if ! rpm -q --whatprovides $package ; then
               missing="${missing:+$missing }$package"
--- a/requirements.txt
+++ b/requirements.txt
@@ -178,7 +178,7 @@ oslo-utils==4.12.2
     #   python-keystoneclient
     #   python-novaclient
     #   python-openstackclient
-packaging==22.0
+packaging==21.3
     # via
     #   ansible-base
     #   build
@@ -207,7 +207,7 @@ pep517==0.11.0
     # via build
 pexpect==4.8.0
     # via teuthology (pyproject.toml)
-pip-tools==6.11.0
+pip-tools==6.4.0
     # via teuthology (pyproject.toml)
 platformdirs==2.4.0
     # via virtualenv
@@ -329,7 +329,7 @@ toml==0.10.2
     # via
     #   teuthology (pyproject.toml)
     #   tox
-tomli==2.0.1
+tomli==1.2.3
     # via
     #   build
     #   pep517
@@ -347,7 +347,7 @@ wcwidth==0.2.5
     # via
     #   cmd2
     #   prettytable
-wheel==0.38.1
+wheel==0.37.1
     # via pip-tools
 wrapt==1.12.1
     # via debtcollector
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,6 +13,7 @@ classifiers =
     Operating System :: POSIX :: Linux
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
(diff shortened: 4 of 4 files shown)


batrick:
I don't see why teuthology should try to be bleeding edge here. It's not necessary AFAICT. I'd prefer we keep the python versions in lock-step.

zmc:
We're not exactly bleeding edge here; python 3.6 was EOLed in December of '21. [RHEL8 can install newer python versions](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_installing-and-using-python_configuring-basic-system-settings) as well.
I do see you're using `cmake`, which teuthology doesn't use - maybe some more context would help. Is someone wanting to run teuthology on RHEL8?

batrick:
I do when building/testing (vstart_runner.py) Ceph.

teuthology/repo_utils.py
```diff
@@ -440,6 +438,9 @@ def fetch_teuthology(branch, commit=None, lock=True):
 
 
 def bootstrap_teuthology(dest_path):
+    # avoid circular imports
+    import teuthology.exporter as exporter
+
     with exporter.BootstrapTime.time():
```
kshtsk, line 444: I'm just thinking, why we use anything from exporter in repo_utils.

zmc: This is what enables the exporter to report the time it takes for bootstrap to complete

zmc:
We spoke about this earlier today, and I logged in to vossi04, where @batrick is having these issues related to the Python version, to take a look. The error message I get when trying to build ceph is:
```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3: Found unsuitable version "3.9.13", but required is
  exact version "3.6" (found /usr/bin/python3, found components: Interpreter)
```
However:
```
[zack@vossi04 ceph]$ ls /usr/bin/python3.[0-9]
/usr/bin/python3.6  /usr/bin/python3.8  /usr/bin/python3.9
[zack@vossi04 ceph]$ ls -l /usr/bin/python3
lrwxrwxrwx. 1 root root 25 Dec 11  2022 /usr/bin/python3 -> /etc/alternatives/python3
[zack@vossi04 ceph]$ ls -l /etc/alternatives/python3
lrwxrwxrwx 1 root root 18 Aug  9 19:18 /etc/alternatives/python3 -> /usr/bin/python3.9
```
Once I update the alternative to `python3.6` via `sudo alternatives --config python3` and select `/usr/bin/python3.6`, the build works again. teuthology's bootstrap has no problem finding the 3.9 interpreter before or after this.

I consider this a bug in ceph's cmake setup since there are other versions installed, the cmake bits have some concept of multiple Python versions, and still just end up only looking at a single interpreter. I tried to look for a fix, but quickly realized I'd have to know a _lot_ more about cmake for that to happen. But the workaround is trivial.

batrick:
> We spoke about this earlier today, and I logged in to vossi04, where @batrick is having these issues related to the Python version, to take a look. The error message I get when trying to build ceph is:
> 
> ```
> CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Python3: Found unsuitable version "3.9.13", but required is
>   exact version "3.6" (found /usr/bin/python3, found components: Interpreter)
> ```
> 
> However:
> 
> ```
> [zack@vossi04 ceph]$ ls /usr/bin/python3.[0-9]
> /usr/bin/python3.6  /usr/bin/python3.8  /usr/bin/python3.9
> [zack@vossi04 ceph]$ ls -l /usr/bin/python3
> lrwxrwxrwx. 1 root root 25 Dec 11  2022 /usr/bin/python3 -> /etc/alternatives/python3
> [zack@vossi04 ceph]$ ls -l /etc/alternatives/python3
> lrwxrwxrwx 1 root root 18 Aug  9 19:18 /etc/alternatives/python3 -> /usr/bin/python3.9
> ```
> 
> Once I update the alternative via `sudo alternatives --config python3` and selecting `/usr/bin/python3.6`, the build works again. teuthology's bootstrap has no problem finding the 3.9 interpreter before or after this.
> 
> I consider this a bug in ceph's cmake setup since there are other versions installed, the cmake bits have some concept of multiple Python versions, and still just end up only looking at a single interpreter. I tried to look for a fix, but quickly realized I'd have to know a _lot_ more about cmake for that to happen. But the workaround is trivial.

Thanks for looking into this Zack. I thought we had done this before but maybe something reset it back to a bad state. I agree the cmake python module needs fixing but I also am not familiar with cmake enough to do it.

batrick:
I created a tracker for this Zack: https://tracker.ceph.com/issues/62428

Thanks for spending time on this!

PR #1858, opened 2023-06-21
exporter: Restart every 24h

zmc:
A design limitation of prometheus-client's multiprocessing mode is that each process creates files to store its own metrics; the exporter then has to read each file, even if the process which created it is dead.

This results in request latency growing over time, to the point of multiple seconds when the file count gets into the thousands. This eventually results in prometheus failing to fetch, leaving gaps in our data.

We can work around this by restarting at a regular interval; 24h seems like a fine place to start.


kamoltat (approved):
LGTM

PR #1863, opened 2023-06-30
FOG._wait_for_ready: Tolerate EOFError

zmc:


teuthology/provision/fog.py
```diff
@@ -298,7 +298,10 @@ def _wait_for_ready(self):
                     try:
                         self.remote.run(args=cmd, timeout=600)
                         break
-                    except ConnectionResetError as e:
+                    except (
+                        ConnectionResetError,
+                        EOFError,
+                    )as e:
```
dmick, line 304: probably should be a space here before 'as'.  I'll see if I can do it here in github

zmc: Nice catch, fixed it

dmick (approved):

PR #1865, opened 2023-07-13
Ansible._handle_failure: Catch all Exceptions

zmc:
Fixes an error first seen here:
https://sentry.ceph.com/organizations/ceph/issues/42041/


dmick (approved):
As long as we come back to this and do better analysis; except Exception is too broad to leave, IMO

zmc:
Yep. Just want to be able to capture the raw records to implement analysis for them.

PR #1869, opened 2023-07-20
orchestra/console: log output from pexpect commands

dmick:
in case anything weird is being noticed and communicated by ipmitool, try to display anything it says

### DIFF
--- a/teuthology/orchestra/console.py
+++ b/teuthology/orchestra/console.py
@@ -1,3 +1,4 @@
+import io
 import logging
 import os
 import pexpect
@@ -35,11 +36,10 @@ class PhysicalConsole(RemoteConsole):
     Physical Console (set from getRemoteConsole)
     """
     def __init__(self, name, ipmiuser=None, ipmipass=None, ipmidomain=None,
-                 logfile=None, timeout=40):
+                 timeout=40):
         self.name = name
         self.shortname = self.getShortName(name)
         self.timeout = timeout
-        self.logfile = None
         self.ipmiuser = ipmiuser or config.ipmi_user
         self.ipmipass = ipmipass or config.ipmi_password
         self.ipmidomain = ipmidomain or config.ipmi_domain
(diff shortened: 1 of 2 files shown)


dmick:
@zmc, please have a look and let's discuss the approach.  I'm not fully comfortable with it but it might be the best we can really do

dmick:
example output (new log lines have 'output:' in them):

```2023-07-19 19:44:25,364.364 INFO:teuthology.orchestra.console:Power off smithi008
2023-07-19 19:44:25,364.364 DEBUG:teuthology.orchestra.console:pexpect command: ipmitool -H smithi008.ipmi.sepia.ceph.com -I lanplus -U inktank -P ApGNXcA7 power off
2023-07-19 19:44:25,380.380 DEBUG:teuthology.orchestra.console:power off output: Chassis Power Control: Down/Off
2023-07-19 19:44:25,381.381 DEBUG:teuthology.orchestra.console:pexpect command: ipmitool -H smithi008.ipmi.sepia.ceph.com -I lanplus -U inktank -P ApGNXcA7 power status
2023-07-19 19:44:25,397.397 DEBUG:teuthology.orchestra.console:check power output: Chassis Power is off
2023-07-19 19:44:25,497.497 INFO:teuthology.orchestra.console:Power off for smithi008 completed```

zmc (commented):
@dmick I think using the `self.logfile` instance variable like this could introduce races in concurrent execution. What I'd change about this approach would be to drop `self.logfile` entirely from `PhysicalConsole` , and `logfile` from its `__init__` method. Since you're already setting the `logfile_read` on the `pexpect.spawn` object, the method calls that produce one can just access that same attr from the returned object. Calling `seek(0)` will still be necessary, but we'd be guaranteed that each `pexpect` session has its own buffer to use.

dmick:
probably a good idea.  I don't think we currently have overlapping spawns, but I did ponder that as a potential problem.  I couldn't figure out any good way to "tie" the pexpect logfile option directly to a logger; I looked into subclassing files to add in a call to log.debug but I just couldn't puzzle it out.

zmc:
Ah, look at `teuthology.task.ansible.LoggerFile`. If that does what you want, we can move it to its own module in `teuthology.task.util`

dmick:
Repushed with suggested changes.

dmick:
jenkins retest please

zmc:
#1870 aims to fix the CI

dmick:
rebased on main after merging https://github.com/ceph/teuthology/pull/1870, tests pass now

zmc (approved):

PR #1870, opened 2023-07-21
Pin PyYAML to fix CI breakage

zmc:


zmc:
The "depedencies" check is failing in the "Initial bootstrap" phase, which is simply reproducing the problem we're fixing here (it bootstraps from `main` first). The "CI" and "integration" checks are what matter here.

dmick (approved):
Lost in a sea of pinned versions because no one can keep backward compatibility, forced into creating technical debt for when the ancient pins finally break.  Business as usual.

dmick:
(obviously it'd be good if the checks passed.)

PR #1879, opened 2023-08-02
PhysicalConsole: Fix UTF-8 related failure

zmc:


zmc:
https://pulpito-ng.ceph.com/runs/zack-2023-08-02_18:04:55-powercycle-reef-distro-default-smithi

yuriw (approved):

PR #1882, opened 2023-08-14
supervisor.reimage(): Improve Sentry reporting

zmc:
Set the `task` tag value to 'reimage' when reporting reimage failures to Sentry, to make searching for them in its UI easier.


zmc:
@dmick this made it easy to provide an org-level saved search: https://sentry.ceph.com/organizations/ceph/issues/searches/25/

We can refine exception names etc. as needed.

PR #1883, opened 2023-08-15
nuke: Avoid a TypeError w/ null node description

zmc:
This avoids a `TypeError: argument of type 'NoneType' is not iterable` when nuking a node whose description is None.

ex: https://sentry.ceph.com/share/issue/91172146663f4c71a6cbfe43725b2e07/

### DIFF
--- a/teuthology/nuke/__init__.py
+++ b/teuthology/nuke/__init__.py
@@ -245,11 +245,11 @@ def nuke(ctx, should_unlock, sync_clocks=True, noipmi=False, keep_logs=False, sh
     with parallel() as p:
         for target, hostkey in ctx.config['targets'].items():
             status = get_status(target)
-            if ctx.name and ctx.name not in status.get('description', ""):
+            if ctx.name and ctx.name not in (status.get('description') or ""):
                 total_unnuked[target] = hostkey
                 log.info(
                     f"Not nuking {target} because description doesn't match: "
-                    f"{ctx.name} != {status['description']}"
+                    f"{ctx.name} != {status.get('description')}"
                 )
                 continue
             elif status.get('up') is False:


teuthology/nuke/__init__.py
```diff
@@ -245,11 +245,11 @@ def nuke(ctx, should_unlock, sync_clocks=True, noipmi=False, keep_logs=False, sh
     with parallel() as p:
         for target, hostkey in ctx.config['targets'].items():
             status = get_status(target)
-            if ctx.name and ctx.name not in status.get('description', ""):
+            if ctx.name and ctx.name not in (status.get('description') or ""):
                 total_unnuked[target] = hostkey
                 log.info(
                     f"Not nuking {target} because description doesn't match: "
-                    f"{ctx.name} != {status['description']}"
```
dmick, line 252: This seems like the actual fix for the exception

zmc: If you visit the Sentry event I linked in the commit message you can see that the error is on line 248

teuthology/nuke/__init__.py
```diff
@@ -245,11 +245,11 @@ def nuke(ctx, should_unlock, sync_clocks=True, noipmi=False, keep_logs=False, sh
     with parallel() as p:
         for target, hostkey in ctx.config['targets'].items():
             status = get_status(target)
-            if ctx.name and ctx.name not in status.get('description', ""):
```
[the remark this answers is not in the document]
zmc: The bug is that we end up trying to iterate over `None`. This happens when `status['description']` has the value `None`. `dict.get` works in such a way that the second arg is returned if there is _no value_ for the requested key, as opposed to there being a Falsey value. So your example would have us hitting the error if `status['description']` were `None` or `""`.

PR #1884, opened 2023-08-17
teuthology/scrape: Fix bad backtrace parsing in Teuthology.log

kamoltat:
Problem:

- confusing warning message stating that the back trace is malformed

- We kept adding to the backtrace buffer even when we exceeded the `MAX_BT_LINES`

Solution:

- Correct the warning message to be "Ignoring backtrace that exceeds MAX_BT_LINES"
- reset the buffer once we exceeded MAX_BT_LINES
- Added some cases where we detect start/end of back trace.

Fixes: https://tracker.ceph.com/issues/62445

### DIFF
--- a/teuthology/scrape.py
+++ b/teuthology/scrape.py
@@ -303,29 +303,34 @@ class Job(object):
         for line in file_obj:
             # Log prefix from teuthology.log
             if ".stderr:" in line:
+                # Only captures the error and disregard what ever comes before
                 line = line.split(".stderr:")[1]
 
-            if "FAILED assert" in line:
+            if "FAILED assert" in line or "__ceph_assert_fail" in line:
                 assertion = line.strip()
 
             if line.startswith(" ceph version"):
                 # The start of a backtrace!
                 bt_lines = [line]
-            elif line.startswith(" NOTE: a copy of the executable"):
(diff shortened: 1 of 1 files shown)


ljflores (approved):
Thanks!

zmc (approved):
Good catch!

zmc:
The docs build failure should be addressed by https://github.com/sphinx-doc/sphinx/pull/11613

kamoltat:
@zmc Thanks for filing the PR to fix the sphinx issue :)

I tried re-running the CI again after https://github.com/sphinx-doc/sphinx/pull/11613 was merged, but it seems to still be failing with the same error :( I wonder if I need to do something more to get the latest update of the patch in my CI

zmc:
@kamoltat they haven't published a release with the fix yet

kamoltat:
> @kamoltat they haven't published a release with the fix yet

@zmc make sense, should we wait for them to publish the release? IMHO since the CI is not really labelled as required, we can just skip it?

zmc:
@kamoltat it's marked as required, but #1885 is a simple fix

kamoltat:
> @kamoltat it's marked as required, but #1885 is a simple fix

Thanks @zmc

PR #1888, opened 2023-09-05
misc._ssh_keyscan: Sort keys before returning any

zmc:
ssh-keyscan's output is unsorted, so this function wasn't deterministic.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1138,9 +1138,11 @@ def _ssh_keyscan(hostname):
         line = line.strip()
         if line and not line.startswith('#'):
             log.error(line)
+    keys = list()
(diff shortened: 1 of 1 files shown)


zmc:
Related: https://sentry.ceph.com/share/issue/fdffc97572b04418b16a6fe929a29503/

zmc:
The failures in the test run don't seem at all related to host key mismatches.

The jenkins job failed because it ran on an 18.04 host which incorrectly has the `bionic` label.

dmick (approved):
LGTM; confirmed that the check failure probably isn't important.  The jenkins labels are inclusive for ceph builds, because ceph uses pbuilder and can build any debian release on any debian release

PR #1892, opened 2023-09-27
Add containers/teuthology-dev

zmc:
This is nearly identical to docs/docker-compose/teuthology, but with some changes to better work with ceph-devstack. The bits in docs/docker-compose should be able to be adapted easily to work with this container.


zmc:
@amathuria this should be ready to merge

amathuria (approved):
Tested this out and it looks good to me.

PR #1897, opened 2023-11-02
config: Use https URLs for paddles and pulpito

zmc:


kshtsk:
@zmc I just wondering why would we prefer to have https references defaults? When the one is installing teuthology in local environment there is no https setup for any of pulpito paddles.

PR #1899, opened 2023-11-10
kill: Don't unlock nodes if killing procs fails

zmc:
... so that we don't unlock nodes while their jobs are running.

### DIFF
--- a/teuthology/kill.py
+++ b/teuthology/kill.py
@@ -70,7 +70,8 @@ def kill_run(run_name, archive_base=None, owner=None, machine_type=None,
     if not preserve_queue:
         remove_beanstalk_jobs(run_name, machine_type)
         remove_paddles_jobs(run_name)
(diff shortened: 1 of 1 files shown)


teuthology/kill.py
```diff
@@ -187,6 +189,7 @@ def kill_processes(run_name, pids=None):
     if len(to_kill) == 0:
         log.info("No teuthology processes running")
     else:
+        survivors = list()
```
kamoltat, line 192: Can I ask why `list()` instead of `[]`. I don't mind `list()`, just curious.

kamoltat (approved):
LGTM, just one minor comment that I'm just curious about.

teuthology/kill.py
zmc: Good question! It wasn't actually intentional, and `[]` is faster, so I'll replace it :)

kamoltat (approved):
Looks good!

PR #1900, opened 2023-11-15
Add systemd units for exporter and dispatcher

zmc:
These are copies of what is currently in use in sepia.

### DIFF
--- a/systemd/teuthology-dispatcher@.service
+++ b/systemd/teuthology-dispatcher@.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Teuthology Dispatcher
+
+Wants=ceph.target
+After=ceph.target
+
+[Service]
+Type=simple
+User=teuthworker
+ExecStart=/home/teuthworker/src/git.ceph.com_git_teuthology_main/virtualenv/bin/python3.8 \
+    /home/teuthworker/src/git.ceph.com_git_teuthology_main/virtualenv/bin/teuthology-dispatcher \
+    -v \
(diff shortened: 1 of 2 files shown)


systemd/teuthology-dispatcher@.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Teuthology Dispatcher
+
+Wants=ceph.target
+After=ceph.target
+
+[Service]
+Type=simple
+User=teuthworker
+ExecStart=/home/teuthworker/src/git.ceph.com_git_teuthology_main/virtualenv/bin/python3.8 \
```
kshtsk, line 10: Just wondering why this refers to git.ceph.com, but not GitHub?
Also maybe there must be a parameter according to ansible or other scripts?

dmick: git.ceph.com is the local mirror that virtually everything in teuthology uses, so it's the one most likely to be present and current.  don't know what parameters you mean exactly, but there is the 'instance' parameter so that one starts teuthology-dispatcher@smithi, for instance

dmick:
I think we're using these now, which is a good recommendation.

dmick (approved):

systemd/teuthology-dispatcher@.service
zmc: These files have several hardcoded elements that someone in a different setting will want to change; I'm not sure if there is a reasonable way to parametrize all of them, but I think it's fine to ship what are effectively templates that users can modify to suit their needs.

PR #1902, opened 2023-11-17
dispatcher: Dont spam the journal

zmc:


dmick (approved):
I want to spend some more time on this, but it doesn't seem like tossing the stdout/stderr from the supervisor process will do any harm; it logs to its own file, and the dispatcher doesn't need to see this, nor does the system journal when the dispatcher runs under systemd

PR #1906, opened 2023-12-20
kill.kill_processes: Fix possibly-unbound variables

zmc:


zmc:
https://pulpito-ng.ceph.com/zack-2023-12-21_00:49:08-orch:cephadm:smoke-main-distro-default-smithi/

VallariAg (approved):
Oh good catch! Tested this branch on docker setup and I was able to kill both running and queued runs.

PR #1909, opened 2024-01-03
Update dependencies

zmc:
This PR:
- Updates all requirements, resolving a number of CVEs
- Moves rarely-used openstack libraries into a variant
- Drops `ansible` because it drastically inflates our install time & size (`ansible-core` has almost all of what we need)
- Enumerates the specific ansible "collections" we require
- Installs collections in a non-default location to avoid versioning conflicts
- Fixes a bug in `repo_utils` that let us DoS ourselves
- Fixes a crash in `dispatcher` and `exporter`

To elaborate on the bloat we're trimming with the ansible changes:
```
$ git checkout main && rm -rf ./virtualenv ./.ansible && time ./bootstrap && du -hs virtualenv
[...]
real    2m34.910s
user    0m55.697s
sys     0m8.713s
626M    virtualenv

$ git checkout deps && rm -rf ./virtualenv ./.ansible && time ./bootstrap && du -hs virtualenv
[...]
real    1m32.248s
user    0m35.911s
sys     0m4.892s
222M    virtualenv
```

So we drop ~65% of our installed size, and ~40% of our install time.

A recent test run using these changes: 
https://pulpito-ng.ceph.com/runs/zack-2024-01-03_21:30:55-orch:cephadm:smoke-main-distro-default-smithi


dmick (approved):
I don't see anything surprising or suspicious.

PR #1913, opened 2024-01-26
supervisor: Do not nuke nodes after jobs finish

zmc:
This was causing a bad race condition, where we could unlock a node, then unlock it again via the nuke process after a different job had locked it.

Fixes: https://tracker.ceph.com/issues/64193


ljflores (approved):
Thanks for the quick fix!

PR #1915, opened 2024-02-02
Removing nuke

zmc:
This PR removes the entire nuke system, save for `clear_firewall()` which is still used by tasks in ceph.git. It contains new implementations of node unlocking and cleanup to better reflect how we use machine resources.


amathuria (approved):
@zmc this looks good to me! I was just wondering if maybe we should also add a reason for removing the nuke command? to the commit message or we can attach a tracker
For future reference.

PR #1918, opened 2024-02-13
teuthology/__init__.py: don't patch threads when running via teuthology_api

VallariAg:
The project [teuthology-api](https://github.com/ceph/teuthology-api) requires threads to be not patched.
Currently, we are using "teuth-api" branch of teuthology where threads are not patched. With this commit, we'll be able to use the "main" branch as a dependency.


zmc (approved):
Thanks!

PR #1931, opened 2024-03-31
scrape: searching backtrace with gzip

NitzanMordhai:
Scrape script that trying to find backtrace in gzip log files can hit TypeError: a bytes-like object is required, not 'str' error and fail to collect results. the gzip file need to be decoded.

Fixes: https://tracker.ceph.com/issues/64402

### DIFF
--- a/teuthology/scrape.py
+++ b/teuthology/scrape.py
@@ -3,7 +3,7 @@
 
 import difflib
 from errno import ENOENT
-from gzip import GzipFile
+import gzip
 import sys
 import os
 import yaml
@@ -361,6 +361,8 @@ class Job(object):
             return
 
         for line in grep(tlog_path, "command crashed with signal"):
+            if not line:
+                continue
             log.debug("Found a crash indication: {0}".format(line))
             # tasks.ceph.osd.1.plana82.stderr
             match = re.search(r"tasks.ceph.([^\.]+).([^\.]+).([^\.]+).stderr", line)
@@ -387,7 +389,8 @@ class Job(object):
                 ))
                 continue
 
-            bt, ass = self._search_backtrace(GzipFile(gzipped_log_path))
+            with gzip.open(gzipped_log_path, 'rt',  errors='ignore') as f:
+                bt, ass = self._search_backtrace(f)
             if ass and not self.assertion:
                 self.assertion = ass
             if bt:
--- a/teuthology/test/test_scrape.py
+++ b/teuthology/test/test_scrape.py
@@ -1,5 +1,7 @@
 from __future__ import with_statement
 
+import glob
+import gzip
 import os
 import shutil
 import tempfile
@@ -12,7 +14,8 @@ class FakeResultDir(object):
     def __init__(self,
                 failure_reason="Dummy reason",
                 assertion="FAILED assert 1 == 2\n",
-                blank_backtrace=False
+                blank_backtrace=False,
+                assertion_osd=False,
     ):
         self.failure_reason = failure_reason
         self.assertion = assertion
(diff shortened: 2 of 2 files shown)


kshtsk:
This PR's been awhile and we gonna celebrate a year birthday. Is there any plans to complete the effort?
I would also suggest to add unitest to cover this. And if there any test results from sepia run would be great to post here.

teuthology/test/test_scrape.py
```diff
@@ -165,3 +175,31 @@ def test_scraper(self):
         assert os.path.exists(os.path.join(d.path, "scrape.log"))
 
         shutil.rmtree(d.path)
+
+    def test_gzip_backtrace_decode(self):
```
kamoltat, line 179: Thanks for adding the decode test to exercise the new code change!

teuthology/scrape.py
```diff
@@ -387,7 +389,10 @@ def _populate_backtrace(self):
                 ))
                 continue
 
-            bt, ass = self._search_backtrace(GzipFile(gzipped_log_path))
+            with gzip.open(gzipped_log_path, 'rb') as f:
+                # Read and decode the contents into strings
+                decoded_content = f.read().decode('utf-8')
+            bt, ass = self._search_backtrace(decoded_content)
```
kamoltat, line 395: I wonder why we cannot do 
`self.backtrace, self.assertion = self._search_backtrace(decoded_content)`

I know that some lines above we are also doing 
`self.backtrace, self.assertion = self._search_backtrace(decoded_content)` but if it found a backtrace it would have returned out of the function and wouldn't reach this line anyways?

kamoltat (approved):
Left a question regarding the return values from `self._search_backtrace` other than that LGTM

kamoltat:
jenkins retest this please

teuthology/scrape.py
kshtsk: @kamoltat maybe because someone wanted to capture the first assertion and continue to look for the backtrace, and finally return only first assertion and the backtrace if any. We would not surmise if the method have corresponding comments or docs what the method is expected to do.

kamoltat:
@NitzanMordhai I think there is a failure here:

```
=================================== FAILURES ===================================
____________________ TestScrape.test_gzip_backtrace_decode _____________________

self = <teuthology.test.test_scrape.TestScrape object at 0x7f9a4cedb050>

    def test_gzip_backtrace_decode(self):
        import gzip, os, glob
    
        with FakeResultDir(assertion="FAILED assert dummy backtrace line",
                        blank_backtrace=True,
                        assertion_osd=True) as d:
    
            with open(os.path.join(d.path, "teuthology.log"), "a") as root_log:
                root_log.write(
                    "command crashed with signal SIGSEGV tasks.ceph.osd.0.host1.stderr\n"
                )
    
            pattern = os.path.join(d.path, "**", "ceph-osd.0.log")
            raws = glob.glob(pattern, recursive=True)
            assert len(raws) == 1, f"expected one raw log, found: {raws}"
            raw_log = raws[0]
            gz_log = raw_log + ".gz"
    
            with gzip.open(gz_log, "wb") as out:
                out.write(open(raw_log, "rb").read())
            os.remove(raw_log)
    
            assert not os.path.exists(raw_log)
            assert os.path.exists(gz_log)
    
            job = scrape.Job(d.path, 1)
>           assert job.get_assertion() == "FAILED assert dummy backtrace line"
E           AssertionError: assert None == 'FAILED assert dummy backtrace line'
E            +  where None = get_assertion()
E            +    where get_assertion = <teuthology.scrape.Job object at 0x7f9a40982c30>.get_assertion
```


I think job.get_assertion() returns a None

teuthology/scrape.py
NitzanMordhai: @kshtsk right, thanks for the explanation!

teuthology/test/test_scrape.py
```diff
@@ -165,3 +175,31 @@ def test_scraper(self):
...
+        import gzip, os, glob
```
kshtsk, line 180: Could you please move imports to the top of the test module.
First of all we need to know about all dependencies as early as possible, from other side, the 'os' is already imported.

teuthology/test/test_scrape.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import with_statement
 
 import os
+import gzip, glob
```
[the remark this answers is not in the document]
NitzanMordhai: Done

kshtsk (approved):

PR #1924, opened 2024-03-07
[DNM] teuthology/suite/run.py: Added some loggings and ScheduleFail exception

kamoltat:
for testing + development purposes

PR #1937, opened 2024-05-08
Update ansible version

VallariAg:
Fixes:
"""
Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/': 'CustomHTTPSConnection' object has no attribute 'cert_file'. 'CustomHTTPSConnection' object has no attribute 'cert_file' 
"""

Blocks Integration tests: https://github.com/ceph/teuthology/pull/1924#issuecomment-2100747323

Run scheduled with this teuthology branch: https://pulpito.ceph.com/vallariag-2024-05-08_17:07:40-smoke-main-distro-default-smithi/

Edit: this fixed the integration test but broke all other CI checks :') 
alternative solution - https://github.com/ceph/teuthology/pull/1938

### DIFF
--- a/docs/docker-compose/teuthology/Dockerfile
+++ b/docs/docker-compose/teuthology/Dockerfile
@@ -11,6 +11,7 @@ RUN apt-get update && \
     python3-pip \
     python3-venv \
     vim \
+    locales-all \
     libev-dev \
     libvirt-dev \
     libffi-dev \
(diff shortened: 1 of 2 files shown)


VallariAg:
Closing because https://github.com/ceph/teuthology/pull/1940 is the correct solution

PR #1939, opened 2024-05-09
Fix CI failures related to Ansible Galaxy

zmc:
This PR bumps the minimum Python version to 3.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,6 @@ jobs:
       matrix:
         include:
           - os: ubuntu-20.04
-            python: 3.8
-          - os: ubuntu-20.04
             python: 3.9
           - os: ubuntu-22.04
             python: "3.10"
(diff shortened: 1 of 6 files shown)


zmc:
ah, seems like 2.16.6 requires a version newer than 3.9

VallariAg:
Looks like according to the ansible-core lifecycle[1], we could use ansible-core 2.15 which supports python3.9-3.11 but it goes EOL in November 2024. Maybe ansible 2.16 and >=python 3.10 is a safer bet?
Also I think ansible collections need to be upgraded as well - https://github.com/ceph/teuthology/blob/main/requirements.yml

[1] https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix

zmc:
Superseded by  #1939, which uses 3.10 instead of 3.9 at @VallariAg's suggestion

PR #1942, opened 2024-05-17
task/install: Fix AssertionError

VallariAg:
Fixes: https://tracker.ceph.com/issues/66093

Before: https://pulpito.ceph.com/vallariag-2024-05-17_08:56:03-rbd:nvmeof-main-distro-default-smithi/
```
2024-05-17T09:14:05.402 INFO:teuthology.run_tasks:Running task install...
2024-05-17T09:14:05.410 DEBUG:teuthology.task.install:project ceph
2024-05-17T09:14:05.410 DEBUG:teuthology.task.install:INSTALL overrides: {'ceph': {'flavor': 'default', 'sha1': '81a3c1e6c488116ab9844fe5ad8db3ce6128ccce'}}
2024-05-17T09:14:05.410 ERROR:teuthology.run_tasks:Saw exception from tasks.
  File "/home/teuthworker/src/git.ceph.com_teuthology_a4b16c58a5d7aa60dee491f6f096b7509e2f1269/teuthology/run_tasks.py", line 109, in run_tasks
    manager.__enter__()
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/teuthworker/src/git.ceph.com_teuthology_a4b16c58a5d7aa60dee491f6f096b7509e2f1269/teuthology/task/install/__init__.py", line 574, in task
    teuthology.deep_merge(extra_system_packages, install_overrides.get('extra_system_packages', {}))
  File "/home/teuthworker/src/git.ceph.com_teuthology_a4b16c58a5d7aa60dee491f6f096b7509e2f1269/teuthology/misc.py", line 1012, in deep_merge
    assert isinstance(b, list)
AssertionError
```

After: https://pulpito.ceph.com/vallariag-2024-05-17_10:54:43-rbd:nvmeof-main-distro-default-smithi/ 
```
024-05-17T11:16:08.809 INFO:teuthology.run_tasks:Running task install...
2024-05-17T11:16:08.816 DEBUG:teuthology.task.install:project ceph
2024-05-17T11:16:08.817 DEBUG:teuthology.task.install:INSTALL overrides: {'ceph': {'flavor': 'default', 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d'}}
2024-05-17T11:16:08.817 DEBUG:teuthology.task.install:config {'flavor': 'default', 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d'}
2024-05-17T11:16:08.817 INFO:teuthology.task.install:Using flavor: default
2024-05-17T11:16:08.842 DEBUG:teuthology.task.install:Package list is: {'deb': ['ceph', 'cephadm', 'ceph-mds', 'ceph-mgr', 'ceph-common', 'ceph-fuse', 'ceph-test', 'ceph-volume', 'radosgw', 'python3-rados', 'python3-rgw', 'python3-cephfs', 'python3-rbd', 'libcephfs2', 'libcephfs-dev', 'librados2', 'librbd1', 'rbd-fuse'], 'rpm': ['ceph-radosgw', 'ceph-test', 'ceph', 'ceph-base', 'cephadm', 'ceph-immutable-object-cache', 'ceph-mgr', 'ceph-mgr-dashboard', 'ceph-mgr-diskprediction-local', 'ceph-mgr-rook', 'ceph-mgr-cephadm', 'ceph-fuse', 'ceph-volume', 'librados-devel', 'libcephfs2', 'libcephfs-devel', 'librados2', 'librbd1', 'python3-rados', 'python3-rgw', 'python3-cephfs', 'python3-rbd', 'rbd-fuse', 'rbd-mirror', 'rbd-nbd']}
2024-05-17T11:16:08.842 INFO:teuthology.task.install:extra packages: []
2024-05-17T11:16:08.843 DEBUG:teuthology.task.install.rpm:_update_package_list_and_install: config is {'branch': None, 'cleanup': None, 'debuginfo': None, 'downgrade_packages': [], 'exclude_packages': [], 'extra_packages': [], 'extra_system_packages': [], 'extras': None, 'enable_coprs': [], 'flavor': 'default', 'install_ceph_packages': True, 'packages': {}, 'project': 'ceph', 'repos_only': False, 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d', 'tag': None, 'wait_for_package': False}
2024-05-17T11:16:08.843 DEBUG:teuthology.packaging:Querying https://shaman.ceph.com/api/search?status=ready&project=ceph&flavor=default&distros=centos%2F9%2Fx86_64&sha1=c2fb93636ebe325b2c93b199bfbcf8e59451a24d
2024-05-17T11:16:08.845 DEBUG:teuthology.task.install.rpm:_update_package_list_and_install: config is {'branch': None, 'cleanup': None, 'debuginfo': None, 'downgrade_packages': [], 'exclude_packages': [], 'extra_packages': [], 'extra_system_packages': [], 'extras': None, 'enable_coprs': [], 'flavor': 'default', 'install_ceph_packages': True, 'packages': {}, 'project': 'ceph', 'repos_only': False, 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d', 'tag': None, 'wait_for_package': False}
2024-05-17T11:16:08.846 DEBUG:teuthology.packaging:Querying https://shaman.ceph.com/api/search?status=ready&project=ceph&flavor=default&distros=centos%2F9%2Fx86_64&sha1=c2fb93636ebe325b2c93b199bfbcf8e59451a24d
2024-05-17T11:16:08.847 DEBUG:teuthology.task.install.rpm:_update_package_list_and_install: config is {'branch': None, 'cleanup': None, 'debuginfo': None, 'downgrade_packages': [], 'exclude_packages': [], 'extra_packages': [], 'extra_system_packages': [], 'extras': None, 'enable_coprs': [], 'flavor': 'default', 'install_ceph_packages': True, 'packages': {}, 'project': 'ceph', 'repos_only': False, 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d', 'tag': None, 'wait_for_package': False}
2024-05-17T11:16:08.847 DEBUG:teuthology.packaging:Querying https://shaman.ceph.com/api/search?status=ready&project=ceph&flavor=default&distros=centos%2F9%2Fx86_64&sha1=c2fb93636ebe325b2c93b199bfbcf8e59451a24d
2024-05-17T11:16:08.848 DEBUG:teuthology.task.install.rpm:_update_package_list_and_install: config is {'branch': None, 'cleanup': None, 'debuginfo': None, 'downgrade_packages': [], 'exclude_packages': [], 'extra_packages': [], 'extra_system_packages': [], 'extras': None, 'enable_coprs': [], 'flavor': 'default', 'install_ceph_packages': True, 'packages': {}, 'project': 'ceph', 'repos_only': False, 'sha1': 'c2fb93636ebe325b2c93b199bfbcf8e59451a24d', 'tag': None, 'wait_for_package': False}
```


teuthology/task/install/__init__.py
```diff
@@ -571,7 +571,7 @@ def task(ctx, config):
         install_overrides = overrides.get('install', {})
         log.debug('INSTALL overrides: %s' % install_overrides)
         teuthology.deep_merge(config, install_overrides.get(project, {}))
-        teuthology.deep_merge(extra_system_packages, install_overrides.get('extra_system_packages', {}))
+        teuthology.deep_merge(extra_system_packages, install_overrides.get('extra_system_packages', []))
```
idryomov, line 574: I'm not sure this is going to work in all cases.

`extra_system_packages` is weird -- it can be either a list or a dictionary (with `deb` and `rpm` lists in it).  So, a dictionary is supplied in the config (i.e. `extra_system_packages` variable at this point is a dictionary instead of of empty list, `deep_merge()` would refuse to merge it with an empty list).

batrick: hmm, this stinks. So if it's a `list` then the packages should be appended to both `deb` and `rpm`? I suppose that's what teuthology should do?

VallariAg: @idryomov, do you think this would cover all cases? 
```suggestion
        teuthology.deep_merge(extra_system_packages, install_overrides.get('extra_system_packages', type(extra_system_packages)()))
```
Even if `extra_system_packages` config is `None`, `deep_merge()` function handles the case if `b` is `None`.

idryomov: As to what the override behavior should be, it's a choice that we need to make.  I see at least three options:
1. Require that the form is the same for the original and the override and otherwise the job is malformed (i.e. check that both are lists or both are dictionaries, then merge).
2. Make the override win, both in form and content (i.e. don't attempt to merge anything).
3. Introduce a preprocessing step and transform both the original and the override into a canonical dictionary first, then merge, then hand over to `install` task.

idryomov: > @idryomov, do you think this would cover all cases?

No, I think it would blow up in case the job specifies e.g. a list in the original and a dictionary in the override (i.e. when the defaults don't kick in at all).

idryomov: > > @idryomov, do you think this would cover all cases?
> 
> No, I think it would blow up in case the job specifies e.g. a list in the original and a dictionary in the override (i.e. when the defaults don't kick in at all).

This is actually option 1 from a previous comment.  If folks think it's good enough, I'm all for it ;)

batrick: Yes, I think that works but perhaps even better would be:

```
overrides_extra_system_packages = install_overrides.get('extra_system_packages')
if overrides_extra_system_packages:
    teuthology.deep_merge(extra_system_packages, overrides_extra_system_packages)
```

It is more verbose but the `type(...)` solution would look baffling to someone unfamiliar with this problem.

VallariAg: Oh right, those are 3 great option! I think option 3 adds some additional flexibility whereas option 1 would throw an error and ask the person editing the config to do the same thing teuthology could (make both config into same `type` format). What do you think?

batrick: works for me

batrick:
Intermediate fix #1943 is done to get QA runs green again.

VallariAg:
Went with option 1 (with proper try/except message) because existing code reads `extra_system_packages` overrides from two place: `overrides.install.ceph.extra_system_packages` and `overrides.install.extra_system_packages` and preprocessing might cause unexpected problems. 

Also fixes: https://tracker.ceph.com/issues/66133

TODO: Need to test with more tests with diff overrides config.

teuthology/task/install/__init__.py
```diff
@@ -565,16 +566,24 @@ def task(ctx, config):
...
-            teuthology.deep_merge(extra_system_packages, overrides_extra_system_packages)
-        repos = install_overrides.get('repos', None)
+        try:
+            install_overrides = overrides.get('install', {})
+            log.debug('INSTALL overrides: %s' % install_overrides)
+            teuthology.deep_merge(config, install_overrides.get(project, {}))
+            overrides_extra_system_packages = install_overrides.get('extra_system_packages')
+            if overrides_extra_system_packages:
+                extra_system_packages = config.get('extra_system_packages')
+                config['extra_system_packages'] = teuthology.deep_merge(extra_system_packages, overrides_extra_system_packages)
```
batrick, line 578:
```suggestion
                teuthology.deep_merge(extra_system_packages, overrides_extra_system_packages)
```

? (This one confused me.)

batrick (changes requested):
otherwise LGTM!

teuthology/task/install/__init__.py
VallariAg: The result of the deep merge should be stored in `config` because later we do [this](https://github.com/ceph/teuthology/pull/1942/files#diff-28d5292d55b79ee4e4e70922a033dbec3a489da3fb4f6672ad8fb64ab053b84cR621) `extra_system_packages=config.get('extra_system_packages', []),`

Using `config.get('extra_system_packages')` is better than directly using variable `extra_system_packages` because when [we log whole config](https://github.com/ceph/teuthology/pull/1942/files#diff-28d5292d55b79ee4e4e70922a033dbec3a489da3fb4f6672ad8fb64ab053b84cR588) there would be no doubt what the actual config would look like.

VallariAg: I think it'll only be the same if `config.get('extra_system_packages')` is a list or a dictionary. If it's `None` (case:`extra_system_packages` not mentioned in install task definition and only in overrides), then `extra_system_packages` wouldn't reference it. 

I think the regression (https://tracker.ceph.com/issues/66133) happened because of the above case:
"python3-pytest" package was defined in `overrides.install.ceph.extra_system_packages` and there was no `extra_system_packages` in install task definition config. So the variable was an empty list which didn't reference the config value. 
Then `teuthology.deep_merge(config, install_overrides.get(project, {}))` added the package to `config["extra_system_packages"]` (but not in `extra_system_packages` variable). We could see the package when config was logged but since we were [using extra_system_package variable](https://github.com/ceph/teuthology/pull/1943/files#diff-28d5292d55b79ee4e4e70922a033dbec3a489da3fb4f6672ad8fb64ab053b84cR612) to makeup the final config, the package never got added to final config and so it was not installed.

PR #1944, opened 2024-05-20
suite: check build_complete for centos9

cbodley:
with the squid release moving to centos 9-based containers, teuthology should prevent scheduling jobs until those container builds finish

Fixes: https://tracker.ceph.com/issues/66096


teuthology/suite/util.py
```diff
@@ -25,7 +25,7 @@
 
 log = logging.getLogger(__name__)
 
-CONTAINER_DISTRO = 'centos/8'       # the one to check for build_complete
+CONTAINER_DISTROS = ['centos/8', 'centos/9'] # the ones to check for build_complete
```
ljflores, line 28:
```suggestion
CONTAINER_DISTROS = 'centos/9' # the ones to check for build_complete
```
This will error out once centos 8 builds start failing and/or are removed (since it will be checking that c8 AND c9 are completed). I think it is okay to move it strictly to centos 9 since we will lose centos 8 very soon.

teuthology/suite/util.py
```diff
@@ -245,7 +245,7 @@ def package_version_for_hash(hash, flavor='default', distro='rhel',
         ),
     )
 
-    if (bp.distro == CONTAINER_DISTRO and bp.flavor == CONTAINER_FLAVOR and 
+    if (bp.distro in CONTAINER_DISTROS and bp.flavor == CONTAINER_FLAVOR and
```
ljflores, line 248: In that case, this change can be removed.

ljflores (commented):
@cbodley I think since c8 will be lost very soon, this patch will still be looking for c8 and will therefore error out.

teuthology/orchestra/opsys.py
```diff
@@ -73,7 +73,7 @@
 DEFAULT_OS_VERSION = dict(
     ubuntu="22.04",
     fedora="25",
-    centos="8.stream",
+    centos="9.stream",
```
cbodley, line 76: in testing i saw that the centos 9 container check wasn't actually happening until i changed this too

zmc:
It looks like we just need to update the centos 8 references here: https://github.com/ceph/teuthology/blob/main/teuthology/suite/test/test_util.py#L254-L267

ljflores (approved):

ljflores:
> It looks like we just need to update the centos 8 references here: https://github.com/ceph/teuthology/blob/main/teuthology/suite/test/test_util.py#L254-L267

Updated!

zmc (approved):

zmc:
Thanks @cbodley and @ljflores!!

PR #1945, opened 2024-05-23
syslog.py: report when ceph works are stuck

lxbsz:
Usually there will be performance bug when hitting the works are stuck:

workqueue: delayed_work [ceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND

URL: https://www.spinics.net/lists/ceph-users/msg78775.html


teuthology/task/internal/syslog.py
```diff
@@ -107,7 +107,7 @@ def syslog(ctx, config):
             stdout = rem.sh(
                 [
                     'egrep', '--binary-files=text',
-                    '\\bBUG\\b|\\bINFO\\b|\\bDEADLOCK\\b',
+                    '\\bBUG\\b|\\bINFO\\b|\\bDEADLOCK\\b|\\bceph\\b.*\\bWQ_UNBOUND\\b',
```
idryomov, line 110: > [Sun Sep 10 07:19:00 2023] workqueue: delayed_work [ceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND
> [Sun Sep 10 08:41:24 2023] workqueue: ceph_con_workfn [libceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND

This regex would match only the first message (ceph), but not the second (libceph).

idryomov:
> Usually there will be performance bug when hitting the works are stuck:
> 
> workqueue: delayed_work [ceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND

This doesn't mean that the work item is stuck -- just that it's taking longer than a configurable threshold.  BTW `WQ_UNBOUND` is not a universal cure -- it can lead to higher latencies:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c25da5b7baf1d243e6612ba2b97e2a2c4a1376f6
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f959325e6ac3f499450088b8d9c626d1177be160

teuthology/task/internal/syslog.py
lxbsz: > > [Sun Sep 10 07:19:00 2023] workqueue: delayed_work [ceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND
> > [Sun Sep 10 08:41:24 2023] workqueue: ceph_con_workfn [libceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND
> 
> This regex would match only the first message (ceph), but not the second (libceph).

Okay, let me run the test again.



> Also, I wouldn't cling to `WQ_UNBOUND` -- it's an implementation detail that could change. If the concern is work items taking longer than expected to run, it would probably be better to match on "workqueue: ... hogged CPU for ...".

I just want to match the ceph ones, this will match all the possible warnings, is that okay ?

lxbsz:
> > Usually there will be performance bug when hitting the works are stuck:
> > workqueue: delayed_work [ceph] hogged CPU for >10000us 4 times, consider switching to WQ_UNBOUND
> 
> This doesn't mean that the work item is stuck -- just that it's taking longer than a configurable threshold.
> 
> BTW `WQ_UNBOUND` is not a universal cure either -- it can lead to higher latencies:
> 
> https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c25da5b7baf1d243e6612ba2b97e2a2c4a1376f6 https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f959325e6ac3f499450088b8d9c626d1177be160

Yeah, you are right. Just this warning could give us a hint that this code could be improved.

teuthology/task/internal/syslog.py
idryomov: > I just want to match the ceph ones

That is totally fine -- I just wanted to say that we shouldn't be matching on name of the internal flag.

PR #1946, opened 2024-05-23
teuthology/suite: use paddles solely for rerun info

batrick:
teuthology.front no longer allows access to the run directories for general users. In fact, it's not required because we are already getting run information from paddles.

This commit makes it more stringent in --rerun handling: if seed/subset/no_nested_subset do not match, exit failure rather than try to continue. Without a subset, many suites are so large that teuthology-suite will run for several minutes before ultimately failing to schedule due to the quantity of jobs. It also doesn't make sense to rerun a suite when these values do not match.

### DIFF
--- a/teuthology/suite/__init__.py
+++ b/teuthology/suite/__init__.py
@@ -5,6 +5,7 @@
 import logging
 import os
 import random
+import sys
 import time
 from distutils.util import strtobool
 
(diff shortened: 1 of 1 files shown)


yuriw (approved):
I tested this and was able to schedule https://pulpito.ceph.com/yuriw-2024-05-23_17:16:53-rados-wip-yuri5-testing-2024-05-15-0804-distro-default-smithi/

zmc (approved):
Looks great. @yuriw does this fix reruns for you as well?

teuthology/suite/__init__.py
```diff
@@ -148,11 +140,59 @@ def main(args):
...
+    seed = None if job0 is None else job0.get('seed')
+    if conf.seed >= 0 and conf.seed != seed:
+        log.error('--seed %s does not match with rerun seed: %s',
+                  conf.seed, seed)
+        sys.exit(1)
+    else:
+        log.info('Using rerun seed=%s', seed)
+        conf.seed = seed
+
+    subset = None if job0 is None else tuple(map(int, job0.get('subset').split('/')))
```
cbodley, line 163: i see `--rerun` failing here with `'NoneType' object has no attribute 'split'`. i generally run the rgw suite without `--subset`

opened https://tracker.ceph.com/issues/66302

PR #1949, opened 2024-05-28
Fixes for teuthology-node-cleanup

zmc:
This mainly gets the command working without having to invoke with `--owner` for each user who has jobs running.

### DIFF
--- a/scripts/node_cleanup.py
+++ b/scripts/node_cleanup.py
@@ -13,13 +13,23 @@ def main():
     stale = query.find_stale_locks(args.owner)
     if not stale:
         return
+    by_owner = {}
+    for node in stale:
+        if args.owner and node['locked_by'] != args.owner:
+            log.warning(
+                f"Node {node['name']} expected to be locked by {args.owner} "
+                f"but found {node['locked_by']} instead"
+            )
+            continue
+        by_owner.setdefault(node['locked_by'], []).append(node)
     if args.dry_run:
         log.info("Would attempt to unlock:")
-        for node in stale:
-            log.info(f"{node['name']}\t{node['description']}")
+        for owner, nodes in by_owner.items():
+            for node in nodes:
+                log.info(f"{node['name']}\t{node['description']}")
     else:
-        names = [node["name"] for node in stale]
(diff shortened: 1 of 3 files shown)


scripts/node_cleanup.py
```diff
@@ -13,13 +13,21 @@ def main():
     stale = query.find_stale_locks(args.owner)
     if not stale:
         return
+    by_owner = {}
+    for node in stale:
+        if args.owner and node['locked_by'] != args.owner:
+            # This should not happen
```
kamoltat, line 19: Should we log this like Error: "node shouldn't be locked by other people"

kamoltat:
Some context as to why you are fixing this in the PR description would be super helpful, thanks!

scripts/node_cleanup.py
zmc: Added

scripts/node_cleanup.py
```diff
@@ -13,13 +13,24 @@ def main():
...
+    by_owner = {}
+    for node in stale:
+        if args.owner and node['locked_by'] != args.owner:
+            log.warning(
+                f"Node {node['name']} expected to be locked by {args.owner} "
+                f"but found {node['locked_by']} instead"
+            )
+            continue
+        owner_nodes = by_owner.setdefault(node['locked_by'], [])
+        owner_nodes.append(node)
```
kamoltat, line 25: I know that owner_nodes is a reference to by_owner, can't we just purely use by_owner? Am I missing something?

kamoltat (approved):
Other than the minor comment, LGTM

scripts/node_cleanup.py
zmc: `by_owner` is a dict where keys are e.g. "scheduled_zack@teuthology", and values are lists of dicts containing node data, whereas `owner_nodes` is a reference to one of those lists - but you're right that there's no real point in holding this reference.

I'll replace the above two lines with:
```
by_owner.setdefault(node['locked_by'], []).append(node)
```

kamoltat (approved):
LGTM, thanks for addressing the comment

PR #1950, opened 2024-05-29
boostrap: expect "FedoraLinux" along with "Fedora"

rishabh-d-dave:
On Fedora 38 -

	$ cat /etc/os-release
	NAME="Fedora Linux"
	VERSION="38 (Workstation Edition)"
	ID=fedora
	VERSION_ID=38

Also see: https://fedoraproject.org/wiki/Changes/Fedora_Linux_in_os-release.

With this "./boostrap install" works fine on Fedora 38 at least.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -100,7 +100,7 @@ Linux)
             fi
         fi
         ;;
-    Fedora)
+    Fedora|FedoraLinux)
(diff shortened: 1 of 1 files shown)


zmc (approved):
Thanks!

PR #1951, opened 2024-05-30
teuthology/suite: use valid subset default for rerun without --subset

batrick:
Fixes: https://tracker.ceph.com/issues/66302


cbodley (approved):
that was fast, thank you! i was able to successfully schedule a --rerun with this change

PR #1952, opened 2024-06-04
docs/docker-compose: use ubuntu jammy for testnode

VallariAg:
Integration test on teuthology repo are failing
because ceph build's support for ubuntu focal on main branch has ended.
This commit fixes this by changing the testnode images from ubuntu focal to jammy.


VallariAg:
@zmc Looks like it was changed from "latest" (or jammy) to "focal" intentionally a couple of years ago: https://github.com/ceph/teuthology/commit/0301e0553272803e57944ee2ae1899968a0c1aa5
Now with jammy,`teuthology:no-ceph` suite seems to be ranning successfuly.

VallariAg:
Better solution: https://github.com/ceph/teuthology/pull/1953

PR #1953, opened 2024-06-04
Various updates to support newer OSes

zmc:
This isn't exhaustive; these are issues I ran into when updating ceph-devstack to work with centos 9 hosts.


VallariAg (dismissed):

zmc:
@VallariAg I found a bug that affected CentOS Stream reimages; the last two commits fix it. Could you review those as well? Thanks!

kamoltat:
Tested on my local docker container setup, works great and fixed the galaxy ansible failure for me

VallariAg (approved):
Looks good - works for me too!

PR #1954, opened 2024-06-05
.github: Add workflow to build dev container

zmc:
It's multi-arch, too!


zmc:
Depends on https://github.com/ceph/teuthology/pull/1953

VallariAg (approved):

PR #1955, opened 2024-06-11
console: Rework power on/off retry mechanism

zmc:
We were only ever issuing the inital request once; Now, after five unsuccessful checks to see that the node is in the desired state, we re-issue the request but do not reset the clock.
We were also checking the power state immediately after sending the request, which is unhelpful as the operation will never complete that quickly. Now, we sleep at least once.


zmc:
A very small run completed successfully: https://pulpito-ng.ceph.com/zack-2024-06-11_22:04:48-orch:cephadm:smoke-small-main-distro-default-smithi/

amathuria (approved):
LGTM

PR #1956, opened 2024-06-12
docker-compose: add teuthology-volume and t-api service

VallariAg:
This PR adds teuthology-volume and t-api service. 
And share pid namespaces btw teuthologyand teuthology_api containers.

This will allow to run teuthology_api container separately instead of inside the teuthology container.


VallariAg:
Wasn't able to reopen this against main, so opened new PR: https://github.com/ceph/teuthology/pull/1957

PR #1958, opened 2024-06-14
suite.Run.schedule_jobs: Unmask error in dry-run

zmc:
Also fix a KeyError.

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -512,13 +512,10 @@ class Run(object):
             log_prefix = ''
             if job in jobs_missing_packages:
                 log_prefix = "Missing Packages: "
-                if (
-                    not self.args.dry_run and
-                    not config.suite_allow_missing_packages
-                ):
+                if not config.suite_allow_missing_packages:
                     util.schedule_fail(
-                        "At least one job needs packages that don't exist for "
-                        "hash {sha1}.".format(sha1=self.base_config.sha1),
+                        "At least one job needs packages that don't exist "
+                        f"for hash {self.base_config.sha1}.",
                         name,
                         dry_run=self.args.dry_run,
                     )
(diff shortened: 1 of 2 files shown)


yuriw (approved):
:+1:

teuthology/suite/run.py
```diff
@@ -512,13 +512,10 @@ def schedule_jobs(self, jobs_missing_packages, jobs_to_schedule, name):
             log_prefix = ''
             if job in jobs_missing_packages:
                 log_prefix = "Missing Packages: "
-                if (
-                    not self.args.dry_run and
-                    not config.suite_allow_missing_packages
-                ):
+                if not config.suite_allow_missing_packages:
                     util.schedule_fail(
```
dmick, line 516: Just thinking about the user impact...it would be better to identify all the jobs clearly that are missing packages when --dry-running.  Would it be better to report something identifying the job and continue, rather than calling schedule_fail here?  In a non-dry-run, failure makes sense (although arguably it ought to fail when there are any jobs missing packages), but in dry-run, reporting as much as possible seems better.  wdyt?

zmc: It felt important to converge the behavior of the two modes, since users seemed pretty surprised that they differed. I do think we could stand to include a little more information though, like what specifically was missing

teuthology/suite/util.py
```diff
@@ -348,15 +348,19 @@ def find_git_parents(project: str, sha1: str, count=1):
...
+        ))
         resp = requests.get(url)
         resp.raise_for_status()
         sha1s = resp.json()['sha1s']
         if len(sha1s) != count:
-            log.debug('got response: %s', resp.json())
-            log.error('can''t find %d parents of %s in %s: %s',
-                       int(count), sha1, project, resp.json()['error'])
+            resp_json = resp.json()
+            err_msg = resp_json.get("error") or resp_json.get("err")
```
kshtsk, line 361: @zmc I wonder why have you left get("error")?

PR #1960, opened 2024-06-25
Finish removing teuthology-worker

zmc:
The dispatcher and supervisor were added in #1546, but code was copied and pasted into the new modules, leaving the worker untouched. Also untouched were the unit tests, meaning that the dispatcher and supervisor were never unit tested. As the copied code changed, the dispatcher and supervisor were not being tested for regressions, while the worker - which wasn't being anymore - had passing unit tests, giving some false sense of security.

This commit removes the old worker code, and adapts the old worker tests to apply to the dispatcher and supervisor. It also splits out teuthology-supervisor into its own command.


zmc:
@kamoltat @VallariAg I was mistaken when I mentioned a test regression in this PR; it's ready for review!

kshtsk:
with docopt it looked nicer imho

kshtsk:
the head of PR: "Finish removing teuthology-worker" does not reflect what is in this patch, in fact there is introducing teuthology-supervisor as separate tool, dropping worker may be probably in separate PR.
I don't know why do say worker unittests are not passing, when I run pytest against it and get `12 passed, 8 warnings in 0.67s`

kshtsk:
Why underscoring unit tests by the way? Following the logic, other modules also must have underscore suffixes?

scripts/dispatcher.py
```diff
@@ -1,35 +1,62 @@
...
 def main():
-    args = docopt.docopt(__doc__)
-    sys.exit(teuthology.dispatcher.main(args))
+    if "--supervisor" in sys.argv:
+        # This is for transitional compatibility, so the old dispatcher can
+        # invoke the new supervisor. Once old dispatchers are phased out,
+        # this block can be as well.
+        sys.argv.remove("--supervisor")
+        sys.argv[0] = "teuthology-supervisor"
+        sys.exit(teuthology.dispatcher.supervisor.main(
```
kshtsk, line 54: if we are splitting supervisor off, then clean this out. Are you trying to keep backward compatibility?

scripts/dispatcher.py
```diff
@@ -1,35 +1,62 @@
...
+        help="if the queue is empty, exit",
+    )
+    return parser.parse_args(argv)
 
 
 def main():
-    args = docopt.docopt(__doc__)
-    sys.exit(teuthology.dispatcher.main(args))
+    if "--supervisor" in sys.argv:
+        # This is for transitional compatibility, so the old dispatcher can
```
kshtsk, line 49: To complicated, do you believe it can really work without errors? :-D

zmc:
@kshtsk I think you're ok with removing the worker at this point, is that the case? Also, responses to your earlier comments:

> with docopt it looked nicer imho

It may, but it's less flexible and harder to maintain

> the head of PR: "Finish removing teuthology-worker" does not reflect what is in this patch, in fact there is introducing teuthology-supervisor as separate tool, dropping worker may be probably in separate PR.

That bit is just adding the _entry point_; the tool has been in use for quite some time, sharing `teuthology-dispatcher`'s entry point

>  I don't know why do say worker unittests are not passing, when I run pytest against it and get 12 passed, 8 warnings in 0.67s

I didn't claim the worker tests weren't passing, only that the dispatcher and supervisor were not being tested.

> Why underscoring unit tests by the way? Following the logic, other modules also must have underscore suffixes?

Sorry I'm not sure what you're asking here

kshtsk:
>> Why underscoring unit tests by the way? Following the logic, other modules also must have underscore suffixes?
> Sorry I'm not sure what you're asking here
> Edit: Oh, if you're referring to the filenames test_dispatcher_.py & test_supervisor_.py, it's so py.test sees them as different from the files in scripts/test/.

if you want to add test for the scripts, I suggest to use test_script_dispatcher.py  and test_script_supervisor.py for them and test_dispatcher.py and test_supervisor.py leave for the modules, otherwise you're breaking name conventions.

kshtsk:
> > with docopt it looked nicer imho
> It may, but it's less flexible and harder to maintain
okay, that sounds reasonable, but then we need to refactor the other scripts too, in scheduled manner, probably someone would love to volunteer to do it

kshtsk (approved):

PR #1961, opened 2024-06-27
orchestra/daemon: add finished() method in DaemonState and CephadmUnit

VallariAg:
The method is used by DaemonWatchdog to determine if DaemonState/CephadmUnit daemons are active or finished.

Use-case example: https://github.com/VallariAg/ceph/commit/8c2e7c00397ed4f7024fd8a492fbc5716f6de4c5

Run example: https://pulpito.ceph.com/vallariag-2024-06-28_12:47:41-rbd:nvmeof-ceph-nvmeof-mon-distro-default-smithi/

### DIFF
--- a/teuthology/orchestra/daemon/cephadmunit.py
+++ b/teuthology/orchestra/daemon/cephadmunit.py
@@ -35,9 +35,9 @@ class CephadmUnit(DaemonState):
 
     def kill_cmd(self, sig):
         return ' '.join([
-            'sudo', 'docker', 'kill',
+            'sudo', 'systemctl', 'kill',
             '-s', str(int(sig)),
-            'ceph-%s-%s.%s' % (self.fsid, self.type_, self.id_),
+            'ceph-%s@%s.%s' % (self.fsid, self.type_, self.id_),
         ])
 
     def _start_logger(self):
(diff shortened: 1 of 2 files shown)


teuthology/orchestra/daemon/cephadmunit.py
```diff
@@ -112,6 +112,9 @@ def running(self):
         """
         return self.is_started
 
+    def finished(self):
+        return self.is_started
```
batrick, line 116: should this not use the `self.status_cmd` to determine if the container is still running?

batrick (changes requested):

teuthology/orchestra/daemon/cephadmunit.py
VallariAg: Made the change, thanks!

batrick (approved):
Code looks right to me but I have not tested it.

Great work!

zmc (approved):
Excellent!

The dev_container failure is unrelated

badone:
Unfortunately, I think this is causing a SIGHUP to end up as a SIGTERM by the time it gets to ceph-osd since the SIGHUP gets sent to the bash shell that is running the unit file. The behaviour is very different to "docker kill", see https://tracker.ceph.com/issues/66883#note-28 and note-29 for our current understanding.

VallariAg:
@badone thank you so much for that analysis. Seems like to me that we should move back to using `docker kill` so OSDs only receive SIGHUP, and not SIGTERM which `systemctl kill` ends up sending.

PR #1972, opened 2024-07-08
teuthology/kill.py: Fix "TypeError: 'NoneType' object is not iterable"

VallariAg:
Fixes: https://tracker.ceph.com/issues/66440#note-1


zmc (approved):
Perfect - thanks so much for adding all these tests!

PR #1976, opened 2024-07-11
find_git_parents: Refresh mirror when required

zmc:
We used to do the refresh every time, which wasn't good. Then we didn't do it at all, which was fine until someone tried to schedule against a commit in the few minutes before the mirror picked it up. Let's try doing it when necessary :)

Fixes: #66893

### DIFF
--- a/teuthology/suite/util.py
+++ b/teuthology/suite/util.py
@@ -347,25 +347,34 @@ def find_git_parents(project: str, sha1: str, count=1):
         log.warning('githelper_base_url not set, --newest disabled')
         return []
 
+    def refresh():
+        url = f"{base_url}/{project}.git/refresh"
+        log.info(f"Forcing refresh of git mirror: {url}")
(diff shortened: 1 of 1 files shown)


dmick (approved):

kamoltat (approved):
As discussed in today's Teuthology meeting, if you can just link a success run or any kind of tests that you did would be great, I went through the code and lgtm.

zmc:
> As discussed in today's Teuthology meeting, if you can just link a success run or any kind of tests that you did would be great, I went through the code and lgtm.

Thanks - this run scheduled as I expected, and this codepath is only hit during scheduling.
https://pulpito-ng.ceph.com/runs/zack-2024-07-17_21:31:01-teuthology:no-ceph-main-distro-default-smithi

zmc:
Failures in the linked test run are fixed by #1975

PR #1977, opened 2024-07-12
Replace deprecated datetime.utcnow() calls

zmc:


teuthology/dispatcher/supervisor.py
```diff
@@ -275,7 +275,7 @@ def unlock_targets(job_config):
 
 
 def run_with_watchdog(process, job_config):
-    job_start_time = datetime.utcnow()
+    job_start_time = datetime.datetime.now(datetime.timezone.utc)
```
kshtsk, line 278: looks gut, I know it must be the same to use shorter form `datetime.UTC`, but why did you prefer longer variant, for the reference:
```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
             python: "3.10"
           - os: ubuntu-22.04
             python: "3.11"
+          - os: ubuntu-24.04
+            python: "3.12"
```
kshtsk, line 23: Love it

teuthology/dispatcher/supervisor.py
zmc: Because they didn't make it available as `datetime.UTC` until 3.11 🙃
https://docs.python.org/3/library/datetime.html#datetime.UTC

teuthology/provision/fog.py
```diff
@@ -227,8 +227,8 @@ def schedule_deploy_task(self, host_id):
         for task in host_tasks:
             timestamp = task['createdTime']
             time_delta = (
-                datetime.utcnow() - datetime.strptime(
-                    timestamp, self.timestamp_format)
+                datetime.datetime.now(datetime.timezone.utc) - datetime.datetime.strptime(
+                    timestamp, self.timestamp_format).replace(tzinfo=datetime.timezone.utc)
```
kshtsk, line 231: Maybe... maybe we can just split this into several lines:
```
created = datetime.datetime.strptime(timestamp, self.timestamp_format)
utc_now = datetime.datetime.now(datetime.timezone.utc)
utc_created = created.replace(tzinfo=datetime.timezone.utc)
time_delta = (utc_now - utc_created).total_seconds()
```

kshtsk: but I'm wondering, it would be way better if python evangelists create three or for incapsulated namespaces, so we can graciously use: `datetime.datetime.datetime.datetime.timezone.utc`

kshtsk (approved):

teuthology/provision/fog.py
zmc: It's a little ridiculous, yeah. For testing purposes I felt the need to import the package, but it would be nice to split it over multiple lines. Still, I don't want to hold up the review so I'll do a follow-up at some point. Thanks!

PR #1980, opened 2024-07-19
kill.kill_job: Get job data from paddles

zmc:
... if we can't get it from the filesystem. This is necessary because of the teuthology archive's restricted permissions on teuthology.front.


idryomov (approved):
This fixes it for me, thanks!

PR #1981, opened 2024-07-25
Disable prometheus when exporter isn't running

zmc:
Also improves the import test performance from 38s to 1s, on my machine.
`teuthology-exporter` on `teuthology.front` is running this branch and behaving normally.
This test run passed unit tests: https://pulpito-ng.ceph.com/runs/zack-2024-07-26_19:28:33-teuthology:no-ceph-main-distro-default-null

### DIFF
--- a/teuthology/dispatcher/__init__.py
+++ b/teuthology/dispatcher/__init__.py
@@ -313,7 +313,10 @@ def lock_machines(job_config):
     fake_ctx = supervisor.create_fake_context(job_config, block=True)
     machine_type = job_config["machine_type"]
     count = len(job_config['roles'])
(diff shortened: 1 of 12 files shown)


teuthology/dispatcher/__init__.py
```diff
@@ -222,7 +222,10 @@ def match(proc):
     procs = {}
     attrs = ["pid", "cmdline"]
     for proc in psutil.process_iter(attrs=attrs):
-        if not match(proc):
+        try:
+            if not match(proc):
+                continue
+        except psutil.AccessDenied:
```
kshtsk, line 228: It is probably better to fix inside match, see corresponding patch #1985

PR #1982, opened 2024-07-30
Remove worker.py

zmc:
Originally, #1960 was intended to do this, but in some review back-and-forth the commit removing the file was dropped. Let's actually remove the file.


zmc:
https://pulpito-ng.ceph.com/runs/zack-2024-07-31_18:04:53-teuthology:no-ceph-main-distro-default-null

PR #1983, opened 2024-07-31
Add job expiration dates

zmc:
This feature has two parts:
* Specifying expiration dates when scheduling test runs
* A global maximum age

Expiration dates are provided by passing `--expire` to `teuthology-suite` with
a relative value like `1d` (one day), `1w` (one week), or an absolute value like
`1999-12-31_23:59:59`.

A new configuration item, `max_job_age`, is specified in seconds. This defaults
to two weeks.

When the dispatcher checks the queue for the next job to run, it will first
compare the job's `timestamp` value - which reflects the time the job was
scheduled. If more than `max_job_age` seconds have passed, the job is skipped
and marked dead. It next checks for an `expire` value; if that value is in the
past, the job is skipped and marked dead. Otherwise, it will be run as usual.

### DIFF
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -33,12 +33,13 @@ class YamlConfig(MutableMapping):
             self._conf = dict()
 
     def load(self, conf=None):
-        if conf:
+        if conf is not None:
             if isinstance(conf, dict):
                 self._conf = conf
-            else:
+                return
+            elif conf:
                 self._conf = yaml.safe_load(conf)
-            return
+                return
         if os.path.exists(self.yaml_path):
             with open(self.yaml_path) as f:
                 self._conf = yaml.safe_load(f)
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -158,6 +158,7 @@ class TeuthologyConfig(YamlConfig):
         'job_threshold': 500,
         'lab_domain': 'front.sepia.ceph.com',
         'lock_server': 'http://paddles.front.sepia.ceph.com/',
+        'max_job_age': 1209600,  # 2 weeks
         'max_job_time': 259200,  # 3 days
         'nsupdate_url': 'http://nsupdate.front.sepia.ceph.com/update',
         'results_server': 'http://paddles.front.sepia.ceph.com/',
--- a/teuthology/dispatcher/__init__.py
+++ b/teuthology/dispatcher/__init__.py
@@ -22,6 +22,7 @@ from teuthology.config import config as teuth_config
 from teuthology.dispatcher import supervisor
 from teuthology.exceptions import BranchNotFoundError, CommitNotFoundError, SkipJob, MaxWhileTries
 from teuthology.lock import ops as lock_ops
+from teuthology.util.time import parse_timestamp
 from teuthology import safepath
 
 log = logging.getLogger(__name__)
@@ -234,6 +235,8 @@ def find_dispatcher_processes() -> Dict[str, List[psutil.Process]]:
 
 def prep_job(job_config, log_file_path, archive_dir):
     job_id = job_config['job_id']
+    check_job_expiration(job_config)
+
     safe_archive = safepath.munge(job_config['name'])
     job_config['worker_log'] = log_file_path
     archive_path_full = os.path.join(
@@ -306,6 +309,31 @@ def prep_job(job_config, log_file_path, archive_dir):
         raise RuntimeError("teuthology branch %s at %s not bootstrapped!" %
                            (teuthology_branch, teuth_bin_path))
     return job_config, teuth_bin_path
+
+
+def check_job_expiration(job_config):
+    job_id = job_config['job_id']
+    expired = False
+    now = datetime.datetime.now(datetime.timezone.utc)
+    if expire_str := job_config.get('timestamp'):
+        expire = parse_timestamp(expire_str) + \
+            datetime.timedelta(seconds=teuth_config.max_job_age)
+        expired = expire < now
+    if not expired and (expire_str := job_config.get('expire')):
+        try:
+            expire = parse_timestamp(expire_str)
+            expired = expired or expire < now
+        except ValueError:
+            log.warning(f"Failed to parse job expiration: {expire_str=}")
+            pass
+    if expired:
+        log.info(f"Skipping job {job_id} because it is expired: {expire_str} is in the past")
+        report.try_push_job_info(
+            job_config,
+            # TODO: Add a 'canceled' status to paddles, and use that.
+            dict(status='dead'),
+        )
+        raise SkipJob()
 
 
 def lock_machines(job_config):
--- a/teuthology/test/test_misc.py
+++ b/teuthology/test/test_misc.py
@@ -1,12 +1,13 @@
 import argparse
-
-from unittest.mock import Mock, patch
-from teuthology.orchestra import cluster
-from teuthology.config import config
-from teuthology import misc
+import pytest
 import subprocess
 
-import pytest
+from unittest.mock import Mock, patch
+
+from teuthology import misc
+from teuthology.config import config
+from teuthology.orchestra import cluster
+from teuthology.orchestra.remote import Remote
 
 
 class FakeRemote(object):
@@ -26,7 +27,7 @@ def test_sh_truncate(caplog):
 
 def test_sh_fail(caplog):
     with pytest.raises(subprocess.CalledProcessError) as excinfo:
-        misc.sh("/bin/echo -n AB ; /bin/echo C ; exit 111", 2) == "ABC\n"
+        misc.sh("/bin/echo -n AB ; /bin/echo C ; exit 111", 2)
     assert excinfo.value.returncode == 111
     for record in caplog.records:
         if record.levelname == 'ERROR':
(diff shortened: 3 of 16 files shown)


teuthology/test/test_misc.py
```diff
@@ -26,45 +27,36 @@ def test_sh_truncate(caplog):
 
 def test_sh_fail(caplog):
     with pytest.raises(subprocess.CalledProcessError) as excinfo:
-        misc.sh("/bin/echo -n AB ; /bin/echo C ; exit 111", 2) == "ABC\n"
+        assert misc.sh("/bin/echo -n AB ; /bin/echo C ; exit 111", 2) == "AB\nC\n"
```
kshtsk, line 30: Why? 'echo -n' means there should not be a new line after AB

zmc: Thank you, this is a good catch. I was fixing the lack of asserts in `test_sh_progress`, and made this similar change here. The test passed, but that's because the `assert` never happens since the call raises. I'll revert these two lines

zmc:
I scheduled two runs to test the feature:
- `--expire 1d` ran and passed: https://pulpito-ng.ceph.com/runs/zack-2024-08-01_20:41:26-teuthology:no-ceph-main-distro-default-null
- `--expire 1s` was skipped as expected: https://pulpito-ng.ceph.com/runs/zack-2024-08-01_20:42:01-teuthology:no-ceph-main-distro-default-null

teuthology/config.py
```diff
@@ -33,12 +33,13 @@ def __init__(self, yaml_path=None):
...
-        if conf:
+        if conf is not None:
             if isinstance(conf, dict):
                 self._conf = conf
-            else:
+                return
+            elif conf:
                 self._conf = yaml.safe_load(conf)
-            return
+                return
```
kshtsk, line 42: Why duplicate returns?

teuthology/dispatcher/__init__.py
```diff
@@ -306,6 +309,31 @@ def prep_job(job_config, log_file_path, archive_dir):
     return job_config, teuth_bin_path
 
 
+def check_job_expiration(job_config):
```
kshtsk, line 312: First of all, I'd like to see inline docs for the method.
Second, since this method is used both in dispatcher and supervisor maybe it is the best to create a separate module for job, and move all related methods there, for example, some of them can be taken from schedule.

teuthology/config.py
kshtsk: nit

teuthology/config.py
```diff
@@ -157,6 +158,7 @@ class TeuthologyConfig(YamlConfig):
         'job_threshold': 500,
         'lab_domain': 'front.sepia.ceph.com',
         'lock_server': 'http://paddles.front.sepia.ceph.com/',
+        'max_job_age': 1209600,  # 2 weeks
```
kshtsk, line 161: nit: A number looks enormous :-) maybe we can switch to minutes instead of seconds

teuthology/util/time.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    try:
+        dt = datetime.fromisoformat(timestamp)
+    except ValueError:
+        dt = datetime.strptime(timestamp, TIMESTAMP_FMT)
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt
+
+def parse_offset(offset: str) -> timedelta:
```
kshtsk, line 23: Add description for the methods.

teuthology/util/time.py
```diff
@@ -0,0 +1,41 @@
...
+    err_msg = "Offsets must either be an ISO-formatted timestamp or " \
+        f"a relative value like '1d', '7h', '45m'. Got: {offset}"
```
kshtsk, line 25: A weak is missing here.

teuthology/util/time.py
```diff
@@ -0,0 +1,41 @@
...
+    unit = match.groups()[1]
+    match unit:
+        case 'm':
+            return timedelta(minutes=num)
+        case 'h':
+            return timedelta(hours=num)
+        case 'd':
+            return timedelta(days=num)
+        case 'w':
+            return timedelta(weeks=num)
```
kshtsk, line 39: Maybe we can add a seconds for convenience?

teuthology/util/time.py
```diff
@@ -0,0 +1,41 @@
...
+    try:
+        dt = datetime.fromisoformat(timestamp)
+    except ValueError:
+        dt = datetime.strptime(timestamp, TIMESTAMP_FMT)
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt
+
+def parse_offset(offset: str) -> timedelta:
+    err_msg = "Offsets must either be an ISO-formatted timestamp or " \
```
kshtsk, line 24: ISO-formatted timestamp, could please elaborate, though, I haven't seen in the code it is parsing ISO format handling, because we fail with ValueError exception if we do not match the relative value.

kshtsk: And did you mean ISO 8601? As far as I understand it is something like `[HH[:MM[:SS[.mmm[uuu]]]]][+HH:MM]`.

kshtsk: Btw, I only could found scans of older 2016 year standard (the recent is 2019) and I guess they use a duration
image

teuthology/test/test_misc.py
```diff
@@ -26,45 +27,36 @@ def test_sh_truncate(caplog):
...
+        misc.sh("/bin/echo -n AB ; /bin/echo C ; exit 111", 2)
     assert excinfo.value.returncode == 111
     for record in caplog.records:
         if record.levelname == 'ERROR':
             assert ('replay full' in record.message or
                     'ABC\n' == record.message)
 
 def test_sh_progress(caplog):
-    misc.sh("echo AB ; sleep 5 ; /bin/echo C", 2) == "ABC\n"
+    assert misc.sh("echo AB ; sleep 0.1 ; /bin/echo C", 2) == "AB\nC\n"
```
kshtsk, line 38: nice, that's probably improve test timing 5 seconds is overkill

teuthology/test/test_misc.py
```diff
@@ -26,45 +27,36 @@ def test_sh_truncate(caplog):
...
     assert ':sh: ' in records[0].message
     assert 'AB' == records[1].message
     assert 'C' == records[2].message
-    #
-    # With a sleep 5 between the first and the second message,
-    # there must be at least 2 seconds between the log record
-    # of the first message and the log record of the second one
-    #
-    assert (records[2].created - records[1].created) > 2
+    assert records[2].created > records[1].created
```
kshtsk, line 43: the comment explaining 5 seconds sleep with at least 2 seconds difference looked confusing for me, but maybe it had some background

teuthology/config.py
zmc: I don't quite understand the question; there are two different return statement for two different conditions, and if neither condition is satisfied the method continues

teuthology/test/test_misc.py
zmc: I agree; I think I wrote it and it still confuses me, so clearly it's of no value!

teuthology/config.py
kshtsk: in both cases it returns nothing, there is no sense to pull return statements into conditionals of is dict type checking.

zmc: I still don't understand the concern here. Are you saying you think there's a bug here, or some sort of style issue?

kshtsk: style, no bug, I've got concern, because it looks unnecessary change and it does not improve anything for my taste, sorry about that, I guess, nit means you can ignore it

teuthology/util/time.py
```diff
@@ -0,0 +1,52 @@
...
+        case 'm':
+            return timedelta(minutes=num)
+        case 'h':
+            return timedelta(hours=num)
+        case 'd':
+            return timedelta(days=num)
+        case 'w':
+            return timedelta(weeks=num)
+        case _:
+            raise ValueError(err_msg)
```
kshtsk, line 52: Suggesting for the followup PR:
```
    m = re.match(r"^\s*"
                 r"(?:(?P<weeks>\d+)\s*w\s*)?"
                 r"(?:(?P<days>\d+)\s*d\s*)?"
                 r"(?:(?P<hours>\d+)\s*h\s*)?"
                 r"(?:(?P<minutes>\d+)\s*m\s*)?",
                 r"(?:(?P<seconds>\d+)\s*s\s*)?$",
                 offset.lower())
    if match is None:
        raise ValueError(err_msg)
    args = {k: int(v or "0") for k, v in m.groupdict().items()}
    delta = datetime.timedelta(**args)
    return delta
```

teuthology/util/test/test_time.py
```diff
@@ -0,0 +1,53 @@
...
+        ["1s", timedelta(seconds=1)],
+        ["1m", timedelta(minutes=1)],
+        ["1h", timedelta(hours=1)],
+        ["1d", timedelta(days=1)],
+        ["1w", timedelta(weeks=1)],
+        ["365d", timedelta(days=365)],
+        ["1x", ValueError],
+        ["-1m", ValueError],
+        ["0xde", ValueError],
+        ["frog", ValueError],
```
kshtsk, line 45: How 'bout the case: "7dwarfs"?

kshtsk (dismissed):

kshtsk:
rebase needed

kshtsk (approved):

PR #1984, opened 2024-08-01
Update docs/laptop

kshtsk:


zmc:
The SSH tunnel feature looks interesting but doesn't feel like it should be in a docs update

kshtsk:
> The SSH tunnel feature looks interesting but doesn't feel like it should be in a docs update

yeah... I need it here at the moment for testing my unexpectedly complicated setup and was going to move it out to a separate PR

zmc (approved):

PR #1985, opened 2024-08-02
Fix access denied from psutil on macOS

kshtsk:
On macOS dispatcher while trying to go through process list gets stuck
on some of the system processes like launchd, logd, systemstats, etc.
and quites unexpectedly with PermissionError and psutil.AccessDenied
exceptions.

Fixes: https://tracker.ceph.com/issues/67313

### DIFF
--- a/teuthology/dispatcher/__init__.py
+++ b/teuthology/dispatcher/__init__.py
@@ -205,6 +205,8 @@ def find_dispatcher_processes() -> Dict[str, List[psutil.Process]]:
     def match(proc):
         try:
             cmdline = proc.cmdline()
+        except psutil.AccessDenied:
+            return False
         except psutil.ZombieProcess:
             return False
         if len(cmdline) < 3:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration.yml
```diff
@@ -6,7 +6,9 @@ jobs:
   test:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
+      - name: Install docker-compose
+        run: sudo apt-get install -y docker-compose
```
zmc, line 11: This didn't really work, and the issue is intermittent. Let's leave it out of this PR and see if a simple retry gets us past the issue for now

kshtsk: I was trying to fix it fast, but right, dropping it for now, am not sure missing docker-compose is gonna be fixed by retrying :-D

kshtsk:
the integration failure is unrelated

zmc (approved):

PR #1987, opened 2024-08-02
Migrate to docker compose v2

zmc:
https://github.com/actions/runner-images/issues/9692
https://docs.docker.com/compose/migrate/


.github/workflows/integration.yml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v2
```
kshtsk, line 9: May we use v4 already?

zmc: I don't understand the question

kshtsk: use `actions/checkout@v4` instead

kshtsk (approved):

PR #1989, opened 2024-08-02
lock/ops: try to stop_node before unlock

kshtsk:
We should try to call stop_node before unlock and don't even try to do this for vms.


zmc:
Does not regress with smithis: https://pulpito-ng.ceph.com/runs/zack-2024-08-07_19:22:49-teuthology:no-ceph-main-distro-default-smithi

zmc (approved):

PR #1991, opened 2024-08-04
bootstrap: fix VENV variable utilization

kshtsk:
Allow bootstrap to use path outside of teuthology repo root directory for building virtualenv.


zmc (approved):
Nice

PR #1994, opened 2024-08-05
teuthology/suite: add len() lupa function

batrick:
Namespacing for this is weird which is why I missed it.


athanatos (approved):

kshtsk:
maybe add 
Fixes: 96f38f2136dd2a4f64cc741919b85dc65a9df536

batrick:
> maybe add Fixes: [96f38f2](https://github.com/ceph/teuthology/commit/96f38f2136dd2a4f64cc741919b85dc65a9df536)

It's not really "fixing" that commit at all. It's just adding another function to the allowlist.

kshtsk:
> > maybe add Fixes: [96f38f2](https://github.com/ceph/teuthology/commit/96f38f2136dd2a4f64cc741919b85dc65a9df536)
> 
> It's not really "fixing" that commit at all. It's just adding another function to the allowlist.

I thought it is follow up commit, because of the comment:
> Namespacing for this is weird which is why I missed it.

I am sorry I was wrong.

PR #1995, opened 2024-08-06
contextutil.safe_while does not sleep between tries if needed to loop infinitely

kshtsk:
This PR provides fixes for several issues introduced in #1816:

1) make use of `tries` argument in `list_lock()` instead of hardcoded infinite loop
2) make `safe_while()` sleep between tries again when `tries` < 0, i.e. in case of infinite retries
3) (unrelated) report number of spent tries by safe_while correctly
4) try to get `list_lock()` only once when we're trying get architectures by given machine type for `teuthology-suite`.

https://tracker.ceph.com/issues/67377

### DIFF
--- a/teuthology/contextutil.py
+++ b/teuthology/contextutil.py
@@ -115,20 +115,20 @@ class safe_while(object):
 
         msg = msg.format(
             action=self.action,
-            tries=self.counter,
+            tries=self.counter - 1,
             total=self.total_seconds,
         )
         return msg
 
     def __call__(self):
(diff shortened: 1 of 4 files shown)


kshtsk:
@zmc what I didn't understand from #1816, was it desired to run tries infinitely by default, or it is just 10 times?

kshtsk:
https://tracker.ceph.com/issues/67377

zmc (approved):

PR #1998, opened 2024-08-14
config: allow reading teuthology config from env var location

phlogistonjohn:
Allow changing the default "user" location of the teuthology configuration yaml using the (optional) TEUTHOLOGY_CONFIG environment variable. This change aids my effort to run a customized local teuthology environment.

### DIFF
--- a/teuthology/config.py
+++ b/teuthology/config.py
@@ -7,6 +7,12 @@ except ImportError:
     from collections import MutableMapping
 
 
+# Configuration constants
+SYSTEM_CONFIG_PATH = '/etc/teuthology.yaml'
+USER_CONFIG_PATH = '~/.teuthology.yaml'
+CONFIG_PATH_VAR_NAME = 'TEUTHOLOGY_CONFIG'  # name of env var to check
+
+
 def init_logging():
     log = logging.getLogger(__name__)
     return log
@@ -135,9 +141,9 @@ class TeuthologyConfig(YamlConfig):
     """
     This class is intended to unify teuthology's many configuration files and
     objects. Currently it serves as a convenient interface to
-    ~/.teuthology.yaml and nothing else.
+    ~/.teuthology.yaml or equivalent.
     """
-    yaml_path = os.path.join(os.path.expanduser('~/.teuthology.yaml'))
+    yaml_path = USER_CONFIG_PATH  # yaml_path is updated in _get_config_path
     _defaults = {
         'archive_base': '/home/teuthworker/archive',
         'archive_upload': None,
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Recursion error looks real and possibly related to this change. Will investigate.

teuthology/config.py
```diff
@@ -285,10 +284,13 @@ def set_config_attr(obj):
 
 
 def _get_config_path():
+    config_path = os.environ.get('TEUTHOLOGY_CONFIG', '~/.teuthology.yaml')
+    config_path = os.path.join(os.path.expanduser(config_path))
```
kshtsk, line 288: Bad style, overriding variable.

kshtsk: Also if we declare the env for the config, we should stop immediately if such file does not exists.

teuthology/config.py
```diff
@@ -137,7 +137,6 @@ class TeuthologyConfig(YamlConfig):
     objects. Currently it serves as a convenient interface to
     ~/.teuthology.yaml and nothing else.
     """
-    yaml_path = os.path.join(os.path.expanduser('~/.teuthology.yaml'))
```
kshtsk, line 140: I guess if drop this line you have to update the comment above about "interface to ~/.teuthology.yaml and nothing else".

teuthology/config.py
```diff
@@ -285,10 +284,13 @@ def set_config_attr(obj):
...
+    config_path = os.path.join(os.path.expanduser(config_path))
     system_config_path = '/etc/teuthology.yaml'
-    if not os.path.exists(TeuthologyConfig.yaml_path) and \
-            os.path.exists(system_config_path):
-        return system_config_path
-    return TeuthologyConfig.yaml_path
+    for path in (config_path, system_config_path):
+        if os.path.exists(path):
+            return path
+    return None
```
kshtsk, line 293: The idea was to always return a meaningful value.

phlogistonjohn:
@zmc is there a primer on how & what basic tests to run locally? 
I tried cleaning this up and testing multiple scenarios on my machine (no env var, env var but no file, etc) but the CI tests still don't like me. I presume it's because this code is doing a bunch of stuff at import time (something I don't care for) and that makes it hard to know how these test suites will handle it.

I'd like to upstream parts of my "hacked up local teuthology" and this is merely the first part. So getting a better handle on how you all test it would hopefully avoid me stumbling over the same problems again and again. Thanks.

kshtsk:
any updates?

teuthology/config.py
```diff
@@ -7,6 +7,12 @@
     from collections import MutableMapping
 
 
+# Configuration constants
+_SYSTEM_CONFIG_PATH = '/etc/teuthology.yaml'
+_USER_CONFIG_PATH = '~/.teuthology.yaml'
+_ALT_PATH_ENV_VAR = 'TEUTHOLOGY_CONFIG'
+
```
kshtsk, line 14: According to PEP8 there is no need to add _ (underscore) in the beginning of the constant name, may we have just `SYSTEM_CONFIG_PATH` and `USER_CONFIG_PATH`?
Also, I think to keep similar naming to env, like `CONFIG_PATH_ENV_NAME`.

phlogistonjohn: I like to use leading underscores on constants that are not meant to be used outside the module defining them. However, I don't feel very strongly about this here, so OK... but that may not always be the case in the future ;-)

phlogistonjohn:
Ready for (more) review

kshtsk (approved):
lgtm

PR #2000, opened 2024-08-20
orchestra: Respect username if set in ssh config

zmc:


dmick (approved):
Looks like a good compromise to me.  I've added code to the sepia-fog-builder job to verify this configuration and error out if not there.

PR #2004, opened 2024-09-18
suite: use owner value for user name

zmc:
pulpito-ng is gaining the ability to kill runs, and for non-admin users will match the user's github username with the test run's owner value to decide if the user is allowed to kill the run (admin users can kill any run). To generate the run name, we normally use the active user account's name.

It could be confusing if a user passed `--owner` but then still sees their unix account name used to name the run; this change lets us use more consistent naming without adding yet another flag (`--user`) to `teuthology-suite`.

Edit: The pulpito-ng PR: https://github.com/ceph/pulpito-ng/pull/51

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -45,10 +45,8 @@ class Run(object):
         # We assume timestamp is a datetime.datetime object
         self.timestamp = self.args.timestamp or \
             datetime.datetime.now().strftime(TIMESTAMP_FMT)
-        self.user = self.args.user or pwd.getpwuid(os.getuid()).pw_name
-
+        self.user = self.args.owner or pwd.getpwuid(os.getuid()).pw_name
         self.name = self.make_run_name()
-
         if self.args.ceph_repo:
             config.ceph_git_url = self.args.ceph_repo
         if self.args.suite_repo:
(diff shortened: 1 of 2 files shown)


VallariAg (dismissed):
If I understand it right, the run's username and owner both will be set to value of `--owner` input with this PR? 

In the current implementation of pulpito-ng's kill-run, we check if run belongs to user by looking at the owner (which is stored at job level) and never decide ownership based on the username. So if you use `--owner zmc` then pulpito-ng should see you as the owner of the run even if the run name starts with "zack-". Let me know if that's not the case!

Here are ownership checks in:
t-api: https://github.com/ceph/teuthology-api/pull/55/files#diff-02d3c9819201f270bdf30499d8e189622dc5d969e96d6582aef93f0e9e029c48R31-R33

pulpito-ng: 
1. https://github.com/ceph/pulpito-ng/pull/51/files#diff-aeb4ab3acaee02c9aa9c960ec5ee9b3e56c316802dc7e7ae0284b5702016c027R59
2. https://github.com/ceph/pulpito-ng/pull/51/files#diff-5db46c41bfe3b6e18f278e8721577a9c75851c2f3e6e519a61b6fe54a74d7b03R39

VallariAg (approved):
Looks great! Tested it when scheduling https://pulpito.ceph.com/vallari-2024-10-01_13:15:47-nvmeof:basic-main-distro-default-smithi/

PR #2005, opened 2024-09-24
teuthology/task: drop packages that are not built

joscollin:
FIxes: https://tracker.ceph.com/issues/68037

PR #2006, opened 2024-10-01
teuthology/task: drop packages that are not built

joscollin:
This is a followup fix of https://github.com/ceph/teuthology/pull/2005
Fixes the error `Unable to find a match: python3-ceph`.

Fixes: https://tracker.ceph.com/issues/68037


zmc (approved):
Thanks @joscollin

PR #2008, opened 2024-10-17
teuthology/suite: pass the base config to scripts

batrick:
This PR is to support: https://github.com/ceph/ceph/pull/60386

Also, this PR fixes a bug: `teuthology/tasks/kernel: default branch should be "distro"`. When the `kernel` dictionary is empty, the `kernel` task should assume `distro`.

### DIFF
--- a/teuthology/suite/merge.py
+++ b/teuthology/suite/merge.py
@@ -5,6 +5,7 @@ import os
 from types import MappingProxyType
 import yaml
 
+from teuthology.config import JobConfig
 from teuthology.suite.build_matrix import combine_path
 from teuthology.suite.util import strip_fragment_path
 from teuthology.misc import deep_merge
@@ -115,6 +116,7 @@ def config_merge(configs, suite_name=None, **kwargs):
     the entire job (config) from the list.
     """
     seed = kwargs.setdefault('seed', 1)
+    base_config = kwargs.setdefault('base_config', JobConfig())
     if not isinstance(seed, int):
         log.debug("no valid seed input: using 1")
         seed = 1
@@ -128,7 +130,7 @@ def config_merge(configs, suite_name=None, **kwargs):
         if suite_name is not None:
             desc = combine_path(suite_name, desc)
 
-        yaml_complete_obj = {}
+        yaml_complete_obj = copy.deepcopy(base_config.to_dict())
         deep_merge(yaml_complete_obj, dict(TEUTHOLOGY_TEMPLATE))
         for path in paths:
             if path not in yaml_cache:
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -9,7 +9,6 @@ import time
 
 from humanfriendly import format_timespan
 
-from tempfile import NamedTemporaryFile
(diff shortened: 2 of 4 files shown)


batrick:
> Also, this PR fixes a bug: `teuthology/tasks/kernel: default branch should be "distro"`. When the `kernel` dictionary is empty, the `kernel` task should assume `distro`.

Example: https://pulpito.ceph.com/pdonnell-2024-10-21_14:21:05-fs-wip-pdonnell-testing-20241019.005706-debug-wip-pdonnell-i66704-default-smithi/

teuthology/suite/merge.py
```diff
@@ -115,20 +115,22 @@ def config_merge(configs, suite_name=None, **kwargs):
...
     """
     seed = kwargs.setdefault('seed', 1)
+    base_config = kwargs.setdefault('base_config', {})
     if not isinstance(seed, int):
         log.debug("no valid seed input: using 1")
         seed = 1
     log.debug("configuring Lua randomseed to %d", seed)
     L.execute(f'local math = require"math"; math.randomseed({seed});')
     new_script = L.eval('new_script')
+    base_config_yaml = yaml.safe_load(str(base_config))
```
zmc, line 125: It's not immediately obvious why we'd need to create a copy of `base_config` here and then again on 133 - could you explain the reasoning?

zmc (commented):
Nice! I just have one question

teuthology/suite/merge.py
batrick: It's to basically enforce a read-only view of the base_config.

batrick: I tried to do `copy.deepcopy(base_config)` below but ran into inscrutable recursion errors that didn't seem worth debugging. So I'm just reloading it as a legit yaml dictionary here.

zmc: I think I would suggest just doing a `yaml_complete_obj = copy.deepcopy(base_config.to_dict())` on 133 and dropping this statement.

batrick: ack

batrick: done

teuthology/suite/merge.py
```diff
@@ -115,6 +115,7 @@ def config_merge(configs, suite_name=None, **kwargs):
     the entire job (config) from the list.
     """
     seed = kwargs.setdefault('seed', 1)
+    base_config = kwargs.setdefault('base_config', {})
```
zmc, line 118:
```suggestion
    base_config = kwargs.setdefault('base_config', JobConfig())
```
And add `from teuthology.config import JobConfig` up top

batrick: thanks, done

batrick:
https://pulpito.ceph.com/pdonnell-2024-10-23_02:34:42-fs-wip-pdonnell-testing-20241019.005706-debug-wip-pdonnell-i66704-default-smithi

batrick:
@zmc good to go?

zmc (approved):

PR #2010, opened 2024-10-25
teuthology: Add support for seek and sync in write_file

chrisphoffman:
Add support for seek and sync in write_file.

This patch will allow for partial writes and sync on completion of write.

This is to support strided writes for cephfs on a multi-mount system.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -265,9 +265,12 @@ class RemoteShell(object):
 
 
     def write_file(self, path, data, sudo=False, mode=None, owner=None,
-                                     mkdir=False, append=False):
+                                     mkdir=False, append=False, bs=None,
+                                     offset=None, sync=False):
         """
         Write data to remote file
+
+        The data written in 512-byte blocks, provide `bs` to use bigger blocks.
 
         :param path:    file path on remote host
         :param data:    str, binary or fileobj to be written
@@ -276,11 +279,20 @@ class RemoteShell(object):
         :param owner:   set file owner if provided
         :param mkdir:   preliminary create the file directory, defaults False
         :param append:  append data to the file, defaults False
+        :param bs:      write up to N bytes at a time if provided, default is 512 in `dd`
+        :param offset:  number of bs blocks to seek to in file, defaults 0
+        :param sync:    sync file after write is complete if provided
         """
         dd = 'sudo dd' if sudo else 'dd'
         args = dd + ' of=' + path
         if append:
             args += ' conv=notrunc oflag=append'
+        if bs:
+            args += ' bs=' + str(bs)
+        if offset:
+            args += ' seek=' + str(offset)
+        if sync:
+            args += ' conv=sync'
         if mkdir:
             mkdirp = 'sudo mkdir -p' if sudo else 'mkdir -p'
             dirpath = os.path.dirname(path)
--- a/teuthology/orchestra/test/test_remote.py
+++ b/teuthology/orchestra/test/test_remote.py
@@ -220,3 +220,15 @@ class TestRemote(object):
         rem2 = remote.Remote(name='jdoe@xyzzy.example.com', ssh=self.m_ssh)
         rem2._runner = m_run
         assert not rem2.is_container
+
+    @patch("teuthology.orchestra.remote.Remote.run")
+    def test_write_file(self, m_run):
+        file = "fakefile"
+        contents = "fakecontents"
+        rem = remote.Remote(name='jdoe@xyzzy.example.com', ssh=self.m_ssh)
+
+        remote.Remote.write_file(rem, file, contents, bs=1, offset=1024)
+        m_run.assert_called_with(args=f"set -ex\ndd of={file} bs=1 seek=1024", stdin=contents, quiet=True)
+
+        remote.Remote.write_file(rem, file, contents, sync=True)
+        m_run.assert_called_with(args=f"set -ex\ndd of={file} conv=sync", stdin=contents, quiet=True)


teuthology/misc.py
```diff
@@ -512,15 +512,18 @@ def gen_addresses():
...
 
     :param remote: Remote site.
     :param path: Path on the remote being written to.
     :param data: Data to be written.
+    :param bs: The block size to be used.
+    :param offset: The number of bytes to seek to in file.
+    :param sync: Sync file.
     """
-    remote.write_file(path, data)
+    remote.write_file(path, data, bs=bs, offset=offset, sync=sync)
```
kshtsk, line 526: This method is left for backwards compatibility, for newer code, please, use only remote context function, so I mean, please, don't change this method.

teuthology/orchestra/remote.py
```diff
@@ -276,11 +277,20 @@ def write_file(self, path, data, sudo=False, mode=None, owner=None,
         :param owner:   set file owner if provided
         :param mkdir:   preliminary create the file directory, defaults False
         :param append:  append data to the file, defaults False
+        :param bs:      block size
```
kshtsk, line 280: Please, add info about default value.

teuthology/orchestra/remote.py
```diff
@@ -276,11 +277,20 @@ def write_file(self, path, data, sudo=False, mode=None, owner=None,
...
+        :param offset:  number of bytes to seek to in file
```
kshtsk, line 281: Add information about defaults.

teuthology/orchestra/remote.py
```diff
@@ -276,11 +277,20 @@ def write_file(self, path, data, sudo=False, mode=None, owner=None,
...
+        :param sync:    sync file
```
kshtsk, line 282: Add default value description.

teuthology/orchestra/remote.py
kshtsk: Also give user more idea about what is the block size how it can be used. Probably this will require add some description above.

teuthology/orchestra/remote.py
kshtsk: Yeah... here is another point to elaborate, so user understand how to use _sync_ argument.

kshtsk:
Interesting idea, I would suggest to address the comments and maybe add unit tests.

kshtsk (changes requested):
Interesting idea, I would suggest to address the comments and maybe add unit tests.

chrisphoffman:
> Interesting idea, I would suggest to address the comments and maybe add unit tests.

Thanks for the review. Comments addressed and tests added. PTAL

teuthology/orchestra/test/test_cluster.py
```diff
@@ -230,3 +230,13 @@ def test_fails_with_invalid_owner(self, m_write_file):
...
+
+    @patch("teuthology.orchestra.remote.RemoteShell.write_file")
+    def test_write_file_offset(self, m_write_file):
+        self.c.write_file("filename", "content", bs=1, offset=1024)
+        m_write_file.assert_called_with("filename", "content", bs=1, offset=1024)
+
+    @patch("teuthology.orchestra.remote.RemoteShell.write_file")
+    def test_write_file_sync(self, m_write_file):
+        self.c.write_file("filename", "content", sync=True)
+        m_write_file.assert_called_with("filename", "content", sync=True)
```
kshtsk, line 242: Those are test not for the `remote.write_file` but for `cluster.remote_write`, I guess there should be new tests added to `test/test_remote.py` module instead.

teuthology/orchestra/remote.py
```diff
@@ -276,11 +277,20 @@ def write_file(self, path, data, sudo=False, mode=None, owner=None,
         :param owner:   set file owner if provided
         :param mkdir:   preliminary create the file directory, defaults False
         :param append:  append data to the file, defaults False
+        :param bs:      write up to N bytes at a time, defaults 512
```
kshtsk, line 280: I guess default is None, maybe just give a reference to `dd`, just add another line right after `Write data to remote file` above, like:

```The data written in 512-byte blocks, provide `bs` to use bigger blocks.```

File: teuthology/orchestra/remote.py [hunk 7]
```diff
@@ -276,11 +277,20 @@ def write_file(self, path, data, sudo=False, mode=None, owner=None,
...
         dd = 'sudo dd' if sudo else 'dd'
         args = dd + ' of=' + path
         if append:
             args += ' conv=notrunc oflag=append'
+        if bs:
+            args += ' bs=' + str(bs)
+        if offset:
+            args += ' seek=' + str(offset)
+        if sync:
+            args += ' conv=sync'
```
kshtsk, line 293: I don't remember exactly if the `conv` can be used multiple times in dd, if not we must take into consideration `notrunct` when `append` needed. Btw, one of use case for a unit test.

teuthology/orchestra/test/test_cluster.py
chrisphoffman: Done.

teuthology/orchestra/cluster.py
```diff
@@ -102,24 +102,28 @@ def sh(self, script, **kwargs):
         remotes = sorted(self.remotes.keys(), key=lambda rem: rem.name)
         return [remote.sh(script, **kwargs) for remote in remotes]
 
-    def write_file(self, file_name, content, sudo=False, perms=None, owner=None):
+    def write_file(self, file_name, content, sudo=False, perms=None, owner=None,
+                   **kwargs):
```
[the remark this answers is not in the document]
chrisphoffman: No need for cluster wide call at this time.

kshtsk (approved):

PR #2012, opened 2024-11-27
[PLACEHOLDER] teuthology: add support for graph walks of QA suites

batrick:
This is a placeholder PR for an upcoming branch I plan to push for my talk at Cephalocon 2024:

https://ceph2024.sched.com/event/1ktVR/the-art-of-teuthology-patrick-donnelly-ibm-inc

I want to link to the PR in the slides but the (preview) PR may not be pushed until the day of the talk (but the slides need to be submitted before then).


teuthology/repo_utils.py
```diff
@@ -95,7 +95,7 @@ def current_branch(path: str) -> str:
     return result
 
 
-def enforce_repo_state(repo_url, dest_path, branch, commit=None, remove_on_error=True):
+def enforce_repo_state(dest_clone, dest_path, repo_url, branch, commit=None, remove_on_error=True):
```
kshtsk, line 98: I'd prefer not to change initial signature and leave repo_url at first position, and add `dest_cone` as an optional argument with defaults value.

batrick: Sure, there will probably be lots of reworking to this code but it was a dirty hack to start because the clones after every QA suite tweak was driving me nuts. That's why it's in this PR.

teuthology/repo_utils.py
```diff
@@ -114,25 +114,100 @@ def enforce_repo_state(repo_url, dest_path, branch, commit=None, remove_on_error
...
         # remove_pyc_files(dest_path)
     except (BranchNotFoundError, CommitNotFoundError):
         if remove_on_error:
             shutil.rmtree(dest_path, ignore_errors=True)
         raise
 
+def bare_repo(git_dir):
+    log.info("bare_repo %s", git_dir)
+    args = ['git', 'init', '--bare', git_dir]
+    proc = subprocess.Popen(args)
```
phlogistonjohn, line 142: If you don't need a lot of the complexity that Popen provides. I heartily recommend using the simpler `subprocess.run` function (avialable since 3.5).  You can use the `check=True` parameter to have it raise exceptions on non-zero exit automatically too.

phlogistonjohn:
I think this is really cool improvement. I have a possibly silly suggestion, based on the nice graphs in your cephalocon presentation, would it be reasonable to have a mode that generates graphvis output that can be used to generate graphical representations of the steps?  This kind of visualization could be handy to debug/understand the steps as they're assembled.

batrick:
> I think this is really cool improvement. I have a possibly silly suggestion, based on the nice graphs in your cephalocon presentation, would it be reasonable to have a mode that generates graphvis output that can be used to generate graphical representations of the steps? This kind of visualization could be handy to debug/understand the steps as they're assembled.

I already played around with a few options (including a terminal representation, which is ... fun to grok). The code is there. I would ultimately want to:

- make a .svg for the path walk that produced a job, in the job dir
- make a .svg for the entire suite graph in the run dir
- make a .svg for a dry-run for inspection prior to scheduling

If you have tooling/library suggestions, please do share.

PR #2014, opened 2024-12-10
teuthology/misc: Add timeout parameter to stop_daemons_of_type for better flexibility

NitzanMordhai:
Updated stop_daemons_of_type to accept a timeout parameter, allowing dynamic control over the timeout value passed to the stop function of each daemon.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1148,15 +1148,17 @@ def ssh_keyscan_wait(hostname):
             log.info("try ssh_keyscan again for " + str(hostname))
         return success
 
-def stop_daemons_of_type(ctx, type_, cluster='ceph'):
+def stop_daemons_of_type(ctx, type_, cluster='ceph', timeout=300):
     """
     :param type_: type of daemons to be stopped.
+    :param cluster: Cluster name, default is 'ceph'.
+    :param timeout: Timeout in seconds for stopping each daemon.
     """
     log.info('Shutting down %s daemons...' % type_)
     exc = None
     for daemon in ctx.daemons.iter_daemons_of_role(type_, cluster):
         try:
-            daemon.stop()
+            daemon.stop(timeout)
         except (CommandFailedError,
                 CommandCrashedError,
                 ConnectionLostError) as e:


rzarzynski (approved):

kshtsk:
Is this tested already?

NitzanMordhai:
> Is this tested already?

i ran few test using my PR issue. didn't riched to timeout, but the param was passed

VallariAg:
@NitzanMordhai the timeout didn't work even though the param was passed? 
I'm not sure if timeouts are actually used in all kinds of daemon's `stop()` method:
It's used here: https://github.com/ceph/teuthology/blob/cca92a2a7751b3ad3aaeb27dfd57b9c09b206299/teuthology/orchestra/daemon/state.py#L131
But I don't see it being used in these overrides of systemd and cephadm daemons: 
https://github.com/ceph/teuthology/blob/cca92a2a7751b3ad3aaeb27dfd57b9c09b206299/teuthology/orchestra/daemon/systemd.py#L199
https://github.com/ceph/teuthology/blob/cca92a2a7751b3ad3aaeb27dfd57b9c09b206299/teuthology/orchestra/daemon/cephadmunit.py#L154

NitzanMordhai:
> @NitzanMordhai the timeout didn't work even though the param was passed?
> I'm not sure if timeouts are actually used in all kinds of daemon's `stop()` method:
> It's used here:

that's the purpose of that PR, to add the timeout so we can use it and not use the default

VallariAg:
@NitzanMordhai This PR relies on passing timeout to `daemon.stop(timeout)`. But I'm not sure if this `stop` method actually uses the timeout param (depending on daemon type).

NitzanMordhai:
it will be pass to teuthology/orchestra/daemon/state.py
that will wait until the daemon stops (or timeout)

VallariAg:
Great, it should work well then!

rzarzynski:
@VallariAg, @NitzanMordhai: is this ready to merge? If it is, let's not hesitate and unblock the PR in `ceph`.

zmc:
Is it the case that we don't want or don't care about timeouts in `CephadmUnit` or `SystemDState`? Those classes' `stop` methods do in fact ignore the `timeout` kwarg, even though they accept it.

NitzanMordhai:
> @VallariAg, @NitzanMordhai: is this ready to merge? If it is, let's not hesitate and unblock the PR in `ceph`.

yes, it is ready

NitzanMordhai:
> Is it the case that we don't want or don't care about timeouts in `CephadmUnit` or `SystemDState`? Those classes' `stop` methods do in fact ignore the `timeout` kwarg, even though they accept it.

@zmc It's more for a daemon/state, where we don't pass any timeout and we do care about it. this is just a parameter reveal, that was coded. but never pass along the tests

VallariAg:
This PR should solve for daemon/state.py type daemons created in [tasks/ceph.py](https://github.com/ceph/ceph/blob/091c7c5400f49e0ca16e8462ede8754b81f17b96/qa/tasks/ceph.py#L1925) so seems good to merge this PR to unblock the linked ceph PR.

NitzanMordhai:
@zmc any other thoughts? can we merge it so https://github.com/ceph/ceph/pull/61021 test can start?

PR #2021, opened 2025-01-20
dispatcher/supervisor: use job config only to unlock targets

kshtsk:
No need to use serializer, because everything needed to unlock machines already present in job config.


kshtsk:
This PR addresses https://tracker.ceph.com/issues/69599

zmc (approved):
Thanks for this!

PR #2023, opened 2025-01-30
orchestra: add ssh tunnelling support for target nodes

kshtsk:


teuthology/orchestra/connection.py
```diff
@@ -79,6 +93,29 @@ def connect(user_at_host, host_key=None, keep_alive=False, timeout=60,
...
+                    continue
+                bastion_host = bastion.get('host')
+                server = sshtunnel.SSHTunnelForwarder(
+                    bastion_host,
+                    ssh_username=bastion.get('user', None),
+                    ssh_password=bastion.get('word', None),
+                    ssh_pkey=bastion.get('identity'),
+                    remote_bind_address=(host, 22))
+                log.info(f'Starting tunnel to {bastion_host} for host {host}')
+                server.start()
```
zmc, line 111: I'm not familiar with this library, but does this not need to get cleaned up at some point?

teuthology/task/ansible.py
```diff
@@ -259,13 +259,61 @@ def generate_inventory(self):
...
+                        if port:
+                            opts += f" -p {port}"
+                        if pkey:
+                            opts += f" -i {pkey}"
+                        if user:
+                            opts += f" {user}@{host}"
+                        else:
+                            opts += f" {host}"
+                        cmd = f"ssh {opts}"
+                        if not host in proxy:
```
zmc, line 295:
```suggestion
                        if host not in proxy:
```

zmc (changes requested):
I'd like to see the majority of this code put in `orchestra.tunnel`, rather than sprinkled about. I also would want to see some tests, config validation and more logging.

PR #2024, opened 2025-01-30
orchestra/run: better handle weird args

kshtsk:
If one of the command args in remote.run is None then there is exception occurs while trying to qoute argument list:

   TypeError: expected string or bytes-like object

This fix addresses the issue and gives user better idea what is happening, and how to fix the error.

Fixes: https://tracker.ceph.com/issues/64452


zmc (approved):
Excellent

PR #2025, opened 2025-02-03
supervisor: Unlock nodes after reimage failure

zmc:
And some debugging which helped me discover this bug.


zmc:
@amathuria ping

amathuria (approved):
Looks good to me!

PR #2026, opened 2025-02-18
task/internal/__init__.py:  Revisit core program extraction regex

badone:
Fixes: https://tracker.ceph.com/issues/61586

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -334,7 +334,8 @@ def fetch_binaries_for_coredumps(path, remote):
                     log.error(e)
                     continue
             try:
-                dump_program = re.findall("from '([^']+)'", dump_out)[0]
+                dump_command = re.findall("from '([^ ']+)", dump_out)[0]
+                dump_program = dump_command.split()[0]
                 log.info(f' dump_program: {dump_program}')
             except Exception as e:
                 log.info("core doesn't have the desired format, moving on ...")


teuthology/task/internal/__init__.py
```diff
@@ -334,7 +334,7 @@ def fetch_binaries_for_coredumps(path, remote):
                     log.error(e)
                     continue
             try:
-                dump_program = re.findall("from '([^']+)'", dump_out)[0]
+                dump_program = re.findall("from '([^\\s']+)", dump_out)[0]
```
idryomov, line 337: Do you intend to match `\t`, `\n`, `\r`, `\f` or `\v` here?  It seems like the intention is to stop at a literal space character, so I'd avoid involving a character class:
```suggestion
                dump_program = re.findall("from '([^ ']+)", dump_out)[0]
```

badone: Yes, let's do that then.

idryomov (approved):

teuthology/task/internal/__init__.py
```diff
@@ -334,7 +334,7 @@ def fetch_binaries_for_coredumps(path, remote):
                     log.error(e)
                     continue
             try:
-                dump_program = re.findall("from '([^']+)'", dump_out)[0]
+                dump_program = re.findall("from '([^ ']+)", dump_out)[0]
```
kshtsk, line 337: or it is just missed `.split()[0]` at the end, however, for better readability maybe elaborate the code, like:
```
dump_command = re.findall("from '([^']+)", dump_out)[0]
dump_program = dump_command.split()[0]
```
If you like you could apply, or just approve the current changes.

badone: @kshtsk updated, thanks for the suggested improvement.

PR #2032, opened 2025-03-05
supervisor: Don't use "safe" unlock methods

zmc:
This fixes a bug where the supervisor would avoid locking its own nodes.

[Example from job 8142206](http://qa-proxy.ceph.com/teuthology/skanta-2025-02-20_02:03:15-rados-wip-bharath1-testing-2025-02-19-0904-distro-default-smithi/8142206/supervisor.8142206.log):
```
2025-03-04T22:01:06.922 ERROR:teuthology.lock.ops:Refusing to unlock smithi087 since it has an active job: skanta-2025-02-20_02:03:15-rados-wip-bharath1-testing-2025-02-19-0904-distro-default-smithi/8142206
```

Note that the job IDs are the same.

### DIFF
--- a/teuthology/dispatcher/supervisor.py
+++ b/teuthology/dispatcher/supervisor.py
@@ -269,7 +269,7 @@ def unlock_targets(job_config):
         return
     if job_config.get("unlock_on_failure", True):
         log.info('Unlocking machines...')
-        lock_ops.unlock_safe(locked, job_config["owner"])
+        lock_ops.unlock_many(locked, job_config["owner"])
 
 
 def run_with_watchdog(process, job_config):


teuthology/dispatcher/supervisor.py
```diff
@@ -269,7 +269,7 @@ def unlock_targets(job_config):
         return
     if job_config.get("unlock_on_failure", True):
         log.info('Unlocking machines...')
-        lock_ops.unlock_safe(locked, job_config["owner"], job_config["name"], job_config["job_id"])
+        lock_ops.unlock_many(locked, job_config["owner"])
```
VallariAg, line 272: Why did we switch to `unlock_many`? It doesn't seem to have any checks before unlocking the nodes.

VallariAg (approved):

jmundack:
@amathuria  - mind taking a look when you have a few mins?

zmc:
I discovered a potential issue with jobs that run too long and get killed; I'll update once I've gotten to the bottom of it.

zmc:
Superseded by #2036

PR #2033, opened 2025-03-05
node-cleanup: Grace period for inactive jobs

zmc:
And some formatting tweaks.


VallariAg:
Filed a fix based on what I understand about the cause of this issue (explained in PR description): https://github.com/ceph/teuthology/pull/2034

PR #2034, opened 2025-03-06
lock/query.py: add grace_time param to node_active_job()

VallariAg:
This is because we only want this grace period
when looking for stale jobs. Not when jobs are
being unlocked by the supervisor by unlock_one_safe().

Fixes: https://github.com/ceph/teuthology/pull/2033#issuecomment-2703733380

### DIFF
--- a/teuthology/lock/query.py
+++ b/teuthology/lock/query.py
@@ -127,17 +127,18 @@ def find_stale_locks(owner=None) -> List[Dict]:
     # running
     result = list()
     for node in nodes:
-        if node_active_job(node["name"]):
+        if node_active_job(node["name"], grace_time=5):
             continue
         result.append(node)
     return result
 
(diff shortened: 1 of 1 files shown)


jmundack:
Removed myself from @ceph/teuthology  - since i am not qualified to review it (was added by default as i setup the team for it)

amathuria:
@batrick @zmc  could one of you have a second look so we can merge this?

VallariAg:
Looks like the integration test is broken on main branch too: https://github.com/ceph/teuthology/actions/workflows/integration.yml

Related thread about this: https://github.com/ceph/teuthology/pull/1773#issuecomment-2692442872

zmc:
I should have mentioned that #2033 depended on #2022, which I'd also requested a review on. Apologies!

PR #2045, opened 2025-04-16
Fix packages duplication when extra system packages requested

kshtsk:
This PR provides fixes combination of two bugs:
- Mistaken appending elements to a list parameter of the function 'rpm': sequential calls of _update_package_list_and_install modified 'rpm' parameter from within the function, because 'rpm' is an object.
- Two attempts to add same packages because of conditional, just do it once after all checks occur.


kshtsk:
@zmc I'd like to ask you to review this fix, and btw, looks like GitHub will discontinue ubuntu-20.04 and we need to drop this case from actions now.

kamoltat (approved):
Yeah even the comments in the package list that you removed is a duplicate.
LGTM

PR #2050, opened 2025-05-28
Revive teuthology-openstack

deepssin:


teuthology/openstack/setup-openstack.sh
```diff
@@ -113,13 +117,19 @@ function apt_get_update() {
 }
 
 function setup_docker() {
-    if test -f /etc/apt/sources.list.d/docker.list ; then
-        echo "OK docker is installed"
-    else
-        sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
-        echo deb https://apt.dockerproject.org/repo ubuntu-trusty main | sudo tee -a /etc/apt/sources.list.d/docker.list
-        sudo apt-get -qq install -y docker-engine
+    local codename=$(lsb_release -sc)
```
kshtsk, line 120: Not many distro's have lsb_release by default, just use `source /etc/os-release`.

deepssin: Updated to use /etc/os-release as suggested, thanks for the pointer!

teuthology/suite/run.py
```diff
@@ -305,7 +305,7 @@ def choose_teuthology_branch(self):
             if not teuthology_branch:
                 teuthology_branch = actual_branch
             teuthology_sha1 = util.git_ls_remote(
-                f"file://{config.teuthology_path}",
+                f"file://$(realpath {config.teuthology_path})",
```
zmc, line 308: I would much rather see something like [pathlib](https://docs.python.org/3/library/pathlib.html#expanding-and-resolving-paths) used here :)

bootstrap
```diff
@@ -114,17 +115,28 @@ if [ -z "$NO_CLOBBER" ] && \
...
 
+PY_MAJOR=$($VENV/bin/python -c "import sys; print(sys.version_info[0])")
+PY_MINOR=$($VENV/bin/python -c "import sys; print(sys.version_info[1])")
+
+# Downgrade pip for older python versions
+if [ "$PY_MAJOR" -eq 3 ] && [ "$PY_MINOR" -lt 6 ]; then
+    $VENV/bin/python -m ensurepip
+    $VENV/bin/pip install pip==20.3.4 --force-reinstall
+else
+    echo "keeping default pip"
```
zmc, line 135: I think this message is unnecessary

requirements.txt
```diff
@@ -200,7 +200,9 @@ zope-event==5.0
     # via gevent
 zope-interface==6.1
     # via gevent
-
+openstacksdk==4.5.0
```
zmc, line 203: Does this need to be added to `setup.cfg` ?

teuthology/openstack/__init__.py
```diff
@@ -221,6 +212,13 @@ class OpenStack(object):
         'ubuntu-16.04-x86_64': 'https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-disk1.img',
         'ubuntu-16.04-aarch64': 'https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-arm64-disk1.img',
         'ubuntu-16.04-i686': 'https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-i386-disk1.img',
+        'ubuntu-18.04-x86_64': 'https://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-amd64.img',
+        'ubuntu-18.04-aarch64': 'https://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-arm64.img',
+        'ubuntu-18.04-i686': 'https://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-i386.img',
+        'ubuntu-20.04-x86_64': 'https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img',
+        'ubuntu-20.04-aarch64': 'https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-arm64.img',
+        'ubuntu-22.04-x86_64': 'https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64.img',
+        'ubuntu-22.04-aarch64': 'https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-arm64.img',
```
zmc, line 221: might as well add 24.04 here as well

zmc (changes requested):
Looks really good. Just some minor changes, I think. I'm curious how you test this as well if you don't mind writing up a short description

teuthology/openstack/__init__.py
deepssin: added

requirements.txt
deepssin: Openstack related dependencies are already there in setup.cfg so it makes sense to add openstacksdk as well , so added.

bootstrap
```diff
@@ -88,8 +89,11 @@ if [ -n "$missing" ]; then
         $install_pkg $missing
     else
         echo "Please install missing packages or run './bootstrap install'"
-        echo "$install_pkg $missing"
-        exit 1
+        echo "$install_pkg $missing , going for automatic install..."
+        $install_pkg $missing || {
```
zmc, line 93: The longstanding behavior of this script is that if it is passed "install" as an argument, like this:
```
./bootstrap install
```
then it will use sudo to install missing packages, whereas if it is run with no args it will not attempt to do anything as root. I don't think we should remove the "don't run things as root" default behavior

bootstrap
```diff
@@ -114,17 +118,26 @@ if [ -z "$NO_CLOBBER" ] && \
...
 if [ -z "$NO_CLOBBER" ] || [ ! -e $VENV ]; then
+    echo "Creating new venv at $VENV"
     $PYTHON_BIN -m venv $VENV
 fi
 
+PY_MAJOR=$($VENV/bin/python -c "import sys; print(sys.version_info[0])")
+PY_MINOR=$($VENV/bin/python -c "import sys; print(sys.version_info[1])")
+
+# Downgrade pip for older python versions
+if [ "$PY_MAJOR" -eq 3 ] && [ "$PY_MINOR" -lt 6 ]; then
```
zmc, line 134: We require python 3.10 because we use features introduced in that version: https://github.com/ceph/teuthology/blob/main/setup.cfg#L29

zmc (changes requested):
Generally looks great - just some bootstrap changes necessary I think

bootstrap
deepssin: I've updated the bootstrap script to restore the original behaviour. To avoid integration test failure , moved the dependencies(jq & curl) to Dockerfile.

zmc (approved):
Looks great! Thanks for making those changes.

PR #2054, opened 2025-06-12
teuthology/schedule.py: update parsed_yaml with base_config

VallariAg:
In `schedule_suite` method,
`self.base_config` gets updated many times while backtracking when using "--newest" flag. These changes were never updated to `parsed_yaml` (job yaml) in `configs`.
As `configs` is intialised before backtracking.

Before, we used to write base_config to a tmp
file and pass that to teuthology-schedule
which used to take care of updates to base_config. But this logic was removed in https://github.com/ceph/teuthology/pull/2008/files so the updates to base_config then never make
it to job yaml anymore.

Fixes: https://tracker.ceph.com/issues/71612

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -514,6 +514,8 @@ class Run(object):
                 '--',
             ])
             arg.extend(self.base_yaml_paths)
+            full_job_config = copy.deepcopy(self.base_config.to_dict())
+            deep_merge(parsed_yaml, full_job_config)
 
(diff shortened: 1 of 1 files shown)


zmc (approved):
thanks so much for tracking this down!

cbodley:
on teuthology's main branch, teuthology-suite is scheduling every job as "centos 9.stream" regardless of the distro specified in the job's yaml. a 'git bisect' showed that the regression was caused by this commit

PR #2055, opened 2025-06-12
suite/run.py: suite_hash should use backtracked sha1

kamoltat:
In schedule_suite() we only try to find backtracked sha1
for ceph sha1 when --newest is provided and packages are
not found in shaman. However, what about suite_sha1?
Currently, we do not use the backtracked sha1 for suite_sha1.
This commit will make sure that suite_sha1 is using
the backtracked sha1 when ceph_branch and suite_branch are the same and
ceph_repo and suite_repo are the same. This ensures that suite_sha1
only use the backtracked sha1 when the user clearly expects
shaman to not have the latest sha1 and want to use the --newest sha1
for both the ceph code base and suite qa code base.

Additionally, the unit-test in this PR will also cover the changes in:
https://github.com/ceph/teuthology/pull/2054

Fixes: https://tracker.ceph.com/issues/71668


amathuria (approved):
LGTM and thanks for adding the test!

zmc (approved):
What @amathuria said :)

PR #2060, opened 2025-07-01
Fix wrong os-type issue

VallariAg:
Revert previous fix for https://tracker.ceph.com/issues/71612 as it created a new regression https://tracker.ceph.com/issues/71911. 

Second commit adds a new fix to the original issue (71612).

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1009,6 +1009,17 @@ def deep_merge(a: DeepMerge, b: DeepMerge) -> DeepMerge:
         return a
     return b
 
+def update_key(key_to_update, a: dict, b: dict):
+    """
+    Update key (`key_to_update`) of dict `a` on all levels
+    to the values of same key in `b` dict.
+    """
+    for key, value in b.items():
+        if key == key_to_update:
+            a[key] = value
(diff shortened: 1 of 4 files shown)


VallariAg:
Just improved some doc strings in new push. 
Tested: https://pulpito.ceph.com/vallariag-2025-07-01_14:32:15-rgw:multifs-main-distro-default-smithi/

teuthology/misc.py
```diff
@@ -1009,6 +1009,17 @@ def deep_merge(a: DeepMerge, b: DeepMerge) -> DeepMerge:
...
+    """
+    Update key (`key_to_update`) of dict `a` on all levels
+    to the values of same key in `b` dict.
+    """
+    for key, value in b.items():
+        if key == key_to_update:
+            a[key] = value
+        elif isinstance(value, dict):
+            if key in a and isinstance(a[key], dict):
+                update_key(key_to_update, a[key], value)
```
kamoltat, line 1022: This recursive call looks safe enough as it only triggers if value is also another dictionary.
Therefore, it should terminate once we hit the last level.

kamoltat (approved):

PR #2061, opened 2025-07-01
opsys.py: Add Rocky 10

djgalloway:


teuthology/orchestra/opsys.py
```diff
@@ -41,6 +41,7 @@
     "rocky": {
         "8.10": "rocky",
         "9.5": "rocky",
+        "10.0": "rocky",
```
kshtsk, line 44: Looks good, I would add 9.6 too

kshtsk (approved):

djgalloway:
I don't think we have any intention of using Alma on the Ceph lab side.  If that's something that is needed, feel free.

Does it make sense / is it possible to just accept any variation of X.Y instead of just adding each individual minor version after every release?

kshtsk:
> I don't think we have any intention of using Alma on the Ceph lab side. If that's something that is needed, feel free.
> 
Teuthology code is used not only in the Sepia lab.

> Does it make sense / is it possible to just accept any variation of X.Y instead of just adding each individual minor version after every release?

I thought the code is more about to early version validation.

PR #2065, opened 2025-07-11
Add flavor to job yaml (at Job creation)

VallariAg:
Currently, when creating jobs, paddles
does not get Job.flavor information.
This leads to "flavor" getting set only
after the job is completed.

This commit adds "flavor" to job yaml config
so Job.flavor is set when creating jobs.


zmc (approved):
Thanks @VallariAg!

PR #2067, opened 2025-07-17
orchestra/remote: add resolve_ip method

kshtsk:
Add utility method to resolve a hostname from within remote host. This is useful to resolve ip address of the remote host itself, because getting ip address from transport object of ssh is not suitable because it may have only bastion host and port, which is not relevant for the purpose of configuring a cluster.

### DIFF
--- a/teuthology/orchestra/remote.py
+++ b/teuthology/orchestra/remote.py
@@ -461,6 +461,46 @@ class Remote(RemoteShell):
                 return
         raise RuntimeError("Could not determine interface/CIDR!")
 
+
+    def resolve_ip(self, name=None, ipv='4') -> str:
+        """
+        Resolve IP address of the remote host via remote host
+
+        Because remote object maybe behind bastion host we need
+        the remote host address resolvable from remote side.
+        So in order to the ip address we just call `host` remotely
+        and parse output like:
+            'smithi001.front.sepia.ceph.com has address 172.21.15.1\n'
+
+        :param name:    hostname to resolve, by defaults remote host itself.
+        :param ipv:     the IP version, 4 or 6, defaults to 4.
+
+        :raises:    :class:`Exception`: when the hostname cannot be resolved.
+        :raises:    :class:`ValueError`: when the ipv argument mismatch 4 or 6.
+
+        :returns:   str object, the ip addres of the remote host.
+        """
+        hostname = name or self.hostname
+        version = str(ipv)
+        if version in ['4', '6']:
+            remote_host_ip = self.sh(f'host -{ipv} {hostname}')
+        else:
+            raise ValueError(f'Unknown IP version {ipv}, expected 4 or 6')
+        # `host -4` or `host -6` may have multiline output: a host can have
+        # several address; thus try and find the first suitable
+        for info in remote_host_ip.split("\n"):
+            if version == '4' and 'has address' in info:
+                (host, ip) = info.strip().split(' has address ')
+                if hostname in host:
+                    return ip
+            elif version == '6' and 'has IPv6 address' in info:
+                (host, ip) = info.strip().split(' has IPv6 address ')
+                if hostname in host:
+                    return ip
+        else:
+            raise Exception(f'Cannot get IPv{ipv} address for the host "{hostname}" via remote "{self.hostname}"')
+
+
     @property
     def hostname(self):
         if not hasattr(self, '_hostname'):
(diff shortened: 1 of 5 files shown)


teuthology/orchestra/remote.py
```diff
@@ -461,6 +461,30 @@ def _set_iface_and_cidr(self):
...
+        and parse output like:
+            'smithi001.front.sepia.ceph.com has address 172.21.15.1\n'
+        """
+        hostname = name or self.hostname
+        if str(ipv) in ['4', '6']:
+            remote_host_ip = self.sh(f'host -{ipv} {hostname}')
+        else:
+            raise Exception(f'Unknown IP version {ipv}, expected 4 or 6')
+        if 'has address' in remote_host_ip:
+            (host, ip) = remote_host_ip.strip().split(' has address ')
```
zmc, line 481: If the name resolves to multiple addresses, we will get:
```
ValueError: too many values to unpack (expected 2)
```

kshtsk:
> I'm curious why we wouldn't prefer to use native python APIs, e.g.:
> 
> ```
> >>> pprint(socket.getaddrinfo('google.com', None))
> [(<AddressFamily.AF_INET6: 30>,
>   <SocketKind.SOCK_DGRAM: 2>,
>   17,
>   '',
>   ('2607:f8b0:400f:802::200e', 0, 0, 0)),
>  (<AddressFamily.AF_INET6: 30>,
>   <SocketKind.SOCK_STREAM: 1>,
>   6,
>   '',
>   ('2607:f8b0:400f:802::200e', 0, 0, 0)),
>  (<AddressFamily.AF_INET: 2>,
>   <SocketKind.SOCK_DGRAM: 2>,
>   17,
>   '',
>   ('142.250.69.238', 0)),
>  (<AddressFamily.AF_INET: 2>,
>   <SocketKind.SOCK_STREAM: 1>,
>   6,
>   '',
>   ('142.250.69.238', 0))]
> ```
> 
> It handles multiple results gracefully, and the same call can return both v4 and v6 addresses with an easy way to filter for one or the other if necessary.

because the idea is to get addresses remotely and not all targets can be resolved from teuthology host, for example, in case of using bastion hosts

teuthology/orchestra/remote.py
kshtsk: hm, didn't test it against google.com, it has indeed strange response:
```
$ host -4 google.com
google.com has address 142.250.185.110
google.com has IPv6 address 2a00:1450:4001:80f::200e
google.com mail is handled by 10 smtp.google.com.
```
I only verified this against target nodes. I guess in this case it is only needed to find first occurrence of "has address" line if any.

kshtsk:
> I'm curious why we wouldn't prefer to use native python APIs, e.g.:
> 
> ```
> >>> pprint(socket.getaddrinfo('google.com', None))
> [(<AddressFamily.AF_INET6: 30>,
>   <SocketKind.SOCK_DGRAM: 2>,
>   17,
>   '',
>   ('2607:f8b0:400f:802::200e', 0, 0, 0)),
>  (<AddressFamily.AF_INET6: 30>,
>   <SocketKind.SOCK_STREAM: 1>,
>   6,
>   '',
>   ('2607:f8b0:400f:802::200e', 0, 0, 0)),
>  (<AddressFamily.AF_INET: 2>,
>   <SocketKind.SOCK_DGRAM: 2>,
>   17,
>   '',
>   ('142.250.69.238', 0)),
>  (<AddressFamily.AF_INET: 2>,
>   <SocketKind.SOCK_STREAM: 1>,
>   6,
>   '',
>   ('142.250.69.238', 0))]
> ```
> 
> It handles multiple results gracefully, and the same call can return both v4 and v6 addresses with an easy way to filter for one or the other if necessary.

I just tried this for google.com in sepia lab:
```
kyr@teuthology:~$ host -4 google.com
google.com has address 142.251.111.138
google.com has address 142.251.111.102
google.com has address 142.251.111.113
google.com has address 142.251.111.101
google.com has address 142.251.111.100
google.com has address 142.251.111.139
google.com has IPv6 address 2607:f8b0:4004:c19::64
google.com has IPv6 address 2607:f8b0:4004:c19::8b
google.com has IPv6 address 2607:f8b0:4004:c19::8a
google.com has IPv6 address 2607:f8b0:4004:c19::66
google.com mail is handled by 10 smtp.google.com.
```
That's interesting, a bunch of addresses of each type, however the introduced function is supposed to be used for target nodes only, which supposed to have only one address correspondingly, does it makes sense to take first found or to return all of them as a list.

zmc (approved):
Thanks for updating to handle the multiple-value-in-response case, and for adding a unit test. Looks good to me!

PR #2068, opened 2025-07-21
tasks/internal/syslog: gather all journald entries

batrick:
Main change here is to gather all of the systemd journal (that's available) rather than just the current boot incarnation.


zmc (approved):
Thanks @batrick!

PR #2071, opened 2025-07-25
teuthology: use test user name from the config

phlogistonjohn:
A few places in the code try to use a configurable name but others effectively hard-code "ubuntu". Fix these cases to support more flexible user naming.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -51,7 +51,13 @@ def host_shortname(hostname):
     else:
         return hostname.split('.', 1)[0]
 
-def canonicalize_hostname(hostname, user: Optional[str] ='ubuntu'):
+# sentinel value to indicate the default user value is to be used.
+# None and the empty string are reserved for specifying no user value
+# to be backwards compatible with older code.
+_default_user = '.'
+
+def canonicalize_hostname(hostname, user: Optional[str] = _default_user):
+    user = get_test_user() if user == _default_user else user
     hostname_expr = hostname_expr_templ.format(
         lab_domain=config.lab_domain.replace('.', r'\.'))
     match = re.match(hostname_expr, hostname)
--- a/teuthology/task/ssh_keys.py
+++ b/teuthology/task/ssh_keys.py
@@ -93,21 +93,22 @@ def tweak_ssh_config(ctx, config):
     """
     Turn off StrictHostKeyChecking
     """
+    user = misc.get_test_user()
     run.wait(
         ctx.cluster.run(
             args=[
                 'echo',
                 'StrictHostKeyChecking no\n',
                 run.Raw('>'),
-                run.Raw('/home/ubuntu/.ssh/config'),
+                run.Raw(f'/home/{user}/.ssh/config'),
                 run.Raw('&&'),
                 'echo',
                 'UserKnownHostsFile ',
                 run.Raw('/dev/null'),
                 run.Raw('>>'),
-                run.Raw('/home/ubuntu/.ssh/config'),
+                run.Raw(f'/home/{user}/.ssh/config'),
                 run.Raw('&&'),
-                run.Raw('chmod 600 /home/ubuntu/.ssh/config'),
+                run.Raw(f'chmod 600 /home/{user}/.ssh/config'),
             ],
             wait=False,
         )
@@ -119,7 +120,7 @@ def tweak_ssh_config(ctx, config):
     finally:
         run.wait(
             ctx.cluster.run(
-                args=['rm',run.Raw('/home/ubuntu/.ssh/config')],
+                args=['rm',run.Raw(f'/home/{user}/.ssh/config')],
             wait=False
             ),
         )


zmc:
This is a good change; let's just get this tested with .e.g. a single scheduled job in sepia

phlogistonjohn:
https://pulpito.ceph.com/phlogistonjohn-2025-07-31_13:58:12-orch-wip-phlogistonjohn-testing-1-2025-07-22-0801-distro-default-gibba/
I performed this run having applied my patch from this PR on top of the teutholgoy main branch - no issues related to my change that I can see, just some known test flakes in the smb orch suite.

teuthology/misc.py
```diff
@@ -51,7 +51,13 @@ def host_shortname(hostname):
     else:
         return hostname.split('.', 1)[0]
 
-def canonicalize_hostname(hostname, user: Optional[str] ='ubuntu'):
+# sentinel value to indicate the default user value is to be used.
+# None and the empty string are reserved for specifying no user value
+# to be backwards compatible with older code.
+_default_user = '.'
```
kshtsk, line 57: Not sure I understand this, why 'None' does not work, in the context of the change it is the same.
Introducing the _default_user with value '.', instead of, for example, 'ubuntu', overcomplicates the code.

teuthology/misc.py
```diff
@@ -51,7 +51,13 @@ def host_shortname(hostname):
...
+
+def canonicalize_hostname(hostname, user: Optional[str] = _default_user):
+    user = get_test_user() if user == _default_user else user
```
kshtsk, line 60: The `get_test_user` already has default value 'ubuntu' for 'test_user' config parameter.

teuthology/misc.py
phlogistonjohn: I guess the comment doesn't explain the situation enough. The issue at hand is twofold - I do not want to use the username `ubuntu` on my test systems, I find it confusing as I mainly test on non-ubuntu systems. I have my teuthology configuration use something like `cephtest`. However, a few places in the code did not honor that existing configuration parameter.

The function `canonicalize_hostname` does not treat `None` as meaning "use the default user from the config" it is something more like "don't include a username component in the resulting hostname string".
I did not want to change this behavior because I didn't want to  have to find all the existing callers and potentially change them.

I did consider making a sentinel object - like `_deafult_user = object()` but that would require a change to the type annotation and I thought that'd be ugly. I thought '.' seems like an obvious and safe sentinel value as it's not a real username and is akin to "here" in unix file systems and some other tools.

kshtsk: And instead of confusing `_default_user` it would be better to user self descriptive naming. Since `_canonicalize_hostname_default_user_sentinel` is too long, maybe just use `_default_user_sentinel` or `_user_sentinel`.

teuthology/misc.py
```diff
@@ -51,7 +51,13 @@ def host_shortname(hostname):
     else:
         return hostname.split('.', 1)[0]
 
-def canonicalize_hostname(hostname, user: Optional[str] ='ubuntu'):
+# sentinel value to indicate the default user value is to be used.
+# None and the empty string are reserved for specifying no user value
+# to be backwards compatible with older code.
+_default_user = '.'
+
+def canonicalize_hostname(hostname, user: Optional[str] = _default_user):
```
kshtsk, line 59: Because the signature is touched, may we have consistent typing style for all parameters and function return value:
```
def canonicalize_hostname(hostname: str, user: Optional[str] = _sentinel_user) -> str:
```

phlogistonjohn: Fair enough, done.

kshtsk:
I guess the new behavior can be covered by a unit test, like adding a case to TestHostnames by https://github.com/ceph/teuthology/blob/main/teuthology/test/test_misc.py#L252-L313

phlogistonjohn:
@zmc as per our call yesterday: I have apparently been doing tests with this patch applied for weeks. I'm doing one more run today with it today after having rebased on teuthology main.

phlogistonjohn:
Also, I'd like to keep the scope of this change fairly small - I only really make use of it in my local teuthology setup. It's one where I don't generally test ubuntu based systems at all and feels weird to create a ubuntu user. I don't really expect to change the sepia lab values and start confusing everyone who is already used to this "quirk". Ultimately, I am just trying to carry fewer custom patches for that setup by up-streaming this. Thanks!

phlogistonjohn:
I just noticed that I'm still carrying this custom patch in my working branch of teuthology. I'd really like to get this PR done and over with but I don't have the time to do much in this space. This fixed my issue and it works for me and it's reasonably minimal.

If I can't submit a minimal change and (let others) iterate and improve things afterward I'll probably have to close the PR (and corresponding issue) as I really have my time occupied by lots of other things. In other words, unless there's something _wrong_ with the change it's a bit of a 'take it or leave it' situation for me. Thanks for understanding.

phlogistonjohn:
I just rebased and it applied cleanly and the tests passed FWIW.

PR #2082, opened 2025-09-09
workflows/dev_container: Use native arm64 runners

zmc:


kshtsk:
It is ridiculously fast now, but the unit tests are hung because of the gevent issue.

PR #2084, opened 2025-09-09
support install repos for rpm

kshtsk:


zmc (approved):

kshtsk:
Just a run with 4 jobs to make sure it does not introduce lame behavior when the feature is not used:
https://pulpito.ceph.com/kyr-2025-10-16_06:42:35-smoke-main-distro-default-smithi/

PR #2086, opened 2025-09-09
openstack: Do not immediately throw ImportError

zmc:
We would rather not have openstacksdk be a requirement for all installs, since
most do not use it. This change will not disturb people who install the
openstack extra, which pulls in that and other packages.

Signed-off-by: Zack Cerza <zack@cerza.org>

### DIFF
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -31,6 +31,7 @@ import os
 import paramiko
 import re
 import socket
+import sys
 import subprocess
 import tempfile
 import teuthology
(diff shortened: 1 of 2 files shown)


kshtsk (approved):
Looks interesting

teuthology/openstack/__init__.py
```diff
@@ -72,14 +76,19 @@ def enforce_json_dictionary(something):
             ' you are encouraged to add a comment if you want it to be'
             ' fixed.')
 
+def create_connection():
+    if openstack_connection is None:
+        raise RuntimeError("Did not find required openstack dependencies!")
```
kshtsk, line 81: Maybe it's better to give a user more ideas which dependencies are needed, like tell about `openstacksdk` or give a hint that teuthology needs to be installed with `[openstack]` option. And no need to scary with bang symbol :-)

zmc: Good idea - updated

kshtsk (approved):

PR #2088, opened 2025-09-15
Fix: handle attached_volumes field when fetching volumes from openstack

deepssin:
In some OpenStack deployments, the server JSON includes 'attached_volumes'
instead of `os-extended-volumes:volumes_attached`. The existing 
get_volumes function only checked the latter, which caused it to always 
return an empty list on these clouds.

This patch updates get_volumes() to:

- Safely access self.info without breaking
- Fallback to 'attached_volumes' if 'os-extended-volumes:volumes_attached'
  is not present
- Return the list of volume UUIDs consistently

With this change, teardown logic (volume cleanup on destroy) works 
correctly across both types of openstack responses.

### DIFF
--- a/teuthology/openstack/__init__.py
+++ b/teuthology/openstack/__init__.py
@@ -128,8 +128,12 @@ class OpenStackInstance(object):
         Return the uuid of the volumes attached to the name_or_id
         OpenStack instance.
         """
-        volumes = self['os-extended-volumes:volumes_attached']
-        return [volume['id'] for volume in volumes ]
+        info = self.info or {}
+        vols = (info.get('os-extended-volumes:volumes_attached')
+                or info.get('attached_volumes') 
+                or [])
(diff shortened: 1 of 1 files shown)


teuthology/openstack/__init__.py
```diff
@@ -113,8 +113,13 @@ def get_volumes(self):
         Return the uuid of the volumes attached to the name_or_id
         OpenStack instance.
         """
-        volumes = self['os-extended-volumes:volumes_attached']
-        return [volume['id'] for volume in volumes ]
+        info = self.info or {}
+        #Some clouds use attached_volumes instead of os-extended-volumes:volumes_attached
```
kshtsk, line 117: Drop this comment, the code below is pretty straightforward. It's better give this explanation  in the commit description.

kshtsk (approved):

zmc (approved):

PR #2095, opened 2025-10-17
Adding check for shaman and chacra build validations

tintumathew10:
This PR is for checking the selected sha1 from github is existing in shaman and chacra.

### DIFF
--- a/teuthology/suite/run.py
+++ b/teuthology/suite/run.py
@@ -5,6 +5,7 @@ import os
 import pwd
 import yaml
 import re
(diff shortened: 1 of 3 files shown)


zmc (changes requested):
I added a couple feedback items but before we get too deep into this, could you explain a little about the reasoning behind adding this instead of modifying or replacing this bit: https://github.com/ceph/teuthology/blob/main/teuthology/suite/util.py#L226-L256
Thanks!

PR #2101, opened 2025-10-31
[dnm] dispatcher: do not put targets in orig config

kshtsk:
There is introduced targets in orig.config.yaml in the dispatcher code, however it should not be here so the original config should be still usable with teuthology (teuthology-run) with arbitrary targets yaml file, which is provided separately.


kshtsk:
Closing in favor of #2196

PR #2137, opened 2026-01-21
orchestra: rocky 9.6 and 10.0 are gone

kshtsk:
Welcome Rocky and Alma Linux 9.7 and 10.1


djgalloway (approved):
Just tried to lock a rocky 10.0 machine and got
```
RuntimeError: Expected trial064's OS to be rocky 10.0 but found rocky 10.1
```
😄

PR #2157, opened 2026-03-17
teuthology: remove ceph-deploy

batrick:


batrick:
@zmc is this test failure my fault?

https://github.com/ceph/teuthology/actions/runs/26756389882/job/78857332160?pr=2157

looks unrelated

zmc:
hadn't seen this before but I like it!

zmc (approved):

PR #2081, opened 2025-09-09
Begin transitioning to uv

zmc:

PR #2161, opened 2026-03-25
Move to uv

kshtsk:


kshtsk:
That's just a rebase branch of the one in #2081

kshtsk:
@zmc have you had a chance to look at this PR?

zmc:
> @zmc have you had a chance to look at this PR?

Not in as much detail yet as I'd like; I did schedule a test run: https://pulpito-ng.ceph.com/runs/zack-2026-03-27_23:19:02-smoke-main-distro-default-trial

bootstrap
```diff
@@ -1,109 +1,74 @@
...
-    has_pkg="rpm -q --whatprovides"
-    install_pkg="sudo dnf install -y"
+    . /etc/os-release
+    OS="${NAME// /}"
     case "$OS" in
     Ubuntu|Debian|LinuxMint)
-        deps=(qemu-utils python3-dev libssl-dev python3-pip python3-wheel python3-venv libev-dev libvirt-dev libffi-dev libyaml-dev build-essential jq curl)
-        has_pkg="dpkg -s"
+        deps=(qemu-utils python3-dev libssl-dev libev-dev libvirt-dev libffi-dev libyaml-dev pipx build-essential jq curl)
+        has_pkg="dpkg -C"
```
deepssin: dpkg -C is ->  
-C|--audit [<package>...]        Check for broken package(s)
in practice dpkg -C <pkg> still exits 0 even when the package is missing (it prints “not installed” to stderr). I verified this in the environment: both an installed and a bogus package name lead the bootstrap loop to treat the dependency as present.

bootstrap
```diff
@@ -1,109 +1,74 @@
...
+        install_pkg="sudo dnf install -y"
+        deps=(libev-devel libffi-devel libvirt-devel "$PYTHON-devel" pipx)
         ;;
     "openSUSE project"|"SUSE LINUX"|"openSUSE"|"openSUSELeap"|"openSUSETumbleweed")
-        PYTHON=python3.12
-        deps=(python312-pip python312-devel python312 libev-devel libvirt-devel libffi-devel)
+        PYTHON=${PYTHON:-"python3"}
+        deps=("$PYTHON" "$PYTHON-devel" "$PYTHON-pipx" libev-devel libffi-devel libvirt-devel)
+        has_pkg="rpm -q --whatprovides"
         install_pkg="sudo zypper install"
```
deepssin: This being non-interactive , I think bootstrap can get stuck here.

deepssin:
Tested this with dispatcher running old code and the repo having PR changes -- trigger went just fine.

bootstrap
kshtsk: Not sure I understand what is the problem here. What is your suggestion change. By design bootstrap should stop immediately when it finds some of the system dependencies are missing.
In this case it just reports users so they can install missing dependencies manually and try again. The "sudo zipper install" is interactive and needs "-y" to auto accept all the packages. Since it is a manual step, guess it is a user choice how to proceed.
Could you please elaborate. 
Btw, this works for me on the latest tumbleweed.

deepssin: There are two paths: plain ./bootstrap already exits when deps are missing—that’s fine. The wrinkle is ./bootstrap install, which runs the package manager for you. On SUSE that’s zypper install without -y, so in non-interactive contexts (like cloud-init, CI or any other automations) it can block on prompts, unlike apt/dnf which already pass -y. Suggest adding non-interactive only for that install path so scripted runs don’t hang; interactive users can still install manually after a plain ./bootstrap.

bootstrap
kshtsk: @zmc what did you have in mind when changing to `-s` to `-C`?

bootstrap
kshtsk: Oh, thanks, I'll add `-y` then.

bootstrap
zmc: I think I just wanted a quieter `-s`, but I will amend this to use `-s` and drop stdout/err. Good catch!!

zmc: or, @kshtsk, if you'd rather just push a small fix to this branch that works as well.

zmc:
When we merge this branch, we should wait for the queue to run through jobs scheduled before the merge before updating the dispatcher

zmc (approved):
I think this has waited more than long enough! Thank you @kshtsk for helping to revive this effort.

PR #2163, opened 2026-03-26
feature: add `PulpProject` class for Pulp-based package repository support

vamahaja:
This PR adds a `PulpProject` class that extends `GitbuilderProject` so teuthology can use Pulp-hosted package repositories. It introduces new config `use_artifacts`, set it to `pulp` (globally in site config or per job) to use Pulp; other values keep the current `Gitbuilder` and `Shaman` behavior.

**Configuration**

- `use_artifacts: 'pulp'` - selects the Pulp backend (alongside existing 'shaman' and the default Gitbuilder path when unset / other values).
- `pulp_host` — Pulp server hostname (documented in siteconfig.rst).

**Key changes**

 - Implement `PulpProject` inheriting from `GitbuilderProject`.
- Override `_get_base_url()` to use Pulp’s repository API.
- Override `_get_distro()` for Pulp’s distro format.
- Override `_get_package_version()` and `_get_package_sha1()` to read metadata from Pulp responses.
- Update `get_builder_project()` so `config.use_artifacts == 'pulp'` returns `PulpProject`.
- Update `siteconfig` (and related docs) for `use_artifacts: 'pulp'` and `pulp_host`.
- Add mock test `TestPulpProject` to `teuthology/test/test_packaging.py`.


kshtsk:
This looks good, do we have an instance running with pulp to check this PR?

vamahaja:
> This looks good, do we have an instance running with pulp to check this PR?

@kshtsk for review, I am working on deploying Pulp server on OCP cluster.

teuthology/openstack/setup-openstack.sh
```diff
@@ -54,7 +54,7 @@ function create_config() {
 
     cat > ~/.teuthology.yaml <<EOF
 $archive_upload
-use_shaman: false
+use_artifacts: 'shaman'
```
kshtsk, line 57: The option `use_shaman: false` means that no _shaman_ should be used. So if we switch to `_use_artifacts`, there should be `None` case to consider.

teuthology/config.py
```diff
@@ -181,8 +181,8 @@ class TeuthologyConfig(YamlConfig):
         'kojiroot_url': 'https://kojipkgs.fedoraproject.org/packages',
         'koji_task_url': 'https://kojipkgs.fedoraproject.org/work/',
         'baseurl_template': 'http://{host}/{proj}-{pkg_type}-{dist}-{arch}-{flavor}/{uri}',
-        'use_shaman': True,
-        'shaman_host': 'shaman.ceph.com',
+        'use_artifacts': 'shaman',
```
kshtsk, line 184: Maybe better to use `artifacts_type` instead?

teuthology/config.py
```diff
@@ -181,8 +181,8 @@ class TeuthologyConfig(YamlConfig):
...
+        'artifacts_host': 'shaman.ceph.com',
```
kshtsk, line 185: I suggest to start using endpoint URI instead of just hostname, so we have some kind of flexibility and avoid hardcoded paths, and for fast switch between artifacts type, maybe it is better to have api configuration separate:
```
shaman:
  endpoint: https://shaman.ceph.com/api
pulp:
  endpoint: http://pulp.example.com/pulp/api/v3
```
What do you think? And you @zmc?

teuthology/config.py
kshtsk: Or even to match teuthology module name use: `packaging_type`?

teuthology/config.py
vamahaja: @zmc as discussed in weekly teuthology call, we can continue with below config -
```
artifacts:
  type: shaman | pulp
  endpoint: "https://<shaman|pulp>.ceph.com"
  username: <shaman|pulp>-username
  password: <shaman|pulp>-password
```
@kshtsk ^^^

kshtsk: I would prefer to stay with separate service tool based config when it comes to authentication, like it is done for fog and maas:
```
packaging_type:  shaman | pulp | None
shaman:
  endpoint: https://shaman.ceph.com/api
  username: 
  password:
pulp:
  endpoint: http://pulp.example.com/pulp/api/v3
  username:
  password: 
```
Why it is better?
Easy switch between each other. Probable benefit from selecting artifact type using overrides.

teuthology/packaging.py
```diff
@@ -1054,13 +1054,113 @@ def _remove_rpm_repo(self):
         )
 
 
+class PulpProject(GitbuilderProject):
+    def __init__(self, project, job_config, ctx=None, remote=None):
+        super(PulpProject, self).__init__(project, job_config, ctx, remote)
+
+        # Set the url for the pulp server.
+        self.pulp_server_url = f'http://{config.artifacts_host}'
```
batrick, line 1062:
```suggestion
        self.pulp_server_url = f'https://{config.artifacts_host}'
```

?

batrick: If you need `http://` for testing, use a config but default to `https://`.

teuthology/run.py
```diff
@@ -390,10 +390,10 @@ def main(args):
...
     config["suite_path"] = fetch_tasks_if_needed(config)
 
-    # If the job has a 'use_shaman' key, use that value to override the global
-    # config's value.
-    if config.get('use_shaman') is not None:
-        teuth_config.use_shaman = config['use_shaman']
+    # If the job has a 'use_artifacts' key, use that value to override the global
+    # config's 'use_artifacts' value.
+    if config.get('use_artifacts') is not None:
+        teuth_config.use_artifacts = config['use_artifacts']
```
batrick, line 396:
```suggestion
        teuth_config.use_artifacts = config['use_artifacts']
        elif config.get('use_shaman') is not None:
             # TODO: Add a logging warning about deprecation here
             teuth_config.use_artifacts = 'shaman' if config['use_shaman'] else None
```

batrick (changes requested):
Looking good!!

teuthology/config.py
zmc: I like it - perhaps `package_source` instead of `package_type`

docs/detailed_test_config.rst
```diff
@@ -300,10 +300,24 @@ Shaman options
...
+
+Pulp is a package manager (https://pulpproject.org/) that could be used to
+build the uri for specified packages based the 'artifacts_host' option.
+
+Options::
+
+    use_artifacts: 'pulp'
+    artifacts_host: 'pulp.example.com'
+    pulp:
+      force_noarch: True # Force to use "noarch" to build the uri
```
kshtsk: As discussed in our meeting the notation is:
```
package_source: pulp
pulp:
  endpoint: http://pulp.example.com/pulp/api/v3
  force_noarch: True
```

docs/siteconfig.rst
```diff
@@ -269,3 +269,12 @@ Here is a sample configuration with many of the options set and documented::
...
     job_threshold: 500
+
+    # Setting for artifacts manager (shaman or pulp)
+    use_artifacts: 'pulp'
+    artifacts_host: 'pulp.example.com'
+
+    # Settings for pulp package manager (https://pulpproject.org/)
+    pulp:
+      username: '<username>'
+      password: '<password>'
```
kshtsk: As discussed in our meeting the notation is:
```
package_source: pulp
pulp:
  endpoint: http://pulp.example.com/pulp/api/v3
  username: '<username>'
   password: '<password>'
  force_noarch: True
```
And I also vote for adding info about new shaman configuration way.

teuthology/openstack/setup-openstack.sh
kshtsk: As discussed this to be changed to: ```package_source: 'shaman'```

teuthology/task/kernel.py
```diff
@@ -374,7 +374,7 @@ def download_kernel(ctx, config):
                 ctx=ctx,
                 remote=role_remote,
             )
-            if teuth_config.use_shaman:
+            if teuth_config.use_artifacts == 'shaman':
```
kshtsk: Change to `teuth_config.package_source == 'shaman'`

teuthology/task/kernel.py
```diff
@@ -1442,7 +1442,7 @@ def process_role(ctx, config, timeout, role, role_config):
         ctx.summary['{role}-kernel-sha1'.format(role=role)] = sha1
 
         if need_to_install(ctx, role, sha1):
-            if teuth_config.use_shaman:
+            if teuth_config.use_artifacts == 'shaman':
```
kshtsk: Same as above

teuthology/packaging.py
```diff
@@ -847,7 +874,7 @@ def _remove_deb_repo(self):
 class ShamanProject(GitbuilderProject):
     def __init__(self, project, job_config, ctx=None, remote=None):
         super(ShamanProject, self).__init__(project, job_config, ctx, remote)
-        self.query_url = 'https://%s/api/' % config.shaman_host
+        self.query_url = 'https://%s/api/' % config.artifacts_host
```
kshtsk: This should be change to match new config approach:
```
shaman:
   endpoint: 'https://shaman.ceph.com/api/'
```
change to:
```
  self.query_url = config.get('shaman', {}).get('endpoint')
```

docs/detailed_test_config.rst
```diff
@@ -300,10 +300,24 @@ Shaman options
...
-packages based the 'shaman_host': 'shaman.ceph.com'.
+packages based the 'artifacts_host' option.
 
 Options::
 
-    use_shaman: True # Enable to use Shaman, False as default
+    use_artifacts: 'shaman'
+    artifacts_host: 'shaman.ceph.com'
     shaman:
       force_noarch: True # Force to use "noarch" to build the uri
```
kshtsk: According to our discussion the new approach is:
```
package_source: 'shaman'
shaman:
  endpoint: 'https://shaman.ceph.com/api/'
```

teuthology/openstack/setup-openstack.sh
vamahaja: done

teuthology/config.py
vamahaja: Updated config using `package_source`

teuthology/config.py
vamahaja: Using `package_source`

teuthology/packaging.py
vamahaja: Changed approach to use endpoints instead of hosts

docs/detailed_test_config.rst
vamahaja: done

docs/siteconfig.rst
vamahaja: Updated based on suggested changes

teuthology/packaging.py
vamahaja: done

docs/detailed_test_config.rst
vamahaja: done

teuthology/task/kernel.py
vamahaja: done

teuthology/task/kernel.py
vamahaja: done

teuthology/run.py
vamahaja: In config, default value is still shaman - https://github.com/vamahaja/teuthology/blob/39c888c05e16dde3f19a4804b731c7575d89da9e/teuthology/config.py#L185

teuthology/packaging.py
```diff
@@ -1054,13 +1098,191 @@ def _remove_rpm_repo(self):
...
+        ref_name, ref_val = list(self._choose_reference().items())[0]
+        labels += f'{ref_name}={ref_val}'
+
+        log.info(f'Search URI: {self._search_uri}')
+        log.info(f'Searching for packages with labels: {labels}')
+        resp = requests.get(
+            self._search_uri,
+            params={'pulp_label_select': labels},
+            auth=(self.pulp_username, self.pulp_password),
+            verify=False
```
batrick: why False?

teuthology/task/kernel.py
```diff
@@ -1442,7 +1442,7 @@ def process_role(ctx, config, timeout, role, role_config):
         ctx.summary['{role}-kernel-sha1'.format(role=role)] = sha1
 
         if need_to_install(ctx, role, sha1):
-            if teuth_config.use_shaman:
+            if teuth_config.package_source == 'shaman':
```
batrick:
```suggestion
if teuth_config.package_source in ('shaman', 'pulp'):
```

I believe?

docs/detailed_test_config.rst
```diff
@@ -296,14 +296,23 @@ specified in ``$HOME/.teuthology.yaml``::
...
+  force_noarch: True
 
-Options::
+Options for pulp:
 
-    use_shaman: True # Enable to use Shaman, False as default
-    shaman:
-      force_noarch: True # Force to use "noarch" to build the uri
+pulp:
+  endpoint: "https://pulp.example.com/pulp/api/v3/"
```
batrick: let's use the actual endpoint for the lab's instance

docs/detailed_test_config.rst
```diff
@@ -296,14 +296,23 @@ specified in ``$HOME/.teuthology.yaml``::
...
+  username: pulp-username
+  password: pulp-password
```
batrick, line 317: do we need a username/password to access the pulp read-only in the lab?

docs/detailed_test_config.rst
```diff
@@ -296,14 +296,23 @@ specified in ``$HOME/.teuthology.yaml``::
...
+  force_noarch: True
```
batrick, line 318: Please add information on how a user can tell if pulp has the packages for their branch.

batrick (changes requested):
I recommend using an LLM to help you write up some docs on this new change.

docs/detailed_test_config.rst
vamahaja: Yes, in the current setup, the ceph-ci-bot user is created with read-only permissions for API access.

teuthology/packaging.py
vamahaja: this was added for testing server, removed now.

teuthology/task/kernel.py
vamahaja: Thanks !

Was not sure if same URL will work for both but as we are following same URL format, this should work.

docs/detailed_test_config.rst
vamahaja: Updated based on recent changes

vamahaja:
> I recommend using an LLM to help you write up some docs on this new change.

The draft documentation is currently available on CephPad: https://pad.ceph.com/p/ceph_artifact_management_via_pulp.

I will migrate this to a wiki page once all associated tasks are complete.

docs/detailed_test_config.rst
vamahaja: I will update the [CephPad document](https://pad.ceph.com/p/ceph_artifact_management_via_pulp) with details on how to fetch the build.

vamahaja:
Smoke tests are passed - http://172.21.5.153:9091/vamahaja-2026-06-05_08:39:01-smoke-tentacle-distro-default-trial

Screenshot From 2026-06-05 19-20-56

djgalloway:
This seems like a breaking change...  While the number of global teuthology users is low, I am personally not a fan of removing options like `use_shaman` that have been available for years.

I'd prefer to see something like

```
use_shaman: false
use_pulp: true
```

But this is just my opinion.  I'd like @zmc to weigh in.

Can the option be set on a per-job/run basis?

kshtsk:
> This seems like a breaking change... While the number of global teuthology users is low, I am personally not a fan of removing options like `use_shaman` that have been available for years.
>

I don't think it is that breaking, @djgalloway to avoid interruption, could you or someone add the config to the server's teuthology yaml already:
```
package_source: shaman
shaman:
      endpoint: "https://shaman.ceph.com/api/"
      force_noarch: True
```
Or at least we need this before the merge. 
 
> I'd prefer to see something like
> 
> ```
> use_shaman: false
> use_pulp: true
> ```
> 
This complicates the code logic and will make potential misuse, what to prefer when both "use_shaman" and "use_pulp" are enable or disabled, it is not possible to predict which is used not looking into the source code. 

> But this is just my opinion. I'd like @zmc to weigh in.
> 
> Can the option be set on a per-job/run basis?

@djgalloway I like the idea of giving package_source on run/job basis though.

vamahaja:
> > This seems like a breaking change... While the number of global teuthology users is low, I am personally not a fan of removing options like `use_shaman` that have been available for years.
> 
> I don't think it is that breaking, @djgalloway to avoid interruption, could you or someone add the config to the server's teuthology yaml already:
> 
> ```
> package_source: shaman
> shaman:
>       endpoint: "https://shaman.ceph.com/api/"
>       force_noarch: True
> ```
> 
> Or at least we need this before the merge.
> 
> > I'd prefer to see something like
> > ```
> > use_shaman: false
> > use_pulp: true
> > ```
> 
> This complicates the code logic and will make potential misuse, what to prefer when both "use_shaman" and "use_pulp" are enable or disabled, it is not possible to predict which is used not looking into the source code.
> 
> > But this is just my opinion. I'd like @zmc to weigh in.
> > Can the option be set on a per-job/run basis?
> 
> @djgalloway I like the idea of giving package_source on run/job basis though.

@djgalloway @kshtsk no changes are needed, default values are already pointing to shaman with the required configs - https://github.com/ceph/teuthology/pull/2163/changes#diff-4525b7398914ab0a466329cd35d0e8b22b35479597fbfdca524baa871ae7d184R184

zmc (commented):
I think this looks great - I see only one outstanding item from @batrick, but other than that the PR seems ready to me!

docs/detailed_test_config.rst
vamahaja: Added new section - `Accessing Pulp Repositories using pulp-cli with Labels`

zmc (approved):
Looks like all review items have been addressed, so from my perspective it looks ready - @batrick would you take a look as well, please?

batrick:
> Yes, in the current setup, the ceph-ci-bot user is created with read-only permissions for API access.

For this to work long-term I don't know how we cannot have pulp be publicly (on the VPN at least) readable?

batrick (approved):
I'm okay with this as-is to start.

kshtsk:
How's pushing the big green button?

PR #2167, opened 2026-04-21
Add tracker suggestions to teuthology log

deepssin:
Log possible Redmine matches on job failure using normalized failure text, optional nested exception detail, and hints from teuthology.log

### DIFF
--- a/teuthology/run.py
+++ b/teuthology/run.py
@@ -272,6 +272,21 @@ def report_outcome(config, archive, summary):
     summary_dump = yaml.safe_dump(summary)
     log.info('Summary data:\n%s' % summary_dump)
 
+    if not passed:
+        try:
+            from teuthology.util.redmine_suggest import log_redmine_suggestions_on_failure
+            log_redmine_suggestions_on_failure(
+                log,
+                summary.get('failure_reason'),
+                config,
+                failure_reason_detail=summary.get('failure_reason_detail'),
+                teuthology_log_path=(
+                    os.path.join(archive, 'teuthology.log') if archive else None
+                ),
+            )
+        except Exception as e:
+            log.debug('Redmine tracker suggestions skipped: %s', e)
+
     if ('email-on-error' in config
             and not passed):
         config_dump = yaml.safe_dump(config)
--- a/teuthology/run_tasks.py
+++ b/teuthology/run_tasks.py
@@ -21,6 +21,32 @@ from teuthology.util import sentry
 log = logging.getLogger(__name__)
 
 
+def _first_nested_exception_message(exc, max_len=300):
+    """
+    Return the first nested exception message from '__cause__'/'__context__'.
+    """
+    if exc is None:
+        return None
+    seen = set()
+    outer_msg = str(exc).strip()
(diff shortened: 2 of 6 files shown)


deepssin:
@zmc I've compared with scrape.py where it actually overlaps: status-124 timeout parsing now goes through TimeoutReason.get_timeout (same regex + path trim). The rest of scrape.py is archive-level (job backtraces, grep, clustering) and isn’t a good fit for live Redmine queries. host/sudo cleanup stays in redmine_suggest because scrape doesn’t really duplicate that.

zmc (commented):
I love this idea, but at a high level there are a few things I want to raise
- We may want a way to disable this conditionally after merge - perhaps at the global config level - unless we can be very confident that something like a redmine outage wouldn't cause crashes.
- `redmine_suggest.py` is hard to follow. I think an OO approach might be helpful here for readability reasons and also in case we, or someone else, wants to implement this for something other than Redmine. A simplistic example for reference could be `teuthology.task.ansible.FailureAnalyzer`. Along these lines I would rather see functions that generate and return messages that the caller could log, than a function that takes a logger and logs messages itself.
- I would like to be able to point this logic at a job manually. I think this would be easiest to implement by looking a an archive filesystem, as opposed to querying paddles, since it needs the job log. A CLI entry-point like `teuthology-bug-search` could serve as an interface.

kshtsk:
> I love this idea, but at a high level there are a few things I want to raise
> 
> * We may want a way to disable this conditionally after merge - perhaps at the global config level - unless we can be very confident that something like a redmine outage wouldn't cause crashes.
> * `redmine_suggest.py` is hard to follow. I think an OO approach might be helpful here for readability reasons and also in case we, or someone else, wants to implement this for something other than Redmine. A simplistic example for reference could be `teuthology.task.ansible.FailureAnalyzer`. Along these lines I would rather see functions that generate and return messages that the caller could log, than a function that takes a logger and logs messages itself.

I can only agree with this, also, maybe teuthology.task.ansible.FailureAnalyzer should go to a separate module as well.
At least the code should be moved out from the `teuthology.util`.
> * I would like to be able to point this logic at a job manually. I think this would be easiest to implement by looking a an archive filesystem, as opposed to querying paddles, since it needs the job log. A CLI entry-point like `teuthology-bug-search` could serve as an interface.

I also agree with this, for sure this "tool" should not be called from teuthology.run, the teuthology should be free from any other teuthology components like paddles, so it can be used independently to run tests.

teuthology/run.py
```diff
@@ -272,6 +272,21 @@ def report_outcome(config, archive, summary):
...
+                summary.get('failure_reason'),
+                config,
+                failure_reason_detail=summary.get('failure_reason_detail'),
+                teuthology_log_path=(
+                    os.path.join(archive, 'teuthology.log') if archive else None
+                ),
+            )
+        except Exception as e:
+            log.debug('Redmine tracker suggestions skipped: %s', e)
+
```
kshtsk, line 289: The way better place to call remdine parsing, out of teuthology.run, either to supervisor or to a separate cli.

PR #2170, opened 2026-04-22
workflows/dev_container: Don't push for PRs

zmc:


zmc:
This is looking good, but I should have filed this from a fork in order to fully test it. I'll re-open.

PR #2173, opened 2026-04-29
dispatcher: Use SIGTERM to stop

zmc:


deepssin (dismissed):
Looks good to me!

teuthology/dispatcher/__init__.py
```diff
@@ -32,32 +33,13 @@
...
-    args.insert(0, sys.executable)
-    os.execv(sys.executable, args)
-
-
-def stop():
+def stop(_sig, _frame) -> None:
     log.info('Stopping...')
     sys.exit(0)
 
+signal.signal(signal.SIGTERM, stop)
```
deepssin: Having `signal.signal(signal.SIGTERM, stop) `at module level means it runs whenever `teuthology.dispatcher `is imported (for e.g. by the exporter), silently overwriting any SIGTERM handler the importer may have set. Would it be worth moving this into main() so the handler only registers when the dispatcher actually starts?

zmc: you're absolutely right, thanks!

batrick (approved):

PR #2184, opened 2026-05-06
uv.lock: Update dependencies

zmc:


kshtsk:
I hope next uv.lock updates won't be so intrusive :-)

kshtsk (approved):

PR #2185, opened 2026-05-08
Fix dependencies CI

zmc:
We had a shell-quoting regression in the bootstrap script. I took this opportunity to clean up the output as well.


kshtsk (approved):
LGTM

PR #2196, opened 2026-05-12
Cleanup dispatcher code

kshtsk:


kshtsk:
@zmc could you please take a look.

teuthology/dispatcher/__init__.py
```diff
@@ -146,15 +134,28 @@ def main(args):
...
-                job_config, teuth_bin_path = prep_job(
-                    job_config,
-                    log_file_path,
-                    archive_dir,
-                )
+                job_config, teuth_bin_path = prep_job(job_config)
             except SkipJob:
                 continue
 
+            job_config_path = os.path.join(job_archive_path, 'init.config.yaml')
```
zmc: Why is the above `init.config.yaml` and not `orig.config.yaml` ?

kshtsk: @zmc Good question, I would say `init.config` is something that is generally include 'targets', `orig.config` is a combination of yaml files which teuthology.run (read teuthology) takes as arguments. Later there is `config.yaml` which also includes internal steps, and is saved only from internal task save_task and could include targets or not, depending on `--lock`. Since there is no documentation which yaml is which, maybe you can shed light on it.

kshtsk: So my idea:
"default" is what taken from queue + job_id
"init" what is provided by worker dispatcher/supervisor updating config, and includes always targets in opposite to what probably user will do, and to distinct what teuthology run updates.
"orig" - is whatever teuthology (and only) creates from arguments.

kshtsk: @zmc actually I just reverted it to 'original.config.yaml'

zmc (approved):

PR #2197, opened 2026-05-13
internal: fix core-dumps/debug-info collection

ronen-fr:
... especially for Crimson OSDs.

Specifically:
- use the value of execfn to identify the correct binary;
- correctly handle absolute paths when constructing the debug symbol path (stripping the leading separator before joining);
- find versioned debuginfo files instead of assuming an exact basename.

### DIFF
--- a/teuthology/task/internal/__init__.py
+++ b/teuthology/task/internal/__init__.py
@@ -299,6 +299,27 @@ def check_conflict(ctx, config):
         raise RuntimeError('Stale jobs detected, aborting.')
 
 
+def get_dump_program(file_info: str) -> str:
+    """
+    Parse the output of the file(1) command on a core dump to extract the
+    program that produced it. Prefers execfn (the actual file passed to
+    execve) over the 'from' field (argv[0], which can be overridden).
+
+    Raises ValueError if neither field can be parsed.
+    """
+    execfn_match = re.findall(r"execfn:\s*'([^']+)'", file_info)
+    if execfn_match:
+        program = execfn_match[0]
+        log.info(f' dump_program (from execfn): {program}')
+        return program
+    from_match = re.findall("from '([^ ']+)", file_info)
+    if from_match:
+        program = from_match[0].split()[0]
+        log.info(f' dump_program: {program}')
+        return program
+    raise ValueError(f"cannot parse program name from file output: {file_info}")
+
+
 def fetch_binaries_for_coredumps(path, remote):
     """
     Pul ELFs (debug and stripped) for each coredump found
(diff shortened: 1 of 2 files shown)


teuthology/task/internal/__init__.py
```diff
@@ -334,34 +334,67 @@ def fetch_binaries_for_coredumps(path, remote):
...
+                # the 'from' field (argv[0], which can be overridden,
+                # e.g. crimson-osd launched with argv[0]='ceph-osd').
+                execfn_match = re.findall(r"execfn:\s*'([^']+)'", dump_out)
+                if execfn_match:
+                    dump_program = execfn_match[0]
+                    log.info(f' dump_program (from execfn): {dump_program}')
+                else:
+                    dump_command = re.findall("from '([^ ']+)", dump_out)[0]
+                    dump_program = dump_command.split()[0]
+                    log.info(f' dump_program: {dump_program}')
```
kshtsk: This looks like wants to get unittested. Can we have this code isolated to a separate method, `get_dump_program(file_info: str) -> str`

teuthology/task/internal/__init__.py
```diff
@@ -334,34 +334,67 @@ def fetch_binaries_for_coredumps(path, remote):
...
+                    ['readlink', '-f', remote_path]).rstrip()
+                if resolved_path:
+                    remote_path = resolved_path
+            except Exception:
+                pass
+
             # Pull Debug symbols:
-            debug_path = os.path.join('/usr/lib/debug', remote_path)
+            debug_dir = os.path.dirname(os.path.join(
+                '/usr/lib/debug', remote_path.lstrip(os.path.sep)))
```
kshtsk, line 390: How binary path can be finalized with `/`?

teuthology/task/internal/__init__.py
ronen-fr: Sure. I'll update the code.

teuthology/task/internal/__init__.py
ronen-fr: Not sure I understand the question. If the path was reached using execfn - it can be absolute, i.e. start with a '/'. Is this what you are asking?

kshtsk (approved):

ronen-fr:
@kshtsk , @zmc - thanks for approving. What tests are required before merging for this repository?
(I have run my own tests - see for example https://pulpito-ng.ceph.com/runs?branch=wip-rf-asokassert-crimson-enabled, which uses this Teuthology branch). Thanks

teuthology/task/internal/__init__.py
kshtsk: I'm sorry I read rstrip instead lstrip.

kshtsk:
Hey @ronen-fr probably would work to see a job against your changes and artifacts with dumps and successful dumps with the cases covering:
1) dump program detected for crimson artifacts, and successfully grabbed the dumps
2) dump program detected for non-crimson artifacts, and successfully grabbed the dumps
3) the case where dump program is not possible to detect, is it possible?

PR #2201, opened 2026-05-15
task/install/deb: Add'l debugging info for held apt lock

djgalloway:
For https://tracker.ceph.com/issues/73843

### DIFF
--- a/teuthology/task/install/deb.py
+++ b/teuthology/task/install/deb.py
@@ -20,10 +20,26 @@ def _retry_if_eagain_in_output(remote, args):
                 return remote.run(args=args, stderr=stderr)
             except run.CommandFailedError:
                 if "could not get lock" in stderr.getvalue().lower():
-                    stdout = StringIO()
-                    args = ['sudo', 'fuser', '-v', '/var/lib/dpkg/lock-frontend']
-                    remote.run(args=args, stdout=stdout)
-                    log.info("The processes holding 'lock-frontend':\n{}".format(stdout.getvalue()))
+                    # Get the PID(s) holding the lock. Set to empty list in case
+                    # apt lock is freed between original apt failure and `fuser` call
+                    try:
+                        pids = remote.sh('sudo fuser /var/lib/dpkg/lock-frontend').split()
+                    except run.CommandFailedError:
+                        pids = []
+                    log.info("The processes holding 'lock-frontend':\n{}".format(pids))
+
+                    # Get detailed info on each PID
+                    for pid in pids:
+                        try:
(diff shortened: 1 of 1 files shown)


teuthology/task/install/deb.py
```diff
@@ -20,10 +20,39 @@ def _retry_if_eagain_in_output(remote, args):
...
-                    args = ['sudo', 'fuser', '-v', '/var/lib/dpkg/lock-frontend']
-                    remote.run(args=args, stdout=stdout)
-                    log.info("The processes holding 'lock-frontend':\n{}".format(stdout.getvalue()))
+                    # Get the PID(s) holding the lock
+                    fuser_out = StringIO()
+                    remote.run(
+                        args=['sudo', 'fuser', '-v', '/var/lib/dpkg/lock-frontend'],
+                        stdout=fuser_out,
+                    )
+                    pids = fuser_out.getvalue().split()
```
kshtsk: Instead of that may we use shortcut:
```
pids = remote.sh('sudo fuser -v /var/lib/dpkg/lock-frontend').split()
```
and avoid using of StringIO.

teuthology/task/install/deb.py
```diff
@@ -20,10 +20,28 @@ def _retry_if_eagain_in_output(remote, args):
...
+                        pids = remote.sh('sudo fuser /var/lib/dpkg/lock-frontend').split()
+                    except run.CommandFailedError:
+                        pids = []
+                    log.info("The processes holding 'lock-frontend':\n{}".format(pids))
+
+                    # Get detailed info on each PID
+                    for pid in pids:
+                        try:
+                            ps_out = remote.sh(
+                                'ps -p {} -o pid,ppid,user,stat,start,etime,cmd'.format(pid)
```
kshtsk: Thanks David for making this change, however, we don't need to use old style format, f-strings are preferable, if you like you could do the same: `remote.sh(f"ps -p {pid} -o pid,ppid,user,stat,start,etime,cmd")`

teuthology/task/install/deb.py
```diff
@@ -20,10 +20,28 @@ def _retry_if_eagain_in_output(remote, args):
...
+                            )
+                            log.info("Process info for PID %s:\n%s", pid, ps_out)
+
+                            cmdline = remote.sh('sudo cat /proc/{}/cmdline'.format(pid))
```
kshtsk: Same here, no need to use format.

kshtsk (dismissed):

deepssin (approved):
Looks good!

PR #2206, opened 2026-05-26
teuthology/orchestra/daemon/cephadmunit: lookup container runtime

batrick:
Fixes: https://tracker.ceph.com/issues/76926

### DIFF
--- a/teuthology/orchestra/daemon/cephadmunit.py
+++ b/teuthology/orchestra/daemon/cephadmunit.py
@@ -1,6 +1,7 @@
 import logging
 
 from teuthology.orchestra.daemon.state import DaemonState
+from teuthology.exceptions import CommandFailedError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


teuthology/orchestra/daemon/cephadmunit.py
```diff
@@ -34,11 +35,21 @@ def _set_commands(self):
         self.status_cmd = self._get_systemd_cmd('status')
 
     def kill_cmd(self, sig):
-        return ' '.join([
-            'sudo', 'docker', 'kill',
-            '-s', str(int(sig)),
-            'ceph-%s-%s.%s' % (self.fsid, self.type_, self.id_),
-        ])
+        try:
+            self.remote.sh('which podman')
```
kshtsk: 'which' is usually not preinstalled, it is better to check podman and docker presence using `podman --version` and `docker --version`.

kshtsk (approved):
LGTM, however, maybe add docstring about thrown exceptions to the `runtime` and `signal` methods.

zmc (approved):

PR #2210, opened 2026-05-28
containers/teuthology-dev: Skip epel during ansible

zmc:
This may have been broken by https://github.com/ceph/ceph-sepia-secrets/pull/1132


djgalloway:
Assuming the problem here is because the containers don't run inside Sepia and we push epel-mirror.front.sepia.ceph.com, a private EPEL mirror, to the testnodes now.  In which case, approved.

djgalloway (approved):

djgalloway:
This broke it: https://github.com/ceph/ceph-cm-ansible/pull/838.  I made a to-do to pass epel-mirror only for testnodes in sepia.  https://ibm.monday.com/boards/18397023430/pulses/12133552498

PR #2212, opened 2026-06-02
repo_utils: fix worktree creation failure when FETCH_HEAD is missing

djgalloway:
When a bare clone exists but has never been successfully fetched (e.g. after bare init but before any fetch completes), FETCH_HEAD does not exist. The previous code would skip fetch_bare_repo if the freshness sentinel was current, then pass 'FETCH_HEAD' to git worktree add, causing:

    fatal: invalid reference: FETCH_HEAD

Fix by checking for FETCH_HEAD's existence explicitly and forcing a fetch if it is absent, regardless of the freshness sentinel state.


joscollin (approved):

batrick:
https://github.com/ceph/teuthology/pull/2213

PR #2214, opened 2026-06-02
scripts/tests: Speed up tests with importlib

zmc:
These take one second instead of 31.


deepssin (approved):
LGTM! Nice win for CI :)

PR #2217, opened 2026-06-04
integration: Use ceph-devstack

zmc:
ceph-devstack now works well enough on Ubuntu that it could replace the `docs/docker-compose` bits for everything I'm currently aware of.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -7,27 +7,51 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
-      - name: Make archive directory
-        run: mkdir /tmp/archive_dir
-      - name: Test using docker-compose
-        run: ./start.sh
-        working-directory: ./docs/docker-compose
-      - name: Rename Directory
-        # Replace ":" with "_" everywhere in directory path.
-        # This needs to be done because GA does not support ":" colon character in artifacts (like in /root-2025-03-06_18:47:26-teuthology:no-ceph-main-distro-default-testnode).
-        # Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
-        if: always()
+      - name: Install uv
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -7,27 +7,50 @@ jobs:
...
+          ceph-devstack config set containers.postgres.count 0
+          ceph-devstack config set containers.teuthology.repo "$PWD"
+      - name: Build teuthology container
+        run: ceph-devstack -v build
+      - name: Pull other required containers
+        run: ceph-devstack pull
+      - name: Start ceph-devstack
+        run: ceph-devstack -v start
+      - name: Wait
+        run: ceph-devstack wait teuthology
```
deepssin: A timeout will help us prevent a hung command

.github/workflows/integration.yml
```diff
@@ -7,27 +7,50 @@ jobs:
...
         uses: actions/upload-artifact@v4
-        if: always()
         with:
-          name: teuthology-logs 
-          path: |
-            /tmp/archive_dir/*
+          name: archive
+          path: /tmp/artifacts/archive.tar
+      - name: Remove
+        run: uv run ceph-devstack -v remove
```
deepssin: Earlier steps call ceph-devstack directly , but the this step uses uv run ceph-devstack.  minor, but may confuse someone reading this later.

.github/workflows/integration.yml
```diff
@@ -7,27 +7,50 @@ jobs:
...
+        if: success() || failure()
         uses: actions/upload-artifact@v4
-        if: always()
         with:
-          name: teuthology-logs 
-          path: |
-            /tmp/archive_dir/*
+          name: archive
+          path: /tmp/artifacts/archive.tar
+      - name: Remove
```
deepssin, line 55: The cleanup step only runs on success.  runners are ephemeral so leftover containers don't persist across jobs, but if a step fails mid-way (e.g., during start or wait) on a retry or in a self-hosted context, state won't be cleaned.
suggest 
```
- name: Remove
  if: always()
  run: ceph-devstack -v remove
```

.github/workflows/integration.yml
zmc: added!

deepssin (approved):
Looks good!

PR #2219, opened 2026-06-08
Small packaging and dependency updates

zmc:


deepssin (commented):
Mostly LGTM.

Only two small things-

1. Removing `exclude = ["*test*"] `may now package teuthology.test and friends into the wheel ,, should we keep the exclude alongside the new include?
2. The `-p no:cacheprovider addopt` from pytest.ini didn’t make it into [tool.pytest] — worth carrying over, and dropping include pytest.ini from MANIFEST.in

kshtsk:
Yeah, we shouldn't package tests. Could you give a bit more context for the patch `pyproject.toml: Fix "Package would be ignored"`

zmc:
Thanks for the comments @deepssin and @kshtsk - I've re-pushed with some slightly different packaging bits to work around a setuptools issue that was causing the tests to be bundled.

deepssin:
@zmc tox.ini is having `posargs:teuthology scripts`, `tox -e py3 `locally might run into problems

deepssin (approved):

PR #2220, opened 2026-06-10
Add type annotations

zmc:


kshtsk:
Do we really needs that change dramatically? 110 files

zmc:
> Do we really needs that change dramatically? 110 files

well I suppose it's debatable, but type hints can really be helpful when making changes - helps with avoiding regressions and also things as simple as code completion

kshtsk:
> > Do we really needs that change dramatically? 110 files
> 
> well I suppose it's debatable, but type hints can really be helpful when making changes - helps with avoiding regressions and also things as simple as code completion

My concern is not about typing itself, I would prefer to go with the changes gradually, file by file if possible. At least module by module, it's very hard to manage all the changes within one big PR.

teuthology/orchestra/connection.py
```diff
@@ -36,29 +38,25 @@ def create_key(keytype, key):
...
+    if timeout is None:
+        timeout = 60
     user, host = split_user(user_at_host)
-    if _SSHClient is None:
-        _SSHClient = paramiko.SSHClient
-    ssh = _SSHClient()
-
-    if _create_key is None:
-        _create_key = create_key
+    ssh = paramiko.SSHClient()
```
kshtsk, line 59: I see many places like that, which goes beyond that just "Add type annotations", signature changed and logic changed.

PR #2229, opened 2026-06-30
provision: add custom repos for downburst

kshtsk:
When provisioning libvirt targets using downburst, many predefined cloud images are using vendor repos or official mirrors. However it appears that default repos may lack reliability, either mirrors are down, or vendor repos are extremely slow, which can cause big timeouts and as a results teuthology cannot
successfully lock nodes.

Having private mirrors are good for:

- Reduce foreign traffic
- Fast package lookup and download
- Independent from upstream schedule of repo outage

### DIFF
--- a/docs/siteconfig.rst
+++ b/docs/siteconfig.rst
@@ -245,6 +245,29 @@ Here is a sample configuration with many of the options set and documented::
     # Used by the [libcloud](https://libcloud.apache.org/) backend
     nsupdate_url: http://nsupdate.front.sepia.ceph.com/update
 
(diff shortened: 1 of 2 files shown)


zmc:
Could we make it more clear what default behavior is?

kshtsk:
> Could we make it more clear what default behavior is?

By default, the should be no action taken in regards of repos in cloud-init, please, give me a hint how can I make it more clear? In code or in the siteconfig documentation?

kshtsk:
@zmc ping

PR #2231, opened 2026-07-09
task/kernel: fix distro kernel boot on BLS legacy-BIOS nodes

djgalloway:
Rocky 10 smithi jobs installed the new distro kernel but rebooted back into the old one; the same code worked on UEFI testnodes.

Three fixes:

- The grubby detection compared the output of `command -v grubby && echo yes` against 'yes', but command -v prints the binary path first, so grubby was never used even when installed.  Test for non-empty output instead.

- Run grub2-mkconfig before grub2-set-default/grubby, not after.  EL's patched grub2-mkconfig and grub.d scripts can rewrite grubenv, which clobbered the default entry we had just set.  This restores the ordering that previously worked (8ce0b025).

- Force GRUB_DEFAULT=saved in /etc/default/grub on RPM distros before regenerating grub.cfg.  grubby/grub2-set-default record the desired entry as saved_entry in grubenv, but the generated config only consults saved_entry when GRUB_DEFAULT=saved.  On images without it (e.g. Rocky 10 legacy-BIOS smithi images, whose BIOS bootloader is hand-installed because Rocky 10 ships no grub2-pc) grub fell back to menu position and booted the old kernel.  Selecting by exact BLS entry id makes the boot deterministic regardless of how the grub build sorts entries.

Note: the lab images' /etc/default/grub can lack both a GRUB_DEFAULT line and a trailing newline, so the setting is applied by deleting any existing GRUB_DEFAULT line and appending with a leading newline; a plain echo >> would glue onto the last line and break grub2-mkconfig ('Invalid output terminal').

| machine type | distro | link |
|--------|--------|--------|
| trial | rocky10 | https://pulpito.ceph.com/dgalloway-2026-07-09_19:59:01-smoke-main-distro-default-trial/ |
| trial | centos9 | https://pulpito.ceph.com/dgalloway-2026-07-09_20:11:56-smoke-tentacle-distro-default-trial/ |
| smithi | rocky10 | https://pulpito.ceph.com/dgalloway-2026-07-09_20:07:18-smoke-main-distro-default-smithi/ |
| smithi | centos9 | https://pulpito.ceph.com/dgalloway-2026-07-09_20:17:14-smoke-tentacle-distro-default-smithi/ |

### DIFF
--- a/teuthology/task/kernel.py
+++ b/teuthology/task/kernel.py
@@ -971,9 +971,10 @@ def _kernel_set_default_bls(remote, newversion, ostype):
     Prefer `grubby --set-default` when available.
     Otherwise fall back to `grub2-set-default <BLS_ENTRY_ID>` using /boot/loader/entries.
     """
-    has_grubby = remote.sh("sudo command -v grubby && echo yes || echo no").strip() == 'yes'
-
-    if has_grubby and ostype == 'rpm':
+    # grubby is only useful on RPM distros; `command -v` prints the path on
+    # success, so test for non-empty output.
+    if ostype == 'rpm' and remote.sh(
+            "sudo command -v grubby", check_status=False).strip() != '':
         vmlinuz = remote.sh(
             f"sudo find /boot -maxdepth 1 -name 'vmlinuz-*' -type f | "
             f"grep -F '{newversion}' | head -n 1"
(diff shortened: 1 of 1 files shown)


teuthology/task/kernel.py
```diff
@@ -971,7 +971,9 @@ def _kernel_set_default_bls(remote, newversion, ostype):
     Prefer `grubby --set-default` when available.
     Otherwise fall back to `grub2-set-default <BLS_ENTRY_ID>` using /boot/loader/entries.
     """
-    has_grubby = remote.sh("sudo command -v grubby && echo yes || echo no").strip() == 'yes'
+    # `command -v` prints the path on success, so test for non-empty output
+    # rather than comparing against a sentinel string.
+    has_grubby = remote.sh("sudo command -v grubby || true").strip() != ''
```
kshtsk: Though this looks pretty intuitive to use `|| true` for basher, there are other options which can be used in teuthology.
Exception based check might look like:
```
try:
   remote.sh("sudo command -v grubby")
   has_grubby = True
except CommandFailedError:
  # No grubby found, maybe log.warning here
  pass 
```
If we don't care of failed command:
```
   has_grubby = remote.sh("sudo command -v grubby", check_status=False) != ''
```
PS: no action required 
PPS: why even try grubby on non rpm os, if it won't be used below?

djgalloway: Fair points.  Fixed.

kshtsk (approved):
lgtm

PR #2234, opened 2026-07-23
ci: Fix integration workflow

zmc:


.github/workflows/integration.yml
```diff
@@ -13,7 +13,8 @@ jobs:
         run: uv tool install ceph-devstack
       - name: Configure ceph-devstack
         run: |
-          ceph-devstack doctor --fix
+          sudo chown $(whoami) /dev/loop-control
```
kshtsk, line 16: maybe append user with ":" to copy group as well?

kshtsk (approved):

PR #2237, opened 2026-08-04
teuthology/run: migrate from docopt to argparse

hasan4791:
- Migrate teuthology (scripts/run.py) from docopt to argparse

Tracker: https://tracker.ceph.com/issues/74760


teuthology/run/cli.py
```diff
@@ -1,38 +1,86 @@
...
+    )
+    parser.add_argument(
+        '-v', '--verbose',
+        action='store_true',
+        help='be more verbose'
+    )
+    parser.add_argument(
+        '--version',
+        action='version',
+        version=teuthology.__version__,
```
kshtsk, line 25: That should not be here. Please keep previous behavior.

teuthology/run/cli.py
```diff
@@ -1,38 +1,86 @@
-"""
-usage: teuthology --help
-       teuthology --version
-       teuthology [options] [--] <config>...
+import argparse
 
-Run ceph integration tests
+import teuthology
```
kshtsk, line 3: Get rid of this import here, we don't need to evaluate version now.

tests/test_run.py
```diff
@@ -249,27 +250,44 @@ def test_main_interactive(
...
+            'name': None,
+            'suite_path': None,
+            'interactive_on_error': True,
+            'config': ['path/to/config.yml'],
+        }
         run.main(args)
         args, kwargs = m_run_tasks.call_args
         fake_ctx = kwargs["ctx"]._conf
-        assert fake_ctx['interactive_on_error'] is True
+        assert fake_ctx['config']['interactive-on-error'] is True
```
kshtsk, line 294: Why the fake context is changing from underscore to dashes?

teuthology/config.py
```diff
@@ -248,9 +248,12 @@ def __init__(self, config_dict=None):
...
         """
-        Makes sure that the keys of config_dict are able to be used.  For
-        example the "--" prefix of a docopt dict isn't valid and won't populate
-        correctly.
+        Makes sure that the keys of config_dict are able to be used. For
+        example the "--" prefix from argparse/docopt dicts isn't valid and
+        won't populate correctly. This method:
+        - Strips "--" prefix from arguments
+        - Strips "<>" from positional arguments
+        - Replaces hyphens with underscores
```
kshtsk: This change is unrelated to the PR's goal and is not necessary, could you please leave it untouched.

tests/test_run.py
```diff
@@ -166,19 +164,22 @@ def test_main(self, m_report_outcome, m_run_tasks, m_fetch_tasks_if_needed, m_ge
...
+            'lock': True,
+            'machine_type': None,
+            'os_type': 'os_type',
+            'os_version': 'os_version',
+            'block': True,
+            'name': 'the_name',
+            'suite_path': 'some/suite/dir',
+            'interactive_on_error': False,
+            'config': ['path/to/config.yml'],
+        }
```
kshtsk: Here and later. I believe the idea of the test was to check if docopt correctly parses the arguments, similar we need to parse the same arguments should be tried to be parsed by your new argparse code.

tests/test_run.py
```diff
@@ -249,27 +250,44 @@ def test_main_interactive(
...
+            'lock': False,
+            'machine_type': None,
+            'os_type': None,
+            'os_version': None,
+            'block': False,
+            'name': None,
+            'suite_path': None,
+            'interactive_on_error': True,
+            'config': ['path/to/config.yml'],
+        }
```
kshtsk: No need to change the test logic, still the arguments:
```[
            "--interactive-on-error",
            "path/to/config.yml",
        ]
```
To check.

scripts/run.py
```diff
@@ -1,38 +1,86 @@
...
-  --os-type OS_TYPE              Distro/OS of machine to run test on.
-  --os-version OS_VERSION        Distro/OS version of machine to run test on.
-  --block                        block until locking machines succeeds (use with --lock)
-  --name NAME                    name for this teuthology run
-  --suite-path SUITE_PATH        Location of ceph-qa-suite on disk. If not specified,
-                                 it will be fetched
-  --interactive-on-error         drop to a python shell on failure, which will
-                                 halt the job; developer can then ssh to targets
-                                 and examine cluster state.
+def _build_parser():
```
kshtsk: I suggest to give name make_parser(), since we already use make_ suffixes elsewhere, and move this argparse code outside of scripts, for example, to teuthology.run so you can import the parser for testing purposes.

kshtsk:
Change PR title from to "script/teuthology: migrate from docopt to argparse"
To give reviewers an idea what is this change about.

kshtsk (changes requested):

tests/test_run.py
```diff
@@ -249,27 +250,44 @@ def test_main_interactive(
...
+            'lock': True,
+            'machine_type': None,
+            'os_type': None,
+            'os_version': None,
+            'block': True,
+            'name': 'the_name',
+            'suite_path': 'some/suite/dir',
+            'interactive_on_error': False,
+            'config': ['path/to/config.yml', 'path/to/config2.yaml'],
+        }
```
kshtsk: Same here, this is a test vector equivalent to `sys.argv[1:]`

teuthology/run.py
```diff
@@ -297,47 +297,47 @@ def get_teuthology_command(args):
...
                 for arg in value:
                     cmd.append(str(arg))
                 continue
+            # convert argparse underscore keys back to --hyphen-flag form
+            flag = "--" + key.replace("_", "-")
+            cmd.append(flag)
             # so we don't print something like --verbose True
             if isinstance(value, str):
                 cmd.append(value)
     return " ".join(cmd)
```
kshtsk: @zmc I barely find this way of dumping of the actual teuthology command executed useful, because what this method does, is just what we've got from the argument parsing, which can distort the real arguments passed to teuthology. I think we can drop this method entirely und just dump "sys.argv" for the purpose of debug.

scripts/run.py
hasan4791: Keeping make_parser() in the same file would keep the cli arguments isolated and less noisy i feel. 
Would that be fine?

kshtsk: I agree, but still we discussed this a bit earlier and we rather move out from the external scripts dir to teuthology code, but in this case will change file layout a bit, thus for the teuthology.run:
```
teuthology/run.py -> teuthology/run/__init__.py
scripts/run.py -> teuthology/run/cli.py
```
but that can happen in the follow up PR

hasan4791: that makes sense. I can take it up in this PR as we have split it up already.

tests/test_run.py
hasan4791: in the yaml config, we are using "interactive-on-error".

teuthology/run/cli.py
hasan4791: Didn't get this one!!

tests/test_run.py
kshtsk: Oh, I see, looks inconsistent and like a bug, I wonder why the test was working earlier? In anyways, leave with dashes that can be changed in follow up PRs.

teuthology/run/cli.py
```diff
@@ -0,0 +1,86 @@
...
+        help=('drop to a python shell on failure, which will halt the job; '
+              'developer can then ssh to targets and examine cluster state'),
+    )
+    return parser
+
+
+def parse_args(argv=None):
+    return make_parser().parse_args(argv)
+
+
```
kshtsk: No need in this proxy method.

teuthology/run/cli.py
```diff
@@ -74,13 +74,9 @@ def make_parser():
         help=('drop to a python shell on failure, which will halt the job; '
               'developer can then ssh to targets and examine cluster state'),
     )
-    return parser
-
-
-def parse_args(argv=None):
-    return make_parser().parse_args(argv)
+    return parser.parse_args(argv)
```
kshtsk: This should still return the parser.

teuthology/run/cli.py
```diff
@@ -74,13 +74,9 @@ def make_parser():
...
 
 
 def main():
-    args = parse_args()
+    args = make_parser()
```
kshtsk: It is supposed to be used like this:
```
args = make_parser().parse_args()
```

teuthology/run/cli.py
hasan4791: Can you clarify the purpose of just returning the parser?

kshtsk: We do not remove parser, the `make_parser` still returns a parser, what I was originally saying, there is no need to have another method just to call the parse_args(argv) from the generated parser.
So, instead of:
```
def parse_args(args):
  parser = make_parser()
  return parser.parse_args(args)
  
.
.
.
args = parse_args(sys.argv)
```
we just use it like:
```
args = teuthology.run.cli.make_parser.parse_args()
```

kshtsk: > Can you clarify the purpose of just returning the parser?

In general if you'd like to have `build_parser` or `make_parser` method, it should correspond the purpose of the method and return a parser. I don't see any problem of having a generator for the parser, it allows to reuse the parser to parse different set of arguments, for example, in testing, no need to build a new object every time. If you don't want to the `make_parser` return parsed arguments, then you need to rename the `make_parser` to `parse_args`.

teuthology/run/cli.py
kshtsk: as mentioned above, either rename _make_parser()_ to _parse_args()_, or use `args = make_parser().parse_args()`, where `make_parser()` returns an `_argparse.ArgumentParser_` object.

teuthology/run/cli.py
hasan4791: Thank you @kshtsk for sharing your thoughts. Will return parser from make_parser()

tests/test_run.py
```diff
@@ -1,16 +1,52 @@
...
 
 
-class TestRun(object):
-    """ Tests for teuthology.run """
+class TestRun(CliTest):
+    """ Tests for teuthology.run and teuthology.run.cli """
+    script_name = 'teuthology'
+
+    def test_all_args(self):
+        args = make_parser().parse_args([
```
kshtsk: Here and after, instead of calling each time the make_parser() you may want to do this once for class:
```
class TestRun(object):
   parser = teuthology.run.cli.make_parser()

  def test_all_args(self):
    args = parser.parse_args([
    .
    .
    .
```

tests/cli_test.py
```diff
@@ -0,0 +1,25 @@
...
+        return import_module(f"teuthology.{name}.cli")
+
+    def test_help(self, capsys: pytest.CaptureFixture[str], module):
+        with pytest.raises(SystemExit):
+            module.make_parser().parse_args([])
+        assert "usage: " in capsys.readouterr().err
+
+    def test_invalid(self, module):
+        with pytest.raises(SystemExit):
+            module.make_parser().parse_args(["--invalid-option"])
```
kshtsk: I'm not sure, why do we need this?

hasan4791: it was ported from existing test script.py

hasan4791: Since we are moving from scripts folder, doing the same from tests/ dir as well.

kshtsk: From which one? I don't think this code is giving any advantage.

hasan4791: this https://github.com/ceph/teuthology/blob/main/tests/scripts/script.py

kshtsk: ah, that one, I'm skeptical about testing the scripts itself, since we are switching to argparse, it must be enough to test only parser, I also don't see any advantage of testing the argparser argument parser class itself on wether it handles wrong options options or not to be honest.

kshtsk: @zmc do we really need all those script tests? My suggestion is to drop it.

tests/test_run.py
hasan4791: I think, I should move this to the baseclass CliTest as the make_parser() will be dynamically imported based on the module.

```
class CliTest(object)
    def parser(self, module):
        return module.make_parser()
```

```
class TestRun(CliTest):
   cli_module = 'teuthology.openstack.cli'

  def test_all_args(self):
    args = self.parser(module).parse_args([
```

tests/cli_test.py
hasan4791: Drop all the tests or only invalid one?

kshtsk: > Drop all the tests or only invalid one?

Test for `--invalid-option`, and avoid porting that CliTest.

hasan4791: > and avoid porting that CliTest.

CliTest would be the base test class for the argparse going forward and we would be removing the Script base class from tests/scripts.

hasan4791:
@kshtsk Are we good with the changes, as per this i would like to proceed with consecutive PRs.

kshtsk:
hey @hasan4791 the changes look good, have head a chance to run some job in sepia against it?
would be great to see if teuthology run is still functional before we can merge it

hasan4791:
> hey @hasan4791 the changes look good, have head a chance to run some job in sepia against it? would be great to see if teuthology run is still functional before we can merge it

I don't have access to the lab. What would be the other options?

kshtsk:
> > hey @hasan4791 the changes look good, have head a chance to run some job in sepia against it? would be great to see if teuthology run is still functional before we can merge it
> 
> I don't have access to the lab. What would be the other options?

I'll kick it for you

kshtsk:
I've run a job for the merge branch:
```
$ git ls-remote https://github.com/ceph/teuthology | grep 2237/merge
656219396f9a2c0b888cf0ac30822c5bcea8ff99	refs/pull/2237/merge
```
https://pulpito.ceph.com/kyr-2026-08-13_12:15:00-smoke-main-distro-default-trial/

```
2026-08-13T12:21:53.366 INFO:root:teuthology version: 0.0.1.dev209+g656219396
```

kshtsk (approved):
looks good

PR #2238, opened 2026-08-04
suite/schedule: migrate from docopt to argparse

hasan4791:
- Migrate teuthology-suite (scripts/suite.py) from docopt to argparse
- Migrate teuthology-schedule (scripts/schedule.py) from docopt to argparse

Tracker: https://tracker.ceph.com/issues/74760


kshtsk:
That is better to split to two separate PR:
- suite: migrate from docopt to argparse
- schedule: migrate from docopt to argpase

tests/test_schedule.py
```diff
@@ -4,19 +4,19 @@
...
+        'last_in_suite': True,
+        'name': 'NAME',
+        'worker': 'tala',
+        'timeout': '6',
+        'priority': '99',
         # TODO: make this work regardless of $PWD
-        #'<conf_file>': ['../../examples/3node_ceph.yaml',
-        #                '../../examples/3node_rgw.yaml'],
+        #'conf_file': ['../../examples/3node_ceph.yaml',
+        #              '../../examples/3node_rgw.yaml'],
```
kshtsk: We do testing literally of command line arguments, the input vector shouldn't be changed.

tests/test_schedule.py
```diff
@@ -39,7 +39,7 @@ def test_basic(self):
 
     def test_owner(self):
         args = self.basic_args
-        args['--owner'] = None
+        args['owner'] = None
```
kshtsk: Same here the args is an equivalent to what `sys.argv[1:0]` returns, it is not a parsed yet dictionary.

kshtsk (changes requested):
fix the tests

hasan4791:
> fix the tests

I need to pick the PR1 changes for the tests to pass!!

kshtsk:
> > fix the tests
> 
> I need to pick the PR1 changes for the tests to pass!!

I meant the comments about the tests in this PR.

hasan4791:
> That is better to split to two separate PR:
> 
>     * suite: migrate from docopt to argparse
> 
>     * schedule: migrate from docopt to argpase

The changes are pretty less. 
Would like to combine 3 to 4 commands in each PR.

scripts/schedule.py
```diff
@@ -1,61 +1,116 @@
...
-  -d <desc>, --description <desc>      Job description
-  -o <owner>, --owner <owner>          Job owner
-  -w <worker>, --worker <worker>       Which worker to use (type of machine)
-                                       [default: plana]
-  -p <priority>, --priority <priority> Job priority (lower is sooner)
-                                       [default: 1000]
-  -N <num>, --num <num>                Number of times to run/queue the job
-                                       [default: 1]
-
-  --first-in-suite                     Mark the first job in a suite so suite
```
hasan4791, line 35: @kshtsk can you help me with this parameter once?
Does it take true/false along with the parameter name?
--first-in-suite true
or
if we specify "--first-in-suite" it will be consider as True?

kshtsk: Those --first-in-suite and --last-in-suite have no argument, both by default are unset or false.

PR #2240, opened 2026-08-05
suite: Avoid creating subprocesses for each job when scheduling

zmc:
This saves 1-2 seconds per job submission, which on large suites adds up quickly.


kshtsk:
@zmc we are stepping each other's toes,  I'm doing equivalent thing in the #2232  https://github.com/ceph/teuthology/pull/2232/changes/7cc9dade35206f21ee7a0d55b55dcd82f64d06c3

Would you like to go first?

kshtsk:
Btw, why do we still need teuthology-schedule I'd get rid of it.

zmc:
@kshtsk oh I had not seen that PR, and as I'm about to be on leave I don't know that I'll be able to take the time to wrap my head around it. What I can say now is I'm fine with this PR being dropped, left as-is, or revised as folks see fit.

PR #2241, opened 2026-08-11
packaging: fix Pulp deb search and improve its error messages

djgalloway:
Two fixes for `PulpProject`, motivated by every deb job failing `check_packages` with:

```
teuthology.exceptions.VersionNotFoundError: Failed to fetch package version from No results found for https://pulp.front.sepia.ceph.com/pulp/api/v3/distributions/deb/apt
```

(e.g. `yuriw-2026-08-10_23:39:44-smoke-wip-yuriw-cve-dryrun_tentacle-distro-default-trial/453537`)

**1. Search with the numeric distro_version for deb distros.** `PulpProject._get_distro` preferred the codename for ubuntu/debian (`distro_version=jammy`), but shaman/chacra and ceph-build's `pulp_upload.sh` all use the numeric version (`22.04`), so the `pulp_label_select` search matched no distributions. Drop the codename preference so the search uses `os_version` like `ShamanProject` does; `os_version` is always normalized to numeric form by `OS.version_codename` even when the job config supplies a codename.

**2. Report the full query in search errors.** The no-results/multiple-results errors only printed the API endpoint, which made the mismatch above needlessly hard to diagnose. Extract the `pulp_label_select` construction into a `_search_labels` property (so the logged query and the actual request can't drift apart) and report the full query:

```
No results found for https://pulp.front.sepia.ceph.com/pulp/api/v3/distributions/deb/apt?pulp_label_select=project=ceph,flavors=default,distro=ubuntu,distro_version=22.04,arch=x86_64,branch=wip-yuriw-cve-dryrun_tentacle
```

`tests/test_packaging.py`: 183 passed, 59 skipped. The existing distributions for in-flight branches already carry `distro_version=22.04`, so no relabeling or rebuild is needed once this lands.

### DIFF
--- a/tests/test_packaging.py
+++ b/tests/test_packaging.py
@@ -930,7 +930,6 @@ class TestPulpProject(TestBuilderProject):
             ('alma', '9.7', None, 'distro_version=9'),
             ('centos', '8.1', None, 'distro_version=8'),
             ('rhel', '7.0', None, 'distro_version=7'),
-            ('ubuntu', '20.04', 'focal', 'distro_version=focal'),
         ],
     )
     def test_search_uses_major_distro_version(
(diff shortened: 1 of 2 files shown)


vamahaja (commented):
instead can we update `GitBuilder` class which is inherited by both Shaman and Pulp class ?

https://github.com/ceph/teuthology/blob/main/teuthology/packaging.py#L985-L991

vamahaja (approved):

PR #2245, opened 2026-08-18
WARN on and ignore Pulp-hosted shaman results when package_source: shaman AND sha1s of package repos lacking a container image

djgalloway:
## Problem

Security builds from the cve-pipeline are pushed to the lab-internal Pulp but still registered in shaman, with `url`/`chacra_url` pointing at Pulp. Shaman lists them first (newest), so for a sha1 that has both a cve build and a regular chacra build, `ShamanProject` picked the Pulp-hosted entry and 404'd fetching `<chacra_url>/repo` from the Pulp API:

```
teuthology-suite -s smoke -m smithi -c tentacle --newest 20 ...
...
INFO:teuthology.task.install.rpm:Pulling from https://pulp.front.sepia.ceph.com/pulp/content/repos/ceph/wip-yuriw-cve-dryrun_tentacle/66f67bca.../rocky/10/flavors/default/
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://pulp.front.sepia.ceph.com/pulp/api/v3/repositories/rpm/rpm/019fedb6-4793-72bd-895b-d0aa8effb7ad/repo
```

Because `teuthology-suite`'s schedule-time package check uses the same `ShamanProject`, `--newest` happily scheduled that sha1 too.

## Changes

**`packaging: ignore Pulp-hosted shaman results`**
- `ShamanProject` filters out search results whose `url`/`chacra_url` are Pulp-hosted and uses the first usable (chacra) one. For the sha1 above, `repo_url` now resolves to `https://3.chacra.ceph.com/repos/ceph/tentacle/66f67bca.../rocky/10/flavors/default/repo` (verified against live shaman).
- With `package_source: shaman`, a sha1 that *only* has a Pulp build is treated as not found; new `pulp_only` property tells the two cases apart.
- `package_source: pulp` is unaffected: `PulpProject` never consults shaman, it searches the Pulp API directly.

**`suite: warn and back off when --newest hits a Pulp-only sha1`**
- `--newest`: a Pulp-only sha1 is skipped with a warning saying so and, if the run doesn't select `package_source: pulp` and `defaults.cephadm.containers.image: quay-int.front.sepia.ceph.com/ceph-ci/ceph`, how to opt in. Non-newest runs get the same hint in the error before failing on missing packages.
- `teuthology-suite` now reads the extra `<config_yaml>` files it's given (previously only forwarded to `teuthology-schedule`) so it can tell what the run selects.
- A job-level `package_source` is honored for the schedule-time checks (mirroring `teuthology.run`), so passing `pulp.yaml` makes `--newest` verify against Pulp instead of shaman. Falls back to the site `package_source` with a warning if no Pulp credentials are configured on the scheduling host.
- Docs / `--newest` help text updated.

**`suite: back off with --newest when a sha1 has packages but no container`**
- Release builds (`ceph-release-pipeline` → `ceph-dev-pipeline` with `CI_CONTAINER=false`) have packages and a `completed` shaman build record, but no ceph-ci container — e.g. tentacle `7f793731…` right now: shaman says ready, `quay.ceph.io/ceph-ci/ceph:7f793731…` is 404. Nothing in shaman records this, so `--newest` would pick it and cephadm jobs would fail to pull.
- For jobs whose tasks use `cephadm`, ask the registry of the configured cephadm image (`defaults`/`overrides.cephadm.containers.image`) whether `<image>:<sha1>` exists (registry v2 manifest HEAD; anonymous bearer token fetched when challenged). If not:
  > `ceph sha1 <sha1> has packages but no container at <image>:<sha1>. Pending release build, or failed container build of a passed package build? Moving on to the next sha1 of <ref>`
  
  and with `--newest` the sha1 is treated like missing packages so we back off. Without `--newest` it's only a warning. If the registry can't be queried, scheduling proceeds as before. Non-cephadm jobs never trigger the lookup, so non-container suites can still test release sha1s.
- Verified live: `quay.ceph.io/ceph-ci/ceph:7f793731…` → missing, `:66f67bca…` → exists; Docker Hub token flow works (`alpine:latest`).

## Reviewer notes

- The Pulp-only detection relies on the cve-pipeline continuing to register its builds in shaman with Pulp URLs in `chacra_url`/`url`. If that ever stops, Pulp-only sha1s just look "not found" (no false positives, but the explanatory warning goes away). Noted in `ShamanProject.pulp_only` and `util.hash_only_in_pulp`.
- The quay-int hostname / pulp source name are module constants in `teuthology/suite/run.py` (`INTERNAL_*`); happy to make them site-config knobs if preferred.

## Testing

`tests/test_packaging.py`, `tests/suite`, `tests/scripts/test_suite.py`, `tests/task/test_install.py`: 321 passed. New tests cover the exact two-entry shaman response from the failing job, the Pulp-only case, config-based detection of pulp/quay-int selection, an end-to-end `--newest` run that skips a Pulp-only sha1 with one warning, the registry helper (200/404/401+token/error), and an end-to-end `--newest` run that skips a container-less sha1.

### DIFF
--- a/docs/detailed_test_config.rst
+++ b/docs/detailed_test_config.rst
@@ -316,3 +316,19 @@ pulp:
   username: pulp-username
   password: pulp-password
   force_noarch: True
+
+``package_source`` may also be set in an extra job yaml passed to
+``teuthology-suite``. It then applies both to the scheduled jobs and to the
+package checks ``teuthology-suite`` performs itself (for example when
+backtracking with ``--newest``), provided the scheduling host has the Pulp
+credentials configured.
+
+Some builds (currently security builds) are published only to the lab-internal
+Pulp and Quay instances. We still register them in Shaman, but their
+``chacra_url`` points to the Pulp repo URL. When using
+``package_source: shaman``, we ignore those repos. If a given sha1 has both a
+chacra *and* a Pulp repo, the chacra build is used. If it only has a Pulp
+build, it is treated as not found; ``--newest`` then logs a warning and moves
+on to the next sha1 unless the run selects ``package_source: pulp`` and the
+internal quay cephadm image (``defaults.cephadm.containers.image:
+quay-int.front.sepia.ceph.com/ceph-ci/ceph``).
--- a/docs/detailed_test_config.rst
+++ b/docs/detailed_test_config.rst
@@ -332,3 +332,11 @@ build, it is treated as not found; ``--newest`` then logs a warning and moves
 on to the next sha1 unless the run selects ``package_source: pulp`` and the
 internal quay cephadm image (``defaults.cephadm.containers.image:
 quay-int.front.sepia.ceph.com/ceph-ci/ceph``).
+
+Similarly, release builds have packages but no CI container. If a job uses
+the ``cephadm`` task, ``teuthology-suite`` asks the registry of the configured
+cephadm image (``defaults.cephadm.containers.image`` in the site config, or
+``defaults``/``overrides`` in the job yaml) whether ``<image>:<sha1>`` exists.
+If it does not, a warning is logged; with ``--newest`` the sha1 is skipped and
+the next one is tried. If the registry cannot be queried, scheduling proceeds
+as if the container existed.
--- a/scripts/suite.py
+++ b/scripts/suite.py
@@ -43,7 +43,14 @@ Standard arguments:
                               Search for the newest revision built on all
                               required distro/versions, starting from
                               either --ceph or --sha1, backtracking
-                              up to <newest> commits [default: 0]
+                              up to <newest> commits [default: 0].
+                              Packages are looked up using the
+                              package_source from <config_yaml> if set.
+                              A sha1 whose packages exist only on the
+                              lab-internal Pulp (e.g. a security build)
+                              is skipped with a warning unless the run
+                              selects package_source: pulp and the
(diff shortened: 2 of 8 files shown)


dmick:
Why are pulp-only builds registered with shaman at all?

djgalloway:
> Why are pulp-only builds registered with shaman at all?

The pulp GUI is half-baked as you know.  Devs would have no way of knowing the status of their branch being built.  We could make shaman interface with pulp directly but, for now, we're tricking shaman by giving it pulp URLs.

djgalloway:
Anyone?

djgalloway:
Opened a tracker forgetting I even had this PR open: https://tracker.ceph.com/issues/80490

The Umbrella RC tanked a bunch of runs again.  e.g., https://pulpito.ceph.com/jenkins-build-2026-09-10_06:07:01-fs-umbrella-distro-default-trial/

kshtsk:
I would avoid changes to override passthrough mechanism now, because this makes changes too complicated.
For processing `--newest` I would suggest to still take into consideration teuthology config as default, and in case one need to override this settings, it is better to use another option for teuthology-suite like --package-source and environment like TEUTH_PACKAGE_SOURCE, instead overrides, override yamls are more like for test config itself. When some task needs to know what package source is used it is naturally taken from the yaml default config, rather than from override method.

PR #2246, opened 2026-08-19
opsys: let a job ask for a major-only Rocky version

djgalloway:
Rocky only maintains its newest point release and ceph's rocky/10 builds are compiled on it (they now need `c-ares >= 1.28`, which only 10.2 ships — see ceph/ceph-cm-ansible#873), so the lab is moving to treating Rocky 10 like CentOS Stream: the FOG image is named by major (`trial_rocky_10`), the capture job keeps it on the current minor, and jobs request `os_version: "10"` instead of `"10.1"`.

Two things stood in the way of `os_version: "10"`:

- `packaging.py` calls `OS.version_codename()` for every job with an `os_version`, and `DISTRO_CODENAME_MAP['rocky']` only had point releases, so `"10"` raised `KeyError: '10 not a rocky version or codename'` at `check_packages`. Add the major, the way centos has `"9"` and `"10"`.
- FOG's `_verify_installed_os()` compares `OS('rocky', '10')` with the node's `/etc/os-release` (`VERSION_ID="10.1"`), and `OS.__eq__` compared versions exactly (only stripping `.stream`), so every deploy would be rejected with `Expected … rocky 10 but found rocky 10.1`. Treat a major-only version as matching any point release of that major; two full versions still have to match exactly.

Nothing else keys on the minor: the FOG image name is built from `os_version` verbatim, reimage-type nodes aren't filtered by OS when locking (`lock/ops.py`), and the shaman/pulp queries already use the major (`rocky/10/x86_64`).

Tests added for the equality rule and the codename lookup; `tests/orchestra/test_opsys.py` + `tests/test_packaging.py` pass.

Companion changes: ceph/ceph-cm-ansible#873 (repos back at `/rocky/10/`), a `trial_rocky_10` FOG image, and `os_version: "10"` in ceph's `qa/distros/all/rocky_10.yaml`.

### DIFF
--- a/teuthology/orchestra/opsys.py
+++ b/teuthology/orchestra/opsys.py
@@ -49,6 +49,10 @@ DISTRO_CODENAME_MAP = {
         "9.8": "rocky",
         "10.1": "rocky",
         "10.2": "rocky",
+        # aliases for latest minor releases
+        "8": "rocky",
+        "9": "rocky",
+        "10": "rocky",
     },
     "centos": {
         "10": "stream",
@@ -277,7 +281,18 @@ class OS(object):
                     codename=repr(self.codename))
 
     def __eq__(self, other):
+        """
+        Equal when the names match and the versions match.  A major-only
+        version matches any minor of that major ("24" == "24.04"); two
+        dotted versions must match exactly ("24.10" != "24.04").  centos's
+        ".stream" suffix is ignored.  See commit message for the why.
+        """
         if self.name.lower() != other.name.lower():
             return False
         normalize = lambda s: s.lower().removesuffix(".stream")
-        return normalize(self.version) == normalize(other.version)
+        mine, theirs = normalize(self.version), normalize(other.version)
+        if mine == theirs:
+            return True
+        if '.' not in mine or '.' not in theirs:
+            return mine.split('.')[0] == theirs.split('.')[0]
+        return False
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -66,6 +66,8 @@ class FOG(object):
         self.shortname = self.remote.shortname
         self.os_type = os_type
         self.os_version = os_version
+        # set when a major-only os_version gets resolved to a point release
+        self.resolved_os_version = None
         self.log = log.getChild(self.shortname)
 
     def create(self):
@@ -163,14 +165,45 @@ class FOG(object):
         name = f"{self.remote.machine_type}_{os_type}_{os_version}"
         if image := do_get(name):
             return image
-        elif os_type == 'centos' and not os_version.endswith('.stream'):
+        if os_type == 'centos' and not os_version.endswith('.stream'):
             image = do_get(f"{name}.stream")
+        elif '.' not in os_version:
+            # A major-only version means the latest minor the lab has an
+            # image for; the deploy is verified against that exact minor
+            image = self._latest_minor_image(name)
         if image:
             return image
-        else:
-            raise RuntimeError(
-                "Fog has no %s image. Available %s images: %s" %
-                (name, self.remote.machine_type, self.suggest_image_names()))
+        raise RuntimeError(
+            "Fog has no %s image. Available %s images: %s" %
+            (name, self.remote.machine_type, self.suggest_image_names()))
+
+    def _latest_minor_image(self, base_name):
+        """
+        The latest captured point-release image for a major-only version,
+        e.g. trial_rocky_10 -> trial_rocky_10.2 over trial_rocky_10.1.
+        Only images FOG has a size for (i.e. a completed capture) count.
+
+        :returns: The image dict, or None
+        """
+        resp = self.do_request(f'/image/search/{base_name}')
+        best = best_key = None
+        for image in resp.json().get('images') or []:
+            img_name = image.get('name') or ''
+            if not img_name.startswith(base_name + '.') or not image.get('size'):
+                continue
+            try:
+                key = tuple(int(p) for p in img_name[len(base_name) + 1:].split('.'))
+            except ValueError:
+                continue
+            if best_key is None or key > best_key:
+                best, best_key = image, key
+        if best:
+            self.resolved_os_version = best['name'].rsplit('_', 1)[-1]
+            self.log.info(
+                "Resolved %s %s to image %s",
+                self.os_type, self.os_version, best['name'],
+            )
+        return best
 
     def suggest_image_names(self):
         """
@@ -375,7 +408,10 @@ class FOG(object):
         )
 
     def _verify_installed_os(self):
-        wanted_os = OS(name=self.os_type, version=self.os_version)
+        wanted_os = OS(
+            name=self.os_type,
+            version=self.resolved_os_version or self.os_version,
+        )
         if self.remote.os != wanted_os:
             raise RuntimeError(
                 f"Expected {self.remote.shortname}'s OS to be {wanted_os} but "
--- a/tests/orchestra/test_opsys.py
+++ b/tests/orchestra/test_opsys.py
@@ -406,3 +406,20 @@ class TestOS(object):
     def test_eq_not_equal(self):
         os = OS(name='ubuntu', codename='trusty', version='16.04')
         assert OS(name='ubuntu', codename='trusty', version='14.04') != os
+
+    def test_eq_major_only_matches_any_minor(self):
+        major = OS(name='ubuntu', version='24')
+        assert major == OS(name='ubuntu', version='24.04')
+        assert major == OS(name='ubuntu', version='24.10')
+        assert major != OS(name='ubuntu', version='22.10')
+        assert OS(name='rocky', version='10') == OS(name='rocky', version='10.2')
+        assert OS(name='rocky', version='10') != OS(name='alma', version='10.2')
+
+    def test_eq_dotted_versions_never_cross_match(self):
+        assert OS(name='ubuntu', version='24.10') != OS(name='ubuntu', version='24.04')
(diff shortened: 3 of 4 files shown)


teuthology/orchestra/opsys.py
```diff
@@ -47,6 +47,10 @@
         "8.10": "rocky",
         "9.7": "rocky",
         "9.8": "rocky",
+        # Rocky only maintains its newest point release and ceph's rocky/10
+        # builds follow it, so jobs ask for the major ("10") and get whatever
+        # 10.x the lab currently images, the way centos "9" means 9.stream.
+        "10": "rocky",
```
kshtsk: I would not let this change merged. First, the 9.stream is actually in format "N.x", here you (or AI) are changing to "10". Also it is not true that rocky changed from 10.x to 10 whatever is latest, it was always a synonym (or just link) to the latest (c.f. https://download.rockylinux.org/pub/rocky/), we've been using the "_latest" suffix in test suites. From other side it is a misguiding practice to refer to the 'lab current images' because no-one knows what is this, I would suggest to get back to stick to the version "N.X" for rocky and alma. Also it is not true that Rocky does not maintain older versions. What does it mean "maintains"? Older versions and repos are still available in vault (i.e. https://download.rockylinux.org/vault/rocky/)

teuthology/orchestra/opsys.py
```diff
@@ -280,4 +284,12 @@ def __eq__(self, other):
         if self.name.lower() != other.name.lower():
             return False
         normalize = lambda s: s.lower().removesuffix(".stream")
-        return normalize(self.version) == normalize(other.version)
+        mine, theirs = normalize(self.version), normalize(other.version)
+        if mine == theirs:
+            return True
+        # A major-only version ("10") means any point release of that major:
+        # a job asking for rocky 10 is satisfied by a node reporting 10.1 or
+        # 10.2. Two full versions still have to match exactly.
```
kshtsk: That's too bad, everyone knows that "10" for rocky means latest minor release, for now it is "10.2", no "10.1" and no "10.0", this may misguide a user and a developer. For centos, 9 is always means "9.stream". There is no "9.1", "9.5", "9.8", teuthology does not operate the centos version like that.

teuthology/orchestra/opsys.py
djgalloway: Teuthology requires a minor version when scheduling runs only on RPM-based distros.

All Ceph packages (and the container image itself) are now built in a container image based on the latest major version of whatever distro we're building for.

See:
https://github.com/ceph/ceph-build/blob/main/scripts/build_container#L47 for the base OS container image of the container itself
https://github.com/ceph/ceph/blob/main/src/script/build-with-container.py#L194 for the container image used to build packages

So one day we're building Ceph `main` and `umbrella` on a Rocky 10.1 container image, the next day, it's silently 10.2 instead.  But teuthology and the `qa/` dir continue to lag behind and we continue to play this game of cat and mouse until someone notices the next Rocky minor version was released and changes all the symlinks and distro codemaps _again_.

Pinning build_container and build-with-container.py would be a step backward.  _Unpinning_ the minor version in teuthology means it follows the same moving target as the rest of the upstream CI.

For CentOS 9.stream, even though `.stream` was technically a minor version of CentOS as far as teuthology was concerned, it was still a moving target.. "stream" was "stream."

kshtsk: I can imagine the container builder is always running on latest 10 and this is confusing. If main is built against new release, what happens to older ceph branches, they are broken immediately if they stick to 10 because of new dependencies or how they are different? We still need to back port bunch of new patches urgently? Why not stick the container building to dotted version?

teuthology/orchestra/opsys.py
kshtsk: > So one day we're building Ceph main and umbrella on a Rocky 10.1 container image, the next day, it's silently 10.2 instead. 

Looks like we developed a reliable process :-)

teuthology/orchestra/opsys.py
djgalloway: If a user is running a packaged-based version of Ceph built on and for Rocky 10.1 and they run a `yum update` to 10.2, Ceph comes with it _precisely because_ we don't pin the container image to a minor version.  (Note that the stock repo files use `$releasever` = `10`)

teuthology/orchestra/opsys.py
djgalloway: > teuthology does not operate the centos version like that.

You're proving my point.  CentOS 9 Stream was a moving target.  9.stream was always the latest packages.  So is "Rocky 10."  We need to stop hardcoding the minor version.

dmick: what is "too bad"?  What's the complaint here?

teuthology/orchestra/opsys.py
kshtsk: I actually don't want to continue arguing, if we need this change I rather approve it, only thing I prefer is to drop this shameful comment because it is a bit confusing, or just leave one liner comment like `# aliases for latest minor releases`, and add the same for "8" and "9" versions.

teuthology/orchestra/opsys.py
kshtsk: that said, I find this code confusing when in the code appears something like this "version(10.2) == version(10.1)" true.

djgalloway: Fair, I'll change it to:

If a user or the qa dir asks for 10.x, they should get a FOG image of 10.x.  If they ask for just "10," ask FOG what the latest is.

In both cases, still require the requested minor version match what actually got deployed.

kshtsk: Since this code is also working for example for Ubuntu:
```
>>> u2410=OS("ubuntu", "24.10")
>>> u2404=OS("ubuntu", "24.04")
>>> u2410 == u2404
False
>>> u24=OS("ubuntu", "24")
>>> u24
OS(name='ubuntu', version='24', codename=None)
>>> u24 == u2404
True
>>> u24 == u2410
True
>>> u2210=OS("ubuntu", "22.10")
>>> u24 == u2210
False

```
Could you please drop this comment and add generalized inline docs for the method.

kshtsk: I mean equal operator method  "__eq__()"

teuthology/provision/fog.py
```diff
@@ -163,14 +165,45 @@ def do_get(name):
         name = f"{self.remote.machine_type}_{os_type}_{os_version}"
         if image := do_get(name):
             return image
-        elif os_type == 'centos' and not os_version.endswith('.stream'):
+        if os_type == 'centos' and not os_version.endswith('.stream'):
             image = do_get(f"{name}.stream")
+        elif '.' not in os_version:
+            # A major-only version means the newest minor the lab has an
+            # image for; the deploy is verified against that exact minor
+            image = self._newest_minor_image(name)
```
kshtsk: We use newest when it comes to git repo sha1 or particular build available on Shaman, wouldn't it be probably better to use _latest_minor_release_image?

yuriw (dismissed):

kshtsk (dismissed):

djgalloway:
Tested https://pulpito.ceph.com/dgalloway-2026-08-19_22:15:01-rados:singleton-main-distro-default-trial/

teuthology/provision/fog.py
djgalloway:
```
diff --git a/tests/provision/test_fog.py b/tests/provision/test_fog.py
index da68ea12..d227a86a 100644
--- a/tests/provision/test_fog.py
+++ b/tests/provision/test_fog.py
@@ -184,7 +184,7 @@ class TestFOG(object):
             # exact major-named image wins when it exists
             ([dict(id=1, name='type1_rocky_10', size='1:2:')],
              'type1_rocky_10', None),
-            # otherwise the newest captured minor is used and remembered
+            # otherwise the latest captured minor is used and remembered
             ([dict(id=2, name='type1_rocky_10.1', size='1:2:'),
               dict(id=3, name='type1_rocky_10.2', size='1:2:')],
              'type1_rocky_10.2', '10.2'),
diff --git a/teuthology/provision/fog.py b/teuthology/provision/fog.py
index fd4a0ed7..64561f4a 100644
--- a/teuthology/provision/fog.py
+++ b/teuthology/provision/fog.py
@@ -168,18 +168,18 @@ class FOG(object):
         if os_type == 'centos' and not os_version.endswith('.stream'):
             image = do_get(f"{name}.stream")
         elif '.' not in os_version:
-            # A major-only version means the newest minor the lab has an
+            # A major-only version means the latest minor the lab has an
             # image for; the deploy is verified against that exact minor
-            image = self._newest_minor_image(name)
+            image = self._latest_minor_image(name)
         if image:
             return image
         raise RuntimeError(
             "Fog has no %s image. Available %s images: %s" %
             (name, self.remote.machine_type, self.suggest_image_names()))
 
-    def _newest_minor_image(self, base_name):
+    def _latest_minor_image(self, base_name):
         """
-        The newest captured point-release image for a major-only version,
+        The latest captured point-release image for a major-only version,
         e.g. trial_rocky_10 -> trial_rocky_10.2 over trial_rocky_10.1.
         Only images FOG has a size for (i.e. a completed capture) count.
         ```

kshtsk (approved):

teuthology/orchestra/opsys.py
dmick: I'm gonna assume you don't know what 'shameful' means, but I have to ask what you could possibly mean by "shameful comment"?

kshtsk: It was a poorly chosen word, and I apologize. Have a nice day.

PR #2251, opened 2026-08-24
teuthology-lock: add --list-images

dmick:
currently only implemented for Fog, but: allow querying reimage server for images known for a machine type.

### DIFF
--- a/teuthology/lock/cli.py
+++ b/teuthology/lock/cli.py
@@ -58,7 +58,7 @@ def main(ctx):
         else:
             # This condition might never be hit, but it's not clear.
             assert ctx.num_to_lock or ctx.list or ctx.list_targets or \
-                ctx.summary or ctx.brief, \
+                ctx.summary or ctx.brief or ctx.list_images, \
                 'machines must be specified for that operation'
     if ctx.all:
         assert ctx.list or ctx.list_targets or ctx.brief, \
@@ -251,6 +251,11 @@ def main(ctx):
         if ctx.desc is not None or ctx.status is not None:
             for machine in machines_to_update:
                 ops.update_lock(machine, ctx.desc, ctx.status)
+
+    elif ctx.list_images:
+        assert ctx.machine_type is not None, \
+            'you must specify a machine type for list_images'
+        ops.list_images(ctx.machine_type)
 
     return ret
 
--- a/teuthology/provision/__init__.py
+++ b/teuthology/provision/__init__.py
@@ -24,10 +24,7 @@ def get_reimage_types():
     return pelagos.get_types() + fog.get_types() + maas.get_types()
 
 
-def reimage(ctx, machine_name, machine_type):
-    os_type = get_distro(ctx)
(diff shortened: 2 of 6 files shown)


teuthology/lock/cli.py
```diff
@@ -252,6 +252,11 @@ def main(ctx):
             for machine in machines_to_update:
                 ops.update_lock(machine, ctx.desc, ctx.status)
 
+    elif ctx.list_images:
+        assert ctx.machine_type is not None, \
+            'you must specify a machine type for list_images'
+    ops.list_images(ctx.machine_type)
```
djgalloway: `\t\t`

teuthology/provision/__init__.py
```diff
@@ -23,11 +23,7 @@ def _logfile(shortname: str, archive_path: str = ""):
 def get_reimage_types():
     return pelagos.get_types() + fog.get_types() + maas.get_types()
 
-
-def reimage(ctx, machine_name, machine_type):
-    os_type = get_distro(ctx)
-    os_version = get_distro_version(ctx)
-
+def get_provisioner_object(machine_type, machine_name='', os_type='', os_version=''):
```
kshtsk, line 27: This is starting to look like a factory method for hypothetical "Provisioner" class.

dmick:
this needs some work; PR was premature.  Will do better.

teuthology/lock/cli.py
dmick: ah, you mean 258 should be indented one more level.  Yep, got it.

teuthology/provision/fog.py
```diff
@@ -205,13 +205,15 @@ def _latest_minor_image(self, base_name):
...
         """
-        Suggest available image names for this machine type.
+        Return available image names for this machine type.
 
         :returns: A list of image names.
         """
-        resp = self.do_request('/image/search/%s' % self.remote.machine_type)
+        if not machine_type:
+            machine_type = self.remote.machine_type
+        resp = self.do_request('/image/search/%s' % machine_type)
```
kshtsk: The method argument immutability is preferable, so maybe:
```
term = machine_type or self.remote.machine_type 
resp = self.do_request(f'/image/search/{term}')
```

dmick:
Fixed up a number of issues, including just plain broken code.  Fixed test, added test case to test list_images() with and without a machine_type.

djgalloway:
> Fixed up a number of issues, including just plain broken code. Fixed test, added test case to test list_images() with and without a machine_type.

Still a Draft or ready for re-review?

teuthology/provision/fog.py
dmick: I don't really see why this is a problem, and I've written and reviewed lots of code that mutates the local copy of the function parameter, but I do think the or is cleaner, so why not.

dmick:
Fixed up kyr's complaint, retested; should be gtg

kshtsk (approved):
lgtm

PR #2253, opened 2026-09-05
bootstrap: tolerate galaxy.ansible.com outages

djgalloway:
Galaxy's recurring 504 spells currently fail any job that runs `./bootstrap` (it killed two sepia-fog-images runs today). Retry the collection install a few times, then fall back to the collections a previous bootstrap already installed on the host before giving up.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -87,5 +87,32 @@ uv sync --quiet --frozen --all-extras
 # To avoid crashing older dispatchers
 ln -sf .venv virtualenv
 
-# Install ansible collections
-uv run ansible-galaxy install -r requirements.yml
+# Install ansible collections.  galaxy.ansible.com has regular outages
+# (504s), so retry, and before giving up fall back to whatever collections a
+# previous bootstrap already installed on this host.
+galaxy_install() {
+    local attempt attempts=3
+    for attempt in $(seq $attempts); do
+        uv run ansible-galaxy install -r requirements.yml && return 0
+        if [ $attempt -lt $attempts ]; then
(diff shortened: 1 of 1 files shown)


bootstrap
```diff
@@ -87,5 +87,30 @@ uv sync --quiet --frozen --all-extras
...
-uv run ansible-galaxy install -r requirements.yml
+# Install ansible collections.  galaxy.ansible.com has regular outages
+# (504s), so retry, and before giving up fall back to whatever collections a
+# previous bootstrap already installed on this host.
+galaxy_install() {
+    local attempt
+    for attempt in 1 2 3; do
+        uv run ansible-galaxy install -r requirements.yml && return 0
+        echo "ansible-galaxy install failed (attempt $attempt); retrying in 15s..." >&2
+        sleep 15
```
kshtsk: I know this is not that important here, but when last attempt finished it's still making extra pause and doesn't do any attempt.

kshtsk: But in general isn't 15 seconds is long enough for the pause? If service has an outage, I guess it is a very unlikely it becomes available shortly.

djgalloway: It worked yesterday when the infra was struggling 🤷‍♂️. I will fix the last unnecessary pause.

djgalloway: Fixed in f23e4461: the sleep only happens between attempts now. Kept 15s: it's the gap that got the smoke runs through the 504s this week, and the fallback to already-installed collections is what actually saves us when galaxy is down for real.

kshtsk: okay, let's give it a try

kshtsk (approved):

PR #2254, opened 2026-09-05
install/deb: stop using apt-key for the autobuild key

djgalloway:
apt-key is deprecated since Debian 11/Ubuntu 22.04 and removed in Ubuntu 26.04 (resolute), where every install task died on `sudo apt-key add -` (first hit: the first 26.04 smoke run). Drop the ASCII-armored autobuild key into `/etc/apt/trusted.gpg.d/ceph-autobuild.asc` instead — supported since apt 1.4, so it works on every apt release we test — and skip when it is already present.

Branch also carries #2253 (bootstrap galaxy retry) so it can be used as-is with `--teuthology-branch`.

### DIFF
--- a/teuthology/task/install/deb.py
+++ b/teuthology/task/install/deb.py
@@ -11,6 +11,28 @@ from teuthology.task.install.util import _get_builder_project, _get_local_dir
 
 log = logging.getLogger(__name__)
 
+
+RELEASE_KEY_URL = os.environ.get(
+    'TEUTH_TASK_INSTALL_RELEASE_KEY_URL',
+    'https://git.ceph.com/?p=ceph.git;a=blob_plain;f=keys/autobuild.asc',
+)
+RELEASE_KEY_PATH = '/etc/apt/trusted.gpg.d/ceph-autobuild.asc'
+
+
+def _ensure_release_key(remote):
+    """
+    Install the Ceph autobuild signing key if it is not already trusted.
+
+    apt-key is deprecated since Debian 11/Ubuntu 22.04 and gone in Ubuntu
+    26.04, so drop the ASCII-armored key straight into trusted.gpg.d
+    (supported since apt 1.4) instead of piping it through apt-key.
+    """
+    remote.sh(
+        "set -o pipefail; "
+        f"test -s {RELEASE_KEY_PATH} || "
+        f"curl -sSfL '{RELEASE_KEY_URL}' | sudo tee {RELEASE_KEY_PATH} > /dev/null"
+    )
+
 def _retry_if_eagain_in_output(remote, args):
     # wait at most 5 minutes
     with safe_while(sleep=10, tries=30) as proceed:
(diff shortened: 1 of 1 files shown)


djgalloway:
Validated: first passing Ubuntu 26.04 smoke run used this branch as `--teuthology-branch wip-resolute` — https://pulpito-ng.ceph.com/dgalloway-2026-09-05_11:49:45-smoke-wip-resolute-distro-default-trial/

teuthology/task/install/deb.py
```diff
@@ -11,6 +11,34 @@
...
+    if r.exitstatus == 0:
+        return
+    remote.run(
+        args=[
+            'wget', '-q', '-O-', RELEASE_KEY_URL,
+            run.Raw('|'),
+            'sudo', 'tee', RELEASE_KEY_PATH,
+        ],
+        stdout=StringIO(),
+    )
```
kshtsk: I wonder why use wget here, the curl installed by default? The one line would be enough in this case:
```
remote.sh(f"curl -s {RELEAsE_KEY_URL} | sudo tee {RELEASE_KEY_PATH}")
```

djgalloway: Good call, switched to curl (`-sSfL` so a bad response fails instead of being written as the key) and squashed into the commit. Kept the args-list form rather than `remote.sh` to match the rest of this file. Also rebased onto the updated #2253 head.

kshtsk: I would still prefer to use `remote.sh` for better readability.
So overall method would look like:

```
remote.sh(
  f"test -s {RELEASE_KEY_PATH} || "
  f"curl -sSfL {RELEASE_KEY_URL} | "
  f"sudo tee {RELEASE_KEY_PATH}"
)
```
If you'd still want to have two separate calls for the key presence and copy, but don't know how to handle it, then it's done by checking CommandFailedError exception:
```
try:
  remote.sh(f"test -s {RELEASE_KEYPATH")
except CommandFailedError:
  remote.sh(f"curl -sSfL {RELEASE_KEY_URL} | sudo tee {RELEASE_KEY_PATH}")
```

teuthology/task/install/deb.py
```diff
@@ -11,6 +11,34 @@
 
 log = logging.getLogger(__name__)
 
+
+RELEASE_KEY_URL = 'https://git.ceph.com/?p=ceph.git;a=blob_plain;f=keys/autobuild.asc'  # noqa
```
kshtsk: I wonder if it makes sense to make it read from the env, like TEUTH_TASK_INSTALL_RELEASE_KEY_URL or something?

bootstrap
```diff
@@ -87,5 +87,32 @@ uv sync --quiet --frozen --all-extras
...
+    print(c['name'] if isinstance(c, dict) else c)
+"); do
+        if ! echo "$installed" | grep -q "^$coll "; then
+            echo "collection $coll is not installed" >&2
+            missing=1
+        fi
+    done
+    return $missing
+}
+galaxy_install
```
kshtsk, line 118: I guess this is here only for fixing some PR check issues, and this change is in another PR, and should go aways once that merged?

teuthology/task/install/deb.py
kshtsk: I don't want to ask you make this change now if it doesn't make sense for you.

bootstrap
kshtsk: Or you like to bring it within this PR?

djgalloway: Yes, the commit should disappear once the first PR is merged I thought.  They're not really related hence the two separate PRs.

kshtsk (dismissed):

djgalloway:
I'm actually 90% sure nothing is using autobuild.asc anywhere anyway...  We use `deb [trusted=yes]` in all our chacra repo files.  Chacra does not sign packages.  I think this key is leftover from the gitbuilder days.  WDYT about just removing `_ensure_release_key` altogether?

kshtsk:
> I'm actually 90% sure nothing is using autobuild.asc anywhere anyway... We use `deb [trusted=yes]` in all our chacra repo files. Chacra does not sign packages. I think this key is leftover from the gitbuilder days. WDYT about just removing `_ensure_release_key` altogether?

This is very interesting, if it is not used for decade I don't mind to drop it, especially if it is hardcoded, however, if we want to test signed repos/packages we need some nice mechanism to inject keys with (probably) handy mechanism to override them, if not env then overriding task.install parameters, no hardcoding. So for now simplest is to drop the code if it is not used.

kshtsk (approved):

PR #2255, opened 2026-09-05
provision/fog: recover from a stale known_hosts entry after reimage

djgalloway:
A freshly-imaged node has new ssh host keys. If the node's old key is still in `~/.ssh/known_hosts` (left behind by an earlier ansible/ssh session run as the same user), paramiko raises `BadHostKeyException` on every connect after the reimage. Today that is handled badly:

- `connection.connect()` retries it like a transient error until `safe_while` gives up (~145 s),
- `FOG._wait_for_ready()` catches the resulting `MaxWhileTries` and loops, power-cycling the node at 50% of `fog_wait_for_ssh_timeout`,
- `teuthology-lock --lock --os-type/--os-version` sits there for hours.

Seen in Sepia on 2026-09-05: `teuthology-lock` runs as `jenkins-build` on soko04 hung ~2 h on trial002/trial065 (and would have on trial150) because that account's known_hosts had entries for 17 trial nodes recorded by earlier `sepia-fog-images` runs. The code comment in `_wait_for_ready` even predicted this case.

Changes:
- `connection.connect()`: `BadHostKeyException` is not retried; it is logged and raised so the caller can decide.
- `FOG._wait_for_ready()`: on `BadHostKeyException`, remove the stale entry with `ssh-keygen -R` (new `misc.ssh_keygen_remove()`) and try again. The node was just reimaged, so a key change is expected.
- Unit tests for both.

Companion change in ceph-build (`sepia-fog-images`, ceph/ceph-build#2705): the job's `phase_prepare` now sets `UserKnownHostsFile /dev/null` for `*.front.sepia.ceph.com` so the agent account stops recording testnode keys in the first place.

### DIFF
--- a/teuthology/misc.py
+++ b/teuthology/misc.py
@@ -1101,6 +1101,23 @@ def ssh_keyscan_wait(hostname):
             log.info("try ssh_keyscan again for " + str(hostname))
         return success
 
+
+def ssh_keygen_remove(hostname):
+    """
+    Remove any entries for a host from the user's ~/.ssh/known_hosts
+    (``ssh-keygen -R``).
+
+    :param hostname: The hostname
+    :returns: True if the command succeeded
+    """
+    args = ['ssh-keygen', '-R', hostname]
+    p = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    if p.returncode != 0:
+        log.warning("%s failed: %s", ' '.join(args),
+                    p.stdout.decode().strip())
+    return p.returncode == 0
+
+
 def stop_daemons_of_type(ctx, type_, cluster='ceph', timeout=300):
     """
     :param type_: type of daemons to be stopped.
--- a/tests/orchestra/test_connection.py
+++ b/tests/orchestra/test_connection.py
@@ -1,4 +1,7 @@
+import paramiko
+
 from mock import patch, Mock
+from pytest import raises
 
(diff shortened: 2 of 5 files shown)


teuthology/misc.py
```diff
@@ -1084,6 +1084,23 @@ def _ssh_keyscan(hostname):
         return sorted(keys)[0]
 
 
+def ssh_keygen_remove(hostname):
+    """
+    Remove any entries for a host from the user's ~/.ssh/known_hosts
+    (``ssh-keygen -R``).  Meant for hosts that were just reimaged, whose host
+    key is expected to have changed.
```
kshtsk: I think the comment "Meant for hosts that.... " we can probably skip here, since this should be only description for what the method is doing, and it is enough to explain why we need to drop all known keys for the host on a caller side.

teuthology/misc.py
```diff
@@ -1084,6 +1084,23 @@ def _ssh_keyscan(hostname):
         return sorted(keys)[0]
 
 
+def ssh_keygen_remove(hostname):
```
kshtsk: Move it above the definition of `ssh_keyscan` or below `ssh_keyscan_wait`, so the `_ssh_keyscan`  and `ssh_keyscan`  and `ssh_keyscan_wait` kept next to each other.

kshtsk (approved):

PR #2256, opened 2026-09-15
[DNM] introduce monkeythreading package

kshtsk:
Instead of monkey patching in teuthology/_init_ locking all modules for patching, let modules be able to work without monkey-patching, and only initiate patching from cli scripts just including teuthology.monkeythreading


kshtsk:
The failure iin integration test is unrelated the ansible error:
```
ERROR! [DEPRECATED]: community.postgresql.postgresql_set has been removed. Use community.postgresql.postgresql_alter_system instead. This feature was removed from community.postgresql in
version 5.0.0. Please update your playbooks.

The error appears to be in '/root/src/github.com_ceph_ceph-cm-ansible_main/roles/paddles/tasks/setup_db.yml': line 2, column 3, but may
be elsewhere in the file depending on the exact syntax problem.
```
only associated with ceph-cm-ansible.

kshtsk:
@djgalloway @zmc does it makes sense to stick the postgresql collection to `<5.0.0` or it is better to fix in ceph-cm-ansbile?

djgalloway:
> @djgalloway @zmc does it makes sense to stick the postgresql collection to `<5.0.0` or it is better to fix in ceph-cm-ansbile?

Made more sense to update the role IMO: https://github.com/ceph/ceph-cm-ansible/pull/892/changes

